Issue #1806, opened 2016-03-04
one minute delay on return (pipe) and a POST-Request


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1806

daghf:
I looked into this.

The issue seems to be that for return (pipe) we don't pass the request body. For pipe this is deferred until we do V1P_Process. This is fine for piped connections where we use a fresh backend connection, but when we get a recycled connection (state == STOLEN) we do VBT_Wait and go on the waiter first.

Thus the backend receives a request indicating a req body and waits around for that. At the same time the vbc is on the waiter expecting data from the backend, and we end up in a deadlock situation until timeout.

A fix could be to always force a fresh backend connection for piped requests (thereby avoiding the waiter), but there might be a more elegant solution?

Krapulat:
+1

fgsch:
Related to #1772.

mbarcia:
Current version on Debian Jessie 4.1.2 (Mar 4th) still has this bug, which was fixed on March 14th. 

Any idea when this fix is going to reach Debian Jessie? Thank you very much.

lkarsten:
@mbarcia We'll do another release when we're done with the backporting efforts, probably in June. After that you have to bug @ssm or another packager to upload the fresh version.

lkarsten:
Strictly speaking, since it is jessie, our official packages covers it and will most likely be out a day or two after the release.

mbarcia:
@lkarsten thank you very much

lkarsten:
backport review: commited to 4.1 head, needs 4.0 review.

Krapulat:
@lkarsten When it will be available the fix in Varnish 4.1 for Ubuntu?

cherouvim:
Hello. Any idea when is this comming to ubuntu? I have a couple of ubuntu 16 servers with migrated sites and until this is fixed I have to change all pipe to pass in the vcls in order for posts to work quickly, but this introduces some other problems.

Thanks

denisbr:
@cherouvim Ubuntu packaging for Xenial is handled by the maintainers of the ubuntu package maintainers. @ssm might be able to guesstimate something, but due to the nature of the process it's not easy.

PR #1877, opened 2016-03-14
Postpone calling VBT_wait until after V1P_Process

daghf:
Fixes a bug where we went on the waiter prior to passing the request
body.

The VBT_Wait call is now handled in vbe_dir_finish.

Test case by thomaslc

Fixes: #1806


bsdphk:
OK for commit.

lkarsten:
Backports handled in #1806 .

Issue #1879, opened 2016-03-15
Wrong set of duplicate haeaders when backend  304 NOT MODIFIED


hermunn:
When a backend supplies duplicate headers in a response, this is cached normally and served correctly to the client. If the object goes into grace, and Varnish has to ask the backend for a fresh copy, the backend may return a "304 NOT MODIFIED". In this case, the backend can choose which headers to send again, as specified by RFC 2616, section 13.5.3.

This is not handled correctly in 4.0 (verified), and seems to be buggy in master as well (not verified). The bugs are limited to cases where headers are duplicated, for example

foo: Alice
foo: Bob

I have written a patch for 4.0, but not tested it. I will attempt to write a varnishtest test for these cases and then resolve any errors.

fgsch:
Fixed in commit d6ed2444b481d73ca51ab2c621827e945f50b389.

PR #1880, opened 2016-03-15
Proper handling of duplicate headers on IMS headers merge

hermunn:
This fixes a problem when a backend replies with 304 Not Modified
(after a http conditional request from varnish) and does not supply a
header that was duplicate in the cached object.

Before this patch, varnish would only supply (by copying from the
expired object) the first instance of a duplicate header.

### DIFF
--- a/bin/varnishd/cache/cache_http.c
+++ b/bin/varnishd/cache/cache_http.c
@@ -992,6 +992,8 @@ HTTP_Merge(struct worker *wrk, struct objcore *oc, struct http *to)
 	const char *ptr;
 	unsigned u;
 	const char *p;
+	unsigned nhd_before_merge;
+	unsigned v;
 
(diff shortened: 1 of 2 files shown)


mbgrydeland:
This pull request is to fix #1879

fgsch:
Since u is an index, can we just reuse it without introducing nhd_before_merge?

hermunn:
I suggest removing the 'v' variable and write 'u' instead. The variable nhd_before_merge should survive, in my humble opinion, as it helps the reader to understand what is going on. I blame my personal code style for the introduction of the 'v' variable, by the way.

fgsch:
For this particular case I find the naming superfluous.
The correct way to document the code, should it call for it, is using comments.

bsdphk:
OK for commit.

lkarsten:
Backports handled in #1879 .

PR #1884, opened 2016-03-18
Only accept C identifiers as acls

### LINKED ISSUE #1883: IP matching with a header crashes the vcc
``server.ip ~ resp.http.foo`` crashe the vcl compilation

## Expected Behavior
Compilation error with error message explaining the behavior is invalid

## Current Behavior
Crash

vtc:
```
varnishtest "Matching on an empty string"

server s1 {
	rxreq
	txresp
} -start

varnish v1 -vcl+backend {
	import std;

	sub vcl_deliver {
		if (server.ip ~ resp.http.foo) {
			set resp.http.wont = "reach";
		}
	}
} -start

client c1 {
	txreq
	rxresp
} -run
```

## Your Environment
* Version used: master
* Operating System and version: arch linux, up-to-date at the time of writing

fgsch:
Fixes #1883.

### DIFF
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -1215,7 +1215,7 @@ vcc_expr_cmp(struct vcc *tl, struct expr **e, enum var_type fmt)
 	    (tl->t->tok == '~' || tl->t->tok == T_NOMATCH)) {
 		not = tl->t->tok == '~' ? "" : "!";
 		vcc_NextToken(tl);
-		ExpectErr(tl, ID);
+		vcc_ExpectCid(tl);
 		vcc_AddRef(tl, tl->t, SYM_ACL);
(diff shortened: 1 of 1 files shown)


bsdphk:
OK for commit.

Dridi:
I believe this is the wrong fix, now that named ACLs have a proper type it is possible to get one from an expression involving a vmod.

fgsch:
That would require more changes that are not specific to the problem at hand or this fix, aka this does not break anything that was working before.

Dridi:
That's a proper fix for the 4.0 branch, but not for  4.1 or master. There should be a way to fail the compilation if the expression's value is not an ACL instead of asserting.

fgsch:
Dridi, no idea what you are talking about. Did you check the code and or run it pre and post fix or you're assuming?

Dridi:
I might, will check.

lkarsten:
backports handled in #1883.

PR #1886, opened 2016-03-21
Add "last updated" timestamp for backend health in varnishadm

aondio:
I've already send this patch to the dev-list and got and OK from PHK.

This PR is to let everyone know what this patch is for:
in 'varnishadm backend.list' it also logs when a backend's health was last changed.


lkarsten:
Re backports; this is a feature extension that changes the output format. Not for 4.0/4.1.

PR #1891, opened 2016-04-01
Enable optional lcov support

Dridi:
I came across this by accident yesterday, but a bunch of people came up with turnkey autoconf magic to enable code coverage using `lcov`.

The integration is painless and non-intrusive, except for the test suite that I run with a 120s `cli_timeout` because I've had failing tests. I realized after bumping the timeout that I had accidentally run with parallel jobs but instead of retrying I figured it could be kept as a safeguard.

Last time I have used varnish-gcov, I had to tweak it so it wouldn't clone on my behalf (or at least make it clone my own local clone given that I was working offline) but having this built-in the source tree makes things a lot easier:

```
./configure --enable-code-coverage && make lcov
[coffee break]
Overall coverage rate:
  lines......: 82.7% (26253 of 31748 lines)
  functions..: 90.4% (1778 of 1967 functions)
  branches...: 57.8% (17035 of 29471 branches)
file:///home/dridi/varnish/src/vc/lcov/index.html
```

Not having to context switch with varnish-gcov is also nice when someone wants to measure the coverage of their code before sending patches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,9 @@ cscope.*out
 /bin/varnishtest/tests/*.log
 /bin/varnishtest/tests/*.log-t
 /bin/varnishtest/test-suite.log
+
+# Code coverage with lcov/gcov
+*.gcno
+*.gcda
+*.info
+/lcov/
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,6 +25,17 @@ distcleancheck_listfiles = \
 	find . -type f -exec sh -c 'test -f $(srcdir)/$$1 || echo $$1' \
 		sh '{}' ';'
 
+# lcov support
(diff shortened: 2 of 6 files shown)


gquintard:
Would be nice to have some lines in README on how to use it.

Looks sane otherwise.

bsdphk:
OK with me.

I run a set of scripts which publish gcov results here:  https://www.varnish-cache.org/static/gcov/

If the script which does that can be made less magic, then I'm all for it.

Dridi:
If you share those scripts with me, I can adapt them to publish something to `/static/lcov` too/instead. At the very least I can get a stab at simplifying them.

Is lcov available on FreeBSD? Since lcov is part of the Linux Test Project I'm not sure, but OTOH lcov is supposed to be only a frontend to gcov.

PR #1894, opened 2016-04-05
Use posix_fallocate if available

fgsch:
If both fallocate and posix_fallocate are available, try the former and fallback to the posix one if the filesystem is not ext4.


fgsch:
Martin has pointed out that in FreeBSD (at least) calling posix_fallocate is not different than the existing code when called with insist set to 1. This might be OK for SHMLOG but definitely a bad idea for storages so withdrawing this PR.

hermunn:
Since the pull request has been redrawn, it is not eligible for back port.

Issue #1856, opened 2016-03-04
LostHeader setting req.url to an empty string


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1856

nigoroll:
We got the same issue with `http_max_hdr` exceeded, so I created https://github.com/varnishcache/varnish-cache/wiki/VIP10-Differentiate-out-%22out-of-workspace%22-errors
@bsdphk @fgsch I could be a good idea to document there the plan you have agreed upon

fgsch:
I don't quite see the point of creating a VIP for something that is already agreed on but OK.

nigoroll:
@fgsch did you notice the fact that there is another aspect too it which we had missed so far? My understanding is that the VIPs are our preferred place for collecting design specs.

fgsch:
As I said, if you want to have a VIP for it, sure. I don't think there is any spec to design though, just sit down and write the code.

PR #1895, opened 2016-04-05
Improve error when calling vrt_do_string with ""

fgsch:
Should address #1856 to some extent.


bsdphk:
We shoudln't mark overflow in that case, should we ?

fgsch:
I followed what we do if we try to set .status to an invalid code. 
I'm not convinced overflow is the way to go but I'm not super keen on ignoring the error.
What are the alternatives?

fgsch:
As discussed on IRC phk would prefer if this adds the 50x abort mechanism in VRT context instead of piggybacking on the overflow.

fgsch:
This will be replaced by a different implementation eventually. 

Closing.

hermunn:

PR #1897, opened 2016-04-07
uint64_t VTIM functions and a trivial benchmark

nigoroll:
to prep for the migration to uint64_t timestamps

### DIFF
--- a/bin/varnishd/mgt/mgt_vcc.c
+++ b/bin/varnishd/mgt/mgt_vcc.c
@@ -299,7 +299,7 @@ mgt_VccCompile(struct cli *cli, const char *vclname, const char *vclsrc,
 	if (status || C_flag) {
 		(void)unlink(vp.libfile);
 		free(vp.libfile);
-		AZ(rmdir(vp.dir));
+		(void)rmdir(vp.dir);
 		free(vp.dir);
 		if (status) {
 			VCLI_Out(cli, "VCL compilation failed");


nigoroll:
Discussed with @bsdphk on irc:
* his idea: could split the 64bit as 36bit seconds + 28bit nsecs
* saves the division / modulo - shift or and+shift instead
* would give ~3.7ns resolution
* would last another 89 years

bsdphk:
So, as you may know I've been moonlighting as a Time Lord since the early 1990ies, and this is a subject close to my ulcer.

There are three problems with timeformats:

1. getting sufficient bits.
2. presenting them to users.
3. efficiently.

The first one is rather interesting, if one wanted to make a numerical time format that could serve all purposes, from quantum mechanics to cosmology, one needs about 100 bits, roughly 60 bits of seconds and 40 bits for fractions of seconds. int128_t here we come.

We can narrow that window somewhat, a hundred years of nanoseconds fits nicely inside 64 bits.

The absolutely wrong way to do this is called "struct timespec".  The british gave up on their mixed radix coinage for a reason.

But that brings us to the next part:  presentation.

To present such a count of nanoseconds as a number of seconds to users is messy.

Forget anything about %T printf extension formats, I spent a lot of time implementing #include "printf.h" in FreeBSD, only to discover that no C-compilers have __printf_like__() which can be taught about it.  And neither can Coverity, FlexeLint or any other code quality tool.  Giving up __printf_like__() is too high a cost.

One could render the number in decimal, and artificially stick a period 9 digits from behind, and throw away six digits if only milliseconds were required and so on. If only seconds were wanted, we could divide by a billion.

This sounds easy, but is messy in actual source code no matter what you do.  People even forget how many zeros are in a billion.

About the best compromise is something like::

  VSLb(val, SLT_FooBar, "At the time of the tone the clock is" VTIMFMT " exactly.", VTIM_STR(t0));

VTIM_STR() may return to values separated by a comma, which makes compiler error messages totally incomprehensible.

Finally computers are not decimal, they're binary and thus the fraction should be binary.

That means that you can derive seconds with a shift instead of a division, and fractional seconds with a mask rather than remainder.  In particular it means that you can add a integer number of seconds with a shift rather than a multiply-add.

But more importantly, it allows you very cheap conversions between different formats.  If you have a 32.8 timestamp, converting to 64.32 is just a shift, and vice versa.

(we use float rather than double for space reasons in struct exp)

The downside to binary fractions is that you have to multiply and shift to present them, and you have to multiply and shift to get from the cruddy time system calls to a binary fraction format.

And all that mess is why I decided to say "Screw it!" and used doubles instead:   Because both the CPU hardware, printf(3) and programmers understands those.

I trust Devons measurements when he says using FP has an impact, but the improvement has to be credible before I see the advantage to messing up our source-code, and I want to see some serious thought go into minimizing its readability impact on the source code.

nigoroll:
thoughts:
* do we really need the computationally most efficient approach of using a base 2 fixed point number, or do we gain 90% by using base 10 as @dhobsd suggested?
* no matter if we stick with doubles or  use a different representation, much auf the printf-conversions are happening to vsl strings, so it could help to move this burden to the vsl reader

bsdphk:
A (partial-)binary VSL format is not a small undertaking, so lets not spend time on that.

As I tried to convey:  My main worry is how this look in source code.

Base 2/10 is only an efficiency issue.

lkarsten:
Nothing commited, nothing to backport.

PR #1902, opened 2016-04-13
 Add an object head condition variable to wait on and use it for ESI

nigoroll:
Add an object head condition variable to wait on and use it for ESI
    
    Previously, ved_include() for ESI include processing retried the
    CNT_Request() FSM step if it hit a busy object, sleeping 10ms
    inbetween attempts. This could lead to starvation symptoms, for
    example in the following scenario when ved_include() "probes", with
    high probability, would always hit the busy case:
    
    - ESI include requires variant A of an object
    - other requests contantly insert variant B with ttl = grace = 0
    - backend requests need long processing time
    
    We now add, next to the waitinglist head, a condition variable which
    gets signalled whenever requests on the waitinglist get woken up in
    hsh_rush(). As all other struct objhead members, it is protected by
    the struct objhead mtx.
    
    ESI threads now block on this condition variable whenver they hit a
    busy object. Parking an ESI processing request on the waitinglist like
    esi_level == 0 requests could be more efficient, but would require
    handling much of the state currently in the ESI processing thread
    differntly, so blocking the ESI thread should be the best viable
    option.
    
    Most of HSH_Lookup() is now in hsh_lookup() to allow retries without
    re-aquiring oh->mtx.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -548,8 +548,6 @@ struct req {
 	const struct transport	*transport;
 	void			*transport_priv;
 
-	VTAILQ_ENTRY(req)	w_list;
-
 	struct objcore		*body_oc;
 
 	/* The busy objhead we sleep on */
--- a/bin/varnishd/cache/cache_ban_lurker.c
+++ b/bin/varnishd/cache/cache_ban_lurker.c
@@ -203,7 +203,7 @@ ban_lurker_test_ban(struct worker *wrk, struct vsl_log *vsl, struct ban *bt,
 			if (i)
 				ObjSendEvent(wrk, oc, OEV_BANCHG);
 		}
-		(void)HSH_DerefObjCore(wrk, &oc);
+		(void)HSH_DerefObjCore(wrk, &oc, 0);
 	}
 }
 
--- a/bin/varnishd/cache/cache_busyobj.c
+++ b/bin/varnishd/cache/cache_busyobj.c
@@ -171,7 +171,7 @@ VBO_ReleaseBusyObj(struct worker *wrk, struct busyobj **pbo)
 
 	if (bo->fetch_objcore != NULL) {
 		AN(wrk);
-		(void)HSH_DerefObjCore(wrk, &bo->fetch_objcore);
+		(void)HSH_DerefObjCore(wrk, &bo->fetch_objcore, -1);
 	}
 
 	VCL_Rel(&bo->vcl);
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -79,7 +79,6 @@ ved_include(struct req *preq, const char *src, const char *host,
 {
 	struct worker *wrk;
 	struct req *req;
-	enum req_fsm_nxt s;
 
 	CHECK_OBJ_NOTNULL(preq, REQ_MAGIC);
 	CHECK_OBJ_NOTNULL(ecx, ECX_MAGIC);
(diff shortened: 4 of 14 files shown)


bsdphk:
We're on the right track here, but putting a CV into every  objhead is a no-go from a memory point of view.

The fundamental problem here is that we let the including request also carry out the included request, and that prevents us from properly go on the waitinglist with the included request.

If we peek ahead to parallel-ESI, we can no longer use this trick anymore anyway, so maybe we should just bite the bullet and do the included request in a separate thread?

That would allow us to put the CV in the 'struct ecx' where it belongs.

The downside is that we leave the including req's worker sleeping on the CV during the include, but that will happen anyway with parallel-ESI.  We have already swallowed a similar impact by the req/backend thread decoupling.

nigoroll:
* this PR: I agree that the memory footprint is a no-go, thank you. On my system, my patch would blow up struct objehead from 104 to 152 bytes. I should have checked this before.
So should we have a dynamically alloc'ed waitinglist head and elements? The head would contain the actual vtailq head and the cv, the waitinglist elements would only point to reqs. This would remove two pointers from struct req and one pointer from the objhead. We'd have a pointer to the waitinglist head IFF a waitinglist exists. We could have a mempool for the two structs, but for waitinglist events (which should not be that frequent) I think that plain malloc/free (via miniobj) should be acceptable.
@bsdphk, Please let me know what you think.
* big picture:
Regarding the big picture of the parallelism required for ESI and H2, I simply don't know your plan. I had sketched my 2 cents in http://www.gossamer-threads.com/lists/varnish/dev/35387 , but I just don't know which route you're following.
If we were to follow my 2 cents, we'd generalize VBO events to also include the waitinglist logic. What I'd see happening for ESI is included in the email. Basically, the read thread would scan bodies for ESI includes to kick off new reqs and create the dependency linkage between them. The write thread would stitch together the response as objects become available.
But I don't think this is the right scope for this PR.

bsdphk:
I need to look at the code before I can say anything on that idea, but it certainly sounds better.

Parallel ESI is very different from H/2 (which is not relevant here).

The best model for parallel ESI is parallel make(1):  the top-esi receives a configured number of tokens, takes one for itself, and any remaining can be used to jump-start esi-includes.  By lending down its own token if necessary, progress is always ensured.

nigoroll:
@bsdphk , I'd appreciate if you could take the time to have a look regarding the waitinglist proposal. I'd like to either succeed with this improvement or stop working on it as early as possible.
The counting-semaphore approach to parallelism makes sense, but I don't understand the big picture you have in mind. This is something I think we should talk about in person if you are interested.

nigoroll:
@bsdphk I've updated the PR to allocate the waitinglist on demand, which should address the memory usage concerns. This removes 8 bytes from objhead compared with current master (second tailq pointer), it is not 96 bytes on my machine (104 before).

nigoroll:
running pre-production now, will go live tomorrow, updated PR to re-check oh->waitinglist under the lock in `HSH_DerefObjHead`

nigoroll:
another update: replace tail-recursion with an actual while loop to avoid running out of stack space when the compiler doesn't optimize away the recursion

bsdphk:
The commit above implements this a much leaner way.

Thanks to Nils for spearheading this.

hermunn:
Backport review: We are not prepared to backport this to 4.1 (or 4.0) unless it becomes part of fixing a crash in varnish.

nigoroll:
@hermunn I'm with you. The changes implemented by @bsdphk will only be a performance improvement for most cases. The issue of ESI request starving while waiting is really quite a corner case and I do not expect that many varnish users to hit it.

PR #1905, opened 2016-04-14
Backend mode in varnishncsa

hermunn:
A new mode, backend mode, in varnishncsa is introduced. It lets you
log trafic from varnish to the backends. Typically you will run this
in addition to varnishncsa in normal mode.


bin/varnishncsa/varnishncsa.c
```diff
@@ -792,8 +795,10 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 
 	for (t = pt[0]; t != NULL; t = *++pt) {
 		CTX.gen++;
-		if (t->type != VSL_t_req)
-			/* Only look at client requests */
+		/* In client mode, only look at client requests.
+		   In bacend mode, only look at backend requests. */
```
gquintard: missing 'k'

fgsch:
I like this but I think we should add support for both -b and -c, and let the user decide without forcing them to run 2 separate processes. This is inline with the rest of the tools.

Forcing the user to run multiple varnishncsa instances is not ideal  If they really want to have different logs, they can specify -b or -c.

bin/varnishncsa/varnishncsa.c
```diff
@@ -134,8 +135,8 @@ struct ctx {
 
 	/* State */
 	struct watch_head	watch_vcl_log;
-	struct watch_head	watch_reqhdr;
-	struct watch_head	watch_resphdr;
+	struct watch_head	watch_reqhdr; /* also breqhdr */
```
fgsch: typo. should be bereqhdr

bin/varnishncsa/varnishncsa.c
```diff
@@ -718,6 +719,8 @@ frag_fields(int force, const char *b, const char *e, ...)
 	frag = va_arg(ap, struct fragment *);
 	for (n = 1, q = b; q < e; n++) {
 		assert(field > 0);
```
fgsch: You could drop this assert since the one below covers both.

bin/varnishncsa/varnishncsa.c
```diff
@@ -809,10 +814,22 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
...
+				   mode. However, assert() is too strong,
+				   so we just ignore such "impossible"
+				   cases. */
 			case SLT_HttpGarbage:
 				skip = 1;
 				break;
 			case SLT_PipeAcct:
+				/* PipeAcct is only possible for requests,
+				   but to be safe we check: */
+				if (CTX.b_opt) break;
```
fgsch: All this if (..) break are not making the cases very readable.

hermunn: This is correct, but I prefered this over other options I could think of. The choice was between
* Making two switch statements, one for each mode. This would result in code duplication, which is very bad.
* Doing asserts. In this case, varnishncsa would die if, for example, we saw a SLT_PipeAcct when in backend mode. This is probably not desired behavior.
* The current solutions, where "impossible" fields are ignored, and as much code as possible is shared between the two modes. In my opinion, this is the shortest path to dual mode, if this is ever going to be implemented.

I may have overlooked a better solution, of course.

fgsch:
2 wrongs don't make one right. I'd take code duplication any time over unreadable code.
My comment does not apply to that particular case either, but all of them.

Also, why there are these if .. break when you have the:

```
+		if ((t->type != VSL_t_req && !CTX.b_opt)
+		    || (t->type != VSL_t_bereq && CTX.b_opt))
```

hermunn:
If I were not to safeguard against bad behavior, then all the if() break; statements would go away. I made a choice, and If the consensus is that it was the wrong one, then I will change it. I guess what I am trying to say is that someone (probably Martin) should just decide how it should be, so that I can implement it and we can move on to something else.

hermunn:
I have thought a lot about this, and think I have found a way to make everyone happy when it comes to readability, maintainability and function. I will update the pull request "soon", and then everyone can look at it again.

However, I need to go bug hunting a bit first, on a different part of the code.

hermunn:
I have now slightly rewritten the most problematic code, and implemented mixed mode (both client and backend requests in the same log). I hope the code is easier to read, easier to maintain, and that the behavior of the old varnishncsa is kept exactly.

The switch construction is a bit hackish, but I hope you will like it anyway.

bin/varnishncsa/varnishncsa.c
```diff
@@ -786,15 +794,24 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
...
+
+		/* Consider client requests only if in client mode.
+		   Consider backend requests only if in backend mode. */
+		if (t->type == VSL_t_req && CTX.c_opt) {
+			CTX.side = "client";
+			be_mark = 0;
+		} else if (t->type == VSL_t_bereq && CTX.b_opt) {
+			CTX.side = "backend";
+			be_mark = BACKEND_MARKER;
+		} else continue;
```
fgsch: Why not simply?

               if ((t->type == VSL_t_req && !CTX.c_opt) ||
                   (t->type == VSL_t_bereq && !CTX.b_opt))
                        continue;

fgsch: Also perhaps use "c" and "b" to be in sync with other tools, e.g. varnishlog?

hermunn: In varnishncsa, the variable CTX already had an a_opt member, so for the two other options, b_opt and c_opt was chosen.

hermunn: The if test

```
if ((t->type == VSL_t_req && !CTX.c_opt) ||
    (t->type == VSL_t_bereq && !CTX.b_opt))
        continue;
```

does not work when both b_opt and c_opt are set. Also, similar constructions would remove safety against "impossible" combinations like SLT_PipeAcct when t->type == VSL_t_bereq . In the current version this bad behavior from varnish would be ignored, and I want to keep the current behavior in the new version.

fgsch: > In varnishncsa, the variable CTX already had an a_opt member, so for the two other options, b_opt and c_opt was chosen.

Not sure what you meant.

> does not work when both b_opt and c_opt are set. Also, similar constructions would remove safety against "impossible" combinations like SLT_PipeAcct when t->type == VSL_t_bereq . In the current version this bad behavior from varnish would be ignored, and I want to keep the current behavior in the new version.

Ugh? How come it will not work?
I don't think the impossible scenarios justify the extra code. If you are concerned about SLT_PipeAcct you can check that particular one.  We have different records (e.g. ReqHeader vs BereqHeader) for that specific purpose.
The code does not handle backend records in client requests either.

fgsch:
> In varnishncsa, the variable CTX already had an a_opt member, so for the two other options, b_opt and c_opt was chosen.

Not sure what you meant.

> does not work when both b_opt and c_opt are set. Also, similar constructions would remove safety against "impossible" combinations like SLT_PipeAcct when t->type == VSL_t_bereq . In the current version this bad behavior from varnish would be ignored, and I want to keep the current behavior in the new version.

Ugh? How come it will not work?
I don't think the impossible scenarios justify the extra code. If you are concerned about SLT_PipeAcct you can check that particular one. We have different records (e.g. ReqHeader vs BereqHeader) for that specific purpose.
The code does not handle backend records in client requests either.

hermunn:
> Not sure what you meant.

I misunderstood, but Dag cleared it up for me, so I changed the reply. You were right all along about "c" and "b" being better than "client" and "backend".

> Ugh? How come it will not work?

if `t->type` is neither `VSL_t_req` nor `VSL_t_bereq`, then test should evaluate to true, but that did not seem to be the case in the quoted code.

> The code does not handle backend records in client requests either.

If a client request is coming in, then `be_mark` will be `0`, and then `tag == SLT_BereqHeader` will be ignored since there is no `case SLT_BereqHeader`, only `case (SLT_BereqHeader + BACKEND_MARK)`. But maybe I misunderstood again?

I just looked at the code immediately below the switch, and realized that

```
			if ((tag == SLT_ReqHeader && CTX.c_opt)
			    || (tag == SLT_BereqHeader && CTX.b_opt))
				process_hdr(&CTX.watch_reqhdr, b, e);
			if ((tag == SLT_RespHeader && CTX.c_opt)
			    || (tag == SLT_BerespHeader && CTX.b_opt))
				process_hdr(&CTX.watch_resphdr, b, e);
```

is wrong, and I should consider `t->type` instead of `CTX.c_opt` and `CTX.b_opt`.

fgsch:
> if t->type is neither VSL_t_req nor VSL_t_bereq, then test should evaluate to true, but that did not seem to be the case in the quoted code.

varnishncsa only support vxid or request grouping mode. No other records are possible.

bin/varnishncsa/varnishncsa.c
```diff
@@ -786,15 +794,24 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 	unsigned tag;
 	const char *b, *e, *p;
 	struct watch *w;
-	int i, skip;
+	int i, skip, be_mark;
 	(void)vsl;
 	(void)priv;
 
+#define BACKEND_MARKER VSL_t__MAX
```
mbgrydeland: I liked this better when BACKEND_MARKER was waaay higher than VSL_t__MAX. If the varnishlog and varnishd binaries are out of sync (not completely unlikely if e.g. one considers a log sent to support), then there can be strangeness observed.

doc/sphinx/reference/varnishncsa.rst
```diff
@@ -34,6 +34,30 @@ The following options are available:
 
 .. include:: ../include/varnishncsa_options.rst
 
+BACKEND MODE
```
mbgrydeland: Backend mode being a title of it own suggests that backend_mode is something added onto rather than being a first class citizen. I'd prefer a title of 'MODES' that discusses the different modes.

doc/sphinx/reference/varnishncsa.rst
```diff
@@ -50,19 +74,23 @@ Supported formatters are:
 
 %b
   Size of response in bytes, excluding HTTP headers.  In CLF format,
-  i.e. a '-' rather than a 0 when no bytes are sent.
+  i.e. a '-' rather than a 0 when no bytes are sent.  For backend mode,
```
mbgrydeland: I'd like these descriptions to be less like backend mode is just something bolted on. Maybe along the lines of:
"In client mode, the number of bytes excluding HTTP headers sent to the client. In backend mode the number of bytes excluding HTTP headers received from the backend. In CLF format..."

bin/varnishncsa/varnishncsa.c
```diff
@@ -1006,6 +1048,8 @@ main(int argc, char * const *argv)
 				usage(1);
 		}
 	}
+	/* default is client mode: */
+	if(!CTX.b_opt) CTX.c_opt = 1;
```
mbgrydeland: Missing SPC after 'if', and preferably on two lines

bin/varnishncsa/varnishncsa.c
```diff
@@ -786,15 +794,24 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
...
+
+		/* Consider client requests only if in client mode.
+		   Consider backend requests only if in backend mode. */
+		if (t->type == VSL_t_req && CTX.c_opt) {
+			CTX.side = "c";
+			be_mark = 0;
+		} else if (t->type == VSL_t_bereq && CTX.b_opt) {
+			CTX.side = "b";
+			be_mark = BACKEND_MARKER;
+		} else continue;
```
mbgrydeland: continue should be on its own line

hermunn:
Pushing again, after editing according to @mbgrydeland's wishes...

hermunn:
Now @mbgrydeland has made a patch that adds logging in varnishd, necessary to log the IP of the backend correctly, and this is now part of this pull request (first commit). The second commit is the backend mode in varnishncsa.

hermunn:
Merged into master and 4.1 manually (fast forward).

lkarsten:
Backport review: In 4.1 as 5a46b8c9932c04d5e91b7c7f2c8d9d95fa86335d and 3b078764fda14c3cedae58028d38c1854803b077. Will be in next VC4.1 release.

This is new functionality and as such not for 4.0.

PR #1906, opened 2016-04-14
Panic address dump

mbgrydeland:
If an address is given to the VAS_Fail function, a memory dump of that
memory area is included in the panic output.

### DIFF
--- a/bin/varnishd/cache/cache_panic.c
+++ b/bin/varnishd/cache/cache_panic.c
@@ -38,6 +38,8 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <signal.h>
+#include <setjmp.h>
+#include <ctype.h>
 
 #include "cache.h"
 #include "cache_transport.h"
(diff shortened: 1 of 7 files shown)


bsdphk:
I am very sympathetic to the idea, but I have a couple of quips with the implementation.

First: I really hate the setjmp thing, it has only one valid use IMO and that is user-space coroutine schedulers and we're not doing that.  Wouldn't mincore(2) be good enough for this ?

Second, the panic calls should be able to specify both an address and a length.  (We might even consider making it varadic so multiple ptr-len pairs can be specified.

Third, we should teach VSB_quote() about HEX+ASCII dumps, it's generally useful.

mbgrydeland:
> First: I really hate the setjmp thing, it has only one valid use IMO and that is user-space coroutine schedulers and we're not doing that. Wouldn't mincore(2) be good enough for this ?

I agree that using setjmp/longjmp should trigger the cringe reflex, but I still maintain that it is the safest way of performing the task. A page marked in core from mincore would be pretty safe to read, though we could still be racing against the malloc implementation that is actively freeing pages in another thread, or the vcl unload or something, and the access would cause an unhandled segfault and the panic message cut short.

nigoroll:
@bsdphk can conclude this one?

bsdphk:
I guess the final discussion/conclusion never made it into this ticket:

The right way to do this is using mincore(2) before poking unknown pointers, not using setjmp to recover from the disasters caused if we do hit a bad location.

I'm closing this ticket as unactionable, pending a patch which does that.

hermunn:

PR #1910, opened 2016-04-18
Ban lurker improvements

nigoroll:
Implement ban lurker efficiency improvements based on the findings discussed on -dev (ban lurker questions).
@mbgrydeland, I'd very much appreciate your review - in particular regarding the `ban_export()` related change

### DIFF
--- a/bin/varnishd/cache/cache_ban_lurker.c
+++ b/bin/varnishd/cache/cache_ban_lurker.c
@@ -36,7 +36,8 @@
 #include "hash/hash_slinger.h"
 #include "vtim.h"
 
-static struct objcore oc_marker = { .magic = OBJCORE_MAGIC, };
+static struct objcore oc_mark_cnt = { .magic = OBJCORE_MAGIC, };
+static struct objcore oc_mark_end = { .magic = OBJCORE_MAGIC, };
 static unsigned ban_batch;
 static unsigned ban_generation;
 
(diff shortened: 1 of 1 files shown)


bsdphk:
I have too much piled up already this week to get to this review.

I suggest you try to run some load on it in the meantime.

nigoroll:
running pre-production now, will go live tomorrow
Updated patch:
* locking in `ban_lurker_getfirst` was wrong

nigoroll:
survived first torture-test with 32k intentionally inefficient artificial bans added in one go (Varnish users: avoid such inefficient regexen)

```
1461162130.968138 14406 -  obj.http.x-url ~ ^/index.php
1461162105.005207     0 C  
1461162074.200886     0 -  obj.http.x-url ~ /some\?.*id=XXXTEST40000.* && obj.http.x-host ~ .*
1461162074.197021     0 -  obj.http.x-url ~ /some\?.*id=XXXTEST39999.* && obj.http.x-host ~ .*
...
```

some stats:

```
MAIN.uptime               4134         1.00 Child process uptime
MGT.uptime                                4134         1.00 Management process uptime
```
```
MAIN.n_object                 17196          .   object structs made
MAIN.n_objectcore             17217          .   objectcore structs made
MAIN.n_objecthead              9522          .   objecthead structs made
MAIN.bans                      8684          .   Count of bans
MAIN.bans_completed               2          .   Number of bans marked 'completed'
MAIN.bans_obj                  8684          .   Number of bans using obj.*
MAIN.bans_req                     0          .   Number of bans using req.*
MAIN.bans_added               45900        11.80 Bans added
MAIN.bans_deleted             37216         9.56 Bans deleted
MAIN.bans_tested              10554         2.71 Bans tested against objects (lookup)
MAIN.bans_obj_killed            793         0.20 Objects killed by bans (lookup)
MAIN.bans_lurker_tested    438724582    112753.68 Bans tested against objects (lurker)
MAIN.bans_tests_tested      35185602      9042.82 Ban tests tested against objects (lookup)
MAIN.bans_lurker_tests_tested    438724586    112753.68 Ban tests tested against objects (lurker)
MAIN.bans_lurker_obj_killed           9221         2.37 Objects killed by bans (lurker)
MAIN.bans_dups                        7970         2.05 Bans superseded by other bans
MAIN.bans_lurker_contention              0         0.00 Lurker gave way for lookup
MAIN.bans_persisted_bytes         15246979          .   Bytes used by the persisted ban lists
MAIN.bans_persisted_fragmentation     12352537          .   Extra bytes in persisted ban lists due to fragmentation
MAIN.n_obj_purged                            0          .   Number of purged objects
```

nigoroll:
Until we hit #1919 (which is not related to these patches to the best of my unerstanding), we survived without issues despite ban lurker banging (create thousands of object bans, create intermediate request bans)

mbgrydeland:
Patch 1: "ban lurker: consider more race cases"

This seems like a lot of work for very seldom if not impossible cases.

It is true that the oc's ban pointer may have changed due to a hsh_lookup evaluation of that OC while the ban lurker was evaluating the same OC against the lurkable bans, and in that case and especially in the presence of a req-ban, the lurker may move the OC down again in the ban list undoing parts of the work done during hsh_lookup.

From the lurker's perspective, and true as long as we only have a single lurker thread, anyone else moving an OC in the ban list would have done so on a more informed basis than the lurker thread. So if we wanted to optimize against the hsh_lookup/ban_lurker same time race, we could have the ban lurker sample the oc->ban pointer before evaluating the lurkable ban list, and then look at it again before moving the OC's ban position. If the oc->ban pointer changed in the mean time, that result is more accurate than the lurkers idea of new ban list position and it should do nothing. This is a simple optimization and removes the need for extra list running loops in the patch to determine position.

nigoroll:
> This seems like a lot of work for very seldom if not impossible cases.
The scenario I was working on was 20k inefficient bans, where the ban lurker would only manage to work the tail of the ban list (because bans got added faster than the ban lurker could process them). 

> So if we wanted to optimize against the hsh_lookup/ban_lurker same time race, we could have the ban lurker sample the oc->ban pointer before evaluating the lurkable ban list, and then look at it again before moving the OC's ban position. If the oc->ban pointer changed in the mean time, that result is more accurate than the lurkers idea of new ban list position and it should do nothing.
Good idea, let's do that. I'll change the patch and update

nigoroll:
On second thought, I am no so sure any more: Later in `ban_lurker_work()`, we want to remove/complete all bans on the `olist`, so we need to make sure that we move any `oc` to `bt` or north of it. Can we really assert that this is always the case?

On performance, with the change I am suggesting
* `oc->ban == bd` (has moved to right place) is already a no-op
* `oc->ban == bt` (has not moved) is already handled without a list search
* so only for the remaining cases that the `oc` has moved "somewhere" is the list being searched.
As you said, a lookup ban check will always move the `oc` to or near the top of the ban list, so the list search should be very efficient.

nigoroll:
@mbgrydeland thank you for your review and the discussion on irc. Patch _Handle races..._ is updated. There is one further optimization commented out and marked with XXX which I am not sure about: I think a concurrent lookup ban check _could_ have started at an older `ban_start`, in which case it would move to a ban on the `olist`. While this optimization could safe some work in the lurker, I'd rather refrain from having it.

nigoroll:
After a detailed discussion with @bsdphk (thank you!), this PR is now down to 2 patches. Current state
* _When loosing the lock race..._: @bsdphk had a suggestion to avoid introducing the contended_oc, but I fail to see how that would help simplifying the code. Might need further discussion
* _avoid the de-tour..._: I'll check if this can be simplified

nigoroll:
btw, the latest push is further simplified and running fine in production

nigoroll:
The cleantail patch is in now, working on another iteration of the lookup race optimization

nigoroll:
last patch is in now. FTR, patches for this PR were
* 4d8aa0c306664ca93194d80de82ba108b712f36e
* 185978d79aed0cb51e3e3343f55d081434cd2e0c
* 1502d90e901b36b9e49a4d0b252c9b9c1ffff2d3
* 8ae34bc7371bcecdd75ee1d3ebf12c73eba788ef

hermunn:
Backport review: As far as I can see, these improvements are strictly performance optimizations. Since they are rather extensive, we will save them for the next release.

nigoroll:
@hermunn correct

Issue #1881, opened 2016-03-15
std.cache_req_body() w/ return(pipe) is broken


daghf:
## Expected Behavior
Varnish should pass the cached request body also in the event that the request was piped.

## Current Behavior
The request body is drained when we do std.cache_req_body(), and since the busyobj for a piped request does not have a bo->req pointer set, it will not be able to recover the cached body.

## Steps to Reproduce
The following test case fails after a timeout since the backend expects a request body that is never transmitted.

<pre>
varnishtest "cache_req_body + pipe does not play well"

server s1 {
	rxreq
	expect req.bodylen == 6
	expect req.http.foo == "true"
	txresp
} -start

varnish v1 -vcl+backend {
	import std;

	sub vcl_recv {
		set req.http.foo = std.cache_req_body(10KB);
		return (pipe);
	}
} -start

varnish v1 -cliok "param.set debug +syncvsl"

client c1 {
	txreq -body "foobar"
	rxresp
	expect resp.status == 200
} -run
</pre>

## Your Environment
* Version used: master

fgsch:
This fixes it. Haven't looked at the implications of this change though.
```
diff --git a/bin/varnishd/cache/cache_req_fsm.c b/bin/varnishd/cache/cache_req_fsm.c
index a2d830c..2caa021 100644
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -559,6 +559,11 @@ cnt_pipe(struct worker *wrk, struct req *req)
        VSLb(req->vsl, SLT_Link, "bereq %u pipe", VXID(bo->vsl->wid));
        THR_SetBusyobj(bo);
 
+       AZ(bo->req);
+       AZ(bo->wrk);
+       bo->req = req;
+       bo->wrk = wrk;
+
        HTTP_Setup(bo->bereq, bo->ws, bo->vsl, SLT_BereqMethod);
        http_FilterReq(bo->bereq, req->http, 0);        // XXX: 0 ?
        http_PrintfHeader(bo->bereq, "X-Varnish: %u", VXID(req->vsl->wid));
```

PR #1913, opened 2016-04-19
cache_req_body should return 413 when body is too large

ehocdet:
cache_req_body close the client connection if body is too big.
This is a server restriction and a 413 "Request Entity Too Large" should be sent.


fgsch:
Why not simply check the return from std.cache_req_body() and act upon? e.g.:

```
if (!std.cache_req_body(110B)) {
   return (synth(413));
}
```

nigoroll:
We should have the option to send the req.body to the backend once when it does not fit the cache size. This is not yet implemented, but returning a 413 in core code would block this route

ehocdet:
@fgsch : the TCP connection is close as a bad client request (expect_close in varnishtest),
and no bypass available when CL is not set.

@nigoroll :  I understand.
My primary use case is to protect web servers from POST flood.
When body is too big, use cases could be:
. bypass the cache
. return 413
. retry with a sfile storage and bigger maxsize (https://github.com/varnishcache/varnish-cache/pull/1914)

nigoroll:
@ehocdet you can't read the body a second time, so trouble is that for "body is too big", atm only "return 413" is possible. I had only seen the bypass option, but putting the rest of the body into a different storage would also be nice.

ehocdet:
@nigoroll yes. I use this patches like this:
```
                if (std.integer(req.http.Content-Length, 134217728) <= 16384) {
                        std.cache_req_body(16384B);
                } else {
                        std.cache_req_body(134217728B, "s0");
                }
```
but chunked requests always use "s0".
With "retry", i wanted to say  use "Transient" with 16k max, so move to "s0" and try with maxsize"  (in the core code)

The vcl syntax could be  cache_req_body(<maxsize>, bypass | error, <storage>)  ?

nigoroll:
This is also related to #1881

nigoroll:
irc:

> (13:39:54) slink: I see the following choices for vcl:
> (13:39:54) slink: - fits buffer: can restart, pipe or pass (core code will send buffer only for pass, and send it first for pipe)
> (13:39:54) slink: - doesnt fit: either error, pipe or pass. core code will send buffer than the client fd, but will refuse restart
> (13:40:15) phk: slink,  that's major work.  Send patch.
> ...
> (13:41:02) slink: The interesting question at this point is if you think that makes sense
> (13:41:09) phk: slink, not really
> (13:41:21) slink: then please explain to my why not
> (13:41:54) phk: slink, I think there is too little gain for that much code.
> (13:42:32) slink: phk, I think this is a major limitation for req body caching
> (13:42:53) slink: What if you _do_ want to support a 4TB upload? Atm all you can do is not cache
> (13:43:09) phk: slink, as I said:  Send patch.

nigoroll:
I'm taking this as a reminder that I got assigned the task of implementing the proper solution because I have suggested it ;)
@ehocdet just asking if you want to work on this?

ehocdet:
@nigoroll  the initial MR is for send a 413 error instead to close **silently** the tcp connection.
I think it's important to report an error in first time.

For the bypass case, i have currently no idea how to do that.

nigoroll:
bugwash discussion:
* unfortunately, this ticket has not received any attention from my side
** if anyone wants to take it, please do and drop a note

design:
* should support partial caching, iow even if `cache_req_body()` fails to consume all of the body, it should be possible to still continue processing by first sending to the backend the partially cached followed by the remainder which has not yet been read from the client
* `cache_req_body()` should at least return a boolean whether or not caching succeeded, maybe OK / PARTIAL / FAIL
* `bereq.can_retry` should indicate if a retry is possible: Either there is no body, or the body had been cached fully

@fgsch did i miss anything?

fgsch:
> @fgsch did i miss anything?

Nothing really except tht cache_req_body() already returns a BOOL. Not sure if we want to distinguish in VCL between partial and fully cached. 
I guess returning a 413 can be entirely decided based on whether it was cached (true).

For other purposes (e.g. a VMOD inspecting the body), I'd want to know between partial and full cache but this is up to VRT_CacheReqBody().

Whoever does this should have this distinction into consideration (std.cache_req_body() calls VRT_CacheReqBody()).

nigoroll:
my comments recapping todays bugwash:
* _partially caching req.body will confuse people_
I do not understand and do not agree. From a VCL standpoint, the body won't be need to appear cached (and as discussed with @fgsch previously, the cache status from VCL should be a boolean, while internally (for vmods) we probably want an interface returning a tri-state), the only point is that, IMHO, we should still be able to allow (from a VCL perspective) to _continue_ the request successfully after we've sucked in all we can.
Right not it's "either the buffer is large enough or we fail". And that's no good because there is chunked encoding. If it was only for the C-L case, admins could check is to be sufficiently small and otherwise not call cache_req_body
* _won't happen for 5.1_ is ok for me

fgsch:
@nigoroll if you don't mind I will take this as there is some other things I'll be working on that depend on the outcome of this ticket.

nigoroll:
@fgsch much appreciated, I'll be away 2 weeks starting next and this week getting current master in production is prio one.

bsdphk:
This issue seems to be going nowhere, will time out 2017-12-01

nigoroll:
@fgsch are you still at it?

fgsch:
@nigoroll Unfortunately I haven't made any progress with this. I just unassigned myself.

Dridi:
Timed out, and already doable in VCL using the return value, we should probably close this one.

nigoroll:
bugwash concensus: closed for now, but @fgsch and myself want optional partial request body caching.

opensubtitles:
I bumped into this "bug", it would be great, if either there is response 413, or to have it somewhere logged, it took me time to find it - what is causing no response, no http code...

nigoroll:
I got woken up by #3205 and have to admit that I did not continue with this ticket because of the lack of priority and funding.

PR #1914, opened 2016-04-19
set a custom storage_hint for cache_req_body

ehocdet:
Cache big req body should not eat the Transient storage.
The use case is to store the body on a dedicated sfile when CL is big.


bsdphk:
This looks like a good idea.

I'm getting to dislike the "hint" thing for storage more and more.  I think we should bite the bullet and teach VCC about storage spaces, and make a VCL_ type for them.  5.0 is a good time for that.

nigoroll:
@ehocdet we got the VCL_STEVEDORE type in the meantime, would you like to adjust your PR to that? I'll close this PR for now, please re-open or create a new one.

ehocdet:
@nigoroll  ok i will check that.

hermunn:
Backport review: This pull has been replaced by #2108 and #2111. Nothing to backport.

PR #1922, opened 2016-04-25
material for least connections director

ehocdet:
I implemented a least-connections director. It needs to access the n_conn value from backend.
The patch introduce vdi_conn_f do to that.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -139,6 +139,20 @@ vbe_dir_healthy(const struct director *d, const struct busyobj *bo,
 	return (VBE_Healthy(be, changed));
 }
 
+static unsigned __match_proto__(vdi_status_f)
+vbe_dir_status(const struct director *d, const struct busyobj *bo,
+    double *changed, double *load)
+{
+	struct backend *be;
+
+	CHECK_OBJ_NOTNULL(d, DIRECTOR_MAGIC);
+	CHECK_OBJ_ORNULL(bo, BUSYOBJ_MAGIC);
+	CAST_OBJ_NOTNULL(be, d->priv, BACKEND_MAGIC);
+	if (load != NULL)
+		*load = be->n_conn;
+	return (VBE_Healthy(be, changed));
+}
+
 static void __match_proto__(vdi_finish_f)
 vbe_dir_finish(const struct director *d, struct worker *wrk,
     struct busyobj *bo)
@@ -353,6 +367,7 @@ VBE_fill_director(struct backend *be)
 	d->vcl_name = be->vcl_name;
 	d->http1pipe = vbe_dir_http1pipe;
 	d->healthy = vbe_dir_healthy;
+	d->status = vbe_dir_status;
 	d->gethdrs = vbe_dir_gethdrs;
 	d->getbody = vbe_dir_getbody;
 	d->getip = vbe_dir_getip;
--- a/bin/varnishd/cache/cache_director.h
+++ b/bin/varnishd/cache/cache_director.h
@@ -45,6 +45,8 @@
 
 typedef unsigned vdi_healthy_f(const struct director *, const struct busyobj *,
     double *changed);
+typedef unsigned vdi_status_f(const struct director *, const struct busyobj *,
+    double *changed, double *load);
 
 typedef const struct director *vdi_resolve_f(const struct director *,
     struct worker *, struct busyobj *);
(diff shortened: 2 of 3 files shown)


Dridi:
This looks good, but I don like the name `nconn` for the methods. Custom backends may not rely on connections per-se, like for instance for *CGI or file backends. Maybe something more generic like `busy`.

The value may not have the same meaning for different kinds of backends (regardless of the name of the method) but I expect one wouldn't put backends of different natures behind a director. Except maybe the last entry for a fallback.

Thoughts?

ehocdet:
@Dridi It make sense.

The purpose of least-connections is to spread the load on the least loaded backend.
I like "busy", the number of connections in current backend, or any other information to report a busy state.
In my first implementation, i used the max_connections to normalize the information, but it's an optional parameter. In generic case this will be not really doable.

I agree, with this view, backends should be homogeneous in a director.
Fallback is a special case, only the first healthy backend report the busy state.

Dridi:
> The purpose of least-connections is to spread the load on the least loaded backend.

Then "load" might even be a better fit than "busy"!

ehocdet:
@Dridi: Why not.

I prefer "busy". It's not really the load of the backend, but an external view. This seems less confusing.
Spread  the load on the least busy backend.

ehocdet:
@Dridi 
Another possibility would be to add a new optional parameter to vdi_healthy_f, to report its level of use.

Thoughts?

Dridi:
Yes, but then I would opt for renaming the method too because load/busyness is not strictly related to health.

ehocdet:
New MR proposition.
Introduce vdi_busy_f, it extend vdi_healthy_f with load parameter.

in next step, vdi_busy_f should replace vdi_healthy_f.

ehocdet:
@Dridi
vdi_busy_f usage on least-connections director:
https://github.com/Gandi/libvmod-unidirectors

Dridi:
I find very interesting how you keep the backend cluster separate from the lookup mechanism, however after a quick glance it looks a bit racy or prone to contention. At the very least the cast of `vd->priv` should happen under the read lock.

For the change in Varnish, I'm not going to merge it myself. I can at least tell it should also be documented before it's merged.

ehocdet:
@dridi Thanks for feedback! indeed, race can arrive when LB method is changed dynamically, i will fix that.

> For the change in Varnish, I'm not going to merge it myself. I can at least tell it should also be documented before it's merged.
ok
If the feature is ok, vdi_busy_f could replace vdi_healthy_f.

Dridi:
Considering how close we are to the next major release, I'm not sure it's a good idea to introduce a new endpoint if we plan to collapse it later. Either we introduce a new one, and it becomes possible to back-port it to the 4.1 branch, or we change the existing healthy method and break forward compatibility.

I have no strong opinion on the busy method also calling the healthy one, but I just noticed that you are not checking `bo`.

FWIW I'd rather not break anything, so IMHO the main thing missing is an update of `directors.rst`.

ehocdet:
@Dridi, ok, thanks.
MR update.

Dridi:
LGTM although that's not what I'd call documentation. :grimacing:
I'll briefly mention the busy method in the docs if it's merged.

ping @bsdphk

ehocdet:
@Dridi
> I have no strong opinion on the busy method also calling the healthy one [...]

vdi_status cold be a better choice

Dridi:
@ehocdet this PR is going to get some attention, but before that you should acquire the backend's lock in order to read `n_conn` (spotted by @fgsch).

ehocdet:
@Dridi, @fgsch  i'm curious about why lock is required to read n_conn? VBE_Healthy read backend->changed and backend->healthy without lock.
I think the usage of this value should not need lock.

In another way:
When i track concurrency issues on my vmod_unidirectors, i noticed potential concurrency issues on original varnish libvmod_directors.
. many vd->n_backend read outside lock.
since https://github.com/varnishcache/varnish-cache/commit/be69a0329c49e46a378599afa48dbd8c10f7bcb4#diff-b4ebbf41e4844812c5500d0710873eef  :
. write after read lock on round-robin resolve
. write after read lock on vdir_pick_be

I miss something?

nigoroll:
@Dridi yours?

ehocdet:
MR update. function is now call vdi_status_f. Add entry in documentation.

bsdphk:
This is probably a good idea in principle, but I don't think this ticket and its patches are in a shape where retaining the ticket helps.

hermunn:
Backport review: Nothing to backport.

PR #1925, opened 2016-04-26
Bind before connect

ehocdet:
the MR is to propose to integrate the bind-before-connect logic in VTCP_connect.
This can be explain in https://idea.popcount.org/2014-04-03-bind-before-connect/

First patch, introduce retry mechanism to VTCP_connect.
It also fix an old misbehavior seen in production:
. VTCP_connect can fail with ETIMEDOUT errno.
. No retry is done and Varnish return 503 instantly

Second integrate the bind-before-connect logic.
VTCP_connect have now a  source suckaddr parameter.

### DIFF
--- a/lib/libvarnish/vsa.c
+++ b/lib/libvarnish/vsa.c
@@ -173,6 +173,20 @@ struct suckaddr {
 
 const int vsa_suckaddr_len = sizeof(struct suckaddr);
 
+static const struct suckaddr ipv4_any = { .magic = 0x4b1e9335,
+					 .sa4 = { .sin_family = AF_INET,
+						  .sin_addr.s_addr = INADDR_ANY,
+						  .sin_port = 0 }
+};
+static const struct suckaddr ipv6_any = { .magic = 0x4b1e9335,
+					 .sa6 = { .sin6_family = AF_INET6,
+						  .sin6_addr = IN6ADDR_ANY_INIT,
+						  .sin6_port = 0 }
+};
+
+const struct suckaddr *vsa_ipv4_any = &ipv4_any;
+const struct suckaddr *vsa_ipv6_any = &ipv6_any;
+
 /*
  * This VRT interface is for the VCC generated ACL code, which needs
  * to know the address family and a pointer to the actual address.
--- a/bin/varnishd/cache/cache_backend_tcp.c
+++ b/bin/varnishd/cache/cache_backend_tcp.c
@@ -220,21 +220,28 @@ VBT_Open(const struct tcp_pool *tp, double tmo, const struct suckaddr **sa)
 {
 	int s;
 	int msec;
+	const struct suckaddr *soa4 = NULL;
+	const struct suckaddr *soa6 = NULL;
+
+	if (FEATURE(FEATURE_BIND_BEFORE_CONNECT)) {
+		soa4 = vsa_ipv4_any;
+		soa6 = vsa_ipv6_any;
+	}
 
 	CHECK_OBJ_NOTNULL(tp, TCP_POOL_MAGIC);
 
 	msec = (int)floor(tmo * 1000.0);
 	if (cache_param->prefer_ipv6) {
 		*sa = tp->ip6;
-		s = VTCP_connect(tp->ip6, msec);
+		s = VTCP_connect(tp->ip6, soa6, msec, cache_param->connect_retry);
 		if (s >= 0)
 			return (s);
 	}
 	*sa = tp->ip4;
-	s = VTCP_connect(tp->ip4, msec);
+	s = VTCP_connect(tp->ip4, soa4, msec, cache_param->connect_retry);
 	if (s < 0 && !cache_param->prefer_ipv6) {
 		*sa = tp->ip6;
-		s = VTCP_connect(tp->ip6, msec);
+		s = VTCP_connect(tp->ip6, soa6, msec, cache_param->connect_retry);
 	}
 	return (s);
 }
--- a/bin/varnishd/cache/cache_backend_tcp.c
+++ b/bin/varnishd/cache/cache_backend_tcp.c
@@ -37,6 +37,7 @@
 #include <stdlib.h>
 
 #include "cache.h"
+#include "common/heritage.h"
 
 #include "vrt.h"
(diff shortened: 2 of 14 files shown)


nigoroll:
shouldn't we have an idea how to configure the source address and VTCs for this feature?

ehocdet:
Last part: bind_before_connect feature bit
Switch to bind-before-connect logic. It permit to have 64k connections per destination (ip:port).

Dridi:
I'm having trouble understanding why one would want to open 64k concurrent connections to a single backend. For the timeout fix, I definitely see the value, but this kind of access pattern to a backend doesn't seem useful.

What an I missing?

Also the number of connect tries should be a `varnish-cli` parameter.

ehocdet:
@Dridi:
With a lot of backends and one source IP address, the limitation is the number of local port.
With 64k local port available and 1000 backends (we have 10k), we can have 64 connections per backend in avg.
Add TIME_WAIT/CLOSE_WAIT states, flapping backends... (not to mention the time spent in the kernel to find a free port)... 
A workaround is to have more sources ip and static routes, but it's very ugly.
The uniqueness of a tcp socket is set with (src_ip, src_port, dst_ip, dst_port)  and not only (src_ip, src_port).

For the retry in varnish-cli parameter, i do that on varnish3.0. But i can't  in VTCP_connect with varnish4.1.  (or y miss something)

Dridi:
10k backends? That's impressive :D

I'm sold on the idea.

Dridi:
> For the retry in varnish-cli parameter, i do that on varnish3.0. But i can't in VTCP_connect with varnish4.1. (or y miss something)

We'd probably need to change the function signature and add a retry argument so that calling code could pass the parameter's value. Or this could be done in the backend subsystem instead.

bsdphk:
We've had this one up once earlier, and being able to specify source address is a requirement.

Ideally we should make it possible to specify that for all outgoing connections, (ie: also -M), but that more or less requires us to come up with a argument representation for specifying it (something IETF should have done but...) and a place to store it, which either makes struct suckaddr twice as big or variable sized.

So I guess at least for now, we should add a "source address" specification to the VCC's backend specs.

Code details:

Please don't mix very different stuff in one pull:  Move the retry stuff to a separate pull request.

vsa_ipv[46]_any should be function calls returning the pointer, so that we are not locked in at compile time (don't ask...).

ehocdet:
patch reworked.
. connect_retry only affect bind-before-connect logic
. introduce vtcp_bind_socket to share code

The first goal is to let the system finding the source ip according to route table, as legacy connect logic do. For that, the source address is set with ANY.
The default source address could be a parameter. (it's not as easy as I thought, with -M ...)
Implementation should be compatible with custom source address on VCC's backend.

lib/libvarnish/vsa.c
```diff
@@ -173,6 +173,24 @@ struct suckaddr {
...
+						  .sin6_addr = IN6ADDR_ANY_INIT,
+						  .sin6_port = 0 }
+};
+
+const struct suckaddr *vsa_ipv4_any() {
+	return (& ipv4_any);
+}
+const struct suckaddr *vsa_ipv6_any() {
+	return (& ipv6_any);
+}
```
Dridi: Instead of functions, `ipv?_any` symbols could be exposed similarly to how `vsa_suckaddr_len` is (and prefixed with `vsa_`).

nigoroll:
@bsdphk I guess this one is waiting for you. In #1935 you mentioned that you're not convinced we need the retry, so I suspect the patch should be changed in this respect. But maybe you'd want to comment on the rest of it?

ehocdet:
PR update.
Add -B option to set a source IP address. VTC added.

@nigoroll @bsdphk
The retry is part of bind-before-connect logic:
With legacy connect, the tcp stack iterate on local ports to find one usable.
With bind-before-connect, connect can fail with EADDRNOTAVAIL. The application should retry on connect to find a usable (src_ip, src_port, dst_ip, dst_port).

bind-before-connect and source ip:
. by default feature work as legacy connect: TCP stack find the source ip according to route table.
. -B option can be use to specify the source address
. code can support custom source address on VCC's backend.

fgsch:
From IRC:

```
12:12 < gofer> #1925: Bind before connect [open] https://github.com/varnishcache/varnish-cache/issues/1925
12:14 < fgs> I think it makes sense to have something like that, not necessarily that implementation though
12:14 < phk> So this is a general sort of issue, where you want to be able to force a src-ip
12:14 < phk> my thinking is that if we want to do it, we want to do it general, so that it works for -a, -T, -M, backend::host and anything else
12:14 < phk> That essentially means coming up with an ASCII syntax for it.
12:15 < phk> (actually: not -a - Duh!)
12:15 < fgs> The reasons will be different though
12:15 < fgs> so while it might be generic, -T and -M are not that important
12:15 < phk> yes, but the desire is the same:  Open a connection to this address [from this address]
12:15 < phk> -T is also not applicable.
12:15 < phk> it's backends and -M
12:16  * phk wonders if he should wake up and start think clearly, or if his alarm tomorrow morning at 07:30 was the better idea.
12:16 < martin> dynamic backends come into play as well i guess
12:16 < phk> yes.
12:17 < phk> but actually if it is only -M and backend::host, then generalizing is not that important.
12:18 < phk> Anybody wants to look at this as a pure backend property ?
12:18 < fgs> I believe they opened a PR before with something around those lines.
12:19 < phk> yeah, but I was looking to see if I could get somebody to own the ticket :-)
12:20 < phk> no takers.
12:20 < martin> guess we'll leave it open as a reminder that we'd like this functionality
```

bsdphk:
I have looked at this again.

My best suggestion is to allow specifications somewhat like:

    .host = "backend_foo/10.0.1.8"

However no matter how we do something like this, it seems to require a rather significant change to backend peddling VMODS, so this is on hold until we know when 6.0 happens.

bsdphk:
Moved to VIP15:

https://github.com/varnishcache/varnish-cache/wiki/VIP15:-Specifying-source-address-for-outgoing-sockets

hermunn:
Backport review: Moved to VIP15, nothing to backport.

PR #1931, opened 2016-04-28
Support reading Last-Modified from response header for IMS requests

rezan:
This came up because a client is modifying Last-Modified in vcl_deliver to compensate for ESI handling. In particular, they are purging an ESI fragment and would like all pages to now reference this time in their Last-Modified.

The PR adds a block checking the resp header before falling back to the object timestamp. Several VTCs fail if we do not fallback to the object timestamp, so the fallback is needed.

### DIFF
--- a/bin/varnishd/cache/cache_rfc2616.c
+++ b/bin/varnishd/cache/cache_rfc2616.c
@@ -256,7 +256,7 @@ rfc2616_weak_compare(const char *p, const char *e)
 int
 RFC2616_Do_Cond(const struct req *req)
 {
-	const char *p, *e;
+	const char *p, *e, *l;
 	double ims, lm;
 
 	/*
@@ -276,6 +276,12 @@ RFC2616_Do_Cond(const struct req *req)
 		ims = VTIM_parse(p);
 		if (!ims || ims > req->t_req)	/* [RFC7232 3.3 p16] */
 			return (0);
+		if (http_GetHdr(req->resp, H_Last_Modified, &l)) {
+			lm = VTIM_parse(l);
+			if (!lm || lm > ims)
+				return (0);
+			return (1);
+		}
 		AZ(ObjGetDouble(req->wrk, req->objcore, OA_LASTMODIFIED, &lm));
 		if (lm > ims)
 			return (0);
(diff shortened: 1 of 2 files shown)


bsdphk:
I don't have any opinion on this being a good/bad idea.

Anybody else ?

rezan:
So here is a good way to think about this: Varnish controls all outbound headers, this is a core Varnish usecase. If we set a standard HTTP header, like Last-Modified, and a browser or CDN doesnt respect it, well, thats a bug, we would expect that to be fixed. So the same applies to Varnish in this scenario. We are setting an important content control header, everything downstream will behave as we expect, minus Varnish, so this header needs to be evaluated to match what everyone else is doing.

nigoroll:
This does make sense to me.

fgsch:
As I mentioned on IRC, wouldn't unsetting the I-M-S header be more correct and/or elegant?
Isn't there a case where you want to change the L-M header without turning the response into a 304? 
I'm not saying this is incorrect but for some reason it feels hacky to me.
From my POV, we have input and output headers. Varnish takes the former and evaluates them in the context of the latter. With this in mind the ones controlling the output are the input ones.

nigoroll:
The main use-case, as I understand it, is to modify LM for esi-constructed objects to a timestamp which is not only determined by the top (outer) object, but also by the includes.
Removing IMS would work to trigger the refresh, but making sure that following requests get a 304 would get hacky in VCL (basically, one would need to implement the LM-time comparison in VCL/vmod to achieve the same).
Respecting a VCL-controlled LM in deliver appears to very much in line with the varnish philosophy of providing _mechanism, rather than policy_.
My understanding is that the current behavior is not following the principle of least astonishment: A VCL author sets LM, but varnish disrespects it.

rezan:
Unsetting IMS will give us a 200. Think about the logic required on the next request to get a 304 response. The user would have to implement IMS logic in VCL via IMS. What is being proposed only requires keeping track of LM. This is a lot more straight forward and simple

nigoroll:
fully agree with @rezan 's last comment

hermunn:
This seems to be cleanly backportable to 4.1, but I am uncertain if this is the type of patch that we port to 4.1. Maybe @lkarsten should have the final word?

hermunn:
The patch from master, 7996af0, has been back ported to 4.1 as 194acd4f0a9e0483b

PR #1934, opened 2016-05-02
Postpone calling VBT_wait until after V1P_Process

hermunn:
Fixes a bug where we wait for an event to fire on the backend socket
prior to passing the request body.

The VBT_Wait call is now handled in vbe_dir_finish.

Test case by thomaslc

Fixes: #1806


hermunn:
The test case is ok, but I will let @daghf have a quick look at it.

daghf:
This should be OK for 4.1 backport

hermunn:
Applied and pushed.

PR #1935, opened 2016-05-02
VTCP_connect retry

ehocdet:
introduce retry mechanism to VTCP_connect.

It fix an old misbehavior seen in production:
. VTCP_connect can fail with ETIMEDOUT errno.
. No retry is done and Varnish return 503 instantly

Retry 3 time is a robust setting.
Set this in varnish-cli parameter, i think it should be necessary to extend the VTCP_connect with a retry parameter. VTCP_connect errno like ETIMEDOUT should not be out of VTCP_connect.

(split from https://github.com/varnishcache/varnish-cache/pull/1925)

### DIFF
--- a/lib/libvarnish/vtcp.c
+++ b/lib/libvarnish/vtcp.c
@@ -264,45 +264,54 @@ VTCP_connect(const struct suckaddr *name, int msec)
 	AN(sa);
 	AN(sl);
 
-	s = socket(sa->sa_family, SOCK_STREAM, 0);
-	if (s < 0)
-		return (s);
+	for (int retry = 0; retry < 3; retry++) {
+		s = socket(sa->sa_family, SOCK_STREAM, 0);
+		if (s < 0)
+			return (s);
 
-	/* Set the socket non-blocking */
-	if (msec != 0)
-		(void)VTCP_nonblocking(s);
+		/* Set the socket non-blocking */
+		if (msec != 0)
+			(void)VTCP_nonblocking(s);
 
-	i = connect(s, sa, sl);
-	if (i == 0)
-		return (s);
-	if (errno != EINPROGRESS) {
-		AZ(close(s));
-		return (-1);
+		i = connect(s, sa, sl);
+		if (msec == 0) {
+			if (i == 0)
+				return (s);
+		}
+		if (errno != EINPROGRESS) {
+			AZ(close(s));
+			return (-1);
+		}
+
+		if (msec < 0) {
+			/*
+			 * Caller is responsible for waiting and
+			 * calling VTCP_connected
+			 */
+			return (s);
+		}
+
+		assert(msec > 0);
+		/* Exercise our patience, polling for write */
+		fds[0].fd = s;
+		fds[0].events = POLLWRNORM;
+		fds[0].revents = 0;
+		i = poll(fds, 1, msec);
+
+		if (i == 0) {
+			/* Timeout */
+			i = -1;
+			errno = ETIMEDOUT;
+			AZ(close(s));
+		} else {
+			i = VTCP_connected(s);
+		}
+		if (i >= 0)
+			return (s);
+		if (errno != ETIMEDOUT)
+			return (-1);
 	}
-
-	if (msec < 0) {
-		/*
-		 * Caller is responsible for waiting and
-		 * calling VTCP_connected
-		 */
-		return (s);
-	}
-
-	assert(msec > 0);
-	/* Exercise our patience, polling for write */
-	fds[0].fd = s;
-	fds[0].events = POLLWRNORM;
-	fds[0].revents = 0;
-	i = poll(fds, 1, msec);
-
-	if (i == 0) {
-		/* Timeout, close and give up */
-		AZ(close(s));
-		errno = ETIMEDOUT;
-		return (-1);
-	}
-
-	return (VTCP_connected(s));
+	return (-1);
 }
 
 /*--------------------------------------------------------------------


fgsch:
Isn't this what bereq.retries / return (retry) is for?

Dridi:
@fgsch you may end up in `vcl_backend_error` for many reasons. This has a tiny scope in comparison.

fgsch:
@Dridi True, though adding a hidden option doesn't make things better and/or more manageable.
What if I want to fail fast?

I don't really like the idea of automagically retrying when there is an error, specially for timeouts.

If anything I'd think something like a .retries attribute in the host definition that is both visible and user configurable makes more sense. Another option would be to expose the error to VCL /  vcl_backend_error{}.

That said, as per the problem description (retry is not done, we return a 503) I still don't understand what this is trying to fix when you can retry.

ehocdet:
@fgsch It's not about a timeout, it's about a failure in poll request to the kernel.

Workaround this problem  with return(retry) implies to do this for all requests.
. It doesn't work for request with body
. it count as a backend error
. retry on a backend is not always a good thing (amplification errors)

For user experience, this problem can generate instant random error...
It's very difficult to understand what's going on.

Dridi:
> If anything I'd think something like a .retries attribute

I have already said that I don't like the hardcoded attribute in #1925. We're on the same page ;)

> Another option would be to expose the error to VCL / vcl_backend_error{}.

It has been discussed, and the idea would be to have an `error` field available somehow in `vcl_backend_response` that should have the same value as the `Error` or `FetchError` VSL record.

ehocdet:
I tried to find the origin of the problem. It not concern a timeout. Failure is always instant.

extract of "man poll":
```
 The timeout argument specifies the number of milliseconds that poll() should block waiting for a file descriptor to become ready.  The call will block until
       either:

       *  a file descriptor becomes ready;

       *  the call is interrupted by a signal handler; or

       *  the timeout expires.
```

I suspected that is related to "the call is interrupted by a signal handler". but i'm not an expert.

fgsch:
> @fgsch it's about a failure in poll request to the kernel.

The poll will return 0 on timeout. This corresponds to the connect_timeout timeout.
Isn't this what you want if you have a low timeout?

> Workaround this problem with return(retry) implies to do this for all requests.
> . It doesn't work for request with body

It will work if you cache the body AFAIK.

> . it count as a backend error

Why is this a problem?

> . retry on a backend is not always a good thing (amplification errors)

Aren't we making this worse by automatically retrying?

I'm not trying to be dense, but I really don't see what you are trying to solve.

fgsch:
If the problem here is really EINTR you might want:

```
diff --git a/lib/libvarnish/vtcp.c b/lib/libvarnish/vtcp.c
index 09bf909..f254be3 100644
--- a/lib/libvarnish/vtcp.c
+++ b/lib/libvarnish/vtcp.c
@@ -293,7 +293,9 @@ VTCP_connect(const struct suckaddr *name, int msec)
        fds[0].fd = s;
        fds[0].events = POLLWRNORM;
        fds[0].revents = 0;
-       i = poll(fds, 1, msec);
+       do {
+               i = poll(fds, 1, msec);
+       } while (i == -1 && errno == EINTR);
 
        if (i == 0) {
                /* Timeout, close and give up */
```

ehocdet:
@fgsch  see my previous post for timeout.
I understand your point about the impact of connect_timeout, it's not this i try to fix in the first place.

http retry and tcp retry is not equivalent, it's complementary. 
tcp retry made a very good job on haproxy, why not in varnish?
since my tcp-retry patch, our production is more stable.

. retry a tcp connection has not the same impact as resend a http request to a backend.
(I see the impact in production before the patch with http retry workaround)
. cache body doesn't work in all case (chunked body and body > maxsize)

ehocdet:
> If the problem here is really EINTR you might want:
the case come with poll return 0, not -1 and errno.

fgsch:
> http retry and tcp retry is not equivalent, it's complementary. 
>[..]

I understand and I'm talking about TCP here.  The problem is that currently there is no way to know if you ended up in `vcl_backend_error{}` due to a connection timeout or something else (e.g. failed fetching the body). Maybe we should expose that. See Driri's comment earlier.

> . cache body doesn't work in all case (chunked body and body > maxsize)

True for body > maxsize. It should work for chunked bodies IIRC.

> If the problem here is really EINTR you might want:
> the case come with poll return 0, not -1 and errno.

EINTR should only be seen for -1. 0 means timeout, always.

ehocdet:
> I understand and I'm talking about TCP here. The problem is that currently there is no way to know if you ended up in vcl_backend_error{} due to a connection timeout or something else (e.g. failed fetching the body). Maybe we should expose that. See Driri's comment earlier.

yes, it would be great!  It would be possible to retry on vcl_backend_fetch in case of connect failure?

> True for body > maxsize. It should work for chunked bodies IIRC.

not yep because  chunked bodies have no CL (related to https://github.com/varnishcache/varnish-cache/pull/1913)

> EINTR should only be seen for -1. 0 means timeout, always.

yes... I suspect bad behavior of the kernel.

fgsch:
> yes, it would be great! It would be possible to retry on vcl_backend_fetch in case of connect failure?

If implemented that would be the idea. Something along the lines:

```
if (error.connect_timeout) {
   return (retry);
}
```

> not yep because chunked bodies have no CL (related to #1913)

That's a different issue. You can still cache chunked bodies as long as they fit.

> yes... I suspect bad behavior of the kernel.

I'm more inclined to believe some network device between your Varnish instance and the backend is at play here. Do you have any load balancer, firewall, or other appliance or server in between?
A tcpdump capture might be useful.

ehocdet:
> I'm more inclined to believe some network device between your Varnish instance and the backend is at play here. Do you have any load balancer, firewall, or other appliance or server in between?
> A tcpdump capture might be useful.

Nothing of the sort. It's on local network, and nothing seen on tcpdump.
It's a very old issue. We will change the platform soon, i will try to reproduce this.

fgsch:
No gateway in between either? 
Anyway, it'd be interesting to check the tcpdump output and/or rule out that is not a matter of increasing the `connect_timeout` timeout.

bsdphk:
1.  This is in waiting position until we know what the specific error situation is.  I'm not aware of TCP connections failing in the way described on any platform and would consider any such platform buggy in any case.

2.  If (for some reason) we implement retries in VTCP_Connect(), it is as a paramter to be passed in, not as a hardcoded '3'.

nigoroll:
@ehocdet i'm closing this for timeout because the last comment had been added May 4th. Please open a new PR when you got more info or re-open this one if you prefer.

hermunn:
Backport review: Timed out

PR #1938, opened 2016-05-05
Add support for TCP Fast Open extension

fgsch:
Disabled by default.

AFAIK this is only available in Linux and on a limited number of browsers (e.g. Chrome for Android).
FreeBSD has the code but it's not compiled in.

Open it in case someone is interested.


fgsch:
Lightly tested locally.

bsdphk:
OK for commit

fgsch:
Applied.

hermunn:
Original patch is 1c662d992d6cc4ed, backported to 4.1 as 65d90e3dbed0a15e533b.

PR #1939, opened 2016-05-05
Correctly handle EOF responses with HTTP/1.1

### LINKED ISSUE #1918: 4.1.x fails to forward HTTP/1.1 response without content-length + missing Transfer-Encoding
Varnish 4.1.2 fails to forward HTTP/1.**1** responses where 'Connection' is set to 'close' and the Content-Length & Transfer-Encoding headers are missing.

Example of such a backend reply:
```
HTTP/1.1 200 OK
Content-Type: text/plain
Connection: close

The quick brown fox
```

## Expected Behavior
AFAIK, such a reply does not violate RFC 7230 and should therefore be handled by Varnish
(But i fully agree that a Backend which sends such a reply isn't that smart.)

## Current Behavior
Varnish currently delivers a zero-sized body to the client:
```
[adrian@ebichu ~]$ curl -v 'http://127.0.0.1:80/'
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.48.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain
< Date: Fri, 22 Apr 2016 06:52:48 GMT
< X-Varnish: 2
< Age: 0
< Via: 1.1 varnish-v4
< Accept-Ranges: bytes
< Content-Length: 0
< Connection: keep-alive
< 
```

This does not happen if a Content-Length is provided or if the HTTP Version in the backend-reply is changed to 1.0

## Possible Solution
Commit a6a6e97a4db6e94b14f81af55fea4d07aa06d0af seems to have introduced this issue.

This patch seems to resolve the problem while (hopefully?) not breaking anything else ;-)
```
--- a/bin/varnishd/http1/cache_http1_proto.c
+++ b/bin/varnishd/http1/cache_http1_proto.c
@@ -307,7 +307,7 @@ http1_body_status(const struct http *hp, struct http_conn *htc)
 		return (cl == 0 ? BS_NONE : BS_LENGTH);
 	}
 
-	if (hp->protover == 11)
+	if (hp->protover == 11 && !http_HdrIs(hp, H_Connection, "close"))
 		return (BS_NONE);
 
 	if (http_HdrIs(hp, H_Connection, "keep-alive")) {
```

1. Build Varnish 4.1.2
2. nc -l -p 8080 -c < http_reply_posted_above.txt
3. configure varnish to forward 0:80 -> 127.0.0.1:8080
4. curl -v 'http://127.0.0.1:80/'

## Your Environment
* Version used: 4.1.2
* Operating System and version: CentOS 7

fgsch:
Fixes #1918.

### DIFF
--- a/bin/varnishd/http1/cache_http1_proto.c
+++ b/bin/varnishd/http1/cache_http1_proto.c
@@ -271,7 +271,7 @@ http1_splitline(struct http *hp, struct http_conn *htc, const int *hf,
 /*--------------------------------------------------------------------*/
 
 static enum body_status
-http1_body_status(const struct http *hp, struct http_conn *htc)
+http1_body_status(const struct http *hp, struct http_conn *htc, int request)
 {
 	ssize_t cl;
 	const char *b;
@@ -301,7 +301,7 @@ http1_body_status(const struct http *hp, struct http_conn *htc)
 		return (cl == 0 ? BS_NONE : BS_LENGTH);
 	}
 
-	if (hp->protover == 11)
+	if (hp->protover == 11 && (request || !http_HdrIs(hp, H_Connection, "close")))
 		return (BS_NONE);
 
 	if (http_HdrIs(hp, H_Connection, "keep-alive")) {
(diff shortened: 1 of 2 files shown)


bsdphk:
OK for commit.

fgsch:
Applied.

hermunn:
The back port to 4.1 is 61a66b7b

PR #1945, opened 2016-05-11
Rush limit

nigoroll:
resurrect a patch from #1902 - see commit msg for detailed description

### DIFF
--- a/bin/varnishd/cache/cache_ban_lurker.c
+++ b/bin/varnishd/cache/cache_ban_lurker.c
@@ -265,7 +265,7 @@ ban_lurker_test_ban(struct worker *wrk, struct vsl_log *vsl, struct ban *bt,
 			if (i)
 				ObjSendEvent(wrk, oc, OEV_BANCHG);
 		}
-		(void)HSH_DerefObjCore(wrk, &oc);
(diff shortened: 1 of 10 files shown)


bsdphk:
I'm not thrilled at spreading this decision all over the code, isn't there any way we make the decision centrally based on the oc's state ?

nigoroll:
@bsdphk I see your point and I had another look, but I am not sure we can make the decision without the context (I am not considering here the idea to add another flag to the oc before calling deref, which would not be much different from adding a deref parameter, IMHO):
Some cases appear simple like `OC_F_PRIVATE` or `OC_F_FAILED`, but for instance when deref'ing an unflagged `stale_oc`, I don't see how `HSH_DerefObjCore` can determine that it should not rush.
One option would be to always explicitly call rush before the deref where needed, but that involves taking `oh->mtx` another time.
your advise?

nigoroll:
@bsdphk I'd very much like to tick this off my list, could you help me with my questions? Or could we see this is an intermediate solution and work towards a different solution as we go forward?

hermunn:
Backport review: As with #1902, we will not backport this unless it fixes a crash in Varnish Cache 4.1.

Issue #1941, opened 2016-05-06
Crash on wrong ESI code when cache is full  ( Wrong turn at cache/cache_esi_deliver.c:344 )


XANi:
Varnish crashes a minute or two after getting to "full" cache, as in "the more memory we give to it the longer it works before the crash. It is hard to guess what exactly caused it as error message does not display ESI part in question (we're heavy user of ESI and trying to migrate away from v3)

Even with wrong ESI part it shouldn't crash the whole server, just 503 the request

Version 4.1.2 (from current debian testing) + softpurge vmod

     12:39:34 blade710 varnishd[17866]: Debug: Child (19345) Started
     12:39:34 blade710 varnishd[17866]: Info: Child (19345) said Child starts
     12:42:12 blade710 varnishd[17866]: Error: Child (19345) died signal=6
     12:42:12 blade710 varnishd[17866]: Error: Child (19345) Last panic at: Fri, 06 May 2016 10:42:12 GMT
     12:42:12 blade710 varnishd[17866]: "Wrong turn at cache/cache_esi_deliver.c:344:
     12:42:12 blade710 varnishd[17866]: ESI-codes: Illegal code
     12:42:12 blade710 varnishd[17866]: thread = (cache-worker)
     12:42:12 blade710 varnishd[17866]: version = varnish-4.1.2 revision 0d7404e
     12:42:12 blade710 varnishd[17866]: ident = Linux,3.16.0-4-amd64,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
     12:42:12 blade710 varnishd[17866]: Backtrace:
     12:42:12 blade710 varnishd[17866]: 0x433815: varnishd() [0x433815]
     12:42:12 blade710 varnishd[17866]: 0x41d3d1: varnishd(VDP_ESI+0xaf1) [0x41d3d1]
     12:42:12 blade710 varnishd[17866]: 0x41aab9: varnishd(VDP_bytes+0x79) [0x41aab9]
     12:42:12 blade710 varnishd[17866]: 0x41afad: varnishd(VDP_DeliverObj+0x11d) [0x41afad]
     12:42:12 blade710 varnishd[17866]: 0x41bf03: varnishd() [0x41bf03]
     12:42:12 blade710 varnishd[17866]: 0x43600f: varnishd() [0x43600f]
     12:42:12 blade710 varnishd[17866]: 0x437596: varnishd(CNT_Request+0x1326) [0x437596]
[...]

PR #1946, opened 2016-05-11
Don't allow partial allocation of ESI memory

hermunn:
This pull request is a hack that fixes a problem (#1941) with allocation of ESI memory.

It is probably not going to be accepted as is, but I hope we can have a discussion on how the real fix should be.


hermunn:
After working with @mbgrydeland a bit, the pull request has been updated to something less hackish. I hope this pull request can be reviewed in the normal way.

The old version is 3fbc544237ee847e59f15411a0ae61ec8aab0fdd, if anyone is interested.

Dridi:
@hermunn I tried to simplify the test case but it doesn't seem possible.

hermunn:
It has been applied to master and back ported to 4.1

hermunn:
The competing pull request has been back ported, removing label.

PR #1959, opened 2016-05-20
Make sure that VFP_Close is always called if and only if VFP_Open was

mbgrydeland:
called

For some (out of workspace or stevedore allocationi failure) errors
VFP_Close would fail to be called when ending up not retrying the
fetch. This would then potentially leak VFP entry context data like
gzip buffers. Fix this by always calling VFP_Close() when changing
fetch state when VFP_Open() has been called.

Also VFP_Close() would for some retries be called twice, causing
double accounting log records print out. This happens when a late
error is caught and streaming is disabled, allowing a retry to be
attempted. Fix this by not unconditionally calling VFP_Close() in the
error state.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -235,8 +235,6 @@ vbf_stp_retry(struct worker *wrk, struct busyobj *bo)
 	bo->do_stream = 1;
 
 	/* reset fetch processors */
-	vfc->failed = 0;
-	VFP_Close(vfc);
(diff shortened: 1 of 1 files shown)


bsdphk:
OK

We should probably enforce this with a panic

mbgrydeland:
This has been pushed to master. Needs 4.1 backport.

hermunn:
This was commited as in master as 2e3149b24b21890.

hermunn:
Backport review: Backported to 4.1 as 0b0bf921f065ff746f38f.

Issue #1955, opened 2016-05-17
4.1.x sometimes duplicates Age and Accept-Ranges headers


blblack:
With Varnish 4.1.x in a production configuration that layers multiple varnish caches requesting from other varnish caches, in the final output we often see duplicate (or even triplicate) `Age:` and `Accept-Ranges:` headers.

Relatedly, we backported d828a042 into our 4.1.x package, and this did seem to resolve the issue for 304 refreshes of stale objects, but not for the general case.

## Expected Behavior
Expect only 1 copy each of `Age:` and `Accept-Ranges:` headers to be emitted by Varnish to the client.

## Current Behavior
Seeing 2-3x of these headers headers sent to client, one for each layer of Varnish cache the request passes through.  They seem to always have identical values, at least in short-TTL cases, but this may be mostly a coincidence.

## Steps to Reproduce (for bugs)
Live example: https://maps.wikimedia.org/ (2-4 layers of varnish caching, depending where you are geographically).
Manual repro: Start 2x varnishd with default VCL on the CLI with the first backending to some standard webserver, and the second backending to the first.  Make a request through the stack and observe output headers:
```
$ /usr/sbin/varnishd -a :3133 -b <IP of some apache server>:80 -p default_ttl=5 -p default_grace=10 -F -T :8188 -n backvarnish &
$ /usr/sbin/varnishd -a :3134 -b localhost:3133 -p default_ttl=5 -p default_grace=10 -F -n frontvarnish &
$ curl -v http://localhost:3134/ >/dev/null
[....]
< Content-Type: text/html; charset=UTF-8
< X-Varnish: 2
< Age: 0
< Via: 1.1 varnish-v4
< Accept-Ranges: bytes
< X-Varnish: 2
< Age: 0
< Via: 1.1 varnish-v4
< Accept-Ranges: bytes
< Transfer-Encoding: chunked
< Connection: keep-alive
<
{ [data not shown]
```

hermunn:
I am working on this now, but I can't reproduce the problem. I have tried the realeases 4.1.1 and 4.1.2, as well as the 4.1 and master branches.

I can see you problem in your live example, but it would be very helpful to have a minimal example (preferably varnishtest) on my own computer where multiple Age headers appear.

I will try again tomorrow.

fgsch:
Here.

```
varnishtest ""

server s1 {
	rxreq
	txresp -hdr "Cache-Control: max-age=1"
} -start

[...]

PR #1962, opened 2016-05-25
Eliminate duplicate Age and Accept-Ranges headers.

hermunn:
When we have several layers of Varnish, in some cases we would send
multiple Age and Accept-Ranges headers to the client. This is fixed in
this commit. The test case demonstrates the problem for the Age
header.

Fixes: #1955

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -91,6 +91,7 @@ cnt_deliver(struct worker *wrk, struct req *req)
 	 * taken before the object entered into cache leading to negative
 	 * age. Truncate to zero in that case).
 	 */
+	http_Unset(req->resp, H_Age);
 	http_PrintfHeader(req->resp, "Age: %.0f",
 	    fmax(0., req->t_prev - req->objcore->t_origin));
 
@@ -278,6 +279,7 @@ cnt_transmit(struct worker *wrk, struct req *req)
 
 		if (cache_param->http_range_support &&
 		    http_IsStatus(req->resp, 200)) {
+			http_Unset(req->resp, H_Accept_Ranges);
 			http_SetHeader(req->resp, "Accept-Ranges: bytes");
 			if (sendbody && http_GetHdr(req->http, H_Range, &r))
 				VRG_dorange(req, r);
--- a/bin/varnishtest/tests/r01955.vtc
+++ b/bin/varnishtest/tests/r01955.vtc
@@ -1,0 +1,36 @@
+varnishtest "Correct handling of Age and headers for chained varnish servers"
+
+# The issue https://github.com/varnishcache/varnish-cache/issues/1955
+# concerns multiple Accept-Ranges headers as well as multiple Age
+# headers, but the problem with multiple Accept-Ranges cannot be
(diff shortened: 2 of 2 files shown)


fgsch:
I have some reservations about this patch.

If you are going to unset Age unconditionally why not just filter it at HTTP_Encode() time?

Also, is this really limited to Age and Accept-Ranges?

For Accept-Ranges, bug aside, I think it does not make sense for us to advertise it for uncacheable objects so this might simplify the patch.

Also I don't really understand the comment on the test. Saying that we cannot test Accept-Ranges should be enough.

hermunn:
> If you are going to unset Age unconditionally why not just filter it at HTTP_Encode() time?

To me that seems very unnatural. First, this would break the name `HTTP_Encode` for a very small optimalization of header space. Second, filtering it at `HTTP_Encode` will make the `Age` header disappear in some VCL code where it is available now. In addition to this, `HTTP_Encode` is in stevedore space, not a place where we do HTTP logic. 

> Also, is this really limited to Age and Accept-Ranges?

Maybe not. Close to setting `Age`, we also set `Via` and `X-Varnish`, but my first idea was that it is correct to have these multiple times (when not removed explicitly by VCL code). I can't spot any other candidates right now.

> For Accept-Ranges, bug aside, I think it does not make sense for us to advertise it for uncacheable objects so this might simplify the patch.

Maybe switching to ForceHeader is a better idea. What do you think?

I am not comfortable with removing the header entirely for some cases (uncachable objects) in this pull request. I think that should be done in a different commit, not in a bug fix like this.

> Also I don't really understand the comment on the test. Saying that we cannot test Accept-Ranges should be enough the rest is unclear / confusing.

I shortened the comment in the test. I hope it is less confusing now.

Dridi:
> we also set `Via` and `X-Varnish`, but my first idea was that it is correct to have these multiple times

I think having VXIDs from another `varnishd` can be very misleading (though helpful) during log analysis.

fgsch:
> To me that seems very unnatural. First, this would break the name HTTP_Encode for a very small optimalization of header space. Second, filtering it at HTTP_Encode will make the Age header disappear in some VCL code where it is available now. In addition to this, HTTP_Encode is in stevedore space, not a place where we do HTTP logic.

I have no idea what you mean with break the name HTTP_Encode() but the header will not disappear from VCL. Also, we already filter things in HTTP_Encode().

> Maybe switching to ForceHeader is a better idea. What do you think?

Yes.

My proposed fix with a simplified test is at: https://gist.github.com/fgsch/f9ed05db8d9c3ca0160780c9cb4a56bc
I believe this is more inline to the original intent and allow us to filter further headers if required.

blblack:
@fgsch +1, your fix does look like the more-general/correct solution pattern for this problem.

hermunn:
@fgsch, your solution looks very nice. Can you put it in your own pull request? I will close this with a reference to your PR when you have made it.

fgsch:
Done. https://github.com/varnishcache/varnish-cache/pull/1968

hermunn:
Closing this pull request as @fgsch's solution #1968 solves the same problem in a better way.

hermunn:
Removed backport tags since #1968 is the one to backport.

Issue #1964, opened 2016-05-26
The builtin.vcl hit-for-pass create logic does not take grace into account


mbgrydeland:
The builtin.vcl for vcl_backend_response contains a test for whether the object we are inserting into the cache has a positive TTL or not. If it doesn't have a positive TTL, the TTL is changed to 120s and a hit-for-pass object is inserted. The purpose of this is to disable request coalescing and enable multiple clients to go to the backend at the same time for this cache point instead of the serialization which would happen otherwise.

Though the calculation does not take grace into account, meaning that a hit-for-pass object is created earlier than strictly necessary. While the object is in grace, we should be able to continue delivering stale cache results while the backend update is performed. Instead we immediately create a hit-for-pass object and request coalescing is deactivated for the next 2 minutes.
## Possible Solution

Change the builtin.vcl to take grace into account. See referenced pull-request.
## Steps to Reproduce (for bugs)

See test case in referenced pull-request.

rezan:
Wouldn't this break VCLs where someone sets `beresp.ttl = 0s` and doesn't touch grace and expects the result to be hit-for-pass? I see this pattern in VCLs often. With this, the object will end up cached, which would be catastrophic.

In the VTC, the scenario seems to be when `Age > TTL`, you get a hit-for-pass. Im guessing this is not desired. Why would a backend specify the Age and no Cache-Control max-age? So now this scenario is `Age > TTL + grace` to get the undesired hit-for-pass (assuming no proper max-age). Is this the reason for the change? If so, we are just tweaking the Age range a bit while introducing a large pitfall.

Apologies if misunderstood something here.

Dridi:
> Wouldn't this break VCLs where someone sets `beresp.ttl = 0s` and doesn't touch grace and expects the result to be hit-for-pass?

Yes, but I suppose breaking would be OK for a major release.

> Why would a backend specify the Age and no Cache-Control max-age? [...] So now this scenario is Age > TTL + grace to get the undesired hit-for-pass (assuming no proper max-age). Is this the reason for the change?

I think @mbgrydeland did it to keep the test simple, otherwise he could have set the TTL and grace via Cache-Control.

> (Edit: I realized Varnish does this... heh)

I was about to comment that varnish could be your backend, and then I saw your edit :)

fgsch:
We could add to the builtin vcl `set beresp.grace = 0s;` in the uncacheable case to avoid breaking things.

fgsch:
[...]

PR #1965, opened 2016-05-26
Include grace time before creating hit-for-pass object

mbgrydeland:
When checking for a non-positive TTL in vcl_backend_response, include
grace before creating a hit-for-pass object.

Fixes: #1964

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -152,7 +152,7 @@ sub vcl_backend_fetch {
 }
 
 sub vcl_backend_response {
-    if (beresp.ttl <= 0s ||
+    if ((beresp.ttl + beresp.grace) <= 0s ||
       beresp.http.Set-Cookie ||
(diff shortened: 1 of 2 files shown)


bsdphk:
Looks OK

mbgrydeland:
@rezan commented in #1964:

> Wouldn't this break VCLs where someone sets beresp.ttl = 0s and doesn't touch grace and expects the result to be hit-for-pass? I see this pattern in VCLs often. With this, the object will end up cached, which would be catastrophic.

This is a concern, and warrants further discussions in VDD. Marking it as such.

Note: If we make this change, it should be in a major release, meaning no backporting.

nigoroll:
FTR, HFM was indroduced later than this PR, see 9f272127c6fba76e6758d7ab7ba6527d9aad98b0

mbgrydeland:
Closing this as OBE by hit-for-miss

PR #1968, opened 2016-05-29
Don't emit multiple Age or Accept-Ranges headers

### LINKED ISSUE #1955: 4.1.x sometimes duplicates Age and Accept-Ranges headers

fgsch:
This might happen during pass'd requests or in a multi-tier setup.
Fixes #1955.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -119,7 +119,7 @@ vbf_beresp2obj(struct busyobj *bo)
 	}
 
 	l2 = http_EstimateWS(bo->beresp,
-	    bo->uncacheable ? HTTPH_R_PASS : HTTPH_A_INS);
+	    bo->uncacheable ? HTTPH_A_PASS : HTTPH_A_INS);
 	l += l2;
 
 	if (bo->uncacheable)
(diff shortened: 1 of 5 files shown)


bsdphk:
Looks OK to me.

fgsch:
Commited.

hermunn:
commit for master is 5d272966fec974 and back port to 4.1 is 925b5f77c5, removing tag.

Issue #1874, opened 2016-03-11
[4.1/master] clock-step related crash


cinek810:
## Expected Behavior

In general child process restart is problematic, because we are loosing whole cache, which can lead to efficiency problems under high load. I'm attaching output of varnishadm panic.show
## Steps to Reproduce (for bugs)

This happened only once after last varnish update, varnish mgmt process was running for 16 days. I don't think it ease to reproduce.
## Your Environment
- Version used: varnish-4.1.1 revision 66bb824 (from https://repo.varnish-cache.org/redhat/varnish-4.1/)
- Operating System and version: CentOS release 6.7 (Final)
## My varnishadm panic.show:

```
Last panic at: Thu, 10 Mar 2016 15:04:48 GMT
"Assert error in SES_Delete(), cache/cache_session.c line 546:
  Condition(now >= sp->t_open) not true.
errno = 11 (Resource temporarily unavailable)
thread = (cache-worker)
version = varnish-4.1.1 revision 66bb824
ident = Linux,2.6.32-573.8.1.el6.x86_64,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
Backtrace:
  0x4328b3: varnishd() [0x4328b3]
  0x43976f: varnishd() [0x43976f]
  0x44e395: varnishd(HTTP1_Session+0x855) [0x44e395]
  0x439dbd: varnishd(SES_Proto_Req+0x5d) [0x439dbd]
  0x448a1d: varnishd() [0x448a1d]
  0x448e4b: varnishd() [0x448e4b]
  0x7f503ed2caa1: libpthread.so.0(+0x7aa1) [0x7f503ed2caa1]
  0x7f503ea7993d: libc.so.6(clone+0x6d) [0x7f503ea7993d]
req = 0x7f4949882020 {
  vxid = 973177293, step = R_STP_MISS,
  req_body = R_BODY_NONE,
  restarts = 0, esi_level = 1,
  sp = 0x7f4d43942020 {
    fd = 85, vxid = 973177236,
    client = 10.3.7.201 34115,
    step = S_STP_H1PROC,
  },
  worker = 0x7f503e6f7bc0 {
    stack = {0x7f503e6f8000 -> 0x7f503e6ec000},
    ws = 0x7f503e6f7db8 {
      id = \"wrk\",
      {s,f,r,e} = {0x7f503e6f7380,0x7f503e6f7380,(nil),+2040},
    },
    VCL::method = MISS,
[...]

PR #1969, opened 2016-05-29
Introduce vtim_real and vtim_mono typedefs to clarify semantics

nigoroll:
This is to prepare for more work to
- use mono time instead of real time
- use a real base timestamp offset with mono time where required/useful

Also relevant for fixing #1874 

from the commit msg:

Durations remain of type double

Exceptions:
- functions agnostic to real/mono time
- vrt interface


nigoroll:
BTW, this has ZERO function changes

fgsch:
Might be worth adding vtim.h to common.h and remove it from other places.
I think this is cleaner, specially now that cache.h and mgt.h need it.

nigoroll:
@fgsch ok, also travis reported a typedef redefinition, so I had produced a duplicate include
fixing now

nigoroll:
as discussed with @bsdphk we want to go the longer route and fix the typing issue for real. Will open a new PR when ready.

lkarsten:
Nothing commited, nothing to backport.

PR #1972, opened 2016-06-01
Provide some documentation for the VTC syntax

gquintard:
The core of the idea is documenting the source with comments looking like this::

        /* SECTION: sectionname.subsectionname title
         * 
         * thoughtful comment and stuff
         */

These are picked up by an awk script that removes the asterisks, reorder the section and output a rst file.


gquintard:
This can obviously be squashed.

bsdphk:
For all I know you are the owner of varnishtest now, so you don't need permission to improve it.

PR #1978, opened 2016-06-12
make the copyright section optional again

nigoroll:


bsdphk:
Why is this a good idea ?

nigoroll:
vmodtool accepted `$Module` as the first line in the past.
So was it intentional to make the first (copyright) section mandatory? If yes, I could prep a patch to emit a proper error message.

nigoroll:
opinions on irc 

> (13:11:03) gquintar: slink: I don't care, but I would argue that it's not vmodtool's place to enforce such things, more so because it's a generic tool used for all vmods, not just the in-tree ones
> (13:14:17) fgs: +1

lkarsten:
Backport review: this is a regression in master, not for 4.1.

Issue #1984, opened 2016-06-15
The counter vsm_cooling never goes down to 0 after switching VCLs


hermunn:
I have already made a patch for this issue, but since the process is to first make an issue, I give you the following VTC:

```
varnishtest "g_cooldown counter test"

server s1 {} -start
varnish v1 -vcl+backend {} -start

# Load and use a new VCL, freeze the first
varnish v1 -vcl+backend {}
varnish v1 -cliok "vcl.state vcl1 cold"

# the VSM_common_cleaner runs every 1.1 seconds.
# Wait for it to update vsm_cooling
delay 2
varnish v1 -expect vsm_cooling > 0

# Wait long enough for all the cooling vsm memory to be transfered to the free list:
delay 62
varnish v1 -expect vsm_cooling == 0
```

The test case demonstrates that vsm_cooling does not go down to zero after the colling period is over.

The test case needs to be run with the parameter `-t70` in `varnishtest.` If #1983 is accepted, this VTC can be changed to run much faster.

hermunn:
Backport review: Back ported to 4.1 as 1c1603e.

PR #1983, opened 2016-06-15
Introduce the parameter vsm_free_cooldown

hermunn:
This new parameter specifies how long freed VSM memory is on the
cooling list before it is actually freed.

This parameter is proposed partly because I want to make a better
VTC for an issue I am submitting soon.

PR #1985, opened 2016-06-15
Make the counter vsm_cooling act according to spec

hermunn:
The counter g_cooling variable is now decremented when memory from the
cooling list is freed, making the counter vsm_cooling update
accordingly.

The test case requires #1983 to work. See #1984 for a test case that
works without #1983.

Fixes: #1984


bsdphk:
OK

hermunn:
Pushed as 3e72fdbbf673a7b4abeb4082c9b6b95ad730c3e9

hermunn:
Backport review: Back ported to 4.1 as 1c1603ea1bef06a5a0003da7a54baa77207cc89a.

PR #1993, opened 2016-06-24
Allow to set (be)resp.body

fgsch:
This is equivalent to synthetic but using the more general set keyword.

This builds on top of #1992.


fgsch:
The checks failing is expected. As the PR message says, it depends on some changes in #1992.

bsdphk:
OK

fgsch:
Committed.

lkarsten:
Backport review: amends VCL syntax. Not for 4.1.

Issue #1990, opened 2016-06-20
Assert error in http_EstimateWS(), cache/cache_http.c line 793


whizzbangs:
```
Last panic at: Mon, 20 Jun 2016 09:56:39 GMT
"Assert error in http_EstimateWS(), cache/cache_http.c line 793:
  Condition((fm->hd[u].b) != 0) not true.
thread = (cache-worker)
version = varnish-4.1.2 revision 0d7404e
ident = Linux,3.10.0-327.13.1.el7.x86_64,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
Backtrace:
  0x4340b6: varnishd() [0x4340b6]
  0x42d3bb: varnishd(http_EstimateWS+0x44b) [0x42d3bb]
  0x422851: varnishd() [0x422851]
  0x423217: varnishd() [0x423217]
  0x44a1f1: varnishd() [0x44a1f1]
  0x44a65b: varnishd() [0x44a65b]
  0x7f0079921dc5: libpthread.so.0(+0x7dc5) [0x7f0079921dc5]
  0x7f007964f28d: libc.so.6(clone+0x6d) [0x7f007964f28d]
busyobj = 0x7efeb048c020 {
  ws = 0x7efeb048c0e0 {
    OVERFLOWED id = \"Bo\",
    {s,f,r,e} = {0x7efeb048df98,+57440,(nil),+57440},
  },
  refcnt = 2,
  retries = 5, failed = 0, state = 1,
  flags = {do_stream, do_pass, uncacheable},
  director_req = 0x7f002ac365f0 {
    vcl_name = default,
    type = round-robin {
    },
  },
  director_resp = 0x7efedd96b438 {
    vcl_name = NAMEXXX,
    type = backend {
      display_name = static.NAMEXXX,
      ipv4 = IPXXX,
      port = 81,
      hosthdr = HOSTNAMEXXX,
      health=healthy, admin_health=probe, changed=1466414686.9,
      n_conn = 0,
    },
  },
  http[bereq] = 0x7efeb048c6a8 {
    ws[Bo] = 0x7efeb048c0e0,
    hdrs {
      \"GET\",
      \"/komentarze/xxxx",
      \"HTTP/1.0\",
      \"Host: www.name\",
      \"Accept: */*\",
      \"Accept-Language: en-us\",
      \"User-Agent: Mozilla/4.0 (compatible; MSIE 11.0; Windows NT)\",
      \"Referer: http://www.name\",
[...]

PR #2001, opened 2016-06-28
Better approach at a fix for #1990

nigoroll:
83c5ead8ff1563823c91eb00c6394b3720911853 was half-hearted, this should fix the root cause


bsdphk:
OK

nigoroll:
merged

hermunn:
Backport review: As #1990 was not backported, neither will this be.

PR #2015, opened 2016-07-13
streamline ttl handling for passes

nigoroll:
It was counter-intuitive that, for pass objects, we logged the
parsed TTL when in fact the ttl visible from vcl was always 0.
In the builtin.vcl the Hit-For-Pass check was also called
for objects already classified as uncacheable.

Now, for passes, we

- do not parse Cache-Control / Expires
  - and, consequently, do not log TTL RFC
- do not call the hfp-code in builtin.vcl
  - and, consequently, do not log TTL VCL

Reflect this in b2.vtc

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -156,15 +156,15 @@ sub vcl_backend_fetch {
 }
 
 sub vcl_backend_response {
-    if (beresp.ttl <= 0s ||
+    if (beresp.uncacheable) {
+        return (deliver);
+    } else if (beresp.ttl <= 0s ||
       beresp.http.Set-Cookie ||
       beresp.http.Surrogate-control ~ "no-store" ||
       (!beresp.http.Surrogate-Control &&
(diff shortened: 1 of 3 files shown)


fgsch:
We should check bereq.uncacheable, no beresp.uncacheable.
That is more correct and if you set beresp.ttl to 0 somehow and beresp.uncacheable to true you will get serialization otherwise.

nigoroll:
Thanks @fgsch yes that's probably safer.

bsdphk:
OK

nigoroll:
Merged manually with d9b0919109b402d60d5978d8ab5f0fe370699785

hermunn:
Backport review: It seems to me that this can break some productions environments if ported to 4.1. This so it will not be backported, unless someone has a compelling argument.

nigoroll:
@hermunn I don't see the risk (the actual behaviour should remain unchanged), but I don't see any reason to backport either.

hermunn:
Thanks @nigoroll for the input. Well leave it as it is for now.

Issue #2010, opened 2016-07-07
varnishadm CLI behaving weirdly


thlc:
Entering commands on the varnishadm CLI tool result in a weird unpredictable behavior as shown by the following extract:

<pre>
# varnishadm
200        
-----------------------------
Varnish Cache CLI 1.0
-----------------------------
Linux,3.2.0-4-amd64,x86_64,-junix,-sfile,-smalloc,-hcritbit
varnish-4.1.3-beta1 revision ce824e2
 
Type 'help' for command list.
Type 'quit' to close CLI session.
 
varnish> backend.list
200        
vcl.show [-v] <configname>
backend.list [-p] [<backend_expression>]
backend.set_health <backend_expression> <state>
ban <field> <operator> <arg> [&& <field> <oper> <arg> ...]
ban.list
 
varnish> backend.list
200        
PONG 1467895240 1.0
varnish>
varnish> backend.list
200        
PONG 1467895243 1.0
varnish> backend.list
200        
Backend name                   Admin      Probe
boot.nur01                     probe      Healthy 5/5
boot.arz01                     probe      Healthy 5/5
boot.agk01                     probe      Healthy 5/5
boot.agk02                     probe      Healthy 5/5
boot.agk01_html                probe      Healthy 5/5
boot.agk02_html                probe      Healthy 5/5
varnish> backend.list
200        
PONG 1467895246 1.0
varnish> backend.list
200        
Backend name                   Admin      Probe
boot.nur01                     probe      Healthy 5/5
boot.arz01                     probe      Healthy 5/5
boot.agk01                     probe      Healthy 5/5
boot.agk02                     probe      Healthy 5/5
boot.agk01_html                probe      Healthy 5/5
boot.agk02_html                probe      Healthy 5/5
varnish> backend.list
200        
Backend name                   Admin      Probe
boot.nur01                     probe      Healthy 5/5
boot.arz01                     probe      Healthy 5/5
boot.agk01                     probe      Healthy 5/5
boot.agk02                     probe      Healthy 5/5
boot.agk01_html                probe      Healthy 5/5
boot.agk02_html                probe      Healthy 5/5
varnish> backend.list
200        
PONG 1467895285 1.0
varnish> backend.list
200        
Backend name                   Admin      Probe
boot.nur01                     probe      Healthy 5/5
boot.arz01                     probe      Healthy 5/5
boot.agk01                     probe      Healthy 5/5
boot.agk02                     probe      Healthy 5/5
[...]

PR #2019, opened 2016-07-15
Add some error checking to child_poker

hermunn:
The child_poker lives in the manager process, and sends pings to the
child. With this patch we check that we actually get a PONG back, and
not some random data. If problems are detected, we kill the child.

Related to: #2010

### DIFF
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -248,6 +248,8 @@ child_listener(const struct vev *e, int what)
 static int __match_proto__(vev_cb_f)
 child_poker(const struct vev *e, int what)
 {
+	char *r = NULL;
+	unsigned status;
 
 	(void)e;
 	(void)what;
@@ -255,9 +257,14 @@ child_poker(const struct vev *e, int what)
 		return (1);
 	if (child_pid < 0)
 		return (0);
-	if (!mgt_cli_askchild(NULL, NULL, "ping\n"))
-		return (0);
-	return (0);
+	if (mgt_cli_askchild(&status, &r, "ping\n") || strncmp("PONG ", r, 5)) {
+		MGT_complain(C_ERR, "Unexpected reply from ping: %u %s",
+		    status, r);
+		if (status != CLIS_COMMS)
+			MGT_Child_Cli_Fail();
+	}
+	free(r);
+	return 0;
 }
 
 /*=====================================================================


bin/varnishd/mgt/mgt_child.c
```diff
@@ -248,16 +248,31 @@ child_listener(const struct vev *e, int what)
...
-	return (0);
+	if (!mgt_cli_askchild(&status, &r, "ping\n")) {
+		if (strncmp("PONG ", r, 5)) {
+			MGT_complain(C_ERR,
+			    "Unexpected reply from ping: %u %s",
+			    status, r);
+			MGT_Child_Cli_Fail();
+		}
+	}
+	else {
```
fgsch: else on the same line as }

bin/varnishd/mgt/mgt_child.c
```diff
@@ -248,16 +248,31 @@ child_listener(const struct vev *e, int what)
 static int __match_proto__(vev_cb_f)
 child_poker(const struct vev *e, int what)
 {
+	char * r = NULL;
```
fgsch: no space between * and variable.

bin/varnishd/mgt/mgt_child.c
```diff
@@ -248,16 +248,31 @@ child_listener(const struct vev *e, int what)
...
+		if (strncmp("PONG ", r, 5)) {
+			MGT_complain(C_ERR,
+			    "Unexpected reply from ping: %u %s",
+			    status, r);
+			MGT_Child_Cli_Fail();
+		}
+	}
+	else {
+		MGT_complain(C_ERR, "Did not get a 200 from ping: %u %s",
+		    status, r);
```
fgsch: isn't this kind of the same error? I'd rather have a single one. 
I don't think the distinction adds any value.

fgsch:
commented inline.

bin/varnishd/mgt/mgt_child.c
hermunn: Thanks, fixing...

bin/varnishd/mgt/mgt_child.c
hermunn: Thanks, fixing

bin/varnishd/mgt/mgt_child.c
hermunn: From a debugging perspective, I think the cases are different enough to produce slightly different errors. But I guess it is a matter of taste.

fgsch: Is it? You get status and r in both cases which covers not getting a pong
and something other than a 200 status.
Both are unexpected replies from the cli pov.
The end user will mostly not understand the difference or care about the
distinction.

hermunn:
Thanks to @fgsch for useful input. I hope the latest version is ok.

fgsch:
Looks good to me.

fgsch:
Merged.w

hermunn:
Backport review: The patch in master is 8322836984bd0b234cf791b804b08232bd85c981 and the patch in 4.1 is dee325d3063b4779ed41df534349f7f6a7d5101f. Thanks, @fgsch, for closing this for me.

PR #2021, opened 2016-07-18
vcc "used before defined" check

nigoroll:
only has an effect on acls atm


Dridi:
With more testing, especially with other kinds of symbols (see r01569.vtc) we could maybe get a better idea of what could be *used-before-defined*, considering we already have the *redefined* and *already-used* errors and we could keep track of the expected type where symbols are *used-before-defined*. The only obvious ambiguity I see is between objects and VMODs that both act as namespaces for functions, but then again we have the `new` and `import` keywords that could remove the ambiguity or confirm a *redefined* or *already-defined* error.

> ```
+# XXX works ok
+#varnish v1 -errvcl {Backend 'b' used before defined.} {
+#	sub vcl_recv { set req.backend_hint = b; }
+#	backend b { .host = "127.0.0.1"; }
+#}
```

I don't understand this comment, it doesn't work for me if I remove the comments.

From the comment I'd expect the VCL to compile, I get this instead:

```
***  v1    0.3 CLI RX  106
**** v1    0.3 CLI RX| Message from VCC-compiler:\n
**** v1    0.3 CLI RX| Symbol not found: 'b' (expected type BACKEND):\n
**** v1    0.3 CLI RX| ('<vcl.inline>' Line 3 Pos 47)\n
**** v1    0.3 CLI RX|         sub vcl_recv { set req.backend_hint = b; }\n
**** v1    0.3 CLI RX| ----------------------------------------------#---\n
**** v1    0.3 CLI RX| \n
**** v1    0.3 CLI RX| Running VCC-compiler failed, exited with 2\n
**** v1    0.3 CLI RX| VCL compilation failed
**   v1    0.3 VCL compilation failed (as expected)
---- v1    0.3 Did not find expected string: ("Backend 'b' used before defined.")
```

Also you can probably remove the stevedore from your TODO-list at the end of the test case, they can neither be *redefined* or *used-before-defined*.

Dridi:
@nigoroll you dragged commits from @fgsch in this pull request, was it intended?

nigoroll:
@Dridi I just merged master.Last time we discussed this we agreed that for PR-branches we did not want a rebase + force-push but rather merge & commit fwd. That's what I did. Before potentially pushing into master I'd rebase.

Dridi:
> Last time we discussed this we agreed that for PR-branches we did not want a rebase + force-push but rather merge & commit fwd.

Did we? That's not what I remember but as long as it's not a mistake it's fine.

nigoroll:
FTR re rebase+force push / merge push update: Discussed at bugwash, I remembered it correctly, but does not seem to matter that much to others.

nigoroll:
Need to work on this in more depth, considering @bsdphk s comments:

> (10:43:36) phk: about the define-before-use VCC thing
> (10:43:43) slink: yes
> (10:44:17) phk: I looked at it and there is no way we can _generally_ use before define, given the all-types-propagate-to-string model we have chosen.
> (10:44:43) phk: the question then is if we insist on _never_ doing it, (re: ACLs)
> (10:44:55) slink: yes that was my understanding and I asked myself the same question
> (10:45:07) phk: or if we should only insist in places were we don't know the type
> (10:45:16) slink: for ACLs it would probably be user-friendlier to keep the old behavior
> (10:45:22) phk: for instance, if you call a VMOD.foo which takes  BLA argument, only BLA will do.
> (10:45:38) phk: and the fold-to-string thing doesn't come into play
> (10:45:55) slink: unless BLA is type string?
> (10:46:05) phk: right
> (10:46:35) slink: I only had quite a shallow look at your vcc re-work, so I was a bit uncertain about this PR
> (10:46:48) slink: all I know for sure is that we should improve at least the errror message
> (10:47:18) slink: What do you think about the strictness question?
> (10:47:21) phk: maybe we should just bite the bullet and say that def-before-use is the rule.
> (10:47:45) phk: it doesn't seem to me that the ACL loophole is that important.
> (10:47:50) slink: ok, then some vcls will break 4.1->5 but that should be ok
> (10:48:03) phk: and if we need it, (for instance for subs) we can always add declarations
> (10:48:10) slink: ack
> (10:48:17) phk: breaking things ->5 is why we release 5
> (10:48:39) slink: :) sorry for mentioning the trivial
> (10:48:40) phk: and we should remember that that many VCL writers are not strictly speaking good programmers
> (10:48:54) phk: so keeping the model simple has value
> (10:49:27) slink: phk so what about the pr? Do you have a better way in mind?
> (10:50:14) phk: I havn't looked at it in any detail.
> (10:50:25) phk: is it watertight and catches all use-before-def cases ?
> (10:50:27) slink: https://github.com/varnishcache/varnish-cache/pull/2021/commits/bd50e8dc569bc02ae1309ba13828f06127e54aae bottom
> (10:51:01) phk: yes, looking at it now
> (10:51:21) phk: obviously we need at least one test-case to provoke it
> (10:51:25) slink: There are other checks elsewhere in the code and I wasn't sure if we should try to unify them
> (10:51:39) slink: yes vtcs are in the pr
> (10:53:37) phk: so which cases are there really ?
> (10:53:44) phk: acl, backend, vmod ?
> (10:53:47) phk: sub ?
> (10:55:00) slink: probe
> (10:57:32) phk: right.
> (10:57:36) slink: We don't run into the used before defined check for those cases where we ask for a type explicitly
> (10:58:04) phk: no, that's what I meant to explain above.
> (10:58:17) slink: and theoretically there could be the stevedore case but that's not relevant before we get stevedore vmods iiuc
> (10:58:26) phk: I'm just wondering if we can actually explain to people why it sometimes work and sometimes doesn't work.
> (10:59:15) phk: the case that bothers me is something like:
> (10:59:24) phk: set req.http.foo = 3.1415 + something;
> (10:59:39) phk: where the result varies wildly with the type of something
> (11:06:15) slink: I think I would need to spend more time with the vcc code to attempt other improvements. Can we make a meta-decision regarding this PR? As in "it should also address these issues" or you want to look at it yourself?
> (11:08:02) phk: slink, I don't have time.  I'm perfectly fine with you doing this, I just wanted us to agree what the goal/end result should be

nigoroll:
rebased

hermunn:
Backport review: The big redesign of VCC is not i 4.1, so this will not be backported.

PR #2029, opened 2016-08-02
Make blacklist mandatory

gquintard:
blacklist is always created for directors using vdir_pick_be (hash and random), so it should never be NULL in vdir_pick_by_weight.


bsdphk:
OK

gquintard:
done by 24270700d367f7b833d601459a1a747ec0430548

hermunn:
This is just hardening the code. No changed behavior, but nice to have in 4.1. Backported as 1b6b671a863.

PR #2030, opened 2016-08-02
Remote variables not necessary to calculate the hit-rate

aondio:
the hit ratio is calculated as: hr / (hr + mr), where:
* hr =  (hit - hitrate.lhit) / dt;
* mr = (miss - hitrate.lmiss) / dt;

therefore, if we substitute hr and mr with the previous definitions we get:
[(hit - hitrate.lhit) / dt] / [(miss - hitrate.lmiss) / dt;]
and we can simplify `dt`

### DIFF
--- a/bin/varnishstat/varnishstat_curses.c
+++ b/bin/varnishstat/varnishstat_curses.c
@@ -95,7 +95,6 @@ struct pt {
 };
 
 struct hitrate {
-	double lt;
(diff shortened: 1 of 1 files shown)


bsdphk:
No idea why this hasn't been committed yet, but go for it!

hermunn:
Backport review: Backported as 0e25e7457a3c.

PR #2032, opened 2016-08-03
Don't overread

### LINKED ISSUE #2031: overread in varnishtest's parse_string
This vtc fails:
```
varnishtest
 server s1
```
(the EOF is right after server, no newline)
To generate it: 
```
echo -n "varnishtest 1
server s1" > foo
```

```
**** top   0.0 extmacro def pwd=/home/gquintard/work/varnish-cache/bin/varnishtest
**** top   0.0 extmacro def localhost=127.0.0.1
**** top   0.0 extmacro def bad_ip=192.0.2.255
**** top   0.0 extmacro def topbuild=/home/gquintard/work/varnish-cache
**** top   0.0 macro def tmpdir=/tmp/vtc.27341.2426f02f
*    top   0.0 TEST foo starting
**   top   0.0 === varnishtest 1
*    top   0.0 TEST 1
**   top   0.0 === server s1
**   top   0.0 === �
---- top   0.0 Unknown command: "�"
*    top   0.0 RESETTING after foo
*    top   0.0 TEST foo FAILED
```

## Expected Behavior
Should pass

## Possible Solution
PR coming

## Your Environment
Tested on master,  but should be the case for pretty much all released version for quite some time.

gquintard:
fix #2031

### DIFF
--- a/bin/varnishtest/vtc.c
+++ b/bin/varnishtest/vtc.c
@@ -264,7 +264,7 @@ parse_string(const char *spec, const struct cmds *cmd, void *priv,
 	struct vsb *token_exp[MAX_TOKENS];
 	char *p, *q, *f, *buf;
 	int nest_brace;
-	int tn;
+	int tn, br = 0;
 	const struct cmds *cp;
 
 	AN(spec);
@@ -298,7 +298,11 @@ parse_string(const char *spec, const struct cmds *cmd, void *priv,
 		/* First content on line, collect tokens */
 		tn = 0;
 		f = p;
-		while (*p != '\0') {
+		while (1) {
+			if (*p == '\0') {
+				br = 1;
+				break;
+			}
 			assert(tn < MAX_TOKENS);
 			if (*p == '\n') { /* End on NL */
 				break;
@@ -377,6 +381,9 @@ parse_string(const char *spec, const struct cmds *cmd, void *priv,
 
 		assert(cp->cmd != NULL);
 		cp->cmd(token_s, priv, cmd, vl);
+
+		if (br)
+			break;
 	}
 }
 


fgsch:
Why not simply leave the existing code and check for *p == '\0'  at the end of the loop?

fgsch:
This:
```
--- a/bin/varnishtest/vtc.c
+++ b/bin/varnishtest/vtc.c
@@ -377,6 +377,9 @@ parse_string(const char *spec, const struct cmds *cmd, void *priv,
 
                assert(cp->cmd != NULL);
                cp->cmd(token_s, priv, cmd, vl);
+
+               if (*p == '\0')
+                       break;
        }
 }
```

gquintard:
Because that breaks other tests (a00008.vtc, for example). buf is modified by setting token ends to '\0', so your if can get wrongly triggered.

fgsch:
Oh, right.  In that case I'd prefer if we change the test to something like this:

```
--- a/bin/varnishtest/vtc.c
+++ b/bin/varnishtest/vtc.c
@@ -262,7 +262,7 @@ parse_string(const char *spec, const struct cmds *cmd, void *priv,
 {
        char *token_s[MAX_TOKENS], *token_e[MAX_TOKENS];
        struct vsb *token_exp[MAX_TOKENS];
-       char *p, *q, *f, *buf;
+       char *e, *p, *q, *f, *buf;
        int nest_brace;
        int tn;
        const struct cmds *cp;
@@ -270,7 +270,8 @@ parse_string(const char *spec, const struct cmds *cmd, void *priv,
        AN(spec);
        buf = strdup(spec);
        AN(buf);
-       for (p = buf; *p != '\0'; p++) {
+       e = buf + strlen(buf);
+       for (p = buf; p < e; p++) {
                if (vtc_error || vtc_stop)
                        break;
                /* Start of line */
```

gquintard:
I voted against that to avoid pointer arithmetic, but a strchr should do the trick and we can get rid of an extra strlen too, thanks for highlighting that.

I'm closing this PR and opening a new one.

fgsch:
I really fail to see the problem with pointer arithmetics, it's one of the great things about C and there are plenty in tree already but wfyb.

hermunn:
Backport review: Pull request retracted in favor of #2033. Nothing to backport.

PR #2040, opened 2016-08-12
Let std.integer also do real to integer conversion

fgsch:
- std.integer can now convert real to integer
- Remove superfluous functions, they can be implemented around existing functions
- Update test

### DIFF
--- a/bin/varnishtest/tests/m00016.vtc
+++ b/bin/varnishtest/tests/m00016.vtc
@@ -1,4 +1,4 @@
-varnishtest "Test real2integer, real2time, time2integer and time2real in std"
+varnishtest "Test std.time2integer and std.time2real"
 
 server s1 {
 	rxreq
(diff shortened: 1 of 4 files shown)


nigoroll:
This simplification is to the expense of a stringification on the workspace and parsing it again, so I don't quite agree it's worth it.

fgsch:
@nigoroll that is a valid concern, indeed, and something that I considered but ultimately decided the simplification was worth it. YMMV.

This aside, and perhaps more important, is the ability to convert real values to integer using std.integer(). People in the training courses have mentioned this in more than one opportunity.
At the moment if you want to extract an integer value from a real you will need to use regsub* or std.real2integer(std.real(.., ..), ..). With d842c67 you can simply use std.integer(.., ..).

nigoroll:
I agree that this is a good feature, and the stringification gives us an implicit polimorphism.
But I see nothing wrong with `std.real2integer(std.real(.., ..), ..)`, so I'd like to keep the efficient `real2*` functions.
Re your first patch, wouldn't it be more efficient to look for a period and `strtol` only on the integer part?

nigoroll:
> Re your first patch, wouldn't it be more efficient to look for a period and strtol only on the integer part?

I guess you're not doing the fp conversion for `round()`. But actually having `std.integer` truncate as opposed to `std.real2*` would make more sense.

fgsch:
Ok, I can do truncate instead. Any other input? @bsdphk any comments?
I'd like to have the change regarding getting an integer from a float in 5.0 if possible.

fgsch:
Superseded by 2761c4362cea5604ff3e27acb5cc1283e503269a.

hermunn:
Backport review: Coversions and changes in type system is Varnish 5 only.

PR #2051, opened 2016-08-23
New assert-like macros `require`, `RN` and `RZ`

Dridi:
There are many places in the code base where assertions are used to
verify an operation relying on a dependency where there is no path for
recovery.

For instance, many `malloc` calls are followed by an `AN` check because
there is no recovery path when the child runs out of memory.  So instead
the classic course of action is to let it panic and rely on the manager
for spawning a new child.

However, `assert` can be turned into a no-op, meaning that an allocation
failure would likely be followed by a segmentation fault slightly harder
to diagnose.  Instead of assimilating dependencies to invariant checks
(assert) the new macros ensure a systematic failures.

These macros should be used for checks on foreign code, free of context.
Assuming that the underlying allocator is bug-free, the only reason to
return a NULL pointer is the lack of memory whereas a failure on a lock
operation would more likely be caused by a misuse of a pthread object
and thus better fits assertions.

Once such patterns are identified they can be applied to the whole code
base conveniently with a Coccinelle semantic patch.

### DIFF
--- a/bin/varnishd/cache/cache_ban.c
+++ b/bin/varnishd/cache/cache_ban.c
@@ -380,7 +380,7 @@ ban_reload(const uint8_t *ban, unsigned len)
 	b2 = ban_alloc();
 	AN(b2);
 	b2->spec = malloc(len);
-	AN(b2->spec);
+	RN(b2->spec);
 	memcpy(b2->spec, ban, len);
 	if (ban[BANS_FLAGS] & BANS_FLAG_REQ) {
 		VSC_C_main->bans_req++;
--- a/bin/varnishd/cache/cache_panic.c
+++ b/bin/varnishd/cache/cache_panic.c
@@ -611,6 +611,12 @@ pan_ic(const char *func, const char *file, int line, const char *cond,
 		    "Incomplete code in %s(), %s line %d:\n",
 		    func, file, line);
 		break;
+	case VAS_REQUIRE:
+		VSB_printf(pan_vsb,
+		    "Unmet requirement in %s(), %s line %d:\n"
+		    "  Condition(%s) not true.\n",
+		    func, file, line, cond);
+		break;
 	default:
 	case VAS_ASSERT:
 		VSB_printf(pan_vsb,
--- a/bin/varnishd/cache/cache_shmlog.c
+++ b/bin/varnishd/cache/cache_shmlog.c
@@ -389,7 +389,7 @@ VSL_Setup(struct vsl_log *vsl, void *ptr, size_t len)
 	if (ptr == NULL) {
 		len = cache_param->vsl_buffer;
 		ptr = malloc(len);
-		AN(ptr);
+		RN(ptr);
 	}
 	vsl->wlp = ptr;
 	vsl->wlb = ptr;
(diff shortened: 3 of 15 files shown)


nigoroll:
aren't these actually cases for XXXAN/XXXAZ or rather proper error handling?

Dridi:
On Aug 24, 2016 13:15, "Nils Goroll" <notifications@github.com> wrote:
>
> aren't these actually cases for XXXAN/XXXAZ or rather proper error
handling?

Not if unrecoverable, that is the idea

slimhazard:
> However, assert can be turned into a no-op, ...

Only if you compile with `-DWITHOUT_ASSERTS`.

Which I suppose someone could do, one would hope with awareness of the consequences, but to be honest I've never heard of anyone building with asserts disabled, after years of working with the Varnish code. I had to go looking in `vas.h` to remember how it's done. And I'm sure I would never recommend it, since I don't see any advantage (certainly no performance gain to speak of), and I can't see how it would be worth the risk of losing good diagnostics if an error is encountered.

So I guess I'm not seeing the problem that needs to be solved. Especially since this brings about a far-reaching change in the code -- do we really want to stop and think about whether to use `AZ` or `RZ` every time?

@Dridi, have you encountered a real-world problem that calls for this change?

Dridi:
> Which I suppose someone could do, one would hope with awareness of the consequences, but to be honest I've never heard of anyone building with asserts disabled, after years of working with the Varnish code.

Yes but `WITHOUT_ASSERTS` is there, so another solution would be to take the too-many-knobs wagon, kill the macro and make all asserts systematic considering their current hybrid use. I'm fine with that solution too but I'd prefer the middle-ground of differentiating actual assertions from external requirements.

> since I don't see any advantage (certainly no performance gain to speak of)

If one doesn't consider correctness to be a performance aspect, then I'm pretty sure (read: I've witnessed) that disabling asserts would yield performance gains. That would remove many branches from the code and be more friendly to instruction caches, especially in functions starting with lots of (I insist, very useful) assertions.

But I've never measured anything *properly* so this is pure speculation. On my machine the test suite takes on average 1m24 vs 48s when I disable asserts, I just tried half a dozen runs of both and all results were always close to the average.

> Especially since this brings about a far-reaching change in the code

No, I've only added the macros (and two replacement examples) and there's certainly no rush in finding out all the places where the check should be systematic. However when someone find such a case then there's a means for it.

> do we really want to stop and think about whether to use AZ or RZ every time?

Conversely wouldn't it be nice to see the difference between invariants and requirements without having to think about it?

> have you encountered a real-world problem that calls for this change?

Actually yes, you know someone who disables asserts. When I work on Varnish I run the test suite without asserts to save time until I have something ready to commit/push. A habit that reduces my mental context switches. Admittedly a very recent habit :)

And then I got puzzled by a failing allocation that turned into a segfault, which isn't obvious to spot when all you have is symbols and hexadecimal offsets in your backtrace.

nigoroll:
> Yes but `WITHOUT_ASSERTS` is there

and while we seem to agree that using it should not be encouraged, adding `RZ`/`RU` would even encourage it more - _oh cool, we can now turn off all the unnecessary assertions only_
This probably is a bikeshed, so I won't make any more comments here. But my impression is that is suggestion is not helpful - while replacing assertions with proper error handling would be.

hermunn:
Backport review: Nothing commited - nothing to backport.

PR #2054, opened 2016-08-24
Modernize the build system for out-of-tree vmods

Dridi:
> It should make the life of VMOD maintainers a lot easier without forcing
them to migrate their build systems right away.  The old macros could be
killed with the next major release.  Ideally they should also build the
`std` and `directors` module because dogfooding. &lt;insert promises of a
unicorns and rainbows and happy princesses here&gt;

### DIFF
--- a/varnish.m4
+++ b/varnish.m4
@@ -1,104 +1,177 @@
 # varnish.m4 - Macros to locate Varnish header files.            -*- Autoconf -*-
-# serial 3 (varnish-4.0)
+# serial 4 (varnish-4.1.4)
 
 # Copyright (c) 2013-2015 Varnish Software AS
 # All rights reserved.
 #
-# Author: Tollef Fog Heen <tfheen@varnish-software.com>
+# Author: Dridi Boukelmoune <dridi.boukelmoune@gmail.com>
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
 # are met:
-# 1. Redistributions of source code must retain the above copyright
-#    notice, this list of conditions and the following disclaimer.
-# 2. Redistributions in binary form must reproduce the above copyright
-#    notice, this list of conditions and the following disclaimer in the
-#    documentation and/or other materials provided with the distribution.
 #
-# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
-# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
-# ARE DISCLAIMED.  IN NO EVENT SHALL AUTHOR OR CONTRIBUTORS BE LIABLE
-# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
-# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
-# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
-# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
-# SUCH DAMAGE.
+# 1. Redistributions of source code must retain the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer.
+# 2. Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials
+#    provided with the distribution.
 #
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
 
-# For compatibility with autoconf < 2.63b
-m4_ifndef([AS_VAR_COPY],
-  [m4_define([AS_VAR_COPY],
-     [AS_LITERAL_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])])
-
-# VARNISH_VMOD_INCLUDE_DIR([])
-# ----------------------------
-
-AC_DEFUN([VARNISH_VMOD_INCLUDES],
-[
-m4_pattern_forbid([^_?VARNISH[A-Z_]+$])
-m4_pattern_allow([^VARNISH_VMOD(_INCLUDE_DIR|TOOL)$])
-# Check for pkg-config
-PKG_CHECK_EXISTS([varnishapi],[],[
-	if test -z "$PKG_CONFIG"; then
-		AC_MSG_FAILURE(
-[The pkg-config script could not be found or is too old.  Make sure it
-is in your PATH or set the PKG_CONFIG environment variable to the full
-path to pkg-config.
-
-To get pkg-config, see <http://www.freedesktop.org/wiki/Software/pkg-config/>.])
-	else
-		AC_MSG_FAILURE(
-[pkg-config was unable to locate the varnishapi configuration data.
-
-Please check config.log or adjust the PKG_CONFIG_PATH environment
-variable if you installed software in a non-standard prefix.])
-	fi
-])
-
-VARNISH_PKG_GET_VAR([VAPI_INCLUDE_DIR], [pkgincludedir])
-_CPPFLAGS="$CPPFLAGS"
-VMOD_INCLUDES="-I$VAPI_INCLUDE_DIR"
-CPPFLAGS="$VMOD_INCLUDES $CPPFLAGS"
-AC_CHECK_HEADERS([vsha256.h cache/cache.h])
-CPPFLAGS="$_CPPFLAGS"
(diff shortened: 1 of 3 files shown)


Dridi:
The diff doesn't help since serial 4 has nothing to do with serial 3, direct link: https://github.com/Dridi/varnish-cache/blob/e9c0f5b/varnish.m4

varnish.m4
```diff
@@ -1,104 +1,177 @@
...
+	AC_SUBST(m4_toupper(VMOD_$1_FILE), [$VMOD_FILE])
+
+	VMOD_IMPORT="querystring from \\\"$VMOD_FILE\\\""
+	AC_SUBST(m4_toupper(VMOD_$1), [$VMOD_IMPORT])
+
+	VMOD_RULES="
+
+vmod_$1.lo: vcc_$1_if.c vcc_$1_if.h
+
+vcc_$1_if.c vcc_$1_if.h vmod_$1.rst vmod_$1.man.rst: .vcc_$1
```
nigoroll: not too happy about this dummy target. Please let's use one of the vmodtool outputs as the main target and have the other ones depend on it

varnish.m4
```diff
@@ -1,104 +1,177 @@
...
+# _VARNISH_VMOD(NAME)
+# -------------------
+AC_DEFUN([_VARNISH_VMOD], [
+
+	AC_REQUIRE([_VARNISH_VMOD_CONFIG])
+
+	VMOD_FILE="\$(abs_builddir)/.libs/libvmod_$1.so"
+	AC_SUBST(m4_toupper(VMOD_$1_FILE), [$VMOD_FILE])
+
+	VMOD_IMPORT="querystring from \\\"$VMOD_FILE\\\""
```
nigoroll: querystring?

varnish.m4
```diff
@@ -1,104 +1,177 @@
...
+
+vmod_$1.3: vmod_$1.man.rst
+	$RST2MAN vmod_$1.man.rst vmod_$1.3
+
+clean: clean-vmod-$1
+
+distclean: clean-vmod-$1
+
+clean-vmod-$1:
+	rm -f vcc_$1_if.c vcc_$1_if.h
```
nigoroll: should ignore errors here (prefix lines with -)

nigoroll: or use CLEANFILES in the first place

varnish.m4
```diff
@@ -1,104 +1,177 @@
...
+
+	AC_REQUIRE([_VARNISH_VMOD_CONFIG])
+
+	VMOD_FILE="\$(abs_builddir)/.libs/libvmod_$1.so"
+	AC_SUBST(m4_toupper(VMOD_$1_FILE), [$VMOD_FILE])
+
+	VMOD_IMPORT="querystring from \\\"$VMOD_FILE\\\""
+	AC_SUBST(m4_toupper(VMOD_$1), [$VMOD_IMPORT])
+
+	VMOD_RULES="
```
nigoroll: looking at libvmod-example/src/Makefile.am, I whether the autogen'ed VMOD_RULES should also contain:
* ..._la_SOURCES
* nodist..._la_SOURCES
* dist_man_MANS
* EXTRA_DIST
* CLEANFILES (but see below)

varnish.m4
```diff
@@ -1,104 +1,177 @@
...
+vmod_$1.lo: vcc_$1_if.c vcc_$1_if.h
+
+vcc_$1_if.c vcc_$1_if.h vmod_$1.rst vmod_$1.man.rst: .vcc_$1
+
+.vcc_$1: vmod_$1.vcc
+	\$(AM_V_VMODTOOL) $PYTHON $VMODTOOL -o vcc_$1_if \$(srcdir)/vmod_$1.vcc
+	@touch .vcc_$1
+
+vmod_$1.3: vmod_$1.man.rst
+	$RST2MAN vmod_$1.man.rst vmod_$1.3
```
nigoroll: the syntax in the current example Makefile is nicer
```
vmod_example.3: vmod_example.man.rst
        ${RST2MAN} $< $@

```

varnish.m4
Dridi: Done.

nigoroll:
I dropped a couple of comments on the diff, but in general I'm very much in favor of @Dridi s work and would like to see it merged. We should then change a couple of vmods (and https://github.com/nigoroll/varnish-modules in particular) and see how what might still needs attention.

varnish.m4
Dridi: Automake doesn't pick them if they're hidden inside a `@BUILD_VMOD_FOO`.

varnish.m4
Dridi: The output here is nicer and would work with both vmod-example and *multi-module* builds.

varnish.m4
nigoroll: :)

varnish.m4
Dridi: No CLEANFILES, see other comment about that.

And `rm -f` doesn't fail with arguments, unless you tamper with the generated files (change permissions?) in which case it would be nice to know and abort the make execution.

varnish.m4
Dridi: Fixed.

varnish.m4
nigoroll: CLEANFILES ok, got it
rm -f : Are you sure this is the case for all systems?

Dridi:
Usage examples (focus on `configure.ac`, `Makefile.am` and `src/Makefile.am`) exist for both single- and multi-module source trees:

- https://github.com/Dridi/libvmod-querystring/tree/feature/build
- https://github.com/Dridi/varnish-modules/tree/feature/build

varnish.m4
Dridi: Misread the comment! I think I had issues with bmake so I did that instead.

varnish.m4
Dridi: Don't know about other platforms but here is what posix mandates on `rm -f`:
> Do not write diagnostic messages or modify the exit status in the case of nonexistent operands.

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/rm.html

slimhazard:
Not sure if this is the right place to suggest this, but here goes anyway. For a while I've been using this trick to get a README.rst generated from the vcc file into the root of a VMOD repo, so that it appears on the project front page of github or gitlab (in Makefile.am in the repo root):

```
doc_DATA = README.rst LICENSE COPYING

README.rst: src/vmod_$VMOD.man.rst
        cp src/vmod_$VMOD.man.rst README.rst

src/vmod_$VMOD.man.rst:
        make -C src vmod_$VMOD.man.rst
```
That is, execute the target to generate the RST from vcc in src, then copy it into the repo root as README.rst.

Then you have to commit the generated README.rst to get it to appear in the project front page, and of course it changes and has to be pushed again whenever the vcc changes.

Checking in a generated file to a source repo is a bit uncouth and I know some people won't like it. My reasons for doing this:

I'm passionate about keeping documentation close to the source that it documents. vcc necessarily has accurate technical information about the VMOD interface, so the docs generated from vcc will always reflect at least that information correctly, even if the programmer falls behind documenting everything else.

I despise maintaining docs about the same material in two different places, meaning that I hate having a separate README for this purpose. The reason for my revulsion is that during the lifecycle of a software project, it's only a matter of time until the one source diverges form the other, and one of the docs becomes inaccurate -- I've seen it happen too many times. In our case, there's too much potential for the interface docs in vcc, which are always correct, to diverge from a separately maintained README. So I very much want to keep vcc as the sole source of technical docs.

Having a README.rst with the entirety of project docs in the front page of a web view of the source repo is likely to be much more informative to first-time visitors of the project. I fear that a lot of people don't take the trouble to click into Files->src->vcc, and even if they do, they then have to read the RST source, instead of getting it all in HTML right up front.

README in the project front page will also provide better information for search engine bots.

For these reasons, checking in a generated file is IMO very much justified.

So I'd recommend adding this to the example VMOD, so that it happens automatically for most VMOD authors (I know, that's not really what @Dridi is doing here -- maybe this should be a PR for libvmod-example). Along with instructions to check in the generated README.rst -- VMOD authors can of course decide not to do that if they don't like it.

Dridi:
@slimhazard, as I went through the same hurdles and have found a solution that completely satisfies me I'll share it with you :)

My README contains a different kind of documentation than the VMOD, that is to say an introduction to its purpose and a general idea of how it works. Then in the VCC file I focus on manual-type documentation and have sections in addition to the ones generated by vmodtool.

The thing about the README is that it can also contain installation instruction that don't belong in a man page installed on your system. Thanks to the material of this PR and some autotools digging I was able to streamline the process from `git clone` to having RPM packages.

See the differences between README and VCC:

* https://github.com/Dridi/libvmod-querystring/blob/feature/build/README.rst
* https://github.com/Dridi/libvmod-querystring/blob/feature/build/README.rst#installation
* https://github.com/Dridi/libvmod-querystring/blob/feature/build/src/vmod_querystring.vcc

And that solve the thing I despise: checking in files generated from other checked-in sources.

varnish.m4
nigoroll: @Dridi thanks and sorry

Dridi:
@lkarsten asked me whether this had support for having a minimum Varnish version. The answer is yes.

The two autoconf public macros can be seen here:
https://github.com/Dridi/varnish-modules/blob/748e7704/configure.ac#L27-L37

The rest is completely hidden from the VMOD maintainer.

I also pushed 67b743f to this pull request to make one more automake variable available.

See current master vs my experiment:
* https://github.com/varnish/varnish-modules/blob/bc4ea8a/src/Makefile.am#L14-L82
* https://github.com/Dridi/varnish-modules/blob/748e770/src/Makefile.am#L25-L56

Hint: one of them fits on one screen.

Automake walkthrough:

* [aclocal](https://github.com/Dridi/libvmod-querystring/blob/842fcb1/Makefile.am#L33)
* [pkg-config flags](https://github.com/Dridi/libvmod-querystring/blob/842fcb1/src/Makefile.am#L32-L33)
* [build](https://github.com/Dridi/libvmod-querystring/blob/842fcb1/src/Makefile.am#L41-L48)
*

fgsch:
Has this been tested in FreeBSD or other non-Linux systems?

Dridi:
Yes, see e9c0f5b although I'll leave OpenBSD to you :smile:. I took advantage of being in a bit passive in Oslo today to run background builds on other platforms (mostly waiting for VMs to be up).

Tested successfully on up to date systems:

- Fedora 24 x86_64
- SunOS 5.11 joyent_20160317T000621Z i86pc i386 i86pc
- FreeBSD freebsd-10-amd64.local
- CentOS 7 x86_64
- CentOS 6 x86_64
- Ubuntu 16.04 x86_64
- Ubuntu 14.04 x86_64

I couldn't make use of it on CentOS 5, or at least not without considerably crippling it and defeating its purpose.

I believe that the macros would work with Varnish 4.0.0, the only requirements are recent enough versions of the autotools suite. So in theory downstream packagers could ship these build macros with their current Varnish 4 version.

Dridi:
@gquintard reported that it worked in his Arch Linux environment.

fgsch:
Am I missing something or anyone using VARNISH_VMOD_INCLUDES, etc. will need to migrate to varnishapi-deprecated?

IMO this is not OK.

Dridi:
Both "modernized" and deprecated macros will be in your `ACLOCAL_PATH`, so old macros would still work out of the box. See e9c0f5b for the whole explanation, but it is said in the very first sentence of this pull request:

> It should make the life of VMOD maintainers a lot easier without forcing them to migrate their build systems right away.

fgsch:
Yup, nevermind, it was late and I misread it.

As long as the old macros are always available, i.e. we do not enforce people to use a particular system, this should be OK (which makes me wonder if we should really have varnish-deprecated.m4 at all).

Dridi:
I'd like to get rid of the old stuff, deprecating it first so that in the Nth [1] next major release it can go away. I also want to later add other macros not targeted at VMODs but VUTs, starting small and improving later. Also as mentioned, dogfooding for std, debug and directors and other autoconf improvements I can see.

[1] where N >= reasonnable threshold

fgsch:
And I'm arguing we shouldn't remove it because you are forcing everyone to use these macros.

Dridi:
I wanted to double-check CentOS 5, and I actually managed to install Varnish 4.1.3 and the branch of vmod-querystring that uses this PR's material.

Although I can't do the autotools dance on el5, building from a dist tarball works fine (I had to sneak in python 2.6 to build Varnish first). So the generated shell scripts and Makefiles also work on el5, not the m4 macros. :astonished:

hermunn:
Backport review: Backported by Dridi as 2b279cd6.

PR #2057, opened 2016-08-25
vcl_deliver read-only access to obj.ttl, obj.age, obj.grace and obj.keep

nigoroll:
Noticed that I've been missing this since the beginning of time when
porting some ancient inline-C.

We already have obj.uncacheable und obj.hits in vcl_deliver, so not
having these seems arbitrary. Copying out the values in vcl_hit just
to have them available in vcl_deliver, as it was possible previously,
implies unnecessary overhead.

One use case is to extend object lifetime downstream when delivering a
stale object, e.g.

```
sub vcl_deliver {
    if (! obj.uncacheable && obj.ttl < 0s) {
	set resp.http.Expires = now + 1m;
	set resp.http.Cache-Control = "public, max-age=60";
	unset req.http.Age;
    }
}
```

### DIFF
--- a/bin/varnishtest/tests/c00011.vtc
+++ b/bin/varnishtest/tests/c00011.vtc
@@ -13,6 +13,13 @@ varnish v1 -vcl+backend {
 	sub vcl_backend_response {
 		set beresp.uncacheable = true;
 	}
+	sub vcl_deliver {
+		set resp.http.o_uncacheable = obj.uncacheable;
+		set resp.http.o_age = obj.age;
+		set resp.http.o_ttl = obj.ttl;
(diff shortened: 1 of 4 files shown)


nigoroll:
rebased

fgsch:
I don't see why not. Looks good to me.

bsdphk:
OK

nigoroll:
closed via 9095c5cfbac7fa3cb48257f6631f24239b30ee45

hermunn:
Backport review: Backported as b76c5f5 b376262 	f6ba857 ce693dc.

PR #2058, opened 2016-08-25
Append a default system-wide location to vcl_path

Dridi:
Of course it can still be removed via the varnish cli.

The *how* and the *why* are detailed in the first commit message.

The *when* dates to the discussions that lead to the replacement of `vcl_dir` by `vcl_path`.

After all the efforts it took me understanding autotools to produce #2054, this one became trivial. I've had ideas of VCL libraries that could fit this model in the past but never implemented any, the one I see fitting best in this model today is @comotion's VSF (Varnish Security Firewall).

In my own `querystring` VMOD I could ship VCL subroutines for turnkey integration with big players (eg. google analytics) or with web stacks that sometimes do cache-unfriendly things with query-strings (although it's been a while since I've last been confronted to a web stack luckily).

In my case it might be overkill since an absolute path would be enough, but in large libraries like VSF with transitive includes that would be a nice alternative to invading the `/etc` tree.

If merged, it needs to be documented and not just in the changelog.

### DIFF
--- a/bin/varnishd/Makefile.am
+++ b/bin/varnishd/Makefile.am
@@ -136,7 +136,7 @@ varnishd_CFLAGS = \
 	-DVARNISHD_IS_NOT_A_VMOD \
 	-DVARNISH_STATE_DIR='"${VARNISH_STATE_DIR}"' \
 	-DVARNISH_VMOD_DIR='"${pkglibdir}/vmods"' \
-	-DVARNISH_VCL_DIR='"${varnishconfdir}"'
+	-DVARNISH_VCL_DIR='"${varnishconfdir}:$(datarootdir)/vcl"'
 
 varnishd_LDFLAGS = -export-dynamic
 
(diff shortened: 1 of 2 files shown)


lkarsten:
vmod_path or vcl_path?

Dridi:
@lkarsten I just s/vmod/vcl/ the title, I had to confuse them at least in one place I suppose :-/

fgsch:
I like this.

bsdphk:
Sounds like a good idea.

Looking over my computers /usr{/local}/share, it looks like the naming in general is ${datadir}/${package-ish}/whatever.

That would imply that we should use ${datadir}/varnish/vcl

Dridi:
> That would imply that we should use ${datadir}/varnish/vcl

I had considered this too, so I'll go for it.

I believe this would break packaging (at least RPM I think) so we'll need to update that too. I'll ping @lkarsten again once it's committed.

Dridi:
I just realized no one commented on installing `devicedetect.vcl`, did you not notice it or is it OK for commit too?

fgsch:
@Dridi It's OK but we should sync it first.

Dridi:
Committed as f3862b4...989b39f.

hermunn:
Backport review: Not a bugfix, so this is not being backported.

PR #2059, opened 2016-08-28
Add a "key" field to each backend.

sesse:
Normally, backends with the same host (as determined by IPv4/IPv6 address pair)
and port will share TCP connections, but if you give them different keys,
they will not. This is useful if your HTTP connections have some kind of state
(e.g. for security) which makes it impossible to reuse them across backends.

This is a (trivial) forward-port of a 4.1 patch we've been running in production.

### DIFF
--- a/bin/varnishd/cache/cache_backend.h
+++ b/bin/varnishd/cache/cache_backend.h
@@ -120,7 +120,7 @@ void VBP_Status(struct cli *cli, const struct backend *, int details);
 
 /* cache_backend_tcp.c */
 struct tcp_pool *VBT_Ref(const struct suckaddr *ip4,
-    const struct suckaddr *ip6);
+    const struct suckaddr *ip6, const char *key);
 void VBT_Rel(struct tcp_pool **tpp);
 int VBT_Open(const struct tcp_pool *tp, double tmo, const struct suckaddr **sa);
 void VBT_Recycle(const struct worker *, struct tcp_pool *, struct vbc **);
--- a/bin/varnishd/cache/cache_backend_cfg.c
+++ b/bin/varnishd/cache/cache_backend_cfg.c
@@ -112,9 +112,9 @@ VRT_new_backend(VRT_CTX, const struct vrt_backend *vrt)
 	Lck_Lock(&backends_mtx);
 	VTAILQ_INSERT_TAIL(&backends, b, list);
 	VSC_C_main->n_backend++;
-	b->tcp_pool = VBT_Ref(vrt->ipv4_suckaddr, vrt->ipv6_suckaddr);
+	b->tcp_pool = VBT_Ref(vrt->ipv4_suckaddr, vrt->ipv6_suckaddr, vrt->key);
 	if (vbp != NULL) {
-		tp = VBT_Ref(vrt->ipv4_suckaddr, vrt->ipv6_suckaddr);
+		tp = VBT_Ref(vrt->ipv4_suckaddr, vrt->ipv6_suckaddr, vrt->key);
 		assert(b->tcp_pool == tp);
 	}
 	Lck_Unlock(&backends_mtx);
--- a/bin/varnishd/cache/cache_backend_tcp.c
+++ b/bin/varnishd/cache/cache_backend_tcp.c
@@ -54,6 +54,7 @@ struct tcp_pool {
 	char			*name;
 	struct suckaddr		*ip4;
 	struct suckaddr		*ip6;
+	char                    *key;
 
 	VTAILQ_ENTRY(tcp_pool)	list;
 	int			refcnt;
@@ -118,12 +119,14 @@ tcp_handle(struct waited *w, enum wait_event ev, double now)
 }
 
 /*--------------------------------------------------------------------
- * Reference a TCP pool given by {ip4, ip6} pair.  Create if it
+ * Reference a TCP pool given by {ip4, ip6, key} pair.  Create if it
  * doesn't exist already.
  */
 
 struct tcp_pool *
-VBT_Ref(const struct suckaddr *ip4, const struct suckaddr *ip6)
+VBT_Ref(const struct suckaddr *ip4,
+    const struct suckaddr *ip6,
+    const char *key)
 {
 	struct tcp_pool *tp;
 
@@ -147,6 +150,16 @@ VBT_Ref(const struct suckaddr *ip4, const struct suckaddr *ip6)
 			if (VSA_Compare(ip6, tp->ip6))
 				continue;
 		}
+		if (key == NULL) {
+			if (tp->key != NULL)
+				continue;
+		} else {
+			if (tp->key == NULL)
+				continue;
+			if (strcmp(key, tp->key))
+				continue;
+		}
+
 		tp->refcnt++;
 		return (tp);
 	}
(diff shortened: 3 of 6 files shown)


nigoroll:
@sesse I find "reuse them across backends" a bit irritating. Do IUC that you are using a backend application which dedicates HTTP/1.1 keepalive connections to some special purpose? As a simplified example I could imagine an application only allowing the same Host: header for all requests of the same connection.
I am not sure if this is useful in the general case, but your PR will probably get discussed during bugwash tomorrow.

sesse:
Lasse can fill in the details, but the basic issue here is that the backend runs each vhost under a different uid and can't switch once the uid is set for the HTTP connection. So yes, your understanding is right. (The other real alternative would be setting Connection: close on each and every backend request, which is suboptimal.)

nigoroll:
@sesse understood, thx
Have you considered assigning the key to tcp pool connections rather than backends?

sesse:
I didn't consider it, no. What would be the advantages? On the disadvantage side would be (AIUI) that it should be somewhat slower; you'd need to scan through a bunch of connections you can't use, comparing the key string against all of them.

nigoroll:
I think it would be advantageous not having to assign keys a-priori. Regarding the linear search, atm we search the pools list linearly, so moving the key lookup to the connection list shouldn't make that much of a difference. That said, we should improve the lookup.

sesse:
Well, the way we use it is that we've also patched gwist to take in a key, so we don't assign keys a priori :-) I don't know to what degree Varnish wants to support dynamic backends without vmods in the future, though.

Dridi:
> This is useful if your HTTP connections have some kind of state (e.g. for security)

Naive question but since HTTP is a stateless protocol, isn't this fundamentally wrong? Do you have a use case that could help me better understand the problem?

I see *a* point but I'd go for fixing the backends instead.

> I don't know to what degree Varnish wants to support dynamic backends without vmods in the future

If I understand the trend since Varnish 3, we are actively removing anything from the core that could live out of tree. I believe dynamic backends (and directors in general) fall in that category.

sesse:
It's not disallowed as per RFC2616, if that's what you ask. The backend is allowed to close a persistent connection at any given time, and clients just have to deal with that. (Otherwise, timeouts wouldn't work.)

As for changing the backend, it's hard to see exactly how one should go about that and still maintain the desired security properties. If the backend has setuid() and then goes and runs some arbitrary user code (with the full privileges of the user running the backend), how would you safely allow it to setuid() back to some other user?

Dridi:
> It's not disallowed as per RFC2616, if that's what you ask.

It's not, it was about actually breaking the statelessness of requests. I was reacting to the word "state" in the text I quoted.

> how would you safely allow it to setuid() back to some other user?

Are you executing arbitrary user code as in calling one of the `exec(3)` or `execve(2)` functions?

In this case I would call `setuid` after the `fork` and before the `exec`, dropping privileges before the call to arbitrary code without affecting the (multi-threaded?) process that received the HTTP request.

sesse:
No, not executing arbitrary code as in forking off a CGI script with suexec; executing arbitrary code as in executing arbitrary code. Anything Apache allows, including an almost full range of modules (which means e.g. mod_perl, which is totally un-sandboxed).

Dridi:
What happens if you `setuid` and run perl code, but while it's running another request handling leads to a `setuid` call too? Your perl execution suddenly switches privileges in the middle of its execution?

sesse:
The process handles only one request at a time (it's a prefork model), so that problem won't happen.

nigoroll:
@sesse if you've also tweaked https://github.com/gquintard/libvmod-gwist, then I really think that _if_ we added something like this, we should really make the key a property which we have available at the time we lookup a tcp connection. working title _backend connection reuse key_
Also I wonder how this will play with H2...

Dridi:
@sesse thank you for the details. I still think that the backend is fundamentally broken if you end up with stateful HTTP sessions.

bsdphk:
I agree that the backend is basically broken relative to the intent and text of the HTTP RFCs.

However, I am not the least surprised, since the majority of the HTTP ecosystem routinely "forgets" that proxies even exist.

If we want to support hacks like that, the sensible way would be to have a per-backend property::

    backend foobar {
        .host = ...
        .share_connections = False;
    }

But then the hard questions muster up, for instance: Are these connections shared across VCLs ?

Dridi:
@bsdphk if you limit connection sharing to a boolean you make the same assumption that incoming requests from a same session come from the same client, discarding proxies in front of Varnish.

So we'd effectively need a way to tag users for session affinity, coming back to square one of this pull request.

> But then the hard questions muster up, for instance: Are these connections shared across VCLs ?

Assuming that the user knows what they do, I'd go with @nigoroll's suggestion to move this at the connection pool level. We currently make the same assumption for request hashing where switching VCLs could effectively "invalidate" all your cache if you break your `vcl_hash`.

nigoroll:
I'm with @Dridi

nigoroll:
@sesse would you be interested in working on a patch to implement the key at the connection pool level?

sesse:
If you feel that is indeed better, then I can do that. It does have the extra cost of comparing strings, but I guess we can live with it.

nigoroll:
@sesse I have no idea if @bsdphk will accept it, but at this point I think this is the only sensible approach. Also, you should be able to do without the gwist tweak.

bsdphk:
@dridi I was addressing only the initial request from @sesse: That connections not be shared across backends.

If you want to make this dynamic, so sharing can be disabled per-whatever, then you still need to answer my question:

Does this state reach across VCLs ?

The new separate VCL thing makes this a very important question:

What if two different vhosts each with its own VCL just happen to pick the same key ("authorized" ?) to what happens to be the same backend-IP (some-random-cloud) should their connections pool ?

The answer seems pretty obvious, but the consequences are stark:  All key'ed connections are local to the VCL that created them, which in the common single-VCL case means: vcl.use drops them all.

Dridi:
> @dridi I was addressing only the initial request from @sesse: That connections not be shared across backends.

Indeed, it makes sense in this context.

> What if two different vhosts each with its own VCL just happen to pick the same key ("authorized" ?) to what happens to be the same backend-IP (some-random-cloud) should their connections pool ?

I will refer to my previous answer and suggest not to care :)

Today VCL labels allow to isolate caching policies in terms of code, making it very convenient to mix unrelated caching policies easily without having them leaking into each other and without repetitive if/elsif dances in all subroutines.

However from the perspective of shared hosting, VCL labels wouldn't prevent hash collisions (think: "a malicious customer may...") and there's nothing to pin a label to a specific storage for instance. So I think we should make it the end-user's problem since everything else is already their problem.

So for a broken-backend-workaround I think it shouldn't go too far.

nigoroll:
@sesse would you close this PR and open a new one when you're ready?

hermunn:
Backport review: Nothing committed, nothing to backport.

sesse:
@nigoroll: I have a few questions related to making this work at the connection pool level instead of making it part of the backend.

1. Presumably this would be part of the bereq, set in vcl\_backend\_fetch; is that right?
2. Won't this mean we need to do many more key checks than with the existing patch? More specifically, on the order of number of connections in the backend's pool instead of on the order of number of backends.
3. Given that the key is a string, how would you recommend we allocate that string? It would presumably be set as part of the busyobj, but the pointer can't be copied from there into the tcp\_pool, since the pool has a longer life than the busyobj (right?). It also can't live as part of the tcp\_pool, since that's long-lived—would we need to refcount it? I suppose we can't give it a fixed max length and just have it be e.g. a char[64].

PR #2060, opened 2016-08-29
Implicit backend cast to boolean

Dridi:
Inspired by a VMOD [1] that does two lookups where one could be enough when there is no backend.

```vcl
sub vcl_recv {
  if (!sd.contains(req.http.host)) {
    return(synth(404));
  } else {
    set req.backend_hint = sd.backend(req.http.host);
  }
}
```

Instead one could do:

```vcl
sub vcl_recv {
  set req.backend_hint = sd.backend(req.http.host);
  if (!req.backend_hint) {
    return(synth(404));
  }
}
```

[1] https://github.com/gquintard/libvmod-stendhal#varnish-stendhal-module

### DIFF
--- a/bin/varnishtest/tests/d00014.vtc
+++ b/bin/varnishtest/tests/d00014.vtc
@@ -1,0 +1,20 @@
+varnishtest "Backend as a boolean expression"
+
+server s1 -start
+
+varnish v1 -vcl+backend {
+	import debug;
+
+	sub vcl_recv {
+		set req.backend_hint = debug.no_backend();
+		if (!req.backend_hint) {
+			return (synth(404));
+		}
+	}
+} -start
+
+client c1 {
+	txreq
+	rxresp
+	expect resp.status == 404
+} -run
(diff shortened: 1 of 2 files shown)


Dridi:
@gquintard ping

nigoroll:
seems like a reasonable thing to do - in particular because
`if (req.backend_hint == "")`
does not work. The current workaround is to assign `req.backend_hin`t to something taking a string and then testing for that thing to be the empty string...

nigoroll:
example for the current hackish workaround to do the same
```
diff --git a/bin/varnishtest/tests/r01501.vtc b/bin/varnishtest/tests/r01501.vtc
index 38f220c..b0f94f8 100644
--- a/bin/varnishtest/tests/r01501.vtc
+++ b/bin/varnishtest/tests/r01501.vtc
@@ -6,11 +6,16 @@ server s1 {
 } -start
 
 varnish v1 -vcl+backend {
-
+       import std;
        import debug;
 
        sub vcl_recv {
                set req.backend_hint = debug.no_backend();
+               std.log(req.backend_hint);
+               set req.http.test = req.backend_hint;
+               if (req.http.test == "") {
+                       std.log("no backend");
+               }
        }
 } -start
 

```

gquintard:
I obviously strongly support this PR.

bsdphk:
OK

hermunn:
Backport review: Type conversion changes and type system update is 5.0 only unless something very special happens.

PR #2062, opened 2016-08-30
Use a switch-case for the fini sequence

gquintard:
This allows to test vgc_inistep only once and make it easier to work for
the compiler.

It doesn't seem to help gcc that's super slow, but clang get a 15% speed boost when compiling with clang on large VCLs (10000 backends).

Performance feedback is welcome.

### DIFF
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -286,7 +286,7 @@ EmitCoordinates(const struct vcc *tl, struct vsb *vsb)
 static void
 EmitInitFini(const struct vcc *tl)
 {
-	struct inifin *p;
+	struct inifin *p, *q = NULL;
 	unsigned has_event = 0;
 
 	Fh(tl, 0, "\n");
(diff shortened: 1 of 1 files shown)


Dridi:
The switch construct looks more appropriate for this kind of jump once then fall through logic :+1:

lib/libvcc/vcc_compile.c
```diff
@@ -321,16 +321,18 @@ EmitInitFini(const struct vcc *tl)
 	Fc(tl, 0, "\nstatic int\nVGC_Discard(VRT_CTX)\n{\n\n");
 
 	Fc(tl, 0, "\t(void)VGC_function_vcl_fini(ctx);\n\n");
+
+	Fc(tl, 0, "\tswitch (vgc_inistep) {\n\n");
 	VTAILQ_FOREACH_REVERSE(p, &tl->inifin, inifinhead, list) {
 		AZ(VSB_finish(p->fin));
```
nigoroll: should we assert maybe that `p->n` is always less than the previous?

lib/libvcc/vcc_compile.c
```diff
@@ -321,16 +321,18 @@ EmitInitFini(const struct vcc *tl)
...
 		if (VSB_len(p->fin)) {
-			Fc(tl, 0, "\t/* %u */\n", p->n);
```
nigoroll: should we assert maybe that p->n is always less than the previous?

nigoroll:
+1 except for the tiny suggestion I made in the diff

bsdphk:
OK with/without nigoroll's suggestion

gquintard:
closed via f70bcbc.

hermunn:
Backport review: I guess @gquintard can backport this if he gets a lot of free time on his hands, but right now it is not a priority.

Issue #2063, opened 2016-08-30
no ctx->ws during init


nigoroll:
## Expected Behavior
simple things like
`vcl_init { std.log(some + thing); }`
should work
## Current Behavior
... but don't because there is no `ctx->ws`
## Possible Solution
init a workspace

Dridi:
Possible alternative solution: VIP4

https://github.com/varnishcache/varnish-cache/wiki/VIP4:-Restrict-VMOD-function-call-sites

fgsch:
Is this in master?

slimhazard:
The same thing is true of `vcl_fini`, for example that it's not possible to use `std.log()` there.

A VMOD can work around the logging restriction in a number of ways. It's always possible to use vxid 0 in `VSL()` and `VSLv()`, which require neither a workspace nor a `ctx->vsl` (and VXID 0 in init and fini kinda makes sense). Failures in `vcl_init` are best handled by setting `VCL_RET_FAIL` in `VRT_handling()`, meaning that the VCL load is caused to fail, in which case you can write an error message to `ctx->msg`, so that you don't need a WS or VSL at all.

Those are ways of handling logging, but of course that doesn't solve any other problems for which you might need a WS in init and fini (although in my experience, logging is the only problem I've ever run into).

A major problem IMO is that many VMODs simply have not implemented ways to handle something like logging from within init or fini, presumably because it never occurred to the VMOD authors to do so. Most notably `std.log()` does nothing at all to handle logging within init and fini, which I think is a serious error -- you can make your Varnish crash that way. The std VMOD, of all things, should handle those situations properly.

@nigoroll, are you missing `ctx->ws` because of logging, or is there something else for which you need a workspace?

If logging is the only real use case, the solution might be not to add a WS, but rather to fix this problem in `std.log()`, and document for VMOD authors how they should handle logging in init and fini. I say that because adding a `ctx->ws` for init and fini might open cans of worms that aren't worth it, if there aren't really any other problems to solve. For example, it would probably become necessary to add a new param `workspace_initfini`, or something like that.

slimhazard:
Looking at this again, your problem wasn't so much logging as it was `some + thing`, because you want a WS to concatenate strings?

Although that could be handled as `VSLv(tag, 0, fmt, va)`, with `fmt` and `va` from the `STRING_LIST`. (It's just that `std.log()` doesn't do that.)

nigoroll:
[...]

PR #2073, opened 2016-09-05
move workspace functions to libvarnish

nigoroll:
Context: #2063 

Checking for possible ways to implement this, I came to the point that any sensible way needs
the ws code outside cache. Reality for most vmods probably is that they rely on the ws functions
anyway, so opening the interface seems the right thing to do.


nigoroll:
sorry libvarnish is not the api.

nigoroll:
needs more attention

hermunn:

PR #2075, opened 2016-09-06
Add a `std.boolean` function

joshuaspence:
Adds a `boolean` function to `vmod_std`. This function is quite similar to `std.integer`, but is used for parsing boolean values rather than integers.

### DIFF
--- a/bin/varnishtest/tests/m00025.vtc
+++ b/bin/varnishtest/tests/m00025.vtc
@@ -1,0 +1,40 @@
+varnishtest "test vmod_std.boolean()"
+
+server s1 {
+    rxreq
+    txresp
+} -start
+
+varnish v1 -vcl+backend {
+    import std;
+
+    sub vcl_deliver {
+        set resp.http.foo = std.boolean(req.http.foo, false);
+    }
+} -start
+
+client c1 {
+    txreq -hdr "Foo: true"
+    rxresp
+    expect resp.http.foo == true
+
+    txreq -hdr "Foo: TRUE"
+    rxresp
+    expect resp.http.foo == true
+
+    txreq -hdr "Foo: false"
+    rxresp
+    expect resp.http.foo == false
+
+    txreq -hdr "Foo: FALSE"
+    rxresp
+    expect resp.http.foo == false
+
+    txreq -hdr "Foo: 0"
+    rxresp
+    expect resp.http.foo == false
+
+    txreq -hdr "Foo: 1"
+    rxresp
+    expect resp.http.foo == true
+} -run
--- a/lib/libvmod_std/vmod.vcc
+++ b/lib/libvmod_std/vmod.vcc
@@ -139,6 +139,16 @@ Example
 	|	...
 	| }
 
+$Function BOOL boolean(STRING s, BOOL fallback)
+
+Description
+	Converts the string *s* to a boolean. If conversion fails,
+	*fallback* will be returned.
+Example
+	| if (std.boolean(req.http.foo, true)) {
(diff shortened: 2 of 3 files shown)


fgsch:
Do you have some RL example that can give more context? 
I cannot immediately see how this will help.

joshuaspence:
Sure. So we have some logic in our `vcl_recv` subroutine which works out whether the request belongs to an authenticated user and sets `req.http.X-Authenticated` to `true` or `false`. The backend then sets an `X-Authenticated` header on the response and we want to do something like this in `vcl_backend_response`:

```
if (bereq.http.X-Authenticated != beresp.http.X-Authenticated) {
  set beresp.uncacheable = true;
}
```

Basically, we want to put in a safeguard to ensure that Varnish doesn't cache a response that was intended for a logged-out user if Varnish thinks that the user is logged in. To do this, we want to ensure that we are flexible and allow the backend to pass `true`, `"true"` or `1` equivalently.

fgsch:
Can you just simply use regex, e.g. `"(?i)true|1"`?

joshuaspence:
Not quite, because we don't care about the exact value. Rather we care about differences in the value. That is, we _could_ do this:

```
if (beresp.http.X-Authenticated ~= "(?i)true|1") {
  set beresp.http.X-Authenticated = true;
} else {
  set beresp.http.X-Authenticated = false;
}

if (bereq.http.X-Authenticated != beresp.http.X-Authenticated) {
  set beresp.uncacheable = true;
}
```

This just seems a bit too verbose and I thought that `std.boolean` made sense given that there is already `std.integer`.

fgsch:
I wonder if we should support something like this: 

`set bereq.http.foo = beresp.http.foo ~ "foo|bar";`

That'd reduce the needed VCL to:

```
set beresp.http.X-Authenticated = beresp.http.X-Authenticated ~ "(?i)true|1";
if (bereq.http.X-Authenticated != beresp.http.X-Authenticated) {
  set beresp.uncacheable = true;
}
```

The difference I see between this and std.integer() and friends is that for the latter there is no alternative (e.g. math ops on values). For treating true/1 as boolean there is.
Others might have a different opinion.

nigoroll:
I'm in favor of the simplest solution and +1 for the PR as is
* the use case is valid
* extending VCC is a larger issue and @bsdphk bandwidth limited
* regexen are not the solution to everything and avoiding them for performance reasons is a valid case

fgsch:
Doing this is rather trivial (I have a patch) and I don't necessarily agree re performance.
Do you have numbers to back this up for this particular case?

nigoroll:
@fgsch 
* I did not intend to start a performance discussion, my statement was of general nature. In general, regexen are #1 performance killer in vcl and I hope we agree that the code presented here will be more efficient than any regexp doing the same.
* If you have a vcc patch then cool.
That's all from me, I do not have any particular interest in this topic - I personally think header presence is enough of a boolean, yet still I think this PR is sensible.

fgsch:
@joshuaspence While we consider this, in the next release you can do `set req.http.foo = req.http.bar ~ "foo|bar"` for your usecase.

nigoroll:
@joshuaspence are you happy with @fgsch suggestion or do you still think your suggestion should go into std?

nigoroll:
closed with timeout. Also relevant in this context: https://github.com/varnishcache/varnish-cache/wiki/vmod-polymorphism-%28for-type-conversions%29

hermunn:
Backport review: Closed for timeout / replaced with various changes to the type system.

Issue #2065, opened 2016-08-31
H/2 code crashes if you specify a low-ish SETTINGS_INITIAL_WINDOW_SIZE


daghf:
varnishd command line:

<pre>
  $ bin/varnishd/varnishd -a :9090 -b localhost:8080 -n /tmp/foo -d
</pre>

Request sent via nghttp command line

<pre>
  $ nghttp http://localhost:9090/ -n -v --no-dep -w 7
</pre>

where `-w 7` sets `SETTINGS_INITIAL_WINDOW_SIZE` to 127.

<pre>
Error: Child (9226) died signal=6
Error: Child (9226) Panic at: Wed, 31 Aug 2016 09:08:38 GMT
"Missing errorhandling code in h2_rxframe(), http2/cache_http2_proto.c line 514:
  Condition(h2->rxf_stream > h2->highest_stream) not true.thread = (cache-worker)
version = varnish-trunk revision e7b05d7
ident = Linux,4.6.0-1-amd64,x86_64,-jnone,-smalloc,-smalloc,-hcritbit,epoll
Backtrace:
  0x4521b0: pan_backtrace+0x20
  0x451fd3: pan_ic+0x363
  0x49af6b: h2_rxframe+0x28b
  0x49a029: h2_new_session+0x489
  0x47fe95: Pool_Work_Thread+0x675
  0x47f6d4: WRK_Thread+0x324
  0x47f37d: pool_thread+0xdd
  0x7f3a26a14464: libpthread.so.0(+0x7464) [0x7f3a26a14464]
  0x7f3a2675330d: libc.so.6(clone+0x6d) [0x7f3a2675330d]
req = 0x7f3a14cf6020 {
  vxid = 65541, transport = H2
  step = 0x0,
  req_body = R_BODY_INIT,
  err_code = 1, err_reason = (null),
  restarts = 0, esi_level = 0,
  sp = 0x7f3a1480f820 {
    fd = 26, vxid = 65541,
    t_open = 1472634518.353233,
    t_idle = 1472634518.353824,
    transport = H2 {
      streams {
        0x00000000 idle  
      }
    }
client = ::1 53890,
  },
  ws = 0x7f3a14cf61f8 {
    id = \"req\",
    {s, f, r, e} = {0x7f3a14cf7ff8, +288, (nil), +57344},
  },
  http_conn = 0x7f3a14cf6128 {
    fd = 26,
    doclose = NULL,
    ws = 0x7f3a1b1eae78,
[...]

PR #2076, opened 2016-09-06
add a workspace for calls into vcl from the cli thread

nigoroll:
first commits are preparations to required code available outside cache

Ref: #2063


Dridi:
I'm wondering whether the workspace code should move away from varnishd. I can see it moving away from the cache directory for its usage in the CLI, but that's about it.

nigoroll:
@Dridi at this point the main issue is that for code in libvarnish we shouldn't depend on code in varnishd/cache - but vcli lives in libvarnish.
On the more technical side, the debug code in cache_ws used the DSL macro for logging, which can only work within cache.

nigoroll:
faling travis job again caused by #2065

lib/libvarnish/vcli_serve.c
```diff
@@ -428,6 +440,12 @@ VCLS_AddFd(struct VCLS *cs, int fdi, int fdo, cls_cb_f *closefunc, void *priv)
 	cfd->cli->vlu = VLU_New(cfd, cls_vlu, *cs->maxlen);
 	cfd->cli->sb = VSB_new_auto();
 	cfd->cli->limit = cs->limit;
+	if (cs->ws_sz > 0) {
+		cfd->cli->ws = malloc(sizeof(*cfd->cli->ws));
```
Dridi: malloc looks wrong, shouldn't it allocate `cs->ws_sz` bytes instead?

What about reading `cache_param->workspace_client` here instead of breaking signatures all the way?

nigoroll: this is the malloc for the struct ws itself. the malloc for the workspace happens in line 447
reading the cache_param here would only be possible if the code lived in cache, but it doesn't. the code here need to work with both cache and mgt.

lib/libvarnish/vcli_serve.c
```diff
@@ -428,6 +440,12 @@ VCLS_AddFd(struct VCLS *cs, int fdi, int fdo, cls_cb_f *closefunc, void *priv)
...
+		assert(cfd->cli->ws != NULL);
+		WS_Init(cfd->cli->ws, "cli",
+		    malloc(cs->ws_sz), cs->ws_sz);
```
Dridi: The usual pattern is to embed the ws structure in the ws allocation itself (hence the question on line 444 that should be ignored).

nigoroll: I actually considered this option, so if that makes you happy, I'd be happy to change it :)

lib/libvarnish/vcli_serve.c
Dridi: I thought that `cache_param` was inherited and read by children but written by the manager on the cli.

nigoroll: look at the calls to VCLS_New to understand it.

lib/libvarnish/vcli_serve.c
Dridi: If we embed the ws everywhere we should maybe do it as part of WS_Init (or introduce a non-breaking WS_Embed function).

nigoroll: yes cool

lib/libvarnish/vcli_serve.c
Dridi: It made things even more confusing :)

* `*heritage.param = mgt_param`
* `heritage.param = VSM_common_alloc(...)`
* `cache_param = heritage.param;`

Shouldn't it be passed as a pointer like `cli_buffer` and `cli_limit` to capture changes over time?

nigoroll: Yes, ideally we should adjust the ws to changing parameters. But I wonder how much sense this makes if we don't even want a specific parameter to tune. @bsdphk 's words on irc went like "we should have a workspace but not guaranteeing a specific size".

bsdphk:
I really don't see the need for so much creative destruction.

Isn't all we have to do is to malloc a single static ws in cache_vcl.c, and plug that into ctx in vcl_{setup|failsafe}_event() ?

nigoroll:
replaced by a much simpler approach in #2079

hermunn:
Backport review: Replaced by #2079, nothing to backport here.

PR #2080, opened 2016-09-08
Do not include storage in coredumps.

lkarsten:
Avoid multi-gigabyte core dumps by not including the stevedore storage.

### DIFF
--- a/bin/varnishd/mgt/mgt_vcl.c
+++ b/bin/varnishd/mgt/mgt_vcl.c
@@ -148,6 +148,12 @@ mcf_find_no_vcl(struct cli *cli, const char *name)
 	return (1);
 }
 
+static int
+mcf_is_label(struct vclprog *vp)
+{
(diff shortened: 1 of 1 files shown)


nigoroll:
Actually I have no idea what happened to it (and whether or not I ever published it), but I still have a branch with this https://gist.github.com/nigoroll/1cd3932172ca8ac03c2f3ef909fc1cf6 hanging around

* not sure if we should make `MADV_DONTDUMP` a default. The overhead may be significant, but I have not benchmarked
* not sure if `#ifdef MADV_DONTDUMP` is portable enough
* I think we need to also undo the `DONTDUMP` upon free because malloc may re-use pages for stuff we do want to dump

lkarsten:
Yes, most definitively it should enable dumping of the memory segment on free.

Another issue is that madvise() requires a page aligned pointer, so we'll need to do some counting.

PR #2082, opened 2016-09-08
Madvise dontdump

lkarsten:
I fumbled the #2080 PR somehow, so here is an updated one that cleans up after itself.

### DIFF
--- a/bin/varnishd/storage/storage_malloc.c
+++ b/bin/varnishd/storage/storage_malloc.c
@@ -96,13 +96,6 @@ sma_alloc(const struct stevedore *st, size_t size)
 
 	p = malloc(size);
 	if (p != NULL) {
-		/* If we are on Linux, avoid gigabyte/terabyte sized core dumps by
-		 * skipping the storage data.
-		 */
-#ifdef MADV_DONTDUMP
-		madvise(p, size, MADV_DONTDUMP);
-#endif
-
 		ALLOC_OBJ(sma, SMA_MAGIC);
 		if (sma != NULL)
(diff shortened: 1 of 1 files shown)


nigoroll:
Re: #2080
* can't see the page alignment addressed
* I still think this should be a debug switch, not just enabled hardcoded

bsdphk:
Please add similar stanza for *BSD's MAP_NOCORE, while you ponder the madness of why they are different.  Then commit.

lkarsten:
Update: We've had production traffic on this in Varnish Plus, with less than optimal results.

Martin spent a while staring a crash due to excessive number of memory maps, and believe this is the reason. In hindsight this (i think) makes sense; somewhere the state of dump/nodump has to be stored.

It would be a nice functionality to have for the malloc stevedore, but unless someone wants to go digging for a way to avoid the memory map issue we should leave it unmerged.

Closing without merge.

nigoroll:
Thank you @lkarsten and @mbgrydeland for the reality check! Actually this is consistent with concerns I had (see also #2080).

delthas:
Additional reason why this might not be possible: if those objects are free()d but the original memory pool isn't returned to the system, you'll have unused pages held by the memory allocator with MADV_DODUMP, that will be dumped. Same issue as here: https://stackoverflow.com/questions/64847205/core-dump-size-is-increased-after-freeing-memory-that-was-excluded-from-the-core#comment114650525_64847205

Seems like we'll never be able to create reasonably-sized memory dumps on systems with large malloc storage backends.

PR #2085, opened 2016-09-11
Switch to globbing on the whole string

### LINKED ISSUE #2022: varnishstat -1 -f field inclusion glob doesn't allow VBE backend fields
## Expected Behavior
When trying to view individual statistics about a backend I would expect to be able to run 
`sudo varnishstat -1 -f "VBE.boot.default.happy"` OR `sudo varnishstat -1 -f VBE\.boot\.default\.happy ` and get the correct results.

This works with similar commands:
`sudo varnishstat -1  -f LCK.wstat.locks
LCK.wstat.locks          55950         0.32 Lock Operations
`

## Current Behavior
When running these commands now I get the error `-f: Wrong number of elements`.

Run varnish with multiple backends, attempt to query individual backend fields with varnishstat -f  command.

## Context
I am trying to configure zabbix autodiscovery and monitoring of individual backends. The ability to run this for each backend would make it easier.

## Your Environment

* Version used: `varnishstat (varnish-4.1.2 revision 0d7404e)` and `varnishstat (varnish-4.0.3 revision b8c4a34)`
* Operating System and version: Tested on both RHEL 7.1 and Archlinux

fgsch:
Fixes #2022

### DIFF
--- a/bin/varnishstat/varnishstat_options.h
+++ b/bin/varnishstat/varnishstat_options.h
@@ -39,13 +39,10 @@
 #define STAT_OPT_f							\
 	VOPT("f:", "[-f <glob>]", "Field inclusion glob",		\
 	    "Field inclusion glob."					\
-	    " A field glob consists of three parts, type, ident and"	\
-	    " name, where ident is optional. Each part can contain a"	\
-	    " '*' character at the end to match a prefix. Use"		\
-	    " backslash to escape characters. If the argument starts"	\
-	    " with '^' it is used as an exclusion glob. Multiple -f"	\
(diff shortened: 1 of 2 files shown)


bsdphk:
Looks OK to me.

It's about time we start to think about regression testing the utils

fgsch:
Yes, I've been thinking about this as well. 

Should we start using uppercase letters?

fgsch:
Merged in 8bb7e90f9adf8502cd0be382cb546a28f9e24390.

hermunn:
Backport review: The workaround is "good enough" for 4.1, so this will not be backported.

Issue #2089, opened 2016-09-15
should we relax DURATION/REAL and DURATION/INT casting?


nigoroll:
I have a performance ocd issue that instead of doing this

```
set beresp.ttl = std.random(beresp.ttl, * 0.9, beresp.ttl);
```

we need to do this

```
set beresp.ttl = std.duration(
    std.random(std.real(beresp.ttl, 10) * 0.9,
               std.real(beresp.ttl, 10)), 10s);
```
which
* stringifies twice just to convert from duration to real and then
* stringifies the result of std.random() just to get back to duration

I'd really be interested to hear if there are good reasons why not cast freely between DURATION and REAL/INT if the de-tour via a string is possible anyway (and will probably be what people would use)

fgsch:
There are 2 points here:

1. Convert a real to duration.
2. Convert a duration to real.

The first one involves a unit. We'll need to relax that requirement (i.e. 10 and 10s should be the same). 
In that regard, your code above will always result in 10s since the first argument doesn't have the unit.

The second should be pretty straight forward.

nigoroll:
Yes, I missed the unit in the second example, thank you. And I agree that _we should assume a lack of unit means seconds_ when parsing duration strings.

But I am not sure if I got the actual suggestion across: I meant to suggest that we should simply allow casting of `INT` and `REAL` to and from `DURATION` to avoid the de-tour via strings.

nigoroll:
for completeness, the correct version of the second exemaple should be
```
 set beresp.ttl = std.duration(
    std.random(std.real(beresp.ttl, 10) * 0.9,
	       std.real(beresp.ttl, 10)) + "s", 10s);
```

nigoroll:
Discussed during bugwash: dropped in favor of
https://github.com/varnishcache/varnish-cache/wiki/vmod-polymorphism-%28for-type-conversions%29

hermunn:
Backport review: Dropped in favor of something else (see above).

PR #2092, opened 2016-09-16
Fix for #2089

fgsch:
This addresses the 2 points from #2089.

### DIFF
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -1304,10 +1304,15 @@ vcc_expr0(struct vcc *tl, struct expr **e, vcc_type_t fmt)
 		}
 		*e = vcc_expr_edit(BOOL, "\v1\v-\n)", *e, NULL);
 	}
-	if (fmt != (*e)->fmt && (fmt == STRING || fmt == STRING_LIST)) {
+	if (fmt == (*e)->fmt)
+		return;
+	if (fmt == STRING || fmt == STRING_LIST) {
 		vcc_expr_tostring(tl, e, fmt);
 		ERRCHK(tl);
-	}
+	} else if (fmt == DURATION && ((*e)->fmt == INT || (*e)->fmt == REAL))
+		(*e)->fmt = DURATION;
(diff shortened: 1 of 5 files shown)


lib/libvcc/vcc_expr.c
```diff
@@ -1304,10 +1304,15 @@ vcc_expr0(struct vcc *tl, struct expr **e, vcc_type_t fmt)
...
-	if (fmt != (*e)->fmt && (fmt == STRING || fmt == STRING_LIST)) {
+	if (fmt == (*e)->fmt)
+		return;
+	if (fmt == STRING || fmt == STRING_LIST) {
 		vcc_expr_tostring(tl, e, fmt);
 		ERRCHK(tl);
-	}
+	} else if (fmt == DURATION && ((*e)->fmt == INT || (*e)->fmt == REAL))
+		(*e)->fmt = DURATION;
+	else if (fmt == REAL && ((*e)->fmt == DURATION || (*e)->fmt == INT))
```
nigoroll: do we need to handle `fmt == INT` also?

lib/libvcc/vcc_expr.c
```diff
@@ -146,13 +146,10 @@ vcc_DoubleVal(struct vcc *tl)
 void
 vcc_Duration(struct vcc *tl, double *d)
 {
-	double v, sc;
```
nigoroll: Yes we should really drop the unit requirement, but accepting numbers as durations makes the syntax ambiguous. What if we kept the durations as-are and just folded `INT` into `DURATION`? IIUC this should be covered by your patch already and should achieve the same?

fgsch: Not sure I follow. Mind sharing some VCL where this is visible?
All examples I can think of require dropping the unit, and in any case we should be consistent.

lib/libvcc/vcc_expr.c
fgsch: We could but first we need to define how we do the conversion from DURATION/REAL to INT (trunc or round). I'd rather handle this in a separate PR.

fgsch:
#2089 is no more.

hermunn:
Backport review: Nothing committed - nothing to backport.

Issue #1865, opened 2016-03-04
Panic accessing beresp.backend.ip in vcl_backend_error{}


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1865

bsdphk:
This is same issue as #1933 where we will track this issue going forward.

fgsch:
This is not related to #1933. Please see testcase below:

```
varnishtest "Accessing beresp.backend.ip does not cause a panic"
 
varnish v1 -vcl {
        backend default {
                .host = "${bad_ip}";
        }
        sub vcl_backend_error {
                set beresp.http.foo = beresp.backend.ip;
        }
} -start
 
client c1 {
        txreq
        rxresp
} -run
```

hermunn:
Backport review: Backported as 0c4c164.

PR #2098, opened 2016-09-22
retire beresp.backend.name

nigoroll:


Dridi:
This looks like a good idea especially after 6487059.

Dridi:
Can we add a "breaking" label for github issues to better keep track of such changes?

nigoroll:
done

fgsch:
I'm not sure I like this. It deviates from the existing idioms and breaks the uniformity of the interface.

I rather change `beresp.backend` to return the director (or not being accessible at all) and keep `beresp.backend.name` for the backend name.

nigoroll:
> I rather change beresp.backend to return the director (or not being accessible at all) and keep beresp.backend.name for the backend name.

`BACKEND` is a VCL type, we use it in vmods and I'd foresee much more interesting use cases than we have today (for example, see discussion on the scope of dynamic `BACKEND` objects).
Right now we have a `BACKEND`->`STRING` cast. If we wanted to retire this (and, consequently, barf a vcc error for use of `BACKEND` in `STRING` context) , that functionality would need to be wrapped in some (vmod) function. `set bereq.backend = vmod.backend()` just to get to `bereq.backend.name` is not a viable option.

So I'd argue that `beresp.backend.*` are not consistent with the rest of the interface and `beresp.backend.ip` should also go. Whether we replace both `.name` and `.ip` by vmod functions or just `.ip` and keep the casting does not matter _that_ much, IMHO, but I think the sensible casts really help making VCL cleaner, so I have a clear preference for the cast. My impression is that this would also be in line with other recent changes - see for example #2089 #2060 #2040

Dridi:
> [...] Whether we replace both `.name` and `.ip` by vmod functions or just `.ip` and keep the casting does not matter that much [...]

I'm not sure `.ip` still makes sense now that we can have non-HTTP/1-over-TCP/IP-backends since Varnish 4.1. How does it even behave on a director? Does it force a backend resolution?

nigoroll:
Good point, @Dridi . For a lazy-resolve director, `.ip` does not make sense unless we require all directors to consistently resolve to the same backend no matter when being asked

fgsch:
If we retire, or move .ip to std (or somewhere else) as well I'm convinced.
Removing .name but leaving .ip would leave us with a non uniform interface IMO.

fgsch:
Touches on #1865

nigoroll:
close until rework done

hermunn:
Backport review: "Close until rework done". I guess this will happen in a different pull request, or @nigoroll will reapply the backport label if something is accepted here.

PR #2100, opened 2016-09-23
RFC: add va_args interface to varargisch vrt api functions

nigoroll:
currently, vmods cannot use the varargish VRT functions for own varargs because we do not yet have the `va_list` variants.

RFC, some thoughts about this:
- this would be the first `#include` in `vrt.h`. I understand we want to avoid creating additional dependencies, but we require the C compiler, so can't we also require some basic system includes?
  - if we wanted to avoid this, starting another `vrt_vmod.h` would be an option
- we'd not need to wrap all the `VRT_l_*_*()` functions, we could also export `vrt_do_string`. For now I think that the first option may be POLA  though
  comments?

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -199,6 +199,8 @@ VRT_String(struct ws *ws, const char *h, const char *p, va_list ap)
 
 /*--------------------------------------------------------------------
  * Copy and merge a STRING_LIST on the current workspace
+ *
+ * no va_list variant because we're merely wrapping VRT_String
  */
 
 const char *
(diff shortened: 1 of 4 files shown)


Dridi:
I think the goal is to have a self-contained VRT single-C-file for VCL compiling, so I'd go for the `vrt_vmod.h` alternative.

nigoroll:
irc discussion: @bsdphk proposed a new VRT STRING_LIST type to replace both the varargish and va interfaces. To transition, the current vrt interface could wrap to the new interfaces and with vrt 6.0 we could change (break) the interface.

hermunn:

PR #2101, opened 2016-09-23
WIP/RFC: relaxed INT/REAL/DURATION parsing/casting

nigoroll:
based on previous work by @fgsch in #2092 and intended to possibly supersede it:

This patch is intended as a prototype of what I had in mind when I opened #2089 

Before I spend more time on it (in particular tests would need more coverage), I'd like to poll for opinions. This may appear a bit radical, but I think it takes a lot of unnecessary trouble out of VCL coding. Yes, there may be cases when truncating real to int may be unexpected and there will be cases when casting from/to duration will be unintentional, but having to call `std.typeconversion()` for simple cases makes VCL clumsy.
And the current workaround of type-casting by stringification is just plain stupid, IMHO (see example in #2089)

### DIFF
--- a/bin/varnishtest/tests/v00020.vtc
+++ b/bin/varnishtest/tests/v00020.vtc
@@ -22,17 +22,84 @@ varnish v1 -errvcl {Found: '0' at} { 0; }
 
 varnish v1 -errvcl {Expected an action, 'if', } " sub vcl_recv { { } { "
 
-varnish v1 -vcl {
-	backend b { .host = "127.0.0.1"; }
+server s1 {
+	rxreq
+	txresp
+} -start
+
+varnish v1 -vcl+backend {
 	sub vcl_backend_response {
 		set beresp.ttl = 1w;
+		set beresp.http.ttl1a = beresp.ttl;
 		set beresp.ttl *= 1.5;
+		set beresp.http.ttl1b = beresp.ttl;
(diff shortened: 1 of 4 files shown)


fgsch:
Besides the truncation, how is this different to #2092?

fgsch:
There are also a number of issues with this PR:

- `now + 1` yields `Sat, 24 Sep 2016 19:33:22 GMT1`
- `1 + 1s` yields `11.000`
- `1s + 1` yields `1.0001`
- `1s + now` yields `1.000Sat, 24 Sep 2016 19:33:22 GMT`
- `1s + "foo"` yields `1.000foo`
- `"foo" + 1s` yields `foo1.000`

DURATION + STRING and STRING + DURATION should continue failing.
INT + DURATION should either fail or return the sum, not sure if as INT or DURATION.
DURATION + TIME should continue failing.
TIME + INT should either fail or return the sum as TIME.

nigoroll:
@fgsch Thank you for the review! I don't get the same results as you but you've raised some points which need to be answered.

nigoroll:
Discussed during bugwash: dropped in favor of
https://github.com/varnishcache/varnish-cache/wiki/vmod-polymorphism-%28for-type-conversions%29

hermunn:
Backport review: Dropped in favor of something else (see above) - nothing to backport.

nigoroll:
continued in #2459

PR #2102, opened 2016-09-25
More changes to VCC

fgsch:
- Try to distinguish between DURATION, REAL and INT automatically.
- Explicitly mention which operands are possible when adding.

Related to #2101.

### DIFF
--- a/bin/varnishtest/tests/v00020.vtc
+++ b/bin/varnishtest/tests/v00020.vtc
@@ -134,8 +134,7 @@ varnish v1 -errvcl {TIME + TIME not possible.} {
 	}
 }
 
-# XXX: error message should say something about DURATION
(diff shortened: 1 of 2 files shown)


nigoroll:
@fgsch it appears to me that this makes sense even with https://github.com/varnishcache/varnish-cache/wiki/vmod-polymorphism-%28for-type-conversions%29 ?

bsdphk:
This looks OK to me.  Sorry for not saying so sooner.

fgsch:
Merged in 6f9595e6563cf53118048bcb5d1591a6b631b84c

hermunn:
Backport review: This is part of the big reorganization of operations in VCC, and not for 4.1.

PR #2108, opened 2016-10-04
Add optional storage parameter to cache_req_body.

ehocdet:
Cache big req body should not eat the Transient storage.
The use case is to store the body on a dedicated sfile when CL is big.

This PR add an optional VCL_STEVEDORE  parameter to cache_req_body.

Related to: https://github.com/varnishcache/varnish-cache/pull/1914

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -935,7 +935,7 @@ void Req_Fail(struct req *req, enum sess_close reason);
 
 /* cache_req_body.c */
 int VRB_Ignore(struct req *);
-ssize_t VRB_Cache(struct req *, ssize_t maxsize);
+ssize_t VRB_Cache(struct req *, ssize_t maxsize, const char *hint);
 ssize_t VRB_Iterate(struct req *, objiterate_f *func, void *priv);
 void VRB_Free(struct req *);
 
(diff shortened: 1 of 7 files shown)


fgsch:
I don't believe this is the right way to do it.
It might make more sense if we can set the hint following the existing semantics, e.g. `set req.storage_hint = "Transient";` before calling `std.cache_req_body()`.

fgsch:
I just saw #1914 and read the comment from @bsdphk stating his preference over a hint.

I wrote the patch in #2111 before seeing that and I still like the idea of req.backend_hint so decided to submit it anyway. YMMV.

ehocdet:
@fgsch I think you have the right way.
perhaps it's time to use VCL_STEVEDORE  and have `set req.storage_hint  = storage.s0;`
But beresp.storage_hint should also use VCL_STEVEDORE for consistency.

fgsch:
That might be a good idea.

We could resolve the storage at compilation time and avoid the problem of
unknown storages.

fgsch:
I'm closing this in favour of the `set req.storage = storage.s0;` implementation I'm working on.
Feel free to open it if you disagree.

hermunn:
Backport review: Replaced by #2114 - nothing to backport here.

PR #2109, opened 2016-10-04
Killing the child proces properly

hermunn:
When debugging #2010 we came to the conclusion that the mgt proces did not have rights to kill the child. The first patch, "Add error checking to confirm bug", was made, and the conclusion was confirmed by @thlc. The next two commits should fix the problem, but I need @nigoroll to look at the case for Sun OS derivates, and I welcome comments, as always.

### DIFF
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -455,10 +455,17 @@ mgt_launch_child(struct cli *cli)
 
 static int
 kill_child(void) {
+	int i, error;
+
+	VJ_master(JAIL_MASTER_KILL);
 	if (MGT_FEATURE(FEATURE_NO_COREDUMP))
-		return (kill(child_pid, SIGKILL));
+		i = kill(child_pid, SIGKILL);
 	else
-		return (kill(child_pid, SIGQUIT));
+		i = kill(child_pid, SIGQUIT);
+	error = errno;
+	VJ_master(JAIL_MASTER_LOW);
+	errno = error;
+	return (i);
 }
 
 static void
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -34,6 +34,7 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#include <errno.h>
 #include <fcntl.h>
 #include <poll.h>
 #include <signal.h>
@@ -583,17 +584,22 @@ mgt_reap_child(void)
 void
 MGT_Child_Cli_Fail(void)
 {
+	int i;
 
 	if (child_state != CH_RUNNING)
 		return;
 	if (child_pid < 0)
 		return;
-	MGT_complain(C_ERR, "Child (%jd) not responding to CLI, killing it.",
-	    (intmax_t)child_pid);
 	if (MGT_FEATURE(FEATURE_NO_COREDUMP))
-		(void)kill(child_pid, SIGKILL);
+		i = kill(child_pid, SIGKILL);
 	else
-		(void)kill(child_pid, SIGQUIT);
+		i = kill(child_pid, SIGQUIT);
+	if (i == 0)
+		MGT_complain(C_ERR, "Child (%jd) not responding to CLI,"
+		    " killing it.", (intmax_t)child_pid);
+	else
+		MGT_complain(C_ERR, "Failed to kill child with PID %jd: %s",
+		    (intmax_t)child_pid, strerror(errno));
 }
 
 /*=====================================================================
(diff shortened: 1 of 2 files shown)


bin/varnishd/mgt/mgt_child.c
```diff
@@ -588,12 +599,26 @@ MGT_Child_Cli_Fail(void)
...
+			break;
+		case ESRCH:
+			MGT_complain(C_ERR, "Child (%jd) not responding to CLI."
+			    " Tried to kill it, but PID did not exist.",
+			    (intmax_t)child_pid);
+			break;
+		default:
+			WRONG("Got wrong error from kill()");
+		}
+	}
```
hermunn: This is an old commit - I was asked to remove verbosity and in the latest version this code is different.

bin/varnishd/mgt/mgt_child.c
```diff
@@ -588,12 +599,26 @@ MGT_Child_Cli_Fail(void)
...
 		return;
-	MGT_complain(C_ERR, "Child (%jd) not responding to CLI, killing it.",
-	    (intmax_t)child_pid);
-	if (MGT_FEATURE(FEATURE_NO_COREDUMP))
-		(void)kill(child_pid, SIGKILL);
-	else
-		(void)kill(child_pid, SIGQUIT);
+	if (kill_child() == 0)
+		MGT_complain(C_ERR, "Child (%jd) not responding to CLI,"
+		    " killing it.", (intmax_t)child_pid);
```
hermunn: I moved the complain under the kill so that we would know if we ran into an error, and the be able to write something slightly different on error. The wording was kept, but I guess it would be better to say "killed".

bin/varnishd/mgt/mgt_child.c
fgsch: Yes, this was a comment for the previous iterations.

bin/varnishd/mgt/mgt_child.c
```diff
@@ -588,12 +601,12 @@ MGT_Child_Cli_Fail(void)
 		return;
 	if (child_pid < 0)
 		return;
-	MGT_complain(C_ERR, "Child (%jd) not responding to CLI, killing it.",
-	    (intmax_t)child_pid);
-	if (MGT_FEATURE(FEATURE_NO_COREDUMP))
-		(void)kill(child_pid, SIGKILL);
+	if (kill_child() == 0)
+		MGT_complain(C_ERR, "Child (%jd) not responding to CLI,"
+		    " killing it.", (intmax_t)child_pid);
```
hermunn: Sure, change into "killed" as above.

bin/varnishd/mgt/mgt_child.c
```diff
@@ -452,6 +453,21 @@ mgt_launch_child(struct cli *cli)
...
+
+	VJ_master(JAIL_MASTER_KILL);
+	if (MGT_FEATURE(FEATURE_NO_COREDUMP))
+		i = kill(child_pid, SIGKILL);
+	else
+		i = kill(child_pid, SIGQUIT);
+	error = errno;
+	VJ_master(JAIL_MASTER_LOW);
+	errno = error;
+	return (i);
```
hermunn: I prefer that kill_child() mimic the behavior of kill(), but YMMV.

fgsch: I'm not sure to understand how this changes the behavior of kill() but OK.

hermunn: This is documented in #2010 and diagnosed (with verification) by martin. Through the jail system the mgt process decreases its rights, and becomes unable to kill the child process. That is why we need to elevate temporarily in this patch.

bsdphk:
OK

bin/varnishd/mgt/mgt_child.c
```diff
@@ -603,7 +603,7 @@ MGT_Child_Cli_Fail(void)
 		return;
 	if (kill_child() == 0)
 		MGT_complain(C_ERR, "Child (%jd) not responding to CLI,"
-		    " killing it.", (intmax_t)child_pid);
+		    " killed it.", (intmax_t)child_pid);
```
Dridi: Do we need an extra commit to fix a message introduced by the same pull request? I would happily squash this change in efcb76f where the message first appeared.

hermunn: Sorry, you were a tad too late, Dridi. But if the end result (ie. the code) is ok, then I am happy.

hermunn:
Commited as 0f291ee97dd6a c189a22703fc6 c41c30aa7f7dfb df4ec11dc46ad4d (oldest to newest).

hermunn:
Backport review: Backported as c6efe045ff f9526a52c5 61cdc9d8416a 2c4ceead04 (oldest to newest).

PR #2111, opened 2016-10-06
Add req.storage_hint

fgsch:
This hints Varnish which storage backend it should use when dealing with the request body (if any).

Supersedes to #2108. Fixes #1914.


bsdphk:
Stevedores are objects in VCL, it doesn't have to be a hint anymore.

The question is if it should still have "hint" semantics though

fgsch:
This will be replaced with the code I'm currently working on. Closing.

hermunn:
Backport review: Will be replaced by upcomming commits or pull requests. Nothing to backport here.

PR #2113, opened 2016-10-13
Ban to vmod and error msg

nigoroll:
* move the ban function to vmod std
* return the error string or the empty string if no error


fgsch:
If we ran out of workspace we won't log anything at all. This is different to the existing behaviour.

nigoroll:
> If we ran out of workspace we won't log anything at all. This is different to the existing behaviour.
Not for those cases where `err` is a fixed string, but otherwise yes.
I don't think this is relevant, but if you think it is, we can buy this by complicating the code a bit

fgsch:
That's one concern. The other is that by moving the code to std people won't be able to ban from other VMODs. 
I reckon we should leave the code in core and either add a new argument or split the function into 2.
I believe this is how we have done it in the past (e.g. rollback).

nigoroll:
@fgsch, good point. vmods should be able to issue bans.
So should we keep `VRT_ban_string` in core with a changed signature (returning the error or NULL) and still move the VCL `ban()` function to _std_ in order to get rid of the extra vcc code?

lkarsten:
Why do you want this to be in a vmod and not in core?

fgsch:
@lkarsten: 

@nigoroll mentioned it on irc. The idea is to be able to obtain the error if the ban failed for whatever reason and communicate it back to the requester via synth.

lkarsten:
I see. I would hope that we could keep VCL compatibility through such a change.

fgsch:
I'd prefer not to break backward compatibility either for point releases.
Maybe have them back-to-back until the next major release?

nigoroll:
@lkarsten @fgsch thank you for your feedback. I'm back now after a one-week break, hope to get back to this soon-ish

nigoroll:
I'll close this one and open a new PR when ready

hermunn:

nigoroll:
continued in #3486

PR #2114, opened 2016-10-13
Add beresp.storage

fgsch:
This supersedes beresp.storage_hint but is kept around for backward
compatibility until the next major release.

### DIFF
--- a/bin/varnishd/cache/cache_vrt_var.c
+++ b/bin/varnishd/cache/cache_vrt_var.c
@@ -333,6 +333,7 @@ VRT_r_beresp_storage_hint(VRT_CTX)
 void
 VRT_l_beresp_storage_hint(VRT_CTX, const char *str, ...)
 {
+	const struct stevedore *stv;
 	va_list ap;
 	const char *b;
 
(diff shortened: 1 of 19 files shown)


bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -347,6 +348,30 @@ VRT_l_beresp_storage_hint(VRT_CTX, const char *str, ...)
...
+}
+
+/*--------------------------------------------------------------------*/
+
+VCL_STEVEDORE
+VRT_r_beresp_storage(VRT_CTX)
+{
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(ctx->bo, BUSYOBJ_MAGIC);
+	if (ctx->bo->storage != NULL)
```
bsdphk: Why the if-then ?  seems pointless...

fgsch: Indeed. WIll correct.

bsdphk:
I've commented on a seemingly pointless if/then in there.

I would also like if one of the testcases did a successful set to storage_hint, so we still exercise that code while we have it.

fgsch:
> I would also like if one of the testcases did a successful set to storage_hint, so we still exercise that code while we have it.

c00078.vtc sets storage_hint, for to an invalid value and then to a valid one.

bin/varnishd/cache/cache_vrt_var.c
fgsch: Done.

bsdphk:
OK

fgsch:
Merged.

hermunn:
Backport review: This will not be backported to 4.1 since it is very much in maintenance mode.

PR #2122, opened 2016-10-28
Master r01874

mbgrydeland:
This pull request adds a runtime parameter to control how much clock step we tolerate before panicking. Also changes the session timestamps checks to take this into account.

Ref: #1874


hermunn:
Backport review: Backported as c855794616fd2, 5ed4b4bee3e2f and 4f58bb832a65d.

PR #2124, opened 2016-11-01
WIP mcf auto-limiting by min/max

nigoroll:
Limit parameter values by the min/max
open questions:
* Do we want the "this parameter has been limited" warning andwhere?
* I am not particularly fond of the recursion cut off using the {{_LIMITING}} flag - any better ideas?

more todos:
* normalize min/max values not just when washing, but always (for those adjusted automatically)
* check for (more) races between interlocked params as in ee4ae94cb53a481afdc0222b7a17feec9155df37
* we leak a bit of memory every time we change a min/max - maybe just replace all compiled-in defaults once with strdups and free every time?

### DIFF
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -443,12 +443,12 @@ init_params(struct cli *cli)
 #endif
 
 	low = sysconf(_SC_THREAD_STACK_MIN);
-	MCF_ParamConf(MCF_MINIMUM, "thread_pool_stack", "%jd", (intmax_t)low);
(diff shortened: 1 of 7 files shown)


nigoroll:
Oh and additional question: is `__match_proto__()` correct?

nigoroll:
brief summary of irc discussion
@bsdphk:
> I think I can say definitively that I don't want to add more "DWIM magic", and if anything, maybe even simplify what is there today.

nigoroll:
I've cherry-picked two minor cleanup commits from this PR, but otherwise will close it until getting back to it.
There is no important TODO currently, this work was motivated by @bsdphk asking if we should auto-limit, so now, for the time being, the answer is now

hermunn:

PR #2132, opened 2016-11-08
Avoid copying in VRT_String if we can

nigoroll:
If we neither need to prepend a header, nor need to concatinate and
the (single non-NULL non-empty) string is already on the correct
workspace, just return it rather than copying.

For the empty string, a const "" not from the workspace is returned.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -252,7 +252,30 @@ const char *
 VRT_String(struct ws *ws, const char *h, const char *p, va_list ap)
 {
 	char *b, *e;
+	const char *q;
 	unsigned u, x;
+	va_list aq;
+
+	while (p == NULL ||
+	       (p != vrt_magic_string_end && *p == '\0'))
+		p = va_arg(ap, const char *);
+
+	if (h == NULL) {
+		q = p;
+
+		if (q == vrt_magic_string_end)
+			p = "";
+		else if (WS_Inside(ws, q, NULL)) {
+			va_copy(aq, ap);
+			do
+				q = va_arg(aq, const char *);
+			while (q == NULL ||
+			       (q != vrt_magic_string_end && *q == '\0'));
+			va_end(aq);
+		}
+		if (q == vrt_magic_string_end)
+			return (p);
+	}
 
 	u = WS_Reserve(ws, 0);
 	e = b = ws->f;
--- a/bin/varnishtest/tests/v00053.vtc
+++ b/bin/varnishtest/tests/v00053.vtc
@@ -1,0 +1,97 @@
+varnishtest "test that strings are only copied when necessary"
+
+server s1 {
+	rxreq
+	txresp -hdr "Connection: close" -body "012345\n"
+} -start
+
+varnish v1 -vcl+backend {
+	import std;
+	import debug;
+
+	sub vcl_recv {
+	    # we try to make this work for 32 and 64 bit by having
+	    # all strings round up to a multiple of 8
+	    # 6 bytes: stringifying the 5-digit free integer
+	    # 14 bytes: wssnap: 12345 + null
+	    # we exploit the implementation detail that std.log does
+	    # not keep a reference to the passed string
+	    set req.http.wssnap = debug.workspace_free(client) - 24;
+	    debug.workspace_snap(client);
+	    std.log(std.integer(req.http.wssnap, 0) - debug.workspace_free(client));
+	    debug.workspace_reset(client);
+
+	    set req.url = req.url;
+	    debug.workspace_snap(client);
+	    std.log(std.integer(req.http.wssnap, 0) - debug.workspace_free(client));
+	    debug.workspace_reset(client);
+
+	    set req.url = "" + "" + req.url;
+	    debug.workspace_snap(client);
+	    std.log(std.integer(req.http.wssnap, 0) - debug.workspace_free(client));
+	    debug.workspace_reset(client);
+
+	    set req.url = "" + "" + req.url + "" + "";
+	    debug.workspace_snap(client);
+	    std.log(std.integer(req.http.wssnap, 0) - debug.workspace_free(client));
+	    debug.workspace_reset(client);
+
+	    set req.url = "/" + "new" + "/st";
+	    debug.workspace_snap(client);
+	    std.log(std.integer(req.http.wssnap, 0) - debug.workspace_free(client));
+	    debug.workspace_reset(client);
+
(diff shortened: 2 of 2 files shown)


fgsch:
This is less useful than one might think it is. Headers will not use it.

So short of .url and perhaps .method and CollectString(), .proto and .reason won't see any difference. Neither storage_hint.

Giving the rather small benefits I have my reservations.

nigoroll:
@fgsch Yes, we can't do anything about headers.
Besides the cases you have already mentioned, vmods have the most potential, I think.
For anything nesting `STRING_LIST` arguments this has the potential to make a significant difference.

fgsch:
> For anything nesting STRING_LIST arguments this may make a significant difference.
View changes

Care to elaborate? Do you have any RL example?

nigoroll:
@fgsch sure:

The motivation to look into this was a vmod we are working on which _may_ rewrite the url. This will look something like this
```
sub vcl_init {
    new validator = validator_vmod.init();
}
sub vcl_recv {
   if (validator.ok(...)) {
        set req.url = validator.url();
   }
}
```
To save workspace with the existing VRT code, we considered the alternative to have `validator.ok` implicitly update `req.url`, but that's un-POLA. With the suggested change, we could have the above nicer interface at no (workspace) cost.

Most vmods use `STRING` arguments, and if the actual argument is `STRING_LIST`, vcc will generate a `VRT_CollectString()` call which calls `VRT_String()` with argument `h = NULL`. So for any `vmod.stringfunc(something + another.func(...))`, the suggested change will save workspace if either `STRING_LIST` component is `NULL` or `""`.

For some cases, VCC still needs STRING hints in the form of `"" + something`. For example (`some_re` being a vmod_re object):
`if (some_re.backref(1, "") != client.ip)`
does not work, one workaround is
`if (some_re.backref(1, "") != "" + client.ip)`
I hope that we will solve this in the VCC type system eventually, but for now it is another case of unnecessary copying.

I am aware of my workspace saving fetish and the usefulness of optimizations can always be easily questioned. But I think that its efficiency is one of the unique features of Varnish and my understanding is that we can only keep up efficiency by continuing to care about the details.

slimhazard:
This is a "side" comment that doesn't speak to the main issue -- I've rolled my own `WS_Contains` for a number of VMODs and would love to see it get into cache_ws.c. But then could we please have:
```
/* Assert that a region of memory lies within a workspace */
static inline void
WS_Contains(struct ws * const restrict ws, const void * const restrict ptr,
            const size_t len)
{
        assert((char *)ptr >= ws->s && (char *)(ptr + len) <= ws->e);
}
```
Then if you just want to check that a pointer lies within the workspace:
```
WS_Contains(ws, ptr, 0);
```

fgsch:
Here's an improved patch rebased against master:

```
diff --git a/bin/varnishd/cache/cache_vrt.c b/bin/varnishd/cache/cache_vrt.c
index c69914c4b..debf66442 100644
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -185,14 +185,33 @@ VRT_StringList(char *d, unsigned dl, const char *p, va_list ap)
 }

 /*--------------------------------------------------------------------
- * Copy and merge a STRING_LIST into a workspace.
+ * Copy and merge a STRING_LIST into a workspace if needed.
  */

 const char *
 VRT_String(struct ws *ws, const char *h, const char *p, va_list ap)
 {
        char *b, *e;
+       const char *q;
        unsigned u, x;
+       va_list aq;
+
+       if (h == NULL) {
+               while (p == NULL || *p == '\0')
+                       p = va_arg(ap, const char *);
+               q = p;
+               if (q == vrt_magic_string_end)
+                       p = "";
+               else if (WS_Inside(ws, q, q + strlen(q))) {
+                       va_copy(aq, ap);
+                       do {
+                               q = va_arg(aq, const char *);
+                       } while (q == NULL || *q == '\0');
+                       va_end(aq);
+               }
+               if (q == vrt_magic_string_end)
+                       return (p);
+       }

        u = WS_Reserve(ws, 0);
        e = b = ws->f;
```

I've been pondering about this and I think we should at the very least skip any trailing NULs and empty strings. Marking it for next bugwash.

nigoroll:
@fgsch thank you for your feedback and suggestion. I have force-pushed an update.
* yes, now that we got `WS_Inside`, we should use it
* good suggestion to save one variable
* I prefer reducing the scope of local variables, but ok

on bits from my original PR which I've kept:
* skipping over initial arguments is not important for the `h != NULL` case, but I think it helps clarity
* We need to check for vrt_magic_string_end because it does not get init'ed and thus `*vrt_magic_string_end == '\0'` may hold true.

fgsch:
> skipping over initial arguments is not important for the h != NULL case, but I think it helps clarity

I disagree. VRT_StringList handles the leading empty strings so making it specific for h == NULL makes the code clearer.

> We need to check for vrt_magic_string_end because it does not get init'ed and thus *vrt_magic_string_end == '\0' may hold true.

If that's the case I'd prefer to follow existing code, e.g:

```
diff --git a/bin/varnishd/cache/cache_vrt.c b/bin/varnishd/cache/cache_vrt.c
index c69914c4b..1f8bcb028 100644
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -192,7 +192,28 @@ const char *
 VRT_String(struct ws *ws, const char *h, const char *p, va_list ap)
 {
        char *b, *e;
+       const char *q;
        unsigned u, x;
+       va_list aq;
+
+       if (h == NULL) {
+               while (p != vrt_magic_string_end &&
+                   (p == NULL || *p == '\0'))
+                       p = va_arg(ap, const char *);
+               q = p;
+               if (q == vrt_magic_string_end)
+                       p = "";
+               else if (WS_Inside(ws, q, q + strlen(q))) {
+                       va_copy(aq, ap);
+                       do {
+                               q = va_arg(aq, const char *);
+                       } while (q != vrt_magic_string_end &&
+                           (q == NULL || *q == '\0'));
+                       va_end(aq);
+               }
+               if (q == vrt_magic_string_end)
+                       return (p);
+       }

        u = WS_Reserve(ws, 0);
        e = b = ws->f;
```

fgsch:
Upon further inspection this optimisation relies on UB.

nigoroll:
* Moved the UB discussion to #2415
* `WS_Inside(ws, q, NULL)` should also suffice

nigoroll:
force-pushed an update

nigoroll:
force-pushed rebase

Issue #2126, opened 2016-11-03
Varnish stuck on VBT_Wait after reused connection (VBC_STATE_STOLEN) failed V1F_SendReq


jpastuszek:
When POST request is canceled by the browser (connection is closed) before content length of data is received by Varnish and the backend connection was reused Varnish will get stuck for backend_idle_timeout .

## Expected Behavior
Varnish should close backend connection and send 400 error response and close client connection.

## Current Behavior
Varnish gets stuck for the backend_idle_timeout since the connection was put for reuse by previous request. When timeout runs out it closes backend connection and sends 503 to the client and closes client connection.

## Possible Solution
The bug is that when we have VBC_STATE_STOLEN connection and V1F_SendReq fails (due to "straight insufficient bytes") Varnish will still call VBT_Wait.
The VBT_Wait blocks until backend_idle_timeout runs out.

In case of fresh connection (VBC_STATE_USED) Varnish will not block and will respond with 503 (should be 400 in this case). The VBT_Wait (VBC_STATE_USED) and V1F_FetchRespHdr (i = -1) calls are skipped all together.

This looks similar to #1806 and probably relates to #1772.

As mentioned in other tickets Varnish should be able to abort backend connection when it is in the waiter or perhaps take it away somehow without the need to wait for full timeout.

## Steps to Reproduce
Make a request - this should put backend connection to reuse.
Make POST request but close the connection (write side is enough) before writing full Content-Length body bytes.
Varnish should get stuck until backend_idle_timeout triggers after it logged "straight insufficient bytes".

```
     98909 Debug          c "geoip2.lookup: No entry for this IP address (192.168.1.115)"
     98910 BackendOpen    b 29 boot.origin 10.1.1.72 80 10.1.1.248 33748
     98910 BackendStart   b 10.1.1.72 80
     98910 BackendReuse   b 29 boot.origin
     98909 Debug          c "RES_MODE 48"

    164271 Debug          c "geoip2.lookup: No entry for this IP address (192.168.1.115)"
    164272 BackendOpen    b 29 boot.origin 10.1.1.72 80 10.1.1.248 33748
    164271 FetchError     c straight insufficient bytes
    164272 BackendStart   b 10.1.1.72 80
    164272 FetchError     b req.body read error: 0 (Success)
    164272 FetchError     b backend write error: 0 (Success)
--- STUCK ---
    164272 BackendClose   b 29 boot.origin
    164272 BerespReason   b Backend fetch failed
[...]

Issue #1772, opened 2016-03-04
first_byte_timeout is ignored when re-using a backend connection (HTTP Keep Alive)


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1772

mbgrydeland:
I've been pondering this and #1806 a bit.

To me it seems that the base problem relates to the inability to forcefully remove a vbc from the waiter code. We used to have that option in the API, but it proved racy and was removed at some point during the 4.1 development cycle.

To fix the race issue, reused vbc's would start off in a STOLEN state. STOLEN here meaning that the waiter has the fd, and has not yet released it from it's in kernel event monitoring. The VBT_Wait function would wait for the waiter to release it, and the fetch code would take care to call VBT_Wait before starting to read or poll on the socket. But for the waiter release to happen, some event would need to register on the fd. The event could either be incoming bytes, a remote hangup or the waiter implemented timeout. The waiter timeout set for vbc's is backend_idle_timeout (default 60s). Lacking any incoming bytes or hangup, this translates to the first_byte_timeout not being used, but instead a timeout up to backend_idle_timeout is used.

To remedy this I suggest taking a slightly different approach. Instead of forcing a waiter exit first and then doing the reads on the vbc, we should use the waiter to do the first read timeout instead. To do this, VBT_Wait would go away, and be replaced with a VBT_Read taking a read timeout as an argument and is tasked with both reading bytes and implementing the read timeout. Fetch code would then call VBT_Read to read bytes, providing first_byte_timeout on the first call, and between_bytes_timeout on subsequent calls.

VBT_Read would upon finding that the vbc is still on the waiter, call into the waiter API to change the timeout. This would require changes to the waiter API, but looks doable (would need to trigger a loop in the waiter thread through the control socket like when inserting new entries). VBT_Read then does a pthread_cond_wait like before waiting for the waiter event. The waiter event would remove the fd from its set, so this slower path would only happen once. When the vbc is not on the waiter, a poll with timeout is done instead.

The waiter API for changing the timeout would have to be able to return failure. This could happen if the call to change the timeout happens at the same time as a socket event happens. The calling code in VBT_Read would need to handle that.

It is my belief that these changes would also fix #1806.

Martin

bsdphk:
Yeah, I've had thoughts along the same lines myself, but it's more complexity than I like...

fgaillot:
Hi,

[...]

PR #2135, opened 2016-11-15
Introduce a parameter value for extrachance retries

ema:
There seem to be conditions in the extrachance logic under which Varnish
could retry a backend request indefinitely. This commit introduces a new
parameter, gethdr_extrachance, allowing to set an upper bound for the
amount of retries in vbe_dir_gethdrs.

The default value for the newly introduced parameter is 1, setting it to
0 disables the feature entirely.

The issue we've observed in production is described here: https://phabricator.wikimedia.org/T150247

Potentially related bugs: https://github.com/varnishcache/varnish-cache/issues/2126 https://github.com/varnishcache/varnish-cache/issues/1806 https://github.com/varnishcache/varnish-cache/issues/1772

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -182,7 +182,7 @@ static int __match_proto__(vdi_gethdrs_f)
 vbe_dir_gethdrs(const struct director *d, struct worker *wrk,
     struct busyobj *bo)
 {
-	int i, extrachance = 1;
+	int i, extrachance = cache_param->gethdr_extrachance;
 	struct backend *bp;
 	struct vbc *vbc;
 
@@ -237,7 +237,7 @@ vbe_dir_gethdrs(const struct director *d, struct worker *wrk,
 		    bo->req->req_body_status != REQ_BODY_CACHED)
 			break;
 		VSC_C_main->backend_retry++;
-	} while (extrachance);
+	} while (extrachance-- > 0);
 	return (-1);
 }
 
--- a/include/tbl/params.h
+++ b/include/tbl/params.h
@@ -494,6 +494,21 @@ PARAM(
 )
 
 PARAM(
+	/* name */	gethdr_extrachance,
+	/* typ */	uint,
+	/* min */	"0",
+	/* max */	"100",
+	/* default */	"1",
+	/* units */	NULL,
+	/* flags */	EXPERIMENTAL,
+	/* s-text */
+	"Maximum number of retries in vbe_dir_gethdrs. "
+	"Setting this to zero disables the feature.",
+	/* l-text */	"",
+	/* func */	NULL
+)
+
+PARAM(
 	/* name */	gzip_buffer,
 	/* typ */	bytes_u,
 	/* min */	"2k",


bsdphk:
If we want this parameter, at the very least we want its minimum >1 so it drains the swamp.

blblack:
@bsdphk - The fixup here is just one way to put some sanity-control on the issue, but I really think the problem runs deeper and may be fixable in other ways.

The main issue is that this is a source of retries that isn't controllable in VCL.  These retries don't loop out through the max_retries mechanisms (and other general logical control abilities) available in e.g. VCL's vcl_backend_response and/or vcl_backend_error.  The code also seems to implicitly assume that extrachance failure->retry will be quick and rare, but there are definitely real-world cases where it is neither (long timeout per retry, and happens almost every time for certain requests, resulting in very long time spent in this loop).

In our example case, the backend's behavior was to consume the request, emit zero bytes of output, and then un-gracefully close the connection after 2 minutes.  As our `first_byte_timeout` was 3 minutes, varnish waiting out those two minutes is legitimate.  However, we would've expected at that point to re-enter VCL and give our VCL logic a chance to control retries/error behavior.  The other confounding factors are that we have a high number of resue-able connections to this backend, and many legitimate healthy requests are flowing over them.  Only one particular request URL with low traffic is triggering the backend into buggy behavior with the 2 minute timeout -> close.  Therefore after each 2 minute timeout, when extrachance loops, the odds of it getting another reused connection to make its attempt is very high, therefore the odds of flipping the extrachance variable to zero is very low, and thus a single request can get stuck in this retry loop for hours.

The reason retry controls are important to us is that our VCL is careful to only retry requests once at our very front-most varnish layer, and only in certain situations (503 and vcl_backend_error).  In all other cases we choose to never retry errors.  This is to avoid a multiplicative storm of retries through our layered varnish tiers assaulting the final application-layer backend.

It's pretty clear that the reason for the extrachance behavior was to catch a rare race, where the backend happened to close a reused connection just as varnish was already sending the request down the wire, but the case it's catching here is not like that at all (it's simply a normal long-ish timeout waiting for the first byte, but within the limits of `timeout_first_byte`).  The code also seems to expect a high probability of the retry using a fresh connection in this case (I guess, unless the authors were expecting many loops here?), but that's also not a valid assumption.

I really think the ideal solution is to get rid of the extrachance retry loop entirely.  There's probably not a reliable way to ensure it only affects the desired race condition, and if the supposedly-rare race condition is hit, it should be treated like any other connection-level error and fall out through vcl_backend_error where max_retries and custom logic can deal with it.

Beyond that, it might help to add a new timeout parameters for backend connection idleness and total lifetime.  These are the two most likely factors for the application to be randomly closing one of varnish's keepalive connections at the "wrong" time.  By exposing that tuning in varnish, an administrator could be sure to tune varnish's backend idle/lifetime timeouts slightly lower than any at the application layer, thus ensuring varnish doesn't hit the race condition at all.

Dridi:
> It's pretty clear that the reason for the extrachance behavior was to catch a rare race

I'm not sure, I remember reading in one of the RFCs that proxies could retry idempotent requests once under certain conditions. I'm saying once but it was a bit more twisted though I don't remember the details.

> Beyond that, it might help to add a new timeout parameters for backend connection idleness and total lifetime.

Aren't you looking for `backend_idle_timeout`? For the total lifetime, I don't think there's any knob available as of today.

blblack:
>> It's pretty clear that the reason for the extrachance behavior was to catch a rare race

> I'm not sure, I remember reading in one of the RFCs that proxies could retry idempotent requests once under certain conditions. I'm saying once but it was a bit more twisted though I don't remember the details.

Proxies certainly *can* retry idempotent requests.  The problem isn't that this violates the RFCs.  The problem is that these retries, unlike all others, aren't controllable from VCL and/or max_retries (which we want for sanity sake in multi-layer varnish setups), and also that they can hang a request indefinitely within the bowels of varnish while indefinitely sending request traffic to an application server.

The comments in the extrachance code mention the race explicitly:
```
                /*
                 * If we recycled a backend connection, there is a finite chance
                 * that the backend closed it before we got the bereq to it.
                 * In that case do a single automatic retry if req.body allows.
                 */
```

The extrachance retry *does* cover that above race condition, it just also covers a bunch of other conditions (some of which involve longer timeouts first), and can retry infinitely (rather than the commentary's "single retry") under the right conditions.

>> Beyond that, it might help to add a new timeout parameters for backend connection idleness and total lifetime.
> Aren't you looking for backend_idle_timeout? For the total lifetime, I don't think there's any knob available as of today.

Ah, I did miss this when I was looking, sorry.  But also, this is only a global setting, it should probably exist per-backend like `first_byte_timeout` and such (as varnish may front several different application stacks with different behaviors...).  Still, the global is there and does offer some control over the potential race situation.

Dridi:
> it should probably exist per-backend like first_byte_timeout and such

And also per-bereq (mentioning it here to get a chance to discuss it next bugwash).

Dridi:
This problem has resurfaced on a different channel, and my year-old comment needed a clarification, so here it is:

> My last comment was about the introduction of a tunable parameter for the extra chances, if we expose one it could also be tunable per-transaction.
>
> Like timeouts have tunable parameters, override-able per backend and per transaction. It was just a reminder to consider the chain of overriding, if it makes sense (for example we can't override `max_retries`) depending on the nature of the backend. I would be perfectly fine with just a global parameter.

daghf:
This was a subject of today's bugwash.

The objective of the "extrachance" logic is to address the race that can occur when we transmit a request over a reused connection while at the same time the backend is closing the connection.

The conclusion of today's discussion is the following:

* Limit extrachance to a single retry
* On the second time around, force a fresh connection
* We do not want to introduce a parameter for number of extrachance retries

Furthermore if both attempts failed it should land us in `vcl_backed_error`, where the VCL user could choose to do `return (retry);` if further attempts are warranted.

PR #2138, opened 2016-11-19
Disable Nagle's algorithm on the backend as well

### LINKED ISSUE #2134: Disable Nagle's
Can we consider disabling Nagle's

This flag (TCP_NODELAY) is an option that can be enabled on a per-socket basis and is applied when you create a TCP socket. This is done for a purpose: Nagle's algorithm is generally useful and helps handle network congestion. I doubt you want to disable it system-wide since your system will probably suffer from this deactivation.

To disable it for a given socket, you can apply the option TCP_NODELAY as explained here and here in C:

int flag = 1;
int result = setsockopt(sock,            /* socket affected */
                        IPPROTO_TCP,     /* set option at TCP level */
                        TCP_NODELAY,     /* name of option */
                        (char *) &flag,  /* the cast is historical cruft */
                        sizeof(int));    /* length of option value */
 if (result < 0)
    ... handle the error ...

fgsch:
Fixes #2134.

### DIFF
--- a/lib/libvarnish/vtcp.c
+++ b/lib/libvarnish/vtcp.c
@@ -279,6 +279,7 @@ VTCP_connect(const struct suckaddr *name, int msec)
 	struct pollfd fds[1];
 	const struct sockaddr *sa;
 	socklen_t sl;
+	int val;
 
 	if (name == NULL)
 		return (-1);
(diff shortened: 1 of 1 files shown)


fgsch:
Committed.

hermunn:
Backport review: The patch in master, 0205cdf9e4, is now backported as 72435ed0e9aa.

PR #2139, opened 2016-11-21
Collect vtest failed tests logs more reliably

Dridi:
Instead of randomly looking for the log files, look for them where we
expect them to land. For instance, after a git pull, a distclean
operation may not remove the dist directory if Varnish's version
changed because it would reconfigure itself on the fly.

Instead of taking a relative file name as a parameter, with the obvious
problems it may create when changing directories, the list of failed
tests is searched in the function's standard input.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,6 +32,9 @@ distcleancheck_listfiles = \
 	find . -type f -exec sh -c 'test -f $(srcdir)/$$1 || echo $$1' \
 		sh '{}' ';'
 
+vtest-clean:
+	$(am__remove_distdir)
+
 # XXX: This is a hack to ensure we have a built source tree when
 # running make dist If we had used non-recursive make we could have
 # solved it better, but we don't, so use this at least for now.
--- a/tools/vtest.sh
+++ b/tools/vtest.sh
@@ -73,6 +73,7 @@ autogen () (
 makedistcheck () (
 	set -e
 	cd "${BUILDDIR}"
+	nice make vtest-clean
 	nice make distcheck
 )
 
--- a/tools/vtest.sh
+++ b/tools/vtest.sh
@@ -77,21 +77,26 @@ makedistcheck () (
 )
 
 failedtests () (
-	for t in `grep '^FAIL: tests/' ${1} | sort -u | sed 's/.* //'`
+	set -e
+
+	REPORTDIR=$(pwd)/_report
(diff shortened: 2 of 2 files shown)


Dridi:
Following discussions on IRC I will change the `awk` command to a `sed` equivalent. But I just realized that this patch introduces a new `WARNING` entry in the report that would need to be handled by the backend first.

bsdphk:
Tried this on a FreeBSD machine and it didn't work, looked like it hung in the "read".

Staring a the patch, it doesn't look like your grep command has any input ?

Dridi:
I just updated the pull request with the missing change, I overlooked it because I had other changes I didn't want to check in. That's basically the function call [1] that needed to turn the positional parameter into a redirection...

[1] https://github.com/varnishcache/varnish-cache/pull/2139/commits/8df29ac#diff-87b6ccb169979a78e050279c356d8d83R172

Dridi:
Because I have an excellent memory:

>11:36 < phk> dridi, looking at your vtest patch, I'm wondering if it wasn't smarter to go in the opposite direction:  Find all .trs files and report errors based on those
11:36 < phk> that is a lot cleaner than grepping things out of make distchecks stdout/err
11:36 < dridi> phk: agreed
11:37 < dridi> I will update my PR
[snip]
12:21 < phk> dridi, update your PR when you have time
12:22 < phk> dridi, Ohh, also, the VTCGITREV line doesn't contain test-name in your current patch

Dridi:
Commit updated, and during my tests I came across another problem solved in cb93756. I had to rebase against master because in the mean time another v50 test appeared.

bsdphk:
This seems to work on FreeBSD so commit it and lets see what happens.

I do have one footnote though:  If we want to get rid of autocrap we will have to revisit this issue when we do the new "make check"

Dridi:
Pushed as 6cd97d1 and d977207.

Dridi:
I blame `bash` in 6e21a2e but I believe it actually happened with the AT&T `sh`. But even with the latter I could not reproduce the problem in an interactive shell or a script. It did happen while I was working on something unrelated, but I'm fine with not digging further if it fixes `vtest.sh` on Solaris.

@nigoroll please restart your continuous integration using the latest master.

hermunn:
Backport review: `vtest` is 4.1 only.

PR #2140, opened 2016-11-21
make req.http.name case-insentive

michbsd:


Dridi:
I'd rather use `std.tolower()` before testing the Host header to promote normalization instead. Otherwise by propagating different cases for the host you will break the default hashing from the built-in VCL.

michbsd:
OK for me. How would you like me to proceed?

Dridi:
You can amend your commit and force-push to your master branch, that will update this pull request.

fgsch:
This sounds like a great idea until you check how the backend parses the query string.
Obviously for this particular case using std.tolower() does not change things but I rather keep the regular expression to avoid any caveats.

Dridi:
> This sounds like a great idea until you check how the backend parses the query string.

@fgsch, not sure to follow you here, it's a `Host:` header check in a snippet in the documentation. I don't see the relation to query-strings here, it's only that a host name is case-insensitive, we're not fiddling with the URL. Did you confuse the two because I mentioned the default hashing?

I will however use this as an occasion to shamefully plug [vmod-querystring](https://github.com/Dridi/libvmod-querystring) for anyone interested in production-grade query-string manipulation. :electric_plug:

michbsd:
Right. So should I change to std.tolower() or leave the regex ?

fgsch:
> @fgsch, not sure to follow you here, it's a Host: header check in a snippet in the documentation. I don't see the relation to query-strings here, it's only that a host name is case-insensitive, we're not fiddling with the URL. Did you confuse the two because I mentioned the default hashing?

Indeed, I did misread it and tried deleting the comment afterwards but my internet connection at the time was flaky. Please ignore it.

Dridi:
@fgsch, the problem with doing that in the built-in VCL is the use of the `std` module.

@michbsd, you missed the indent issue on the import statement. Don't worry about that, I will manually squash and merge your commits. I may go a bit further because the regex used are also not ideal...

hermunn:
Backport review: The file `doc/sphinx/users-guide/vcl-separate.rst` is new in 5.0.0 - nothing to backport.

PR #2141, opened 2016-11-21
more argument processing before daemonizing

nigoroll:
or am I overlooking a good reason why these should happen after daemonizing?

### DIFF
--- a/bin/varnishd/mgt/mgt_cli.c
+++ b/bin/varnishd/mgt/mgt_cli.c
@@ -570,7 +570,7 @@ mgt_cli_telnet(const char *T_arg)
 	AZ(error);
 	AZ(VSB_finish(vsb));
 	if (VSB_len(vsb) == 0)
-		ARGV_ERR("-T %s could not be listened on.", T_arg);
+		ARGV_ERR("-T %s could not be listened on.\n", T_arg);
 	/* Save in shmem */
 	mgt_SHM_static_alloc(VSB_data(vsb), VSB_len(vsb) + 1, "Arg", "-T", "");
 	VSB_destroy(&vsb);
--- a/bin/varnishd/mgt/mgt_cli.c
+++ b/bin/varnishd/mgt/mgt_cli.c
@@ -506,21 +506,17 @@ mgt_cli_secret(const char *S_arg)
 
 	VJ_master(JAIL_MASTER_FILE);
 	fd = open(S_arg, O_RDONLY);
-	if (fd < 0) {
-		fprintf(stderr, "Can not open secret-file \"%s\"\n", S_arg);
-		exit(2);
-	}
+	if (fd < 0)
+		ARGV_ERR("Can not open secret-file \"%s\"\n", S_arg);
+
 	VJ_master(JAIL_MASTER_LOW);
 	mgt_got_fd(fd);
 	i = read(fd, buf, sizeof buf);
-	if (i == 0) {
-		fprintf(stderr, "Empty secret-file \"%s\"\n", S_arg);
-		exit(2);
-	}
-	if (i < 0) {
-		fprintf(stderr, "Can not read secret-file \"%s\"\n", S_arg);
-		exit(2);
-	}
+	if (i == 0)
+		ARGV_ERR("Empty secret-file \"%s\"\n", S_arg);
+	if (i < 0)
+		ARGV_ERR("Can not read secret-file \"%s\"\n", S_arg);
+
 	AZ(close(fd));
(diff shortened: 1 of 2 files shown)


nigoroll:
advise from @bsdphk during bugwash

> there are things which need to happen after daemon() under all circumstances, so we cannot do it the other way around anyway.

may be more complicated than it looks

bsdphk:
I've looked at this at regular intervals and every time I have shaken my head and dropped it.

The problem is that there is initialization that we should do after daemon() and error messages we cannot get back from there.

What really need to happen is that we implement our own half-daemon(3), so that we can:

    User starts varnishd (pid=AAA)

    If -d is given:
      AAA performs argument processing, setup and turns stdin/stdout into CLI
    else:
      AAA does a "half-daemon(3)" which detaches (pid=BBB)
        BBB's stdout+stderr is a pipe back to AAA
      BBB performs argument processing, setup and starts the child process.
      If any of this fails, a suitable error message is sent down the pipe to AAA
        AAA emits the error on stderr and exits(>0)
        NB: AAA cannot see BBB's non-zero exit code, BBB is detached by half-daemon(3).
      On successfull startup BBB sends a magic "OK" token down the pipe
        AAA and BBB closes pipe, and AAA exit(0)

This is way more complex than I've been comfortable with.

It may be time now.

bsdphk:
I've done most of the above.  Please evaulate if this closes the ticket.

nigoroll:
quick review:
```
$ /tmp/sbin/varnishd -a 127.0.0.1:8888 -b 127.0.0.1:8080 -S /sfd
Error: Cannot open -S file (/sfd): No such file or directory
```
good

```
$ /tmp/sbin/varnishd -a 127.0.0.1:8888 -b 127.0.0.1:8080 -T 8.9.10.11:80
Debug: Platform: Linux,3.16.0-4-amd64,x86_64,-jnone,-smalloc,-smalloc,-hcritbit
Error: -T 8.9.10.11:80 could not be listened on.
```
good

```
$ /tmp/sbin/varnishd -a 127.0.0.1:8888 -b 127.0.0.1:8080 -M bazz.bar.boom:80
Debug: Platform: Linux,3.16.0-4-amd64,x86_64,-jnone,-smalloc,-smalloc,-hcritbit
Debug: Child (1437) Started
```
do we care?

nigoroll:
possibly related:

https://github.com/varnishcache/pkg-varnish-cache/blob/master/debian/varnish.init and other tools use `-C` to execute a "config test" with other arguments like so:
```

$ /tmp/sbin/varnishd -C -a :81 -f empty.vcl 
Error: Could not get socket :81: Permission denied
```

Do we want this behavior (and change init scripts) or should other arguments remain unchecked with `-C`?

fgsch:
fwiw, that example is not 100% accurate. If you have `-a :81` you will need privs both when running with and without -C.

Whether we want to error out if -C and other arguments are present (or ignore them) is unrelated to that particular error.

nigoroll:
@fgsch where exactly is the example inaccurate?

to compare with 4.1:
```
$ /tmp/41/sbin/varnishd -C -a :81 -f ~/Devel/varnish-git/varnish-cache/empty.vcl 2>&1 | head 
/* ---===### include/vdef.h ###===--- */

/*-
 * Copyright (c) 2006 Verdens Gang AS
 * Copyright (c) 2012 Fastly Inc
 * Copyright (c) 2006-2015 Varnish Software AS
 * All rights reserved.
 *
 * Author: Poul-Henning Kamp <phk@phk.freebsd.dk>
 * Author: Rogier 'DocWilco' Mulhuijzen <rogier@fastly.com>
slink@haggis:/tmp/varnish-cache (4.1)$ /tmp/41/sbin/varnishd -V
varnishd (varnish-4.1.4 revision aee2522)
Copyright (c) 2006 Verdens Gang AS
Copyright (c) 2006-2015 Varnish Software AS

```

fgsch:
@nigoroll It was impossible to tell by your example that this is an actual change in behaviour from 4.1/5.0, in particular considering the context of the debian init scripts.

I stand corrected.

nigoroll:
I see, sorry.

bsdphk:
Looks OK.

nigoroll:
IRC note: /me to create new PRs for the follow-up issues documented herein

nigoroll:
to be continued in #2217

hermunn:
Backport review: I have decided that argument processing will not be backported to 4.1 unless there is a clear bug that needs to be fixed.

PR #2145, opened 2016-11-25
Set default.backend in vcl_init{}

Dridi:
This patch set introduces a new optional `default.backend` variable that can be set to a backend or a director in `vcl_init`. It also allows compilation of VCLs without backends, but such VCLs won't load if they don't have a `default.backend` when they leave `vcl_init`.

The last patch fixes an alleged bug in label selection by relying on the `default.backend`.

It is not documented yet, except if you consider the two test cases proper documentation. I'd like a review first.

### DIFF
--- a/bin/varnishd/cache/cache_priv.h
+++ b/bin/varnishd/cache/cache_priv.h
@@ -108,7 +108,7 @@ void VSL_ChgId(struct vsl_log *vsl, const char *typ, const char *why,
 void VSL_End(struct vsl_log *vsl);
 
 /* cache_vcl.c */
-struct director *VCL_DefaultDirector(const struct vcl *);
+const struct director *VCL_DefaultDirector(const struct vcl *);
 void VCL_SetDefaultDirector(struct vcl *, const struct director *);
 const struct vrt_backend_probe *VCL_DefaultProbe(const struct vcl *);
 void VCL_Init(void);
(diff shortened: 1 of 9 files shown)


Dridi:
@bsdphk as you requested on IRC I have pondered the idea of the alternate syntax you suggested:

```
new default_backend = [...];
```

It will overload the `new` operator, which in my opinion adds more magic than the original intention of removing some. The operation we are doing here is really a simple assignation, not calling a constructor, so I would go for `set`.

You also used the `default_backend` notation while the consensus before I submitted this PR was `default.backend` with the idea of having a namespace-like notation. I think I like `default_backend` better especially because of how we treat namespaces. This PR doesn't actually introduce a full-fledge `default` namespace...

Anyway, renaming the variable is not the hardest thing to do, this PR is still relevant otherwise.

Dridi:
Commits updated, rebased against current master because  of conflicts that appeared in the mean time.

bsdphk:
I am really uncomfortable with the object-lifetime-issue in this patch.

Right now, all backends from VMODs have per-transaction scope, whereas any VMOD backend set as default backend has eternal scope ?

How does J.Random.VMOD know which it is ?

I don't see how we can do this, without refcounting the backends.

And I we are going there, I want us to refcount all VMOD objects rather than just backends, because the same bloody issue will keep popping up as as we relax the restrictions on VMODs.

Dridi:
Setting a default backend/director is not going to happen before we deal with global backends and refcounting. Closing.

hermunn:

PR #2152, opened 2016-12-02
add std.getenv()

nigoroll:


lib/libvmod_std/vmod_std.c
```diff
@@ -248,3 +248,12 @@ vmod_strstr(VRT_CTX, VCL_STRING s1, VCL_STRING s2)
...
 	return (strstr(s1, s2));
 }
+
+VCL_STRING __match_proto__(td_std_getenv)
+vmod_getenv(VRT_CTX, VCL_STRING name)
+{
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	if (name == NULL || *name == '\0')
+		return (NULL);
+	return (getenv(name));
```
Dridi: This should be copied in the workspace.

nigoroll: why?

Dridi: The prototype is `char *getenv(const char *name);`, it's not safe to assume it won't change during the transaction. A VMOD/library call may call `setenv` too.

nigoroll: it's not even save to assume that it won't change during the memcpy. So why bother?

Dridi: Transactions last orders of magnitude longer than a `memcpy`, but I agree it wouldn't be 100% safe either.

bsdphk:
Dridi is right in that you cannot trust the constancy of the getenv(3)'s return if the process changes its own environment.

However, anybody who changes the environment of varnishd while it runs, deserve what they get as a result

OK

nigoroll:
69838a5cc3250b79ad03d264ba49f37f814c6f8f

hermunn:
Backport review: Backported as 503044881785

PR #2153, opened 2016-12-05
New rollover director

gquintard:
I'd like to introduce a new type to director to vmod-director: the rollover one.

This is basically the fallback director, except it doesn't revert back if a backend with a higher priority goes healthy again. This has be requested by some clients as it can help with consistency when backends are not totally in sync.

### DIFF
--- a/bin/varnishtest/tests/d00027.vtc
+++ b/bin/varnishtest/tests/d00027.vtc
@@ -1,0 +1,68 @@
+varnishtest "rollover director"
+
+server s1 {
+	rxreq
+	expect req.url == "/qux"
+	txresp
+} -start
+
+server s2 {
+	rxreq
+	expect req.url == "/foo"
+	txresp
+} -start
+
+server s3 {
+	rxreq
+	expect req.url == "/bar"
+	txresp
+
+	rxreq
+	expect req.url == "/baz"
+	txresp
+} -start
+
+varnish v1 -vcl+backend {
+	import directors;
+
+	sub vcl_init {
+		new vd = directors.rollover();
+		vd.add_backend(s1);
+		vd.add_backend(s2);
+		vd.add_backend(s3);
+	}
+
+	sub vcl_recv {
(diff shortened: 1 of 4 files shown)


Dridi:
I would change the semantic of `vdir_remove_backend` instead of repeating the complete logic in the rollover director.

`unsigned vdir_remove_backend(struct vdir *, VCL_BACKEND be);`

Currently it returns the number of backends after the removal, but this value is never used [1] by existing directors. So instead I would return the index of the removed backend, or `-1` if none were found. This way you can decrement the current backend if the removed index was smaller. It would change the prototype to:

`int vdir_remove_backend(struct vdir *, VCL_BACKEND be);`

[1] For reference:
```
$ git grep vdir_remove_backend
lib/libvmod_directors/fall_back.c:      (void)vdir_remove_backend(fb->vd, be);
lib/libvmod_directors/hash.c:   (void)vdir_remove_backend(rr->vd, be);
lib/libvmod_directors/random.c: (void)vdir_remove_backend(rr->vd, be);
lib/libvmod_directors/round_robin.c:    (void)vdir_remove_backend(rr->vd, be);
lib/libvmod_directors/vdir.c:vdir_remove_backend(struct vdir *vd, VCL_BACKEND be)
lib/libvmod_directors/vdir.h:unsigned vdir_remove_backend(struct vdir *, VCL_BACKEND be);
```

fgsch:
Without looked at the code, couldn't we add a parameter to the fallback constructor to indicate whether it should be sticky?

Dridi:
@fgsch, both @gquintard and I agree!

And my comment on `vdir_remove_backend` would still be relevant.

nigoroll:
I'm with @fgsch . To me adding stickyness to other directors would appear to be the most obvious solution.

gquintard:
Cool, thanks for the feedback, I'll propose a new Product based on this.

hermunn:

PR #2156, opened 2016-12-06
Sticky options for fallback director

gquintard:
Add a "sticky" argument to the fallback director. If set, the director doesn't go back to a higher priority backend coming back to health.

I also took the opportunity to rename variables in a more consistent way in the last commit.

### DIFF
--- a/lib/libvmod_directors/vmod.vcc
+++ b/lib/libvmod_directors/vmod.vcc
@@ -95,13 +95,17 @@ Example
 	set req.backend_hint = vdir.backend();
 
 
-$Object fallback()
+$Object fallback(BOOL sticky = 0)
 
 Description
 	Create a fallback director.
 
 	A fallback director will try each of the added backends in turn,
 	and return the first one that is healthy.
+
+	If ``sticky`` is set to true, the director will keep using the healthy
+	backend, even if a higher-priority backend becomes available. Once the
+	whole backend list is exhausted, it'll start over at the beginning.
 
 Example
 	new vdir = directors.fallback();
--- a/bin/varnishtest/tests/d00027.vtc
+++ b/bin/varnishtest/tests/d00027.vtc
@@ -1,0 +1,72 @@
+varnishtest "Sticky fallback director"
+
+server s1 {
+	rxreq
+	expect req.url == "/qux"
+	txresp
+} -start
+
+server s2 {
+	rxreq
+	expect req.url == "/foo"
+	txresp
+} -start
+
+server s3 {
+	rxreq
+	expect req.url == "/bar"
+	txresp
+
+	rxreq
+	expect req.url == "/baz"
+	txresp
+} -start
+
+varnish v1 -vcl+backend {
+	import directors;
+
+	sub vcl_init {
+		new vd = directors.fallback(sticky = true);
+		vd.add_backend(s1);
+		vd.add_backend(s2);
+		vd.add_backend(s3);
+	}
+
+	sub vcl_recv {
+		set req.backend_hint = vd.backend();
+		return(pass);
+	}
+
+} -start
+
+varnish v1 -cliok "backend.set_health s1 sick"
+
+client c1 {
+	txreq -url /foo
+	rxresp
+	expect resp.status == 200
+} -run
+
+varnish v1 -cliok "backend.set_health s2 sick"
+
+client c1 {
+	txreq -url /bar
+	rxresp
+	expect resp.status == 200
+} -run
+
+varnish v1 -cliok "backend.set_health s1 healthy"
+
+client c1 {
+	txreq -url /baz
+	rxresp
+	expect resp.status == 200
+} -run
+
+varnish v1 -cliok "backend.set_health s3 sick"
+
+client c1 {
+	txreq -url /qux
+	rxresp
+	expect resp.status == 200
+} -run
--- a/bin/varnishtest/tests/d00028.vtc
+++ b/bin/varnishtest/tests/d00028.vtc
@@ -1,0 +1,85 @@
+varnishtest "Sticky fallback director, removing backends"
+
+server s1 {
+	rxreq
+	expect req.url == "/foo"
+	txresp
+} -start
+
+server s2 {}
+
+server s3 {
+	rxreq
+	expect req.url == "/bar"
+	txresp
+
+	rxreq
+	expect req.url == "/baz"
+	txresp
+
+	rxreq
+	expect req.url == "/qux"
+	txresp
+} -start
+
+server s4 {}
+
+varnish v1 -vcl+backend {
+	import directors;
+
+	sub vcl_init {
+		new vd = directors.fallback(sticky = true);
+		vd.add_backend(s1);
+		vd.add_backend(s2);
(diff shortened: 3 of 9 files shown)


ehocdet:
I see a potential race in vmod_fallback_resolve:

(fb->cur =  fb->vd->n_backend-1)
Thread A:  ++fb->cur
Thread B: fb->vd->backend[fb->cur]; 
```
	vdir_rdlock(fb->vd);
	if (!fb->st)
		fb->cur = 0;
	for (u = 0; u < fb->vd->n_backend; u++) {
		be = fb->vd->backend[fb->cur];             <=== Thread B
		CHECK_OBJ_NOTNULL(be, DIRECTOR_MAGIC);
		if (be->healthy(be, bo, NULL))
			break;
		if (++fb->cur == fb->vd->n_backend)    <==== Thread A
			fb->cur = 0;

```

Dridi:
Didn't we already discuss those assignments under the read lock? I don't remember why we kept it like this.

@slimhazard any idea?

Dridi:
My bad, it's actually introduced by this patch. Good catch @ehocdet.

gquintard:
Indeed, since then there was no modified global variable.

Thanks @ehocdet.

gquintard:
Changed the rdlock to a wrlock.

Now that only the rr director uses rdlock, do we actually need them? ping @slimhazard

slimhazard:
We added the rdwr locks when backends became dynamic, and we added the remove_backend() methods.

Before that, most of the directors in the distribution used common code which had the backends in a list from which a director could take a pick. The lists were immutable, since backends were fixed, so each thread could read from the list without any mutexing at all.

remove_backend(), for those directors with the common code base, changed the list, so accessing the list had to be mutexed. It felt a little iffy at the time, since backend threads had never before blocked each other while a director was choosing backends. But since changes in the list would presumably happen very rarely, whereas reading from the list to pick a backend would go on all day long, we decided to use a shared lock for reads and an exclusive lock only when the list had to be changed.

I believe that since then, it's turned out that directors blocking each other has turned out to be less of a problem than we feared. OTOH if at least the rr director can still work that way, why not keep the shared lock?

ehocdet:
With "cur" as a local variable, rdlock is ok:
```
	cur = fb->cur;
	for (u = 0; u < fb->vd->n_backend; u++, cur++) {
		if (cur >= fb->vd->n_backend)
			cur = 0;
		be = fb->vd->backend[cur];
		CHECK_OBJ_NOTNULL(be, DIRECTOR_MAGIC);
		if (be->healthy(be, bo, NULL))
			break;
	if (fb->st)
		fb->cur = cur;
```

In vmod_fallback_remove_backend, fb->cur  is changed outside any lock.
Reset it when a backend is remove could be a good trade-off?

gquintard:
Right,that was the reason for duplicating remove_backend in the first place, totally forgot about it, and that would still me my preferred way of doing it.

gquintard:
Cleaned and rebased.

Dridi:
In 0b04237 the function still returns `unsigned` values, it's fixed by the next commit.

In 38f40de we should `assert` that `*cur` is within bounds if provided at the beginning of `vdir_remove_backend`. I would also use the type `VCL_BOOL` for the `sticky` parameter.

In 6cc34e3 the test case doesn't check the response status, the test would pass even if the client requests never reach the servers (`expect` on the server side is not enough).

That's it for my review, only nitpicks.

bsdphk:
I'll pass on this one, I have no input on director policy choices

lib/libvmod_directors/vmod.vcc
```diff
@@ -95,14 +95,18 @@ Example
 	set req.backend_hint = vdir.backend();
 
 
-$Object fallback()
+$Object fallback(BOOL sticky = 0)
```
fgsch: 0 should be false (a boolen) no?

Dridi: Yes, it works too.

gquintard: Doesn't appear to work.

gquintard:
Rebased to take care of nitpicks, except the "BOOL sticky = 0" one, that seems to be a vcc issue.

Barring complaints I'll push in a few days.

fgsch:
> Rebased to take care of nitpicks, except the "BOOL sticky = 0" one, that seems to be a vcc issue.

Just to be clear, my comment was about using BOOL sticky=false.
That does not work?

gquintard:
d00001.vtc (and presumably all vtc using fallback) get me:

```
**** v1    0.2 CLI RX|Message from C-compiler:
**** v1    0.2 CLI RX|vgc.c:1522:9: error: use of undeclared identifier 'false'
**** v1    0.2 CLI RX|        false);
**** v1    0.2 CLI RX|        ^
**** v1    0.2 CLI RX|1 error generated.
**** v1    0.2 CLI RX|Running C-compiler failed, exited with 1
```

nigoroll:
> Just to be clear, my comment was about using BOOL sticky=false.
> That does not work?
No, the defaults need to be given in C notation. I had documented this some time ago with 4dd0ab6805ce9c543ded51ae16516068f81ba16d

Dridi:
My bad, it did compile the VMOD but I didn't run the test. It's the same as using `NULL`, the generated VCL doesn't include anything so default values must either be references created by the VMOD or literals...

gquintard:
closed with c9fcdf2bd10642b8fda3402ca8f9656554504103

hermunn:
Backport review: If I am not mistaken, this is master only. I assume that @gquintard will tell me if he disagrees.

gquintard:
@hermunn : I confirm

ibreger:
Sometimes after failover, when higher-priority backend is back to service, after some verification we'd like to switch back to this one (higher-priority) again. How it may be performed manually, for example by varnishadmn ?

Dridi:
Seeing your comment I realize that what we introduced is rather sticky round-robin: only when sick, move to next healthy, and cycle through the list.

I think we should move this behavior to the round-robin director and have this behavior instead: only when sick, use the first healthy backend in the list. The difference with regular fallback being that when a "higher-priority" backend is back, we stay to the current one as long as it's healthy.

@ibreger regarding what you are looking for, I'd say regular fallback is the solution (sticky=false).

sanjeevrn:
@Dridi - how could the same be achieved at runtime using the same definition (i.e. with sticky=true)? It does not seem useful to still keep skipping a higher priority backend once original issue (that caused it to go offline) has been found/fixed. At the very minimum, it seems a runtime facility needs to exist to reset things back to normal?

Dridi:
@sanjeevrn, yes, having a means to reset a "sticky" director would definitely be useful, I'm not ruling that out.

JonathanHuot:
I'm wondering if the fallback "sticky" state could be stored in backend instead of the director itself. For example, if the director marks the higher priority backend to "admin" sick once it becomes sick, the operator will be able to fallback by doing the regular varnishadm command.

Dridi:
Short answer is no. A backend can belong to several directors, and you can stack directors and don't have the guarantee that as a director you're dealing directly with backends.

PR #2158, opened 2016-12-08
export missing VSB symbols from libvarnishapi

slimhazard:
Now that libvarnish is no longer shared, this makes the remainder of VSB available from libvarnishapi.

It's worth pointing out that the man page ``vsb.3`` that lives in ``lib/libvarnish`` has fallen very far behind (it hasn't been maintained for years). I'd offer to update it, but I don't know roff markup (and would probably do it by copy-pasting), and maybe it's better to just retire it.


Dridi:
@slimhazard put the symbols in the 1.6 revision instead, it hasn't been released yet.

While on this topic, now that `libvarnish` has been folded into `libvarnishapi`, we should consider the remaining symbols from `libvarnish` that are available in `make install`'ed headers. I think many of the `V*` functions have been stable enough to be considered.

Dridi:
You can also bump the `-version-info` to `1:6:0`, we forgot it in the 5.0 release...

nigoroll:
@slimhazard I was totally unaware of the fact that we have (manually crafted?) roff in the tree. The source should really be rst and processed as for the other man pages

slimhazard:
@Dridi, done.

I agree that some more V* interfaces can be considered for inclusion in libvarnishapi (I personally would be in favor of VMB and VPF), but this is just VSB for now. I thought my apps that had been linking to libvarnish would run into a great deal more trouble, but it turns out that ``VSB_quote()`` and ``VSB_bcat()`` were all I needed (``VSB_delete()`` was missing as well, but that could be replaced with ``VSB_destroy()``).

So I think that libvarnishapi as an "official" replacement for libvarnish is much further along than I would have expected.

Dridi:
Probably a forsaken artifact inherited when VSB was stolen from FreeBSD's equivalent.

slimhazard:
@nigoroll, man pages documenting V* interfaces would be revolutionary in this project, but I wouldn't count on it.

Why maintain a doc for VSB in another format, after it's been unmaintained for years as it is? And why have a doc for VSB, but not for everything else that starts with a V?

I think it would be more accepting of reality to just remove ``vsb.3``.

fgsch:
@slimhazard +1 removing vsb(3).

bsdphk:
I'm OK with exposing the rest of VSB in libvarnishapi.

vsb(3) was added for any devs not familiar with FreeBSD's sbuf() API, but I think we have enough examples of VSB usage now that we don't need that any more.

OK

hermunn:
Backport review: There is no need to backport these to a stable release.

PR #2163, opened 2016-12-13
export missing VTIM symbols from libvarnishapi

slimhazard:
Exports the remaining VTIM symbols: ``VTIM_parse``, ``_timespec`` and ``_timeval``. Also bumps the LIBVARNISHAPI version to 1.7 (soname 1.0.7).

If the PR is accepted, I promise to merge it to master the right way this time. %^)

### DIFF
--- a/lib/libvarnishapi/Makefile.am
+++ b/lib/libvarnishapi/Makefile.am
@@ -9,7 +9,7 @@ AM_CPPFLAGS = \
 
 lib_LTLIBRARIES = libvarnishapi.la
 
-libvarnishapi_la_LDFLAGS = $(AM_LDFLAGS) -version-info 1:6:0
+libvarnishapi_la_LDFLAGS = $(AM_LDFLAGS) -version-info 1:7:0
 
 libvarnishapi_la_SOURCES = \
 	vsm_api.h \
--- a/lib/libvarnishapi/Makefile.am
+++ b/lib/libvarnishapi/Makefile.am
@@ -9,7 +9,7 @@ AM_CPPFLAGS = \
 
 lib_LTLIBRARIES = libvarnishapi.la
 
-libvarnishapi_la_LDFLAGS = $(AM_LDFLAGS) -version-info 1:7:0
+libvarnishapi_la_LDFLAGS = $(AM_LDFLAGS) -version-info 1:6:0
 
 libvarnishapi_la_SOURCES = \
 	vsm_api.h \
--- a/lib/libvarnishapi/libvarnishapi.map
+++ b/lib/libvarnishapi/libvarnishapi.map
@@ -171,10 +171,6 @@ LIBVARNISHAPI_1.6 {
 	VSB_vprintf;
 	VSB_delete;
 	VSB_indent;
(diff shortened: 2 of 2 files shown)


Dridi:
We still haven't done any release with of the current master/1.6, there's no point bumping to 1.7, move them to 1.6 with the rest.

Are you doing this on purpose to get a merge situation to master? :trollface:

slimhazard:
@Dridi, I suspected you might have asked for that, I just went for the new version to be on the safe side (changing an existing version, if that's what it is, can be troublesome). Will do.

fgsch:
Looks good to me.

bsdphk:
OK.

Any chance we can trick you into doing a comprehensive review of what we should export ?

daghf:
Reminds me of a discussion at the VDD in Oslo last year, where we spent some time going through which bits should be exposed to VMODs.

See https://etherpad.wikimedia.org/p/VDD15Q3 starting at line 118, copy-pasted below for convenience.

This should probably go into a VIP ?

<pre>
VMOD API mumble mumble
----------------------

(Think about this in context of VMOD version-check/dependency
 Does generate.py grep all (known) *.h's for version strings ?)

vdef.h
   pull miniobj into vdef ?
   --//-- vas --//-- ?
   
vqueue.h  \
vsha256.h  \
vtree.h     \  libvarnish,
vbm.h       /  as today
vtim.h     /
vsb.h     /
      
Base API ("cache_pub.h" ?)
    typedef VCL_*
    struct vrt_ctx
    gethdr stuff
    magic string
    get/set headers (selecthttp/GetHdr/SetHdr)
    WS access
    vmod_priv
    vmod_event
    CacheReqBody
    purge/ban ?
    re
    vrt_obj.h
    VRT_String*()
    RFC2616*()
    vcl.h less VCL_conf
    object iterator (?)  ("read only body access")
    Obj* ("read only")
    VSL
    req/busyobj/wrk/ws/sess/... are all opaque here.

Backend/director/VFP API ("reach out of machine")
    backend
    director
    protocol (?)
    vtcp
    vss
    vsa

VDP API ("modify body")
    push VDP
    push "the usual VDPs" (range, esi, gzip/gunzip)

Kitchensink API ("cache.h")
    internals of req/busyobj/...
    obscure functions (VSM, BAN, ...)
    
Non-API ("cache_priv.h")
    THR_*
    *_Init()
    Stuff which is not *ever* APIs
    

</pre>

slimhazard:
Pushed

slimhazard:
@daghf, thanks, I was trying to remember exactly that VDD conversation. I agree that we should take those notes at a starting point, let's not let that good VDD work go to waste.

@bsdphk, holidays coming up, then I'll be away until January 8th, and I've already agreed to look into having RST as the source for VMOD interfaces rather than the .vcc file (would like to get something started with that this week). So if I'm the one driving this subject, it won't be until a week into January.

hermunn:
Backport review: Unless there is explicit demand for this in 4.1, it will not be backported. If you disagree, comment here and reapply the **bacport-review-4.1** tag.

PR #2164, opened 2016-12-13
HPACK Huffman review

Dridi:
It was brought back to my attention when I saw bc2ba0a pop on the commits mailing list that I had a pending review in dridi/cashpack#3 that looked a lot familiar. The good news was that being on call for the weekend and having to stay close to the laptop was a good excuse to play with bitwise operators and do shifty things.

So I did review it and couldn't end up with the same lookup tables, and after tedious manual checks I believe that the changes I'm submitting are correct. This is for `varnishtest`.

A while back I also reviewed `varnishd`'s HPACK implementation, and couldn't find anything worth reporting. A few nitpicks here and there, but mostly the lack of support for CONTINUATION frames in h2 that seemed intentional.

@mbgrydeland's HPACK implementation uses a single table for decoding, halfway between what I initially did in cashpack and what @gquintard did in `varnishtest`. I can't remember how it works, the code generator is hard to follow (probably as hard as mine) but the generated table is self-documented and I came to the conclusion months ago that it was correct. I don't have the courage to try again.

As a side note, my intuition tells me that @gquintard trades more space for more speed, @mbgrydeland's algorithm needs more jumps and branches (in both cases for the worst case). I haven't tried any benchmark.


mbgrydeland:
> A while back I also reviewed varnishd's HPACK implementation, and couldn't find anything worth reporting. A few nitpicks here and there, but mostly the lack of support for CONTINUATION frames in h2 that seemed intentional.

Its great that you reviewed it, much appreciated! I am not sure I understand what you mean by lack of support for CONTINUATION frames though. It was my impression that the code is very much ready for CONTINUATIONs, in that it can accept a break in the input bytes at any byte. The built-in unit testing also tests this when run in the 1IN mode, where the input is fed 1 byte at a time, analogues to the headers being split among the maximum number of CONTINUATIONs possible.

It is true that the Varnish HTTP/2 frame handling still doesn't take CONTINUATIONs into account. But when it learns about those, the HPACK decode part should be ready to accept them. If you have spotted something amiss, please elaborate.

> As a side note, my intuition tells me that @gquintard trades more space for more speed, @mbgrydeland's algorithm needs more jumps and branches (in both cases for the worst case). I haven't tried any benchmark.

Yes, my emphasis was the benign common case. The 74 most common characters decode in a single lookup, which covers most characters used in headers. Though every single character produced will cause bit shifting of the input bits, which is kind of heavy handed. It would be possible to improve on that, trading a much larger table for only bit shifting every other input byte. Not sure that is worth it though. On the obscure side of the table, which is where a DOS attack would focus, there is a bit of table jumping before reaching the right output byte. This could be an area for improvement.

Dridi:
> I am not sure I understand what you mean by lack of support for CONTINUATION frames though.

I meant in the h2 code, not HPACK.

Dridi:
I was able to tweak my generator to reach the same result, and actually both set of tables are correct. I will commit the polish and keep the table change out. Closing.

hermunn:
Backport review: HPACK and H/2 code is 5.0 only, so nothing to backport.

PR #2170, opened 2016-12-20
Document default timeout for varnishtest

rnsanchez:
To avoid surprises on the user side, document the default timeout for `varnishtest`.

Found while testing a VCL that used `delay 300`.


gquintard:
Merged, thanks!

PR #2176, opened 2017-01-02
A `write_body` VTC command

Dridi:
Following my campaign for empowering out-of-tree uses of `varnishtest`, an escape hatch to delegate body (XML, JSON, etc) checks to external programs.

This is also slightly breaking the VTC syntax as part of a general cleanup, `non-fatal` was renamed to `non_fatal`.

It includes opinionated changes to `libvarnish.a`, hopefully not controversial.

### DIFF
--- a/bin/varnishtest/tests/a00015.vtc
+++ b/bin/varnishtest/tests/a00015.vtc
@@ -1,9 +1,14 @@
 varnishtest "Write a body to a file"
 
 server s1 {
+	# First, HTTP checks
 	rxreq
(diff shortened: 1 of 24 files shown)


Dridi:
I will add more commits, I forgot the HTTP/2 streams...

Dridi:
Added `write_body` for HTTP/2 streams too, rebased against master. Ready for review.

hermunn:
Backport review: It it is breaking, it will surely not be backported.

Issue #2187, opened 2017-01-16
vsm_free stat is never incremented


ehocdet:
vsm_free stat is never incremented.
detected with varnish 4.1

$ varnishstat -n cache -1 -f "MAIN.vsm_*" 
MAIN.vsm_free     18446744071652838288          .   Free VSM space
MAIN.vsm_used        137578048          .   Used VSM space
MAIN.vsm_cooling             0          .   Cooling VSM space
MAIN.vsm_overflow            0          .   Overflow VSM space
MAIN.vsm_overflowed            0         0.00 Overflowed VSM space

fgsch:
Fixed in b8dec3334e739a07a22dcbeb39aafcd45ed15fb8.

hermunn:
Backport review: Backported (1978ebff506439) in the corresponding Pull Request.

PR #2188, opened 2017-01-16
fix vsm_free (never incremented)

ehocdet:
a proposal fix for the issue: https://github.com/varnishcache/varnish-cache/issues/2187


Dridi:
For some reason beyond me the build fails on travis while this patch was applied on an up to date master, so I wouldn't be bothered by that. I've seen another build of master passing.

LGTM, well spotted!

fgsch:
I'd prefer this is done in VSM_common_free() together where g_used is decremented.

Dridi:
Catching up with unread commits, I now believe that a29fca7 breaks pull request builds.

fgsch:
Merged. Thanks for the PR!

hermunn:
Backport review: Patch b8dec3334e739a backported as 1978ebff506439.

PR #2196, opened 2017-01-19
Fix/Improve H1 Expect & 100 Continue / Logging / Start H2 Expect handling

nigoroll:


bin/varnishd/http2/cache_http2_deliver.c
```diff
@@ -92,6 +92,44 @@ h2_bytes(struct req *req, enum vdp_action act, void **priv,
...
+	const char	*p;
+	size_t		l;
+	unsigned	s;
+	const char	*r;
+} h2_simple_response_msg[] = {
+	[R_400] = { "\x92", 1, 400, "Bad Request"},
+#define SRESP(s, r) [R_##s] = { "\x18\x03" #s, 5, s, r}
+	SRESP(100, "Continue"),
+	SRESP(417, "Expectation Failed")
+};
```
Dridi: This part, I don't understand. This `SRESP` doesn't have to be the same as the H/1 version, you can add the `"\x??"` prefixes in the macro signature. You can also have `H1_SRESP` and `H2_SRESP` macros if you want to avoid confusion.

Also you forgot to `#undef` it here. Nitpick about the other `SRESP`, I would `#undef` it inside the curly braces.

nigoroll: Nitpicking addressed (I would rebase before merge).
Other than that I don't know what you mean, sorry. Could you illustrate with code?

Dridi: Nevermind, there were two differences I didn't spot.

bsdphk:
Thanks for looking at this.

The trouble with 100-Continue are at top of page 35 in rfc7231:

```
   o  A client that sends a 100-continue expectation is not required to
      wait for any specific length of time; such a client MAY proceed to
      send the message body even if it has not yet received a response.
      Furthermore, since 100 (Continue) responses cannot be sent through
      an HTTP/1.0 intermediary, such a client SHOULD NOT wait for an
      indefinite period before sending the message body.
```

By instantly sending 100 Continue, as our current code does, one can avoid all the races contained there, and the subsequent session state issues, on all usable TCP connections.

rfc7231 has about another page of text about what happens if you don't do that, which uses troublesome text such as "send an immediate response".

The traditional varnish interpretation of "immediate" is "before we do anything with a backend", which again means "trying to connect or send on a socket." and that is feasible, by complicating cache_req_body.c slightly and not sending 100-C if VRB_Ignore() is called.

However, that leaves the client connection in Schrödings box:  either the client started sending the body anyway - or it didn't, and we can only find out by looking at the next bytes received.

That gets really ugly if the body of the request was a HTTP request.

RFC7231 has the following suggestion for that:

```
   o  A server that responds with a final status code before reading the
      entire message body SHOULD indicate in that response whether it
      intends to close the connection or continue reading and discarding
      the request message (see Section 6.6 of [RFC7230]).
```

In other words:  "Connection: close"

One can imagine further embellishments:   If there is a C-L and it's small, sending 100-C and swallowing the body would salvage the connection, but then, if it is chunked all bets are off.

I'd like to avoid new parameters for this, so here is a strawman:

Unless vcl_recv{} supresses 100-C (req.dont_continue_yet = True), 100-C gets sent immediately and unconditionally when vcl_recv{} returns.

If vcl_recv{} supresses 100-C, it only gets sent if VRB_Cache() or VRB_Iterate() gets called, whereas VRB_Ignore() becomes a no-op.  It is the responsibility of the vcl to make responses "Connection: close" if need be.

nigoroll:
@bsdphk thank you for looking at this and the detailed response.

### Intro / Motivation

First of all, I should probably document herein the motivation, because so far I had shared it on irc only. The scenario is varnish behind a CDN, where http-level filtering (marketing-buzzword: WAF) is implemented in varnish but not in the CDN. The case at hand is that the CDN forwards a POST request which varnish decides is invalid (for that URL) and responds with a 405. In this case, the CDN may run into a read error and emit a 503 because varnish does not finish reading the body. So when I asked the CDN to fix the response status, their response was that the issue was due to the 100 followed by a 405 and connection close.

Sending a 100 irrespective of the whether or not we'll actually consume a request body is contradictory if we later decide to refuse it, for instance with a 405 response -- the semantics are _yeah, send me that request body_ but then _ah nah, don't want it_.

The way I read 7231, it does not forbid this behavior, but I see a valid point that always sending a 100 is not exactly right when we won't actually read the request body.

### response to phk
I have trouble understanding the race you see: varnish behavior for reading the request body does not depend on whether or not we've sent a `100 Continue` and, as you rightly pointed out, the client doesn't need to care either. In other words, I regard `100 Continue` as an optional advisory.
The previous code generated the `100 Continue` response right after parsing the request, and I am suggesting to move this to the time when we actually know whether or not we're going to read it. I think this is totally in line with the original intention behind Expect: `100-continue` -- to save resources on the client side for constructing a request body which might never be consumed:

(rfc 7231)
> This allows the client to wait for an indication that it is worthwhile to send the
> message body before actually doing so, which can improve efficiency
>  when the message body is huge or when the client anticipates that an
>  error is likely

So while I agree that this may add some latency for some scenarios (like small bodies over high latency connections), it actually is the only interpretation of _immediate_ in the presence of `Expect: 100-continue` which I would consider valid -- otherwise, the whole idea was rendered useless. Also, as you've pointed out also, it is the client's choice whether or not to use this mechanism.

Regarding `Connection: close`, I have also added the tiny bits (which I had in production with a previous varnish version) to instruct core code to close the connection from `vcl_synth`. The option to not drain-read the body is a important security option, IMHO.

Please note that we also did not implement this bit correctly, and that should be fixed also:

>   o  A server that receives a 100-continue expectation in an HTTP/1.0
>      request MUST ignore that expectation.

nigoroll:
FYI: branch rebased

nigoroll:
but @bsdphk , you are right that vcl should be able to trigger the `100 Continue` once the decision has been made to process a request body.

nigoroll:
@bsdphk added `std.send_100_continue`. If anyone wanted the original behavior back, they could just call it first in `vcl_recv`.

nigoroll:
notes on my homework:

> (22:57:05) slink: OK, can we settle on 100-C or close?
> (22:57:22) phk: I'm still trying to graph my way through this.
> (22:57:54) slink: ok can you send email/PR note when you got news and/or can we continue later?
> (22:57:57) phk: I think we only need a "std.late_100_continue()" function.
> (22:58:07) slink: Need to do the early morning kids service station tomorrow
> (22:58:14) phk: if we end up in vcl_synth{} we can send final response and be done.
> (22:58:22) slink: oh std.late_100_continue() in principle, yes
> (22:58:28) phk: if we do anything else, we will know to send 100-C or not, depending on VRB
> (22:58:46) slink: but I see the point that for VCL clarity, you'd want to change your mind.
> (22:58:56) phk: and we will (or at least _can_) know that, before doing socket ops with backend.
> (22:59:12) slink: as in: assume late because we could bail out, but then for the POST paths, do early again
> (23:00:24) phk: but we need to figure out the Connection:close situation, and a way to veto it.
> (23:01:29) phk: Here's for you to ponder:  If you get std.late_100_continue(), and it only sends 100-C if we (will) do something with VRB, would you be able to remove the unwanted Connection:close from VCL, if it were automatically applied ?
> (23:01:50) phk: or would the C:close be unwanted in your case ?
> (23:02:05) slink: I'd take this as homework for the coming days
> (23:02:09) phk: check.
>

bsdphk:
Quick summary of IRC discussion:

1. We always optimize Varnish based on the assumption that traffic is legit, so early/immediate 100-C will continue to be the default.

2. Adding a std.late_100_continue{} which postpones the decision until we have VRB-activity to guide us before doing socket-ops to backend seems feasible.

3. Not sending 100-C, be it vcl_synth{} or because we contact backend, exposes us to the "did they or didn't they send the body anyway" race.  We need a transport->method() (->dont_send_100c() ?) which for H1 causes "Connection:close" on the response to the client.

4. Preferably it should be possible to remove that "Connection: close" again from VCL, if it is known that the client will be patient (enough) about the 100-C.

5. For H2, the stream is not reused and the transport->method() is a functional no-op as seen from VCL, even though activity happens at H2 frame-level.

nigoroll:
to be continued in  #2215

hermunn:
Backport review: This is moved to #2215, and will not be backported here.

PR #2198, opened 2017-01-25
Added VTC for different directors in a same stack (fallback+rrb)

JonathanHuot:
Hi,
An existing VTC already exists for directors stack (d00006.vtc), but it involves only a single director's type (round_robin).

This PR brings another very typical use-case of two directors inside a single stack: a fallback director containing two round-robin directors.
Typical usages are to use it alongs regional load-balancing (primary pool, secondary pool), or when using active/passive load-balancing.

Because I think this use-case is standard, later on, an example could be added into the documentation.

### DIFF
--- a/bin/varnishtest/tests/d00027.vtc
+++ b/bin/varnishtest/tests/d00027.vtc
@@ -1,0 +1,90 @@
+varnishtest "Test vmod.directors round robin and fallback directors stacked"
+
+server s1 {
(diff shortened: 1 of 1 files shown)


JonathanHuot:
I don't understand how I could have miss d00005.vtc ! I'm closing the PR.

hermunn:

Issue #2154, opened 2016-12-05
Add the current VXID to the Begin log record


Dridi:
There is currently no way that I know of to use a vxid in a VSL query, because no record exist in transaction with a field containing the transaction's id.

The closest one is `Begin`:

```
%s %d %s
|  |  |
|  |  +- Reason
|  +---- Parent vxid
+------- Type ("sess", "req" or "bereq")
```

In order not to break, I would suggest this change:

```
%s %d %s %d
|  |  |  |
|  |  |  +- Current vxid
|  |  +---- Reason
|  +------- Parent vxid
+---------- Type ("sess", "req" or "bereq")
```

It shouldn't break existing tools and allow lookups like `varnishlog -q "Begin[4] == $vxid"`.

fgsch:
You can already do this using grouping (e.g. -g request).

Dridi:
I don't know how you'd query a specific vxid.

fgsch:
This will bring the whole request but should do it: `varnishlog -g request -q "Begin[2] == $vxid"`.

Dridi:
`Begin[2]` would be the parent transaction, not the one I'm looking for.

fgsch:
It will match on the child so you will get the parent. 
That said, to cover all cases you also need to check Link[2].

Dridi:
Yes, I can either query a parent or a child vxid, not the *current* one. So depending on the grouping I need I may not get the results I expect.

What I'm proposing is to fix having to work around the main problem. I pick an `X-Varnish` header on an interesting transaction and want to query one of the vxids it's showing (eg. using `-r` or `-d`) but today I have to deal with relatives and proper grouping.

bsdphk:
Dridi, the current VXID is present in all VSL records, but it is a binary field which is not part of the text and not normally rendered.

I'll leave it to Martin to find out if the query language should be extended or if rendering it in begin's text is smarter.

Dridi:
I know, you get them with `varnishlog -v` but it doesn't solve the query limitation. Assigning to @mbgrydeland.

mbgrydeland:
Moved to different TODO list. Closing ticket due to it not being a bug.

hermunn:
Backport review: Nothing commited, nothing to backport.

PR #2199, opened 2017-01-26
Introduce a `vxid` pseudo tag for VSL queries

Dridi:
Following #2154 the consensus was to extend the query language by supporting `vxid` in expressions.

Before it can be merged (if OK'd) I need a response for the XXX question, and the other XXX comment is something I'm not satisfied with and I plan to remove the magic. Also there is absolutely no documentation, if OK'd I plan to mention it in the `vsl-query` manual.


lib/libvarnishapi/vxp_parse.c
```diff
@@ -112,14 +112,20 @@ vxp_expr_lhs(struct vxp *vxp, struct vex_lhs **plhs)
 
 	while (1) {
 		/* The tags this expression applies to */
-		if (vxp->t->tok != VAL) {
+		if (vxp->t->tok == VXID) {
+			i = SLT__VXID;
+			vsl_vbm_bitset(SLT__VXID, (*plhs)->tags);
+		}
+		else if (vxp->t->tok != VAL) {
```
fgsch: nitpick: else should be in the same line as closing bracket.

lib/libvarnishapi/vxp_parse.c
```diff
@@ -112,14 +112,20 @@ vxp_expr_lhs(struct vxp *vxp, struct vex_lhs **plhs)
...
 			VSB_printf(vxp->sb, "Expected VSL tag name got '%.*s' ",
 			    PF(vxp->t));
 			vxp_ErrWhere(vxp, vxp->t, -1);
 			return;
 		}
-		i = VSL_Glob2Tags(vxp->t->dec, -1, vsl_vbm_bitset,
-		    (*plhs)->tags);
+		else {
```
fgsch: same here.

fgsch:
@Dridi Nitpicks aside I have not had time to review the code in any detail yet.

bsdphk:
This one is martins call.

include/vapi/vsl_int.h
```diff
@@ -91,6 +91,7 @@ enum VSL_tag_e {
 	SLT__Bogus = 0,
 #define SLTM(foo,flags,sdesc,ldesc)	SLT_##foo,
 #include "tbl/vsl_tags.h"
+	SLT__VXID = 253, /* XXX: does it belong after SLT__Reserved? */
```
mbgrydeland: It belongs after SLT__Reserved. But I don't want to change the value of SLT__Reserved, as that will introduce compatibility issues. So I think you should avoid adding the VXID as a pseudo-header altogether, and use a flag in struct vex_lhs to show that the left hand side of the expression is the VXID of the trans, instead of a log record.

Dridi: I suspected this kind of answer, that's why I asked in the first place. I will try a new patch, thanks.

mbgrydeland:
Having looked over the code, the proposed solution looks a bit too heavy handed. It will for every transaction processed create a string that looks like a record containing the vxid. While this will work with the least amount of code, it also means that (since we can assume the test will be an integer type test) the string will be converted back to an integer for matching. Especially in -g raw this will cause a lot of unnecessary processing.

I think that instead of reusing the vslq_test_rec() function to match against something that looks like a record, we should add a new test funciton vslq_test_vxid(). The execution code then knows to call vslq_test_vxid() instead of vslq_test_rec() when the vex_lhs is marked as a vxid instead of a tag bitmap (the parsing step creates the right type of vex_lhs, see also comment on the XXX line in the patch).

The vslq_test_vxid then should do simple integer match against the vex_rhs, when the right hand side is of type int. The question is if we need to support the other rhs types as well for matching on vxid's. If we want that (not sure I see the use cases), it will then need to do the conversion of the input to the right type (float, or string if the rhs is string compare or regex). If we don't want that, the parser should probably error if a different rhs type than int is given in the expression.

The parsing step should also have some error checking for validity of the lhs when we introduce the new vxid type there. E.g. specifying a prefix for a vxid lhs wouldn't make sense.

Your thoughts on that proposal?

Dridi:
> create a string that looks like a record containing the vxid

The rationale was that all records were already stringly typed with possible conversions depending on operators. So this was more of a "not making assumptions" regarding how it will be used.

> The question is if we need to support the other rhs types as well for matching on vxid's. If we want that (not sure I see the use cases),

For instance `vxid ~ "(id1|id2)"` as a lazy shorthand for the `or` operator. If we start with a subset of the operators, we can always add more later, not the other way around.

> E.g. specifying a prefix for a vxid lhs wouldn't make sense.

I did think about that (including `{depth}` and `[field]` operators) and figured it would be simpler to support the general cases, and only document the sensible cases. Regarding the prefix, I could argue that it only makes sense for a subset of the tags.

> Your thoughts on that proposal?

For a first dive in the VSL code without much hints I'm OK with that patch, like you said: "this will work with the least amount of code". I will attempt a new patch with all your suggestions, but then I have a question:

What operators should `vxid` support as a bare minimum? I can only see the `==` operator as a sensible operator for `vxid`. Either to get a specific transaction (eg. from a guru meditation synthetic page or an `X-Varnish` header) or to filter out transactional records when grouping with `raw`.

mbgrydeland:
> What operators should vxid support as a bare minimum?
The bare minimum I'd say are the integer based operators. I agree that the only sensible ones are == and !=, but I do not see any reason not to be complete on integers (<, >, <=, >=).

It is right like you say that some query features like the prefix only makes sense for a subset of the records, and a silly query using prefix matching on a non-prefix type record would not give any error messages and just not match anything.

The vxid matching is different when it comes to executing queries, but will present to the user just like the other records. So I think we either mimic the behaviour of the record matching ones completely (even when they are silly and wouldn't ever match on a vxid), or we give a parse time error. So the behaviour of e.g. the field is a non-match if the field is anything but 0 (complete string) or 1 (first element). If a prefix is given on a vxid would then be always non-match. But I wouldn't like a (silly but otherwise valid) query like 'vxid ~ "^0$"' to silently just not match anything. So we either then have to on demand convert it to a string and run the regex matching, or give a parse error. I think I prefer parse error.

Dridi:
@mbgrydeland patch updated, sill no documentation. I will do that if it passes your review.

Dridi:
Actually it seems that I broke something.

Dridi:
I pushed a fix, copy-paste accident...

```diff
diff --git i/lib/libvarnishapi/vxp_parse.c w/lib/libvarnishapi/vxp_parse.c
index e788136..d39cdd7 100644
--- i/lib/libvarnishapi/vxp_parse.c
+++ w/lib/libvarnishapi/vxp_parse.c
@@ -330,7 +330,7 @@ vxp_expr_cmp(struct vxp *vxp, struct vex **pvex)
        vxp_expr_lhs(vxp, &(*pvex)->lhs);
        ERRCHK(vxp);
 
-       if (&(*pvex)->lhs->vxid) {
+       if ((*pvex)->lhs->vxid) {
                vxp_vxid_cmp(vxp);
                ERRCHK(vxp);
        }
```

Dridi:
I will squeeze in a doc update today!

Dridi:
83cbb7f
936c339
0930c1e
b28c352

PR #2208, opened 2017-02-09
Rework once more how +/- is handled

fgsch:
If the types are different and we are in a string context we need a string first or in-between the types.

### DIFF
--- a/bin/varnishtest/tests/v00020.vtc
+++ b/bin/varnishtest/tests/v00020.vtc
@@ -77,7 +77,6 @@ varnish v1 -vcl {
 		set req.http.foo = now - 1s;
 
 		set req.http.foo = now - now;
-		set req.http.foo = 1 + 1s;
 
 		set req.http.foo = 1 + 1;
 		set req.http.foo = 1 - 1;
@@ -170,6 +169,13 @@ varnish v1 -errvcl {TIME + TIME not possible.} {
 	}
 }
 
+varnish v1 -errvcl {TIME + INT not possible.} {
+	backend b { .host = "127.0.0.1"; }
+	sub vcl_recv {
+		set req.http.foo = now + 1;
+	}
+}
+
 varnish v1 -errvcl {INT + STRING not possible.} {
 	backend b { .host = "127.0.0.1"; }
 	sub vcl_backend_response {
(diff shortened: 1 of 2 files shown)


fgsch:
To be discussed in the next bugwash.

fgsch:
Some background
=============

Before d25ce0b57ae73e8163e8c3376e91996242a135dd (i.e. before very recently and in 4.0, 4.1 and 5.0), the add (+) operation in a STRING context will work depending on the first operand [1]. e.g. these fail:

```
set req.http.foo = now + 1
set req.http.foo = now + 1.0
```

OTOH, these works:

```
set req.http.foo = 1 + now
set req.http.foo = 1.0 + now
```

The value will be `1Sat, 11 Feb 2017 21:16:34 GMT` and `1.000Sat, 11 Feb 2017 21:16:34 GMT`.

In addition, before we failed statements like ```set req.http.foo = 1s + now``` while now results in ```1.000Sat, 11 Feb 2017 21:29:11 GMT```. 

Perhaps the more confusing is the previously not supported ```set req.http.foo = 1 + 1s``` which now results in ```11.000```.  Note that at first sight the result looks like a DURATION.

In a DURATION context the previous statement fails because adding an INT and a DURATION is not allowed.

This PR
======

This somewhats revert d25ce0b57ae73e8163e8c3376e91996242a135dd and also forbids some statements like ```set req.http.foo = 1 + 1s``` and ```set req.http.foo = 1 + now```.
For them to work, one needs to "instruct" vcc that it wants a STRING by ensuring the first operand is folder to a STRING, e.g. to get the same results as before:

```
set req.http.foo = now + "" + 1
set req.http.foo = now + "" + 1.0
set req.http.foo = 1 + "" + 1s
```

Making statements such as ```set req.XXX = 1s + 1``` (XXX being ttl or some header) invalid regardless of the context.

This removes some ambiguity and it's more intuitive than the existing behaviour IMO.
I also believe it's easier to communicate and for the users to understand.

1. Except for the allowed list available at https://github.com/varnishcache/varnish-cache/blob/master/lib/libvcc/vcc_expr.c#L1005

fgsch:
Discussed at length today. We will stick with the current code.
The rules have been distilled to:

- If we have A = B op C, and "B op C" is not legal, we convert (if possible) B to type of A.
- if that makes "A(B) op C" legal, then we're cool, if not we fail.

Until the next time, case closed.

PR #2215, opened 2017-02-15
Fix/Improve H1 Expect & 100 Continue / Logging / Start H2 Expect handling

nigoroll:
supersedes #2196 

on @bsdphk 's comments:

> We always optimize Varnish based on the assumption that traffic is legit, so early/immediate 100-C will continue to be the default.

done, but I have fixed the default behavior:
* No 100-C if we already read some of a body
* No 100-C for < HTTP/1.1

> Adding a std.late_100_continue{} which postpones the decision until we have VRB-activity to guide us before doing socket-ops to backend seems feasible.

Done.

> Not sending 100-C, be it vcl_synth{} or because we contact backend, exposes us to the "did they or didn't they send the body anyway" race. We need a transport->method() (->dont_send_100c() ?) which for H1 causes "Connection:close" on the response to the client.

I have only implemented this for the synth case and kept the 100-C for the fetch case for now. I _think_ that for fetch with late 100, we should actually add back the Expect header and let the backand decide. The alternative would be to read the backend response headers before starting VRB handling.

The Expect header now gets added back for pipe with late 100.

> Preferably it should be possible to remove that "Connection: close" again from VCL, if it is known that the client will be patient (enough) about the 100-C.

(edited, see additional comment below):I don't think this is related to the client, but rather to VCL policy: For a synth response, do we want to just send our response and be done, or do we want to drain-read the client body? I've now made the former the default and the latter possible.

>    For H2, the stream is not reused and the transport->method() is a functional no-op as seen from VCL, even though activity happens at H2 frame-level.

I've kept the H2 code as-is for now. DATA support is missing anyway and I'd like to avoid wasting more time in case the H1 part needs more attention.

### DIFF
--- a/bin/varnishd/cache/cache_req_body.c
+++ b/bin/varnishd/cache/cache_req_body.c
@@ -87,7 +87,7 @@ vrb_pull(struct req *req, ssize_t maxsize, objiterate_f *func, void *priv)
 	yet = req->htc->content_length;
 	if (yet != 0 && req->want100cont) {
 		req->want100cont = 0;
-		(void) req->transport->sresp(req, 100);
+		(void) req->transport->minimal_response(req, 100);
 	}
 	if (yet < 0)
 		yet = 0;
--- a/bin/varnishd/cache/cache_req_body.c
+++ b/bin/varnishd/cache/cache_req_body.c
@@ -37,6 +37,7 @@
 #include "vtim.h"
 #include "hash/hash_slinger.h"
 #include "storage/storage.h"
+#include "cache_transport.h"
 
 /*----------------------------------------------------------------------
  * Pull the req.body in via/into a objcore
@@ -84,6 +85,10 @@ vrb_pull(struct req *req, ssize_t maxsize, objiterate_f *func, void *priv)
 	AZ(req->req_bodybytes);
 	AN(req->htc);
 	yet = req->htc->content_length;
+	if (yet != 0 && req->want100cont) {
+		req->want100cont = 0;
+		req->transport->sresp(req, R_100);
+	}
 	if (yet < 0)
 		yet = 0;
 	do {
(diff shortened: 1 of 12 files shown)


nigoroll:
I just realized that we can be a bit more flexible in cnt_synth

bsdphk:
What happens/should happen if the client sends C-L: 0 and Expect: 100-continue ?

There is no need to have http1_simple_response_msg{}[], we already have the strings once (include/tbl/...) and we have no intrinsic reason to limit the xport->sresp() method to just those three.  Furthermore, that design means we'd have to replicate a similar table in h2, h3, h4...

"sresp()" is not very intuitive, something like "minimal_response()" or "blank_response()" might make more sense.

with "l = write(req->sp->fd, m->resp, strlen(m->resp));" you should use the VTCP_ function to check for "acceptable errnos", so that we don't overlook something horrible like EBADF here.

nigoroll:
@bsdphk , again thank you for this blazing fast review :)

I've added more commits in the hope that these will make additional reviews easier, but for a final merge I'd squeeze them.

> What happens/should happen if the client sends C-L: 0 and Expect: 100-continue ?

Thank you for spotting this horrible oversight on my side!

I hope to have also reflected your other comments appropriately.

nigoroll:
had to rebase for production build

bsdphk:
OK for commit.

We didn't use to VSL the 100 responses, but as I read the patch we do now.  We should probably check with Martin if that is a good idea from a varnishlog point of view.  If it is a problem, we can make the VSL'ing conditional on the status being >= 200.

nigoroll:
@mbgrydeland could you please review phks comment in VSL? What's your opinion, should we log the 100? I think we should...

nigoroll:
@mbgrydeland I'm going to merge anyway, guess this has time to be changed later before release

nigoroll:
reworked the patches and manually merged with the following commits
5a0817d1446c64467745e50fb9d9ab0e791e9bec
70ba8e03d8f2ba9986756abb5cc6fb50857dbea5
5af2605607a28f62efbc70203f65cc5fb154fbdc
88d1d1dce4426584003c6b690436c96c3c3f28fa

hermunn:
Backport review: Since the code is very much influenced by H2, this will not be back ported to 4.1 now, and probably not ever.

PR #2216, opened 2017-02-15
Make sure Age is always less than max-age

rezan:
By rounding Age down, we make sure Age < max-age while the object
is fresh. Otherwise, we can prematurely get Age == max-age and Varnish
will calculate that as a 0s TTL and create a pass scenario.

More info can be found in this dev list email:

https://www.varnish-cache.org/lists/pipermail/varnish-dev/2016-December/009079.html


rezan:
Not sure why Travis is failing here, that vtc is unrelated and passes before and after on my local.

gquintard:
I restarted the job, and it seems to pass. It was probably just an issue with acquiring ports.

bsdphk:
OK.

I worry that the test-case may be too timing sensitive, so keep an eye on Vtest until we know.

hermunn:
Backport review: The commit 104af2ae has been backported as 14ce4804.

PR #2218, opened 2017-02-16
RFC: carve objects from power-of-two chunks

nigoroll:
This is a simple idea I've been carrying around with me for a long time, and it popped up again while trying to understand weird jemalloc behavior:
While, from a performance perspective, minimizing the number of IO vectors we need to deliver some object is optimal, from a memory footprint perspective minimizing the memory allocator overhead should be optimal.
So shouldn't we offer an option to try to help the allocator by only ever requesting allocations in sizes of powers of two?
I do not mean to propose this patch as-is - instead we should make this configurable at least for the stevedore, maybe from vcl. But to do so, we'd probably need to hand the allocation flags down to the stevedore.
Comments?

### DIFF
--- a/bin/varnishd/storage/storage_simple.c
+++ b/bin/varnishd/storage/storage_simple.c
@@ -45,6 +45,27 @@
 /* Flags for allocating memory in sml_stv_alloc */
 #define LESS_MEM_ALLOCED_IS_OK	1
 
+// https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+static inline unsigned int
+rnddn_pow2(unsigned int v)
+{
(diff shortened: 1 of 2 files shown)


bsdphk:
Interesting idea.

Best way to find out is to run it and measure...

nigoroll:
I'd want to run a canary with otherwise equally configured and used servers side by side when #2215 or anything comparable is in

nigoroll:
This won't get any attention from myself for the next three weeks, so closing.
The sole purpose of the PR was to get feedback anyway.

hermunn:
Backport review: Nothing committed, nothing to backport.

PR #2220, opened 2017-02-17
Add a setenv command to varnishtest

daghf:
This lets you set environment variables in a test case. Useful for test cases for VMODs using libraries that are configured via environment variables.

Usage is

    setenv "FOO" "bar"

which will set FOO=bar if FOO does not already exist.

There is also an optional '-force' argument, which will overwrite an
existing variable if already set:

    setenv -force "FOO" "baz"

The force argument corresponds to passing a non-zero third argument to
setenv(3).

### DIFF
--- a/bin/varnishtest/tests/m00028.vtc
+++ b/bin/varnishtest/tests/m00028.vtc
@@ -1,0 +1,73 @@
+varnishtest "Test setenv"
+
+setenv FOO "BAR BAZ"
+
+varnish v1 -vcl {
+	import std from "${topbuild}/lib/libvmod_std/.libs/";
+
+	backend dummy { .host = "${bad_ip}"; .port = "9080"; }
+
+	sub vcl_recv {
+		return(synth(200));
+	}
+
+	sub vcl_synth {
+		set resp.http.X-FOO = std.getenv("FOO");
+	}
+} -start
+
+client c1 {
+	txreq
+	rxresp
+	expect resp.http.X-FOO == "BAR BAZ"
+} -run
+
+varnish v1 -stop
+
+setenv -ifunset FOO QUUX
+
+varnish v2 -vcl {
+	import std from "${topbuild}/lib/libvmod_std/.libs/";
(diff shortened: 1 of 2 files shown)


bsdphk:
Do we need the -force argument ?  It sounds like a recipe for test unpredictability if it were ever used...

Dridi:
@bsdphk, are you suggesting to always or never override?

daghf:
@bsdphk The idea was that there could be case where someone would want permit overriding from the command line,

<pre>
$ FOO=bar varnishtest test.vtc
</pre>

I don't personally use this, and I'd be OK with just dropping `-force` and make it always enabled.

hermunn:
What if the default is to `-force` and have an option `-noforce` or similar? That should make it very unlikely that someone overrides by mistake.

Dridi:
@hermunn why not but I prefer the other way around. By default I can do some manual fiddling unless there was a good reason (-force) not to.

bsdphk:
Are there any examples of what kind of environment variables we talk about, and what they are used for?

daghf:
@bsdphk,

for my particular case, it's OpenSSL's `SSL_CERT_DIR` / `SSL_CERT_FILE`, which are used for specifying custom CA certificates.

bsdphk:
I think I prefer if we make "respect previous value" the option and default to setting them hard.

`setenv [-ifunset] var val
`
maybe ?

daghf:
+1 to that.

@Dridi @hermunn comments?

If not, I'll change the pull request to reflect the suggestion in @bsdphk latest comment.

Dridi:
As long as we keep the mechanism, I'm fine.

daghf:
@bsdphk  PR updated

bsdphk:
OK

daghf:
Merged.

hermunn:
Backport review: Backported dcdb2504e5c06 as 09ce321d10e797bcb.

PR #2221, opened 2017-02-17
Introduce a ZERO_OBJ macro similar to bzero

Dridi:
`explicit_bzero(3)` is coming to glibc and I was thinking that it might be the occasion to hint compilers that we really want the erasure of magic numbers. Instead of having different interfaces for zeroing I hid the details behind a `ZERO_OBJ` macro.

I haven't actually tested this on a system that has the function, but I believe FreeBSD does. I can only confirm the fallback works so far.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -194,6 +194,7 @@ AC_CHECK_HEADERS([pthread_np.h], [], [], [#include <pthread.h>])
 AC_CHECK_HEADERS([priv.h])
 
 # Checks for library functions.
+AC_CHECK_FUNCS([explicit_bzero])
 AC_CHECK_FUNCS([nanosleep])
 AC_CHECK_FUNCS([setppriv])
 AC_CHECK_FUNCS([fallocate])
--- a/include/miniobj.h
+++ b/include/miniobj.h
@@ -5,6 +5,12 @@
  *
  */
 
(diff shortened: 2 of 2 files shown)


bsdphk:
I don't see the point of doing it in INIT_OBJ() If the compiler can see that far, I want its optimization.

Doing it in FREE_OBJ() makes good sense, that's an error detection thing.

Dridi:
> I don't see the point of doing it in INIT_OBJ() If the compiler can see that far, I want its optimization.

I'm not trying to force `explicit_bzero` during init, as I said it was that:

> Instead of having different interfaces for zeroing I hid the details behind a ZERO_OBJ macro.

I did so assuming that this function only prevents "dead" assignment removal, not other kind of optimizations. If that is the case, I'm disappointed in &lt;insert compiler&gt; and will update my patch to leave init alone. For example, I'm assuming that a call to `explicit_bzero` could be avoided for the magic number in favor of a simple assignment, considering it's only zeroing a simple field.

Dridi:
Updated patch leaving init alone

bsdphk:
OK

Dridi:
983555d

hermunn:
Backport review: 983555d7 backported as 768a00c1.

PR #2222, opened 2017-02-19
Fix typo in unadulterated

dunglas:


fgsch:
Applied. Thanks for the PR.

PR #2223, opened 2017-02-20
add WS_Assert_Allocated: assert that ws->s <= a memory region < ws->f

slimhazard:
Verify that the region (ptr, ptr + len) lies strictly in the part of a workspace previously allocated with one of WS_Alloc, WS_Printf and friends. If len < 0, then check with len = strlen(ptr) + 1 (includes a terminating null byte in the check).

My use case for this is checking PRIV_TASK data that should have been previously allocated in the task's workspace, especially to ensure that subsequent uses of WS_Alloc and friends don't write over it.

```
if (priv->priv == NULL) {
    priv->priv = WS_Alloc(ctx->ws, sizeof(widget));
    // [...]
}
else {
    WS_Assert_Allocated(ctx->ws, priv->priv, sizeof(widget);
   // [...]
}
```


bsdphk:
OK

slimhazard:
Merged

hermunn:
This was back ported as 456ee023f117b, 68f15e0e402958f some time ago.

PR #2232, opened 2017-02-24
Introduce varnishd -f [name[,label]=]file (discussion, not to be merged)

Dridi:
This is a response to d306f91 that I reviewed today:

> The plan is to also allow both specification of an (optional) label and
which VCL is to be the active one, but I want to think about if we should
shoehorn that into -f or use a different option letter.

I've done little thinking on this and think that using a different flag would make things too complicated because we'd need to correlate `boot${N}` names to labels. Instead I'd like to suggest using a syntax somewhat similar to storage options, allowing to set both the name and the label.

I chose to put the label before the equal sign because shoehorning it there was easier to quickly do. But having it after the file name (like `-a`, `-s`, `-j` etc) could also help decide which one is to be active. With this PR, it's still the last `-f`.

In the absence of a name, it falls back to the `boot${N}` automatic name. It also makes things less convoluted regarding VCL dependencies. Supposing a `-L` option to declare labels, one would need to jump between flags but links would also look more explicit:

```
# prod references the labels www and static (eg. independent sub-domains)

# this PR syntax
varnishd -f www1,www=www.vcl -f static1,static=static.vcl -f prod=prod.vcl

# the -L example
varnishd -f www1=www.vcl -f static1=static.vcl -L www=www1 -L static=static1 -f prod=prod.vcl

# you may later relink www to www2, ww3 and so on...
```

Other parameters have comma-separated arguments at the end which would be fine too, but to start the discussion I took the easy route.

Pros: you can start Varnish with any VCL (including the active one) referencing labels. Name collisions are rejected, and the error message could be improved (once again, quick'n'dirty).

Cons: users can no longer freely use VCL file names containing an equal sign or a comma. I can live with that :)

The reason why I baked this quickly was to hijack the discussion on the `-C` option too (ping @nigoroll). Now that we can have multiple `-f` flag, I think that `-C` concatenates every single C file in the output. I haven't checked but that's how it looks.

Instead I would like to suggest adding an optarg to `-C`: a glob expression matching VCL names. I don't think that dumping them all by default would help, as myself a user wanting to inspect generated code I would likely want to see the C code for one VCL at a time (and it would now work even when labels are involved).

As a side note,  ~~current master leaks `struct f_arg.farg`, and~~ I found a bit of a conflict (see 917773a).

### DIFF
--- a/bin/varnishd/mgt/mgt.h
+++ b/bin/varnishd/mgt/mgt.h
@@ -182,8 +182,8 @@ char *mgt_VccCompile(struct cli *, struct vclprog *, const char *vclname,
     const char *vclsrc, const char *vclsrcfile, int C_flag);
 
 void mgt_vcl_init(void);
-void mgt_vcl_startup(struct cli *, const char *vclsrc, const char *origin,
-    const char *vclname, int Cflag);
+int mgt_vcl_startup(struct cli *, const char *vclsrc, const char *vclname,
+    const char *vcllabel, const char *origin, int Cflag);
 int mgt_push_vcls_and_start(struct cli *, unsigned *status, char **p);
 void mgt_vcl_export_labels(struct vcc *);
 int mgt_has_vcl(void);
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -440,7 +440,8 @@ mgt_f_arg(struct f_arg *fa)
 	if (p != NULL) {
 		q = memchr(fa->farg, ',', p - fa->farg);
 		if (q != NULL) {
-			fa->label = strndup(q, p - q);
+			/* Off by one to avoid the comma */
+			fa->label = strndup(q + 1, p - q - 1);
 			AN(fa->label);
 		}
 		else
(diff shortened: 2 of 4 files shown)


Dridi:
In the rush I forgot to summarize my actual answer to d306f91:

- `-f [<name>=]file[,label=<label>][,state=<auto|cold|warm|active>]`
- `-C <glob>`

bsdphk:
The reason I didn't do anything like this is that it gets too complex.

I'm currently leaning towards:

1.  An argument (-I ?) which specifies a file containing CLI commands.

2. Interpret arguments after flags as CLI commands.

`varnishd -n ...  "vcl.load foo /etc/foo.vcl" "vcl.label bar foo" ...`

In both cases it is probably a good idea to adopt make(1)'s "if it starts with minus it's ok to fail" convention.  It may even make sense to let this be the case for all CLI commands.

bsdphk:
I bit the bullet and implemented -I cli_file (9e27227550179cb2cc0e8619fc507079efe29130)

hermunn:
Backport review: This is new functionality that will not make it into 4.1.

Issue #1206, opened 2016-03-04
Pass shouldn't honor IMS/INM if the backend doesn't


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1206

Issue #2042, opened 2016-08-16
Send a 304 response for a just-gone-stale hitpass object when appropriate


slimhazard:
When the hit-for-pass TTL for an uncacheable object expires (that is, on the first request after Varnish stops passing, so that we have a cache miss and go to fetch ), the backend might send a response with an ETag header that matches If-None-Match in the client request. In that case, there's no reason why the client response shouldn't be 304. But evidently resp.status is always 200 in this case.

Presumably the same thing happens when req.http.If-Modified-Since < resp.http.Last-Modified just after the TTL elapses for a hitpass object (but I haven't tried it).
## Expected Behavior

We set beresp.uncacheable to true and beresp.ttl > 0 in vcl_backend_response -- meaning we set the object to hit-for-pass.

After the beresp.ttl elapses, suppose that the backend sets the ETag header, which in turn sets resp.http.ETag, and suppose further that its value matches req.http.If-None-Match. Then I expect client response status 304.
## Current Behavior

Under these conditions (beresp.ttl just elapsed for a hitpass object, and req.http.If-None-Match==resp.http.ETag), Varnish apparently always sets resp.status to 200, and sends the response body.

I'll post a gist with a VTC to illustrate what happens.

Note that while beresp.ttl is unelapsed, so that the object is still in hitpass, the If-None-Match header is passed along to the backend, and if the backend responds with 304 (because INM matches ETag), then status 304 is passed along to the client. The effect is that clients get 304 and no response body most of the time, but suddenly get 200 at random-seeming intervals, despite INM matching ETag.
#1858 fixed the backend request, so that INM and IMS are not sent to the backend for just-elapsed hitpass objects. That's correct, and it may lead to the fetch after the miss getting a 200 response from the backend along with the beresp body. But there's no reason not to set the client response to 304 when the conditions in the client request and response allow it.
## Possible Solution

The 304 client response is set by this code at the end of cnt_deliver:

```
        if (!(req->objcore->flags & OC_F_PASS)
            && req->esi_level == 0
            && http_IsStatus(req->resp, 200)
            && req->http->conds && RFC2616_Do_Cond(req))
                http_PutResponse(req->resp, "HTTP/1.1", 304, NULL);
```

At the moment I don't see why the 304 response isn't set. After the hitpass-TTL elapses, the OC_F_PASS should not be set, so that shouldn't be what's preventing it. Maybe req->http->conds is false?
## Steps to Reproduce (for bugs)

[...]

PR #2234, opened 2017-02-27
Roll back the fix for #1206, so that client delivery solely decides if resp.status==304

slimhazard:
whether to set client response 304, regardless of the pass state of
the backend fetch. That is, set resp.status 304 (at ESI level 0)
iff req.http.{INM,IMS} matches resp.http.{ETag,Last-Modified}.

This means that the client will get a 304 response in the pass
cases where the backend sets ETag or Last-Modified to match INM/IMS,
but nevertheless declines to set beresp.status to 304. This behavior
is consistent for the three kinds of pass: "ordinary" pass (return(pass)
in vcl_recv), beresp.uncacheable ("hit-for-miss"), and the recently
revived hit-for-pass (return(pass(DURATION)) in vcl_backend_response).

r02042.vtc shows how to restore the behavior from #1206, by unsetting
INM/IMS in vcl_pass (for ordinary pass) or vcl_miss (for hit-for-miss).


slimhazard:
This and PR #2235 are two possible fixes for issue #2042 -- we'll need one or the other of them, not both. See #2042 for details.

fgsch:
That's a hell of a large test. Do we really need it to be that large?
Also, can you normalize whitespace in the test?

slimhazard:
Yeah, you're right on both counts. The test is in part to clarify some things, and since the docs will want to have some VCL snippets, this way we know what works. But not all of it has to live in Varnish forever. And sorry about the whitespace.

slimhazard:
We decided for this one in today's bugwash, fixed in  6b518c4.

hermunn:
Backport review: For 4.1 the patch goes through, but the test case for #2042 does not compile, and it is not immediately clear to me how to change the test case. Because of this I will not backport it now, but I will leave the backport review tag for now.

hermunn:
We will keep this out of 4.1 - keeping consistent behavior is more important than the feature this PR implements.

PR #2235, opened 2017-02-27
Consider setting resp.status=304 in delivery if the fetch was hit-for-pass

slimhazard:
... as recently revived (i.e. if OC_F_HFP is set).


slimhazard:
This and PR #2234 are two possible fixes for issue #2042 -- we'll need one or the other of them, not both. See #2042 for details.

slimhazard:
We decided for #2234 in today's bugwash, against this one, fix in  6b518c4.

hermunn:
Backport review: This was not commited, so only #2234 may be backported,

PR #2243, opened 2017-03-02
Vsl query booleans

nigoroll:
I'm suggesting this after having failed to train my grey matter to think `and` when wanting to express `&&` in vsl


nigoroll:
I overlooked a local test failure - broke the existing boolean verbs

Dridi:
I'm not convinced by the POLA-ness of this change, but having both should be OK.

nigoroll:
bugwash: @mbgrydeland prefers to reserve these operators for later use (possibly and/or operators with slightly different semantics)

hermunn:

PR #2244, opened 2017-03-03
Send PROXY headers with probes

### LINKED ISSUE #2151: Health probes do not honor a backend's PROXY protocol setting
I'm using varnish-5.0.0 with PROXY protocol enabled on the backends thanks to the new `.proxy_header` attribute. This works great as long as the backends don't have health probes defined. When a backend has both a `.proxy_header` attribute and a probe defined, varnish fails to properly detect the backend health state and marks the backend as sick.

The problem seems to be that varnish does not use the PROXY protocol when sending health probes. If the backend correctly implements the PROXY protocol, it is required to drop the connection, as it cannot identify it as valid PROXY protocol v1 nor v2. Thus, from varnish's pov the backend is unreachable.

## Possible Solution
Health probes should honor the backend's `.proxy_header` attribute and send the correct PROXY header where appropriate.

Dridi:
Fixes #2151, ~~probably a good candidate for 4.1 too~~.

### DIFF
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -43,6 +43,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include <netinet/in.h>
+#include <netinet/tcp.h>
+
 #include "binary_heap.h"
 #include "vcli_serve.h"
 #include "vrt.h"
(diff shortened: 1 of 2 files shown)


fgsch:
Looks good to me.

We should document this too, perhaps something like this will suffice:

```
   ``.proxy_header``
     The PROXY protocol version Varnish should use when connecting to
-    this backend. Allowed values are ``1`` and ``2``.
+    this backend, including probes. Allowed values are ``1`` and ``2``.
```

hermunn:
Backport review: We will keep this a Varnish 5 only feature unless someone makes a really compelling argument.

PR #2246, opened 2017-03-06
Add documentation about cache misses and uncacheable content to the User's Guide

slimhazard:
This started out as documentation of the varieties of pass for the 5.1 release: pass from vcl_recv, hit-for-miss and hit-for-pass. Which I think we need now that we have two kinds of hit-for-something.

That in turn required some background about what happens for misses, especially request coalescing and the fact that fetches on cache misses cannot be conditional. So I went ahead and added a chapter just about misses, also documenting how you can help in VCL by setting grace and keep.

This has gone into "Achieving a high hitrate", because that seemed to me to be the place in the existing docs where the content seems to fit best. But it might be odd for that chapter to have a lot info about subjects that are not cache hits. If that's too odd, we can move the new content into a new document.

Also I noticed that the "Cookies" section had subsections that aren't about cookies: Cache-Control, Age, Authorization etc. So I changed the header levels, so that "Cookies" and those other sections are subsections of "The role of HTTP headers".

Still to do: go over the rest of the docs where there are references to hit-for-pass, and update them to mention both of the hit-for-somethings, or distinguish them properly.

Note that these updates are committing us to that terminology (we were still referring to hit-for-miss as hit-for-pass in the 5.0 docs), so now's the time to object if we don't want that.

### DIFF
--- a/doc/sphinx/users-guide/increasing-your-hitrate.rst
+++ b/doc/sphinx/users-guide/increasing-your-hitrate.rst
@@ -442,6 +442,23 @@ validated for a fairly long time. If the backend can send a 304
 response long after the TTL has expired, you save bandwith on the
 fetch; if not, then it's no different from any other cache miss.
 
+If, however, you would prefer that backend fetches are not
+conditional, just remove the If-* headers in ``vcl_backend_fetch``::
+
+  sub vcl_backend_fetch {
+    # To prevent conditional backend fetches.
+    unset bereq.http.If-None-Match;
+    unset bereq.http.If-Modified-Since;
+  }
+
+That should only be necessary if the conditional fetches are
+problematic for the backend, for example if evaluating whether the
+response is unchanged is too costly for the backend app, or if the
+responses are just buggy. From the perspective of Varnish, 304
+responses are clearly preferable; fetches with the empty response body
+save bandwidth, and storage does not have to be allocated in the
+cache, since the existing cache object is re-used.
+
 To summarize, you can improve performance even in the case of cache
 misses by:
 
@@ -517,6 +534,22 @@ default ``vcl_recv`` gets executed; so take a close look at ``vcl_recv``
 in ``builtin.vcl``, and duplicate any part of it that you require in
 your own ``vcl_recv``.
 
+As with cache hits and misses, Varnish decides to send a 304 response
+to the client after a pass if the client request headers and the
+response headers allow it. This might mean that Varnish will send a
+304 response to the client even after the backend saw the same request
+headers (``If-Modified-Since`` and/or ``If-None-Match``), but decided
+not to respond with status 304, while nevertheless setting the
+response headers ``ETag`` and/or ``Last-Modified`` so that 304 would
+appear to be warranted. If you would prefer that Varnish doesn't do
+that, then remove the If-* client request headers in ``vcl_pass``::
+
+  sub vcl_pass {
+    # To prevent 304 client responses after a pass.
+    unset req.http.If-None-Match;
+    unset req.http.If-Modified-Since;
+  }
+
 hit-for-miss
 ~~~~~~~~~~~~
 
@@ -542,6 +575,10 @@ are just like cache misses, meaning that:
 * fetches cannot be conditional, so ``If-Modified-Since`` and
   ``If-None-Match`` headers are removed from the backend request.
 
+When Varnish sees that it has hit a hit-for-miss object on a new
+request, it executes ``vcl_miss``, so any custom VCL you have written
+for cache misses will apply in the hit-for-miss case as well.
+
 ``builtin.vcl`` sets ``beresp.uncacheable`` to ``true``, invoking the
 hit-for-miss state, under a number of conditions that indicate that
 the response cannot be cached, for example if the TTL was computed to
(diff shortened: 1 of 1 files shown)


slimhazard:
Added some docs to reflect the rollback of #1206 in #2042, and how to restore the behavior of #1206 with VCL means. Generally, this part now documents how to prevent conditional backend fetches, and how to prevent the decision to send a 304 client response at delivery time, by removing the If-* request headers in the appropriate vcl sub.

Dridi:
Damn you @slimhazard, this is not the PR I accepted to review what with you adding more commits and stuff 😠

slimhazard:
@Dridi wha? what'd I do?

doc/sphinx/users-guide/increasing-your-hitrate.rst
```diff
@@ -451,3 +451,146 @@ misses by:
...
+cannot be seen until Varnish receives the backend response, so
+``vcl_recv`` can't know about it.
+
+By default, if a request is not passed and the backend response turns
+out to be uncacheable, the cache object is set to "hit-for-miss", by
+setting ``beresp.uncacheable`` to ``true`` in
+``vcl_backend_response``.  A minimal object is saved in the cache, so
+that the "hit-for-miss" state can be recognized on subsequent
+lookups. (The cache is used to remember that the object is
+uncacheable.)  In that case, no request coalescing is performed, so
```
Dridi: > (The cache is used to remember that the object is uncacheable.)

For a limited period of time.

doc/sphinx/users-guide/increasing-your-hitrate.rst
```diff
@@ -451,3 +451,146 @@ misses by:
...
+that fetches can run concurrently. Otherwise, fetches for hit-for-miss
+are just like cache misses, meaning that:
+
+* the response may become cacheable on a later request, for example
+  if it sets a positive TTL with ``Cache-Control``, and
+
+* fetches cannot be conditional, so ``If-Modified-Since`` and
+  ``If-None-Match`` headers are removed from the backend request.
+
+``builtin.vcl`` sets ``beresp.uncacheable`` to ``true``, invoking the
```
Dridi: Maybe mention "for 2 minutes" too.

Dridi:
@slimhazard I was kidding, and "complaining" that you made the scope of my review larger by adding one more commit. My comments are mainly nitpicks and overall I find this great material!

slimhazard:
@Dridi many thanks. Any thoughts about this part being in the section titled "Achieving a high hitrate", when it fact it's all about what happens when there's no cache hit? I'm still wondering if this should go into a separate chapter.

Dridi:
I agree "Uncacheable content" it could get its own file, I would keep "Cache misses" where it is.

doc/sphinx/users-guide/increasing-your-hitrate.rst
```diff
@@ -383,3 +383,71 @@ So if you *really* need to vary based on 'User-Agent' be sure to
...
+by default it performs a fetch from the backend on the hypothesis that
+the response might be cached. This has two important consequences:
+
+* Concurrent backend requests for the same object are *coalesced* --
+  only one fetch is executed at a time, and the other pending fetches
+  wait for the result. This is to prevent your backend from being hit
+  by a "thundering herd" when the cached response has expired, or if
+  it was never cached in the first place.  If it turns out that the
+  response to the first fetch is cached, then that cache object
+  satisfies the other pending requests.
```
slimhazard: Yeah, the :ref: is two paragraphs further down ("Misbehaving servers"), which is a pretty good writeup on grace. Hm, will think about where it should go.

doc/sphinx/users-guide/increasing-your-hitrate.rst
slimhazard: That's true of hit-for-pass, in the sense that hfp has a "TTL", but not hit-for-miss -- the object may become cacheable after the next fetch. That was the point of hfm, there was too much POLA about responses being uncacheable for two minutes, even though everything about them allowed for cacheability.

Dridi: Hit-for-miss inherits the `beresp.ttl`, which is set to 2 minutes in the built-in VCL.

slimhazard:
@Dridi the latest commit reflects your comments, and I also added paragraphs about the effect of the TTL in the hit-for-miss and hit-for-pass cases.

doc/sphinx/users-guide/increasing-your-hitrate.rst
```diff
@@ -649,8 +674,12 @@ headers, then remove the headers from the client request in
 ``vcl_pass``, as shown above for pass.
 
 The hit-for-pass state ends when the "hit-for-pass TTL" given in the
-``return`` statement elapses. The next request for that object is then
-handled as an ordinary miss.
+``return`` statement elapses. As with passes, the response to a
+hit-for-pass fetch is never cached, even if it would otherwise fulfill
+conditions for cacheability.  So unlike hit-for-miss, it is not
+possible to end the hit-for-pass state ahead of time with a cacheable
```
Dridi: > So unlike hit-for-miss, it is not possible to end the hit-for-pass state ahead of time with a cacheable response.

I'm not sure it's impossible to escape hit-for-pass, what about `hash_always_miss`?

slimhazard: Ah, well maybe, offhand I don't know either (would have to test it). But I don't think this part of the docs should try to explore the whole combinatorial space of everything that could happen in VCL, it would distract from the point of the text. And besides, the statement with this qualification is correct in any case:

> ... it is not possible to end the hit-for-pass state ahead of time **with a cacheable response**.

slimhazard:
Pushed to master

Dridi:
@slimhazard see 06f9176, we can amend the docs to point the hitpass escape hatch to the part of the manual where `hash_always_miss` is documented.

hermunn:
Backport review: This is definitely not accurate for 4.1, so we skip it. Maybe the 4.1 docs should get some of the changes, but that will be a separate effort if someone wants to put in the time.

hermunn:
Small documentation patches that are correct also for 4.1 are typically backported, but I will not spend much time to pick the parts that are relevant, at least not now.

PR #2248, opened 2017-03-07
Log hits-for-pass and -miss with HitPass and HitMiss tags

slimhazard:
The HitPass tag was already there, but evidently was not used in 5.0. The HitMiss tag is added.


mbgrydeland:
Looks good to me.

Martin

slimhazard:
Pushed to master

Issue #2251, opened 2017-03-09
varnishapi.pc and varnishconfdir


rezan:
This is more of a discussion right now, but I am proposing adding 2 new entries to `varnishapi.pc`:

```
sysconfdir=@sysconfdir@
varnishconfdir=@varnishconfdir@
```

Basically, im running into a situation where I need to know where the `vcl_dir` is (or `vcl_path`). The problem is that this path is prefixed when you install from source, but unprefixed when you install from package. In both cases, you are left guessing where is the actual location. Im thinking this info should be in the `varnishapi.pc`. Thoughts? Is there a better way to find the `varnishconfdir` after Varnish has been installed on a system?

rezan:
Im now thinking to just query `varnishadm` directly and then fallback to `varnishconfdir` detection. However, this problem is 2 parts. We want to drop vcls into vcl_dir, simple enough. We also want to compile a VMOD which default loads configuration from vcl_dir. For the latter, we need some kind of auto config var.

Dridi:
@rezan we've already had this discussion for 5.0 and as a result #2058 was submitted (but not merged as is). You already have a blessed directory for packaging where you can install VCL files and expect them to be readily available unless the user changes their `vcl_path`.

With pkg-config that would be `${pkgdatadir}/vcl/`.

It is documented in the changelog for 5.0 as "New system wide VCL directory: `/usr/share/varnish/vcl/`". There's maybe room for improvement in the documentation, but it's mentioned in varnishd(1) too:

http://varnish.org/docs/5.0/reference/varnishd.html#vcl-path

rezan:
ok, that worked too. I did this, since it matched the [Makefile.am in varnishd](https://github.com/varnishcache/varnish-cache/blob/e5b99ec618d95cdd11160ccfbdc6aa408c5333f4/bin/varnishd/Makefile.am#L149):

```
${datarootdir}/varnish/vcl
```

Still have a need to put a VMOD config file into `varnishconfdir`. The idea here is all the VMOD conf file sits next to the Varnish conf files and VCL. So this path would need to be known at compile time for the VMOD. Any thoughts on this one?

Dridi:
Sorry, I was too quick to close this. The reason why the Makefile doesn't use pkgdatadir is because it wasn't already there. I'd still recommend to take it from pkg-config (it's its purpose) and nothing prevents us from feeding automake a pkgdatadir variable in the varnish source tree.

[...]

PR #2252, opened 2017-03-11
Improvement for out-of-tree builds

Dridi:
See #2251, I haven't tested anything yet, feedback appreciated. I'll get around testing this at worst on Monday, it already includes documentation so browsing the diffs should give all context.

ping @rezan

### DIFF
--- a/bin/varnishd/Makefile.am
+++ b/bin/varnishd/Makefile.am
@@ -148,7 +148,7 @@ varnishd_CFLAGS = \
 	-DVARNISHD_IS_NOT_A_VMOD \
 	-DVARNISH_STATE_DIR='"${VARNISH_STATE_DIR}"' \
 	-DVARNISH_VMOD_DIR='"${pkglibdir}/vmods"' \
-	-DVARNISH_VCL_DIR='"${varnishconfdir}:${pkgdatadir}"'
+	-DVARNISH_VCL_DIR='"${pkgsysconfdir}:${pkgdatadir}"'
 
 varnishd_LDFLAGS = -export-dynamic
 
--- a/configure.ac
+++ b/configure.ac
@@ -505,8 +505,8 @@ VARNISH_STATE_DIR='${localstatedir}/varnish'
 AC_SUBST(VARNISH_STATE_DIR)
 
 # Default configuration directory.
-varnishconfdir='${sysconfdir}/varnish'
-AC_SUBST(varnishconfdir)
+pkgsysconfdir='${sysconfdir}/varnish'
+AC_SUBST(pkgsysconfdir)
 
 # Check for linker script support
 gl_LD_VERSION_SCRIPT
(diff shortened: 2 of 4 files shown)


Dridi:
I have done some testing earlier than I thought I would, it works.

rezan:
I think im still seeing the same problem. However, I would have to test how this builds into a package to be certain. Its possible this is just because I am building from source:

```
> ./configure --prefix=/usr
...
> cat varnishapi.pc
prefix=/usr
...
sysconfdir=${prefix}/etc
pkgsysconfdir=${sysconfdir}/varnish
...
```

So I cant say for sure, but it looks like we still have a prefixed `${sysconfdir}`, and in this case:

```
/usr/etc/varnish
```

rezan:
I think this is a packaging thing, cause `varnishd` is built with this:

```
-DVARNISH_VCL_DIR='"/usr/etc/varnish:/usr/share/varnish/vcl"' 
```

So the `varnishapi.pc` is correct.

Dridi:
It looks like your autoconf is broken, I get this:

```
$ grep -e prefix= -e sysconf varnishapi.pc
prefix=/usr
exec_prefix=${prefix}
sysconfdir=/etc
pkgsysconfdir=${sysconfdir}/varnish
```

Your varnishapi.pc is neither correct nor wrong, it is consistent with your autoconf results. What do you get with the short `configure.ac` file I showed you earlier?

https://github.com/varnishcache/varnish-cache/issues/2251#issuecomment-285796667

rezan:
I get this:

```
> cat configure.ac
AC_INIT([foo])
AC_CONFIG_SRCDIR([configure.ac])
echo prefix: ${prefix}
echo sysconfdir: ${sysconfdir}
> autoconf
> ./configure
prefix: NONE
sysconfdir: ${prefix}/etc
> ./configure --prefix=/opt
prefix: /opt
sysconfdir: ${prefix}/etc
> ./configure --prefix=/usr
prefix: /usr
sysconfdir: ${prefix}/etc
```
I am running Ubuntu 14.

As long as the packages build with the correct information, we should be good. In the meantime, I have a proper workaround for the installation detection problem since this likely won't land in the wild for quite a few release iterations.

Dridi:
I'm assuming you mean 14.04, I will have a look on Monday. Otherwise you still have the option to build Varnish with `--prefix=/usr --sysconfdir=/etc` (and probably `/var` needs to be fixed too). This is usually enforced during packaging, at least I can say so for Fedora.

However you are testing it wrong.

What we want to make sure is that you can install configuration files inside Varnish's `pkgsysconfdir`. So really what you want to check in this case is that your VMOD installs its config in `/usr/etc/varnish` alongside `default.vcl` and its VCL in `/usr/share/varnish/vcl`. In other words, that your VMOD project can properly inherit configuration from Varnish.

Dridi:
Discussed offline with @rezan on Saturday, this solution works for him. Rebased against current master, most of it is polish or documentation.

mbgrydeland:
+1 for inclusion in the release of Varnish 5.1

hermunn:
Backport review: This great effort is varnish 5 only.

Issue #2270, opened 2017-03-22
Newly loaded auto VCLs don't get their go_cold timer set


mbgrydeland:
A newly loaded VCL that is never set active should go cold after the VCL cooldown period has elapsed.

PR will follow

bsdphk:
See #2271, we don't need two issues for this.

PR #2271, opened 2017-03-22
Set the go-cold timer on newly loaded warm auto VCLs

mbgrydeland:
Newly loaded VCLs didn't get their go cold timer set even when they
were set to auto, causing them to never go cold automatically.

Fixes: #2270

### DIFF
--- a/bin/varnishtest/tests/r02270.vtc
+++ b/bin/varnishtest/tests/r02270.vtc
@@ -1,0 +1,19 @@
+varnishtest "Test that never used VCLs go cold automatically"
+
+server s1 {
+	rxreq
+	txresp
+} -start
+
+varnish v1 -arg "-p vcl_cooldown=1" -vcl+backend {
+} -start
+
+varnish v1 -expect VBE.vcl1.s1.happy >= 0
+
+varnish v1 -cliok {vcl.inline vcl2 "vcl 4.0; backend s1 {.host = \"${s1_addr}\"; .port = \"${s1_port}\"; }"}
+
+varnish v1 -expect VBE.vcl2.s1.happy >= 0
+
+delay 5
+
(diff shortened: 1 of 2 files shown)


bin/varnishtest/tests/r02270.vtc
```diff
@@ -0,0 +1,19 @@
...
+varnish v1 -arg "-p vcl_cooldown=1" -vcl+backend {
+} -start
+
+varnish v1 -expect VBE.vcl1.s1.happy >= 0
+
+varnish v1 -cliok {vcl.inline vcl2 "vcl 4.0; backend s1 {.host = \"${s1_addr}\"; .port = \"${s1_port}\"; }"}
+
+varnish v1 -expect VBE.vcl2.s1.happy >= 0
+
+delay 5
```
Dridi: That's a long delay, shouldn't you get away with a 1s delay followed by a dummy CLI command? (e.g. ping)

bsdphk:
This behaviour was partly intentional.

The assumption was that people wouldn't vcl.load something if they don't expect to use it, and consequently it didn't make sense to set the newly loaded VCL cold before they had a chance to use it.

I'm not saying that is (still) a valid assumption...

mbgrydeland:
I see your point that if the VCL is loaded that is on the assumption that it will be used. But the vcl_cooldown covers this in my opinion. The newly loaded VCL will still be warm, and stay that way for 10 minutes (default vcl_cooldown). So its just the wrongly loaded VCLs that weren't used after all that would then go cold after the cooldown.

bsdphk:
Bugwashed, consensus of three non-random old farts was that cooling down newly loaded VCL's is most consistent.

hermunn:
This was backported to 4.1 as 72a6d1e.

Issue #2275, opened 2017-03-22
Condition(v1l->niov < v1l->siov) not true


joshuaspence:
We are seeing panics on Varnish 5.1.1. Here is the stack trace:

```
Panic at: Wed, 22 Mar 2017 22:54:00 GMT
Assert error in V1L_Chunked(), http1/cache_http1_line.c line 274:
  Condition(v1l->niov < v1l->siov) not true.
version = varnish-5.1.1 revision de38712, vrt api = 6.0
ident = Linux,4.4.0-62-generic,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
now = 86524.289610 (mono), 1490223240.320110 (real)
Backtrace:
  0x438b37: /usr/sbin/varnishd() [0x438b37]
  0x45bbf5: /usr/sbin/varnishd(V1L_Chunked+0x135) [0x45bbf5]
  0x458fc9: /usr/sbin/varnishd(V1D_Deliver+0x3d9) [0x458fc9]
  0x43c541: /usr/sbin/varnishd(CNT_Request+0x221) [0x43c541]
  0x45a867: /usr/sbin/varnishd() [0x45a867]
  0x451b69: /usr/sbin/varnishd() [0x451b69]
  0x451feb: /usr/sbin/varnishd() [0x451feb]
  0x7fc6cd6896ba: /lib/x86_64-linux-gnu/libpthread.so.0(+0x76ba) [0x7fc6cd6896ba]
  0x7fc6cd3bf82d: /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7fc6cd3bf82d]
thread = (cache-worker)
thr.req = 0x7fc4a09f9020 {
  vxid = 1842984, transport = HTTP/1 {
    state = HTTP1::Proc
  }
  step = R_STP_TRANSMIT,
  req_body = R_BODY_NONE,
  restarts = 0, esi_level = 0,
  sp = 0x7fc587ef5a20 {
    fd = 89, vxid = 1842983,
    t_open = 1490223239.797381,
    t_idle = 1490223239.797381,
    transport = HTTP/1 {
      state = HTTP1::Proc
    }
    client = 10.155.20.108 50237,
  },
  worker = 0x7fc6c192fc70 {
    stack = {0x7fc6c1930000 -> 0x7fc6c1924000},
    ws = 0x7fc6c192fe70 {
      id = \"wrk\",
[...]

PR #2276, opened 2017-03-23
Include `libvmod_debug.so` in the Varnish package

joshuaspence:
Currently, `libvmod_debug.so` isn't included in the resulting `varnish` package. Although it does somewhat make sense to exclude `libvmod_debug.so`, it makes it difficult for users to write tricky `varnishtest` test cases. Specifically, I am trying to write a regression test for #2275, but I am able to add `import debug` fails. I was thinking that I could solve this by installing the `varnish-dbg` package, but no such package exists. Given that VMODs aren't loaded automatically but rather are only loaded explicitly within VCL, I think that it's reasonable to bundle `/usr/lib/varnish/vmods/libvmod_debug.so` with the rest of the Varnish package.

### DIFF
--- a/lib/libvmod_debug/Makefile.am
+++ b/lib/libvmod_debug/Makefile.am
@@ -10,8 +10,7 @@ vmoddir = $(pkglibdir)/vmods
 vmod_srcdir = $(top_srcdir)/lib/libvmod_debug
 vmodtool = $(top_srcdir)/lib/libvcc/vmodtool.py
 vmodtoolargs =
-
-noinst_LTLIBRARIES = libvmod_debug.la
+vmod_LTLIBRARIES = libvmod_debug.la
 
 libvmod_debug_la_CFLAGS = \
 	@SAN_CFLAGS@


bsdphk:
I beat you to the test-case for #2275, and yes, that took a bit of extra work on vmod_debug :-)

I'll leave the question of shipping vmod_debug open so we can bugwash it.

joshuaspence:
I'm curious, how did you calculate the magic number that you passed to `debug.allocate_workspace` in your regression test? I'm struggling to find the correct magic value by trial and error.

Dridi:
Since I'm not attending next bugwash, I'll comment now.

I ran into situations where vmod-debug would have helped out-of-tree testing a lot so I'm strongly in favor of its installation. However, it being an internal thing today, it could be the occasion to polish the VCC before it goes "public". And last point, we need to make sure we in pkg-varnish-cache that the VMOD and its manual are part of the -dev (debian) or -devel (rhel) subpackages.

bsdphk:
Discussed on bugwash.  We can ship vmod_debug only if it is not importable if vcc_unsafe_path is false.

Preferably the VCL files importing vmod_debug should contain a unescapable warning.

Ideas:

```
import debug from "unsafe_at_any_speed/";

sub vcl_init {
    debug.init("Yes, I know what I'm doing, thanks for asking.");
}
```

fgsch:
Installing it in a non-standard path should do the trick. I'm not sure about the warning though.
Having to do both the `import debug from ...` and setting vcc_unsafe_path should be plenty.

Dridi:
It looks like the discussion drifted away from the use case, which is availability of the VMOD for out-of-tree `varnishtest` usage.

While I have no objection with an installation outside of the default `vmod_path`, I think we should also keep in mind that we don't want it to be a PITA to simply `import debug` in test cases. My suggestion would be to drop it in `${vmod_dir}-devel` and do an extra lookup there if the `vtc_mode` debug flag is set.

fgsch:
> edit: so obviously I'm against the idea of having to debug.init() it

I agree. Seems overkill to me.

Dridi:
From today's bugwash:

We will introduce a new `vmod-vtc` containing the interesting things for out-of-tree testing (eg. like in-VCL barrier syncing) and keep nonsensical stuff in `vmod-debug`. I will first propose a list of functions to move away from `vmod-debug` and then submit a patch once we reach consensus.

joshuaspence:
Yep, that all sounds reasonable to me.

denisbr:
Will someone let me know when I should add the vmod-debug stuff to the packages?

Dridi:
@denisbr this will be for the September release, and I will let you know if packaging work is required.

Dridi:
I'm taking advantage of #2318 to suggest the contents of vmod-vtc. This is based on the exhaustive list of existing symbols in the current vmod-debug, in order of appearance in the VCC descriptor.

### Things to keep in vmod-debug

```
$Event event_function

$Function STRING author(...)
$Function VOID test_priv_call(...)
$Function VOID test_priv_vcl(...)
$Function STRING test_priv_task(...)
$Function STRING test_priv_top(...)

$Object obj(...)
$Method VOID .enum(...)
$Method STRING .string()
$Method STRING .number()
$Method STRING .foo(...)
$Method TIME .date()
$Method VOID .test_priv_call(...)
$Method VOID .test_priv_vcl(...)
$Method STRING .test_priv_task(...)
$Method STRING .test_priv_top(...)

$Function VOID rot52(...)
$Function STRING argtest(...)
$Function INT vre_limit() (more on this one below)
$Function VOID register_obj_events(...)
$Function VOID fail()

$Object dyn(...)
$Method BACKEND .backend()
$Method VOID .refresh(...)

$Function VOID vcl_release_delay(...)
$Function BOOL match_acl(...)
$Function VOID test_probe(...)
```

### Things to introduce via a new vmod-vtc

```
$Function VOID panic(...)
$Function BLOB str2blob(...)
$Function STRING blob2hex(...)
$Function BACKEND no_backend()
$Function STEVEDORE no_stevedore()
$Function VOID sleep(...)
$Function VOID workspace_allocate(...)
$Function BOOL workspace_overflowed(...)
$Function VOID workspace_overflow(...)
$Function INT workspace_free(...)
$Function VOID workspace_snap(...) (not sure about this one)
$Function VOID workspace_reset(...) (not sure about this one)
$Function BOOL barrier_sync(...)
$Function INT typesize(...)
```

Regarding `vre_limit`, it only returns the value of the `pcre_match_limit` parameter so I would rather leave it vmod-debug. However, we could also provide a set of functions to access any parameter (one per type) from VCL. But has anyone ever ran into this need besides the `vre_limit` occurrence?

Also, if you are a varnishtest user and ever needed something not figuring in this list, this is the place to let us know!

fgsch:
I agree with the suggested split. We should move forward and do any fine-tuning in the tree.

bsdphk:
This is overtaken by #2362

PR #2287, opened 2017-03-29
Added std.file_exists("/foo") to check if "/foo" exists.

JonathanHuot:
I'm suggesting to add a new function to test the presence of a file.

For Varnish3, it has already be done via a VMOD (see https://github.com/thomsonreuters/libvmod-utils) but since, Varnish4 has introduced more and more functions (e.g. `fileread`), so I guess the best place for `file_exists()` should be in vmod_std instead of a separate VMOD.
Also, I used `stat(2)` and not the other variants e.g. `lstat(2)`, but I guess that's OK.

For us, the main purpose of this function is the ability to check if a local file exists, and reply 200 or 503 accordingly. Then, a loadbalancer in front of Varnish can send an healthcheck to Varnish and can exclude or include it accordingly to the presence of the file.

If you're OK in principle, let me know if anything can be improved into this PR to get accepted.

Thanks

### DIFF
--- a/bin/varnishtest/tests/m00029.vtc
+++ b/bin/varnishtest/tests/m00029.vtc
@@ -1,0 +1,83 @@
+varnishtest "Test file_exists for std VMOD"
+
+shell {
+	touch "${tmpdir}/m00029_file_empty"
+	touch "${tmpdir}/m00029_file_noperm"; chmod 000 "${tmpdir}/m00029_file_noperm"
+	printf "Content" > "${tmpdir}/m00029_file_content"
+}
+
+server s1 {
+	rxreq
+	expect req.http.exists == "true"
+	txresp
+	rxreq
+	expect req.http.exists == "true"
+	txresp
+	rxreq
+	expect req.http.exists == "true"
+	txresp
+	rxreq
+	expect req.http.exists == "false"
+	txresp
+} -start
+
+varnish v1 -vcl+backend {
+	import std;
+
+	sub vcl_recv {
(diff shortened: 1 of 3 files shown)


nigoroll:
ok for me. In a similar private implementation I'm caching the stat result for a configurable amount of time, but reflecting on it, that's probably overkill.

fgsch:
Overall this looks ok to me. 

The tests could be less specific however. Using file_exists in 2-3 scenarios will be enough to exercise the code without any assumption about its usage, e.g:

```
sub vcl_deliver {
    set resp.http.non-existent = std.file_exists("/non/existent");
    set resp.http.ok = std.file_exists("${tmpdir}/v1/_.vsm");
    set resp.http.no-perms = std.file_exists("/something/without/perms");
}
```

Also, do we want to restrict this to regular files or anything goes as long as you have perms?

fgsch:
@JonathanHuot would you mind rebasing so it does not conflict?

JonathanHuot:
Hi @fgsch, I rebased and applied your suggestion about the VTC's content. Note that I've removed the test on a non-permission file which has no added-value in my opinion. Correct me otherwise.
Also, Travis failed because of tests/t02007@clang, which is unrelated. I assume this is a test which (periodically|recently) failing.

fgsch:
Merged with minor tweaks and all the commits squashed.

Thanks for the PR!

hermunn:
Backport review: dff3c2be8 backported as 59b47f86.

PR #2290, opened 2017-03-30
statistics aggregation overhaul

nigoroll:
not to be merged before 5.1.2, but I'd appreciate earlier reviews
the first two commits are only to get support for fixed size mempools
the actual patch is the third commit, copied description:

    Previously, statistics aggregation was a three-level process:
    
    1) workers collected per-thread statistics
    
    2) When wthread_stats_rate was reached or a worker went idle,
       statistics were aggregated on the pool level
    
    3) Only when a worker was idle were the pool level statistics added to
       the global aggregation
    
    This could lead to long delays between the time some statistic was
    generated and it being reported, in particular under full load. Global
    statistics were never updated as long as all worker threads were busy
    at all times.
    
    We now simplify this into a two-level process: As before, workers keep
    their own statistics, but when it is time to aggregate, we send them
    off to a separate thread, which periodically adds all received stats
    to the global counters.
    
    The parameter thread_stats_rate is replaced by thread_stats_latency,
    which specifies a latency goal for statistics aggregation. The default
    of 1s is motivated by the default varnishstat update interval.

### DIFF
--- a/bin/varnishd/Makefile.am
+++ b/bin/varnishd/Makefile.am
@@ -21,6 +21,7 @@ varnishd_SOURCES = \
 	cache/cache_cli.c \
 	cache/cache_deliver_proc.c \
 	cache/cache_director.c \
+	cache/cache_dstat.c \
 	cache/cache_esi_deliver.c \
 	cache/cache_esi_fetch.c \
 	cache/cache_esi_parse.c \
@@ -112,6 +113,7 @@ varnishd_SOURCES = \
 noinst_HEADERS = \
 	builtin_vcl.h \
 	cache/cache_ban.h \
+	cache/cache_dstat.h \
 	cache/cache_esi.h \
 	cache/cache_obj.h \
 	cache/cache_pool.h \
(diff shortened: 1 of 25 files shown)


nigoroll:
the travis failure is re t02006

nigoroll:
rebased to hopefully get rid of the travis check maked failing

nigoroll:
rebased and resolved conflict with f895383ded3f24f2a42e7224567d31ec9a2af535

nigoroll:
I appreciate @bsdphk working on dynamic stats and think that the new mechanics proposed here make even more sense with more stats because
* we avoid a second copy step
* on the global level we might have a different stats structure than on the worker level, so loading this off to a separate thread might help

rebased onto current master

Dridi:
@nigoroll in case you forgot about that one, we now have the dynamic stats in place, maybe it's time to revisit this? If I understand correctly this only changes the internals, not the API, so maybe something that could be back-ported to a 6.0.x release if the timing is to short for 6.0.0.

nigoroll:
this needs to be re-done now that we got the new vsm/vsl world order. Closing for now, this will hopefully come back soon

Issue #2292, opened 2017-03-30
Documentation for std.collect might be bad


simonvik:
https://varnish-cache.org/docs/trunk/reference/vmod_std.generated.html#func-collect

Contains an example that says: 
```
Example
std.collect(req.http.cookie);
```

Multiple cookie headers should not be sent over HTTP/1.1 and if they were then they should be joined by "; " and not "," as the module does.

h2-clients might send multiple cookie headers and that's reported in : #2291

hermunn:
Backport review: I do not think there is a need for this in 4.1, so we will keep it Varnish 5 only unless I hear strong protests.

PR #2294, opened 2017-03-30
Collapsing cookies correctly

### LINKED ISSUE #2291: Collect cookie-headers when running h2
When running h2 its common that the client sends multiple cookie headers to increase compression.

According to https://tools.ietf.org/html/rfc7540#section-8.1.2.5 a h2 -> h1 proxy should concat multiple cookies into one delimited by "0x3B 0x20"

## Expected Behavior
That I only see one single req.http.cookie header delimited by "; "

## Current Behavior
Varnish sends multiple Cookie-headers to the backend.

## Possible Solution
1. Collect cookie-headers default when running h2.
2. Get std.collect( to take a delimiter option. 
     `std.collect(req.http.cookie, "; ");`

Dridi:
Fixes #2291 and #2292 and improves a bit further.

@simonvik, please try this out too.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -783,6 +783,7 @@ void http_CopyHome(const struct http *hp);
 void http_Unset(struct http *hp, const char *hdr);
 unsigned http_CountHdr(const struct http *hp, const char *hdr);
 void http_CollectHdr(struct http *hp, const char *hdr);
+void http_CollectHdrSep(struct http *hp, const char *hdr, const char *sep);
 void http_VSL_log(const struct http *hp);
 void HTTP_Merge(struct worker *, struct objcore *, struct http *to);
 uint16_t HTTP_GetStatusPack(struct worker *, struct objcore *oc);
--- a/bin/varnishd/cache/cache_http.c
+++ b/bin/varnishd/cache/cache_http.c
@@ -1,6 +1,6 @@
 /*-
  * Copyright (c) 2006 Verdens Gang AS
- * Copyright (c) 2006-2015 Varnish Software AS
+ * Copyright (c) 2006-2017 Varnish Software AS
  * All rights reserved.
  *
  * Author: Poul-Henning Kamp <phk@phk.freebsd.dk>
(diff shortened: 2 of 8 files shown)


bsdphk:
I'm OK with this, subject to it actually working.

API-wise it should be safe for 5.1.2.

Not clear if the Collect should happen in C or (builtin-)VCL, nor if there is any reason to make it H2 only ?

Doc in .vcc file says 'comma' but it really is 'comma + space'

Dridi:
I will comment tomorrow, because I won't be present during next bugwash to discuss this.

Dridi:
> I'm OK with this, subject to it actually working.

I had a look at the failures on Travis, usual tests flakiness (same h2 tests that sometimes fail on vtest too these days). I restarted the failing job and it passed the second time.

> API-wise it should be safe for 5.1.2.

See e5097a4, it is. I also added a placeholder in `vrt.h` for the release.

> Not clear if the Collect should happen in C or (builtin-)VCL

Definitely in C, because this is a transport-specific matter. I also added a comment (see 9c261d1) for RFC coverage. An alternative to doing it before `vcl_recv` (this PR) could be after `vcl_backend_fetch`.

> nor if there is any reason to make it H2 only ?

I didn't want to do it in H1 because it isn't supposed to happen. Cookies can easily be expensive and collapsing them can eat much workspace. There's no reason for a Varnish user to see this behavior in the wild, unless they have a legacy client. In that case they can use `std.collect`.

> Doc in .vcc file says 'comma' but it really is 'comma + space'

Indeed, as I said to @simonvik on IRC I thought that `std.collect` already had a `", "` separator, but that was a side effect of keeping leading blanks. So when I fixed that in ad9623d you can notice that the only affected test case was the one created in this PR. That's why I forgot this detail in the docs: it didn't break existing behavior. Will push a docfix too.

Dridi:
> I'm OK with this, subject to it actually working.

I had a look at the failures on Travis, usual tests flakiness (same h2 tests that sometimes fail on vtest too these days). I restarted the failing job and it passed the second time.

> API-wise it should be safe for 5.1.2.

See e5097a4, it is. I also added a placeholder in `vrt.h` for the release.

> Not clear if the Collect should happen in C or (builtin-)VCL

Definitely in C, because this is a transport-specific matter. I also added a comment (see 9c261d1) for RFC coverage. An alternative to doing it before `vcl_recv` (this PR) could be after `vcl_backend_fetch`.

> nor if there is any reason to make it H2 only ?

I didn't want to do it in H1 because it isn't supposed to happen. Cookies can easily be expensive and collapsing them can eat much workspace. There's no reason for a Varnish user to see this behavior in the wild, unless they have a legacy client. In that case they can use `std.collect`.

> Doc in .vcc file says 'comma' but it really is 'comma + space'

Indeed, as I said to @simonvik on IRC I thought that `std.collect` already had a `", "` separator, but that was a side effect of keeping leading blanks. So when I fixed that in ad9623d you can notice that the only affected test case was the one created in this PR. That's why I forgot this detail in the docs: it didn't break existing behavior. Will push a docfix too.

simonvik:
Seems to solve my issue, running in production for a while now.

bsdphk:
Some comments:

Strictly speacking we cannot just update copyright year for adding a single C prototype, that is a trivial change relative to the rest.

Also: Don't use <ctype.h> for HTTP headers, use "vct.h"

Also: Let http_CollectHdrSep() do the NUl/NULL check on sep.

Dridi:
Done:

* 8dc398a
* 9ff704f
* 2708889
* c81bb85
* 01f991e
* 090653e

bsdphk:
I'm still not convinced that we should only collect cookies for H2 ?

Also, don't bump the API version until 5.1.2 is out.  Vmods will have to wait a cycle before calling this.

Dridi:
Merged without the VRT bump:

* 84e98d8
* 99efebb
* eadcc45
* 58b2d30
* d26a828
* 5087989

hermunn:
Backport review: There is no need to collapse cookies in 4.1, so we don't.

Issue #2103, opened 2016-09-26
4.1: Unexpected reply from ping: 400 CLI communication error


rnsanchez:
Varnish (4.1.4-beta1 245ec6c2f17263f6772c7b7d220da81f8c869e93) is dying without traces of abort().  It silently restarts, wiping out cache.

This happened only on instances running with dynamic backends.  Logs are rather terse on describing the issue:

```
Sep 14 22:55:03 syslog varnishd[63464]: Child (8919) not responding to CLI, killing it.
Sep 14 22:55:03 syslog varnishd[63464]: Unexpected reply from ping: 400 CLI communication error (hdr)
Sep 14 22:55:12 syslog varnishd[63464]: Child (8919) not responding to CLI, killing it.
Sep 14 22:55:12 syslog varnishd[63464]: Unexpected reply from ping: 400 CLI communication error (hdr)
Sep 14 22:55:12 syslog varnishd[63464]: Child (8919) not responding to CLI, killing it.
Sep 14 22:55:12 syslog varnishd[63464]: Unexpected reply from ping: 400 CLI communication error
Sep 14 22:55:12 syslog varnishd[63464]: Child (8919) died signal=3
Sep 14 22:55:12 syslog varnishd[63464]: Child (7915) said Child starts
```

Apparently from:

```
bin/varnishd/mgt/mgt_child.c:       VSB_printf(vsb, " signal=%d", WTERMSIG(status));
```

There is no panic.  We can help test/trace the issue, so please inform what additional information may be needed.

What seems to trigger the issue is changing the VCL.  It may take several minutes until the issue is actually triggered.

Environment: 4.1.4-beta1 245ec6c2f17263f6772c7b7d220da81f8c869e93, Slackware64-14.1, Linux-4.4.14.

rnsanchez:
@Dridi this behavior started after fixes for #2008 were released.  It happens in production, does not happen always, and usually takes a few minutes to trigger---when it does.

Dridi:
I don't see why this is happening, yet. Can you please monitor your raw logs with `varnishlog -g raw` too and see what's being said there? Not everything goes to syslog when enabled, only CLI traffic I believe.

Dridi:
It may also be related to #2019 since #2008 commits didn't directly touch the varnish-cli.

rnsanchez:
@Dridi I'm trying to obtain the output you requested.

In the meanwhile, could it be related to timeout?

[...]

PR #2296, opened 2017-03-31
PR to fix #2295

### LINKED ISSUE #2295: Spinning loop in VBE_Poll causes master to kill child on CLI timeout
In VBE_Poll() there is a loop that deletes old dynamic backends after a cooling period. This loop bugs out and becomes a kind of long running spinlock when the backend to be cleaned has a non-zero n_conn (due to very slow backend fetch). This causes the CLI thread to spin in a tight loop, possibly for a very long time, causing the master process to kill the child on CLI timeout.

PR and test case will follow.

mbgrydeland:


Dridi:
Also related to #2103, for which I had `VBE_Poll` as a suspect but don't remember why I never submitted a fix for this function.

hermunn:
Backport review: 667256f1 backported as 04b4e380.

PR #2298, opened 2017-03-31
Don't reset memory out of `struct http` at http_Teardown

catap:
Usual code to create a `struct http` looks lie:
```
nhttp = (uint16_t)cache_param->http_max_hdr;
hl = HTTP_estimate(nhttp);

req->http = HTTP_create(p, nhttp);
p += hl;
p = (void*)PRNDUP(p);
assert(p < e);

req->http0 = HTTP_create(p, nhttp);
p += hl;
p = (void*)PRNDUP(p);
assert(p < e);
```

Well, when someone do `http_Teardown(req->http)` it will reset memory inside `req->http0` because `req->http->hdf` is a pointer to last `struct http` in allocated/reserved memory.

### DIFF
--- a/bin/varnishd/cache/cache_http.c
+++ b/bin/varnishd/cache/cache_http.c
@@ -186,7 +186,6 @@ http_Teardown(struct http *hp)
 	AN(hp->shd);
 	memset(&hp->nhd, 0, sizeof *hp - offsetof(struct http, nhd));
 	memset(hp->hd, 0, sizeof *hp->hd * hp->shd);
-	memset(hp->hdf, 0, sizeof *hp->hdf * hp->shd);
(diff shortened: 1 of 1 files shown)


bsdphk:
I have looked high and low for any overlap between http and http0 and cannot find any.

To make doubly sure, in 6c9021d9f5ed89ebcd597d7ef144b1886e370111 I added an assert to make sure no such overlap happens due to alignment mistakes on our part.

Can you describe the symptoms you are trying to cure with this patch ?

Ie: what goes wrong, and how does it lead you to this patch ?

catap:
@bsdphk, sorry, I confirm that the issue was in my code.

PR #2314, opened 2017-04-18
Install VUT headers, plus cleanups

Dridi:
I just realized that the headers were deemed private, despite the fact that they were added in version 1.5 of libvarnishapi. While at it I had a quick look and polished things here and there in the VUT area.


fgsch:
> Use argv[0] as the VUT.progname

Can you name some tools that will output the full path in the usage?

Dridi:
`httpd`

fgsch:
Any others? I don't really like this change tbh.

Dridi:
I figured you didn't like it, and I hadn't thought of other tools doing that prior to doing this change. `httpd` was my top-of-mind try, here are more:

```
$ /usr/bin/cat --help | grep -i usage
Usage: /usr/bin/cat [OPTION]... [FILE]...
$ /usr/bin/head --help | grep -i usage
Usage: /usr/bin/head [OPTION]... [FILE]...
$ /usr/bin/cut --help | grep -i usage
Usage: /usr/bin/cut OPTION... [FILE]...
$ /usr/bin/dash --help | grep -i usage
/usr/bin/dash: 0: Illegal option --
$ /usr/bin/zsh --help | grep -i usage
Usage: /usr/bin/zsh [<options>] [<argument> ...]
$ /usr/bin/emacs --help | grep -i usage
Usage: /usr/bin/emacs [OPTION-OR-FILENAME]...
```

fgsch:
Hmm, might be a linux-only.  I cannot see this in any of the BSDs I have access to.

fgsch:
FWIW, my preference is to not use argv[0] but keep the existing behaviour, if anything because I don't see the benefit.

Dridi:
Bugwash summary:

- @mbgrydeland agrees in general with exposing VUT symbols
- regarding argv[0] nobody objected among attendees (it seemed to be more about using a pristine argv[0] or its basename).
- @bsdphk wants to hold the merge back for at least a couple weeks

At this point we want to make sure it's safe to expose this API, look at all the bells and whistles (coverage, coverity, flexelint...) and avoid shipping an inappropriate interface to `libvarnishapi`consumers. It also means that if we break the VUT ABI before exposing it we will need a 2.0 bump of the soname.

bsdphk:
This is the kind of stuff I was thinking of, when I talked about making sure VUT was properly polished before we expose it:

```
During Specific Walk:
  File varnishhist.c line 590: VUT_Arg(?, 0) #2
../../lib/libvarnishapi/vut.c  160  Warning 418: Passing null pointer to
    function 'strtol(const char *, char **, int)', arg. no. 1 [Reference: file
    varnishhist.c: line 590]
varnishhist.c  590  Info 831: Reference cited in prior message
```

bsdphk:
Another observation:

Functions like VUT_Error() which sometimes return (status == 0) and sometimes don't (status != 0) are generally a bad idea.

Split it into VUT_Warning() which always return and VUT_Error() or VUT_Fatal() which never does.

Dridi:
Well flexelint was mistaken so I helped it in 978e82a. And because in another scenario it could be right I pushed 63fcdd6 too.

I had a quick look and I think `getopt` handling for `-V` is not OK yet because it prints Varnish's version. This is easy to circumvent.

Dridi:
FYI this evening I started a tutorial in the form of a repository similar to libvmod-example:

https://github.com/dridi/varnish-template

It's making use of the current VUT headers from Varnish 5.1.2 (for now).

bsdphk:
There were only one VUT_Error() use outside vut.c with status zero (varnishstat), so I have made VUT_Error() always fatal.

I'm a bit torn on the fprintf's about acquiring SHM, maybe they should only happen under some kind of verbose flag ?

Dridi:
How about a `FILE *log` member in `struct VUT`? Those messages are fprintf'd only if it's non NULL and as far as in-tree VUTs are concerned we set it to stderr.

Dridi:
And we could wrap the NULL check+fprintf in a `VUT_Printf` function.

Dridi:
Updated to current master (started 328 commits ago!) and ready to resume the discussion any time.

gquintard:
+1 for getting this in the september release

Dridi:
New elements from IRC discussions and changes since the 328 commits ago:

- libvarnishapi already breaks, so we can safely break the VUT API
- it is currently a global API not fit for concurrent VSM accesses

It wouldn't be hard to make it usable on multiple `varnishd` instances, or even multiple times on a single `varnishd` instance. This is a trivial change, we may want to make the `struct` opaque, and interactions with the API could look (error handling omitted) like this:

```
struct VUT *vut = VUT_Init(argc, argv, &vopt_spec);

while ((opt = getopt(argc, argv, vopt_spec.vopt_optstring)) != -1) {
        switch (opt) {
        [...]
        default:
                VUT_Arg(vut, opt, optarg);
        }
}

VUT_Setup(vut, dispatch_f, dispatch_priv);
VUT_Main(vut);
VUT_Fini(vut);
```

The main problem is that `getopt` is MT-Unsafe, so we need to document that clearly since the VUT heavily relies on it. Or we could wrap the `getopt` handling in the VUT API and pass a callback+priv pair to VUT_Init do enforce the locking/reset of the getopt state if that's possible (and still document it to prevent users from running `getopt` concurrently).

Dridi:
I'd like to take a step back after further IRC discussions.

First, I'm no longer in favor of making this API concurrent-able, because `getopt` gets in the way and we have no control over it. Also because this is clearly a convenience for single-purpose VSL consumers like `varnishlog` and `varnishncsa`.

Not only is the API single-process oriented as shown by the use of `getopt`, it is also meant to generate RST bits of the manual via the `-x` option: the synopsis and the list of options.

On the other hand if someone needs to build something more involved than a single-purpose VSL consumer, they have access to the rest of `libvarnishapi` and full control over all things VUT can't do by design.

This would make it a realistic target for the September release again.

carlosabalde:
We've recently developed a VUT that processes VSL transactions in order to generate (1) real-time stats similar to the ones generated by varnishstat but fully customizable and friendly with monitoring tools like Zabbix, Nagios, OpenMS, etc.; and (2) ad-hoc JSON logs friendly with log analysis tools like ELK, Splunk, etc.

I don't think I can provide any useful feedback about the internals being discussed here, but I want to share my two cents and vote for exposing the  VUT functionality in the next release. It would help a lot to simplify VUT development focusing on what really matters (and not on parsing arguments, daemonizing processes, etc.).

bsdphk:
Dridi:  I simply don't understand what you are saying about getopt ?

As far as I can tell, getopt() is not called anywhere from VUT ?

How a multi-varnishd-util goes about initializing it's multiple VUT instances is entirely up to the person who writes it, and I don't see anything preventing that from being::

    vab -n /tmp/vi1 -n /tmp/vi2 -n /tmp/vi3

or similar ?

Dridi:
> Dridi: I simply don't understand what you are saying about getopt ?
>
> As far as I can tell, getopt() is not called anywhere from VUT ?

The VUT setup relies on `getopt` options, including the RST generation for the manual. So either you use `getopt` to manage your options and fall back to `VUT_Arg` for the common options or you roll your own `getopt`-like parsing of the arguments. But at some point you need to deal with the `vopt_optstring` one way or another.

It also encourages using the same options as `varnishlog` and others to keep a consistent/familiar experience with foreign tools.

> How a multi-varnishd-util goes about initializing it's multiple VUT instances[...]

I get the point, and I was in favor of it until I tried to bend the API to try such a case. That is when I changed my mind to keeping VUT for single-purposed single-varnish VSL consumers: a convenience that does a great job for this case while more involved tools can rely on the underlying libvarnishapi directly.

The VUT API would be a great gateway to out-of-tree VUT development like in @carlosabalde's case, and all the VSL consumers (not including ours) I have seen, worked on, or heard of are single-purposed. Adding the VUT headers doesn't remove functionality, it removes a lot of boilerplate for the most common case I encountered.

Also, if you need to graduate from a simple tool to something more advanced, looking at the VUT code shows what's needed to set up a VSL consumer and makes the transition easy.

bsdphk:
I still don't buy it.

He'res how I would probably do it::

     while (getopt())
         collect -n options, create a VUT for each
     optreset=1
     while (getopt())
        ignore '-n'
        all other options, apply to all VUTs

either way, using the VUT API for single varnishd apps do not suffer because we get rid of the global VUT variable, so my decision stands.

Dridi:
It doesn't work if your program may spawn VUTs concurrently, unless you fork them. `getopt` is not MT-safe.

bsdphk:
then you will have to pthread_mutex_lock your usage of getopt(3)

But that still doesn't matter...

Dridi:
Updated, rebased against current master that was conflicting since the last update.

The API no longer relies on a global VUT symbol, but command-line options with a global impact are kept in the API to still benefit from reuse. The fact that those options are global is made explicit.

Signal handling adds an indirection, so programs need to forward signals to the VUTs they maintain.

Error handling is done via a new callback signature, by default it uses a callback that prints to `stderr` and exits.

I don't consider this to be done, because `struct VUT` is not opaque, but I'd like some feedback before working on that.

fgsch:
Does not build in OSX. 
From https://travis-ci.org/varnishcache/varnish-cache/jobs/274525706:

```
In file included from vut.c:55:
../../include/vut.h:80:17: error: a parameter list without types is only allowed in a function definition
void VUT_Signal(sighandler_t);
                ^
vut.c:224:12: error: unknown type name 'sighandler_t'
VUT_Signal(sighandler_t sig_cb)
```

Dridi:
That's on me, I didn't notice in the manual that `sighandler_t` is a GNU extension, thanks! I updated the patch and I will press the merge button unless the next Travis build fails.

Dridi:
Hopefully for the last time I updated this pull request (rebased against latest master).

The new commits are: 7dcb762 (autoconf macro) and 704e186 (release notes)

I ported my template project to 5.1.2 [1] and by adding the macro to `configure.ac` [2] all I need to generate the synopsis and list of options for the manual is a one-liner [3] for building and another one [4] for cleaning.

[1] https://github.com/Dridi/varnish-template/commit/267f0e5
[2] https://github.com/Dridi/varnish-template/blob/41df3e8/configure.ac#L45
[3] https://github.com/Dridi/varnish-template/blob/41df3e8/src/Makefile.am#L63
[4] https://github.com/Dridi/varnish-template/blob/41df3e8/src/Makefile.am#L76

Issue #2313, opened 2017-04-18
cannot link to varnishapi, symbols missing


dmatetelki:
`libvarnishapi.map` contains:
```
VSM_iter0;
VSM_intern;
```
out of sync with `vapi/vsm.h`, which contains:
```
VSM__iter0;
VSM__itern;
```

Currently it's possible to compile against those symbols, as they are found in `vapi/vsm.h`, but cannot link to, as `libvarnishapi.map` does not expose them.

Dridi:
This happened in f394f0adba53e7a60a516c85968a0f4c95095423, so quite a while ago. There may be other missing symbols from the API change, and the breakage appears to have happened between the inauguration of `libvarnishapi.map` and the first release shipping it.

Workaround: `#define` ?

dmatetelki:
Two possible directions emerged:

1. These 2 symbols meant to be low level and not exposed, hence the double underscores in the names.

d

Dridi:
We can also both have the foreach macro and promote (and possibly rename) those symbols. However, I wouldn't retroactively put them in 1.0 but rather inaugurate the next soname version. Older builds with the wrong names wouldn't expose the symbols anyway.

dmatetelki:
@Dridi : do you mean in `libvarnishapi.map` 

```
LIBVARNISHAPI_1.7 {
  global:
	VSM__iter0;
	VSM__itern;
} LIBVARNISHAPI_1.0;
```

Dridi:
Confirmed

hermunn:
Backport review: As this is not very critical, I will leave it to @dmatetelki to decide if he wants it in 4.1. Send me an email with the port, and I can review it.

PR #2315, opened 2017-04-18
cannot link to varnishapi, symbols missing

dmatetelki:
fixes https://github.com/varnishcache/varnish-cache/issues/2313

### DIFF
--- a/lib/libvarnishapi/libvarnishapi.map
+++ b/lib/libvarnishapi/libvarnishapi.map
@@ -40,8 +40,8 @@ LIBVARNISHAPI_1.0 {
 	VSM_Head;
 	VSM_Find_Chunk;
 	VSM_Close;
-	VSM_iter0;
-	VSM_intern;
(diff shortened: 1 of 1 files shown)


Dridi:
I don't think we should fix this by retroactively *changing* what's in the 1.0 ABI.

dmatetelki:
I wonder if anyone ever used it, since it never worked.

Dridi:
The pattern for API "symbols" seems to discard double underscores. For example `Lck` functions have a `__` in their names, but macros are provided with single `_` equivalents. We may not want to expose them as-is.

mbgrydeland:
These two symbols aren't fully exposed. They are not part of the documented API, but the VSM_FOREACH macro is part of the documented API and it uses these two functions. So the pattern is similar to that of the Lck API, and keeping the double underscores is the right "message" to send regarding their semi-exposed nature.

I'll merge this PR as is.

PR #2326, opened 2017-05-08
Updated instructions for building on SmartOS

RantyDave:
Varnish needs a couple of configure flags to make everything go well on SmartOS.

I also have an SMF (a manifest for SmartOS's boot system) at https://gist.github.com/RantyDave/1d67e4fa571e675c3eced6896c6872ee but considered it impolite to reference off-repo in the patch itself. Feel free to do whatever, but it would certainly help new installs if they could get it from somewhere...

Cheers,
Dave

### DIFF
--- a/doc/sphinx/installation/install.rst
+++ b/doc/sphinx/installation/install.rst
@@ -119,12 +119,16 @@ Optionally, to rebuild the svg files:
 Build dependencies on a SmartOS Zone
 ------------------------------------
 
-As of SmartOS pkgsrc 2015Q4, install the following packages::
+As of SmartOS pkgsrc 2017Q1, install the following packages::
 
 	pkgin in autoconf automake libedit libtool ncurses \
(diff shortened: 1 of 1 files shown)


doc/sphinx/installation/install.rst
```diff
@@ -119,12 +119,16 @@ Optionally, to rebuild the svg files:
 Build dependencies on a SmartOS Zone
 ------------------------------------
 
-As of SmartOS pkgsrc 2015Q4, install the following packages::
+As of SmartOS pkgsrc 2017Q1, install the following packages::
 
 	pkgin in autoconf automake libedit libtool ncurses \
 		 pcre py27-sphinx python27 gmake gcc49 pkg-config
+		 
+For the `configure` step (below) use::
```
Dridi: The \`configure\` bit looks like a markdownism

Dridi:
The SMF file could be featured in the pkg-varnish-cache repo, although we don't support this platform packaging-wise.

PR #2331, opened 2017-05-22
support ncurses with a split tinfo library

stefantalpalaru:
Some Linux distributions like Gentoo can install ncurses with a separate libtinfo. This pull request adds support for that scenario by simply updating m4/ax_with_curses.m4 to the latest version from https://www.gnu.org/software/autoconf-archive/ax_with_curses.html and replacing the deprecated variable CURSES_LIB with CURSES_LIBS in the affected Makefile.am files.

In order to delete the related Makefile.in files, you might have to do a "git clean -x".

later edit: Travis fails with "error: possibly undefined macro: PKG_CHECK_EXISTS". Is pkg-config installed in the test environment?

### DIFF
--- a/bin/varnishhist/Makefile.am
+++ b/bin/varnishhist/Makefile.am
@@ -18,4 +18,4 @@ varnishhist_LDADD = \
 	$(top_builddir)/lib/libvarnishapi/libvarnishapi.la \
 	-lm \
 	@SAN_LDFLAGS@ \
-	@CURSES_LIB@ ${RT_LIBS} ${PTHREAD_LIBS}
+	@CURSES_LIBS@ ${RT_LIBS} ${PTHREAD_LIBS}
--- a/bin/varnishstat/Makefile.am
+++ b/bin/varnishstat/Makefile.am
@@ -19,4 +19,4 @@ varnishstat_CFLAGS = \
 varnishstat_LDADD = \
 	$(top_builddir)/lib/libvarnishapi/libvarnishapi.la \
 	@SAN_LDFLAGS@ \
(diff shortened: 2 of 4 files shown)


Dridi:
> In order to delete the related Makefile.in files, you might have to do a "git clean -x".

Technically we need to regenerate the whole thing once we touch m4 code so running `autogen.des` should be enough.

> Is pkg-config installed in the test environment?

That reminds me of problems with old versions of `pkg-config` that we often had in VMOD projects. I can have a look next week.

Dridi:
So rather than the next week I had a look at this the next month, sorry for taking this long.

While I was trying to fix this by bundling `pkg.m4` in our build tree I eventually realized that we are already using `PKG_*` macros in our build system so this is already something we can take for granted today (unlike before [1]). So yes, either pkg-config or pkgconf is installed in our test environments (Travis isn't our only CI facility).

For some reason curses detection fails, and on closer inspection it is using an internal macro `_PKG_CONFIG` not part of the public API. So I'm not keen on merging this and working around whatever is failing.

Could you maybe get in touch with upstream autoconf-archive and ask them not to rely on private macros? Once they release a new serial I will consider upgrading to that.

[1] https://github.com/varnish/libvmod-example/pull/18

stefantalpalaru:
> Could you maybe get in touch with upstream autoconf-archive and ask them not to rely on private macros?

No, there's a limit to the amount of time I'm throwing down the drain on unwanted contributions.

Dridi:
I understand, and I didn't have time myself to look at this sooner. I also don't want to spend too much time on this, especially since phk's been talking about getting rid of autotools when we were working on the 5.1 release. And I often see problems with macros from the autoconf-archive, things that assume gmake for example.

PR #2339, opened 2017-06-05
Universal failure for VRT_purge

Dridi:
This logs an error and fails the transaction when an illegal purge is performed (that is, outside `vcl_{hit,miss}`). Currently such cases are silently ignored, potentially leaving an end-user thinking that the plan is going according to plan.

How `VRT_purge` should actually behave (eg. ignore or fail) has yet to be discussed. I also need to understand the side effects of calling it with a non-zero TTL, this is why coverage is only for the `return (purge)` equivalent so far. Knowledgeable people in this area are welcome to enlighten me and save me some research/testing :)

This function is for VMODs, and what I'd like to do instead of simply changing `VRT_purge` is to write a very simple VMOD with three functions:

```
$Module purge 3 Varnish Purge Module

$Function INT hard()
Equivalent to return(purge) but called from vcl_{hit,miss}

$Function INT soft(DURATION grace = -1, DURATION keep = -1)
Set the TTL to 0 and preserve grace and keep by default.

$Function INT rearm(DURATION ttl, DURATION grace, DURATION keep)
Reset all three properties, I haven't pondered default values yet.
```

This would mean a purge batteries included in 6.0, negative values would fall back to the current object's values. You may also notice that they'd return an INT instead of nothing, that would be the number of affected objects implying an ABI change. Fortunately for 6.0 we are already likely to break VRT. I'm adding the _breaking_ label to this issue for this hypothetical VMOD. Such a VMOD would mostly be documentation with a bit of actual code.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -539,11 +539,17 @@ VRT_purge(VRT_CTX, double ttl, double grace, double keep)
 {
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+
+	if (ctx->method != VCL_MET_HIT && ctx->method != VCL_MET_MISS) {
+		VSLb(ctx->vsl, SLT_VCL_Error,
+		    "purge can only happen in vcl_hit{} or vcl_miss{}");
+		VRT_handling(ctx, VCL_RET_FAIL);
+		return;
+	}
+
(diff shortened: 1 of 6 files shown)


Dridi:
Discussed during last week's bugwash:

- we want to fail illegal purges
- changing the VRT_purge to know the number of "purged" objects is OK
- a vmod-purge would be OK in-tree if it passes the review

In the mean time I will add a changelog entry and merge this PR, dealing with the other points in a future PR.

PR #2346, opened 2017-06-14
fix small bug (http wasn't checked)

jambon69:
fix a little bug. only the fm[0] was checked for http


bsdphk:
Good catch!

I have fixed the original array index (copy&paste) mistake in aa8f6a1b9828677, because for performance reasons we prefer not to call any of those str*(3) functions which might pull in the entire UNICODE/rune mightmare when we deal with strict ASCII fields.

PR #2347, opened 2017-06-16
Make sure first_byte_timeout is applied on reused backend connection

### LINKED ISSUE #1772: first_byte_timeout is ignored when re-using a backend connection (HTTP Keep Alive)

fgaillot:
This avoids getting stuck with a stalled backend connection until
reaching backend_idle_timeout.

We now wait for the waiter to release the fd for a limited amount of
time (first_byte_timeout). In case we reach the timeout, the backend
connection is killed to make sure we are not going to reuse it.

Fixes #1772.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -157,7 +157,7 @@ vbe_dir_finish(const struct director *d, struct worker *wrk,
 	CAST_OBJ_NOTNULL(vbc, bo->htc->priv, VBC_MAGIC);
 	bo->htc->priv = NULL;
 	if (vbc->state != VBC_STATE_USED)
-		VBT_Wait(wrk, vbc);
+		AZ(VBT_Wait(wrk, vbc, 0));
 	if (bo->htc->doclose != SC_NULL || bp->proxy_header != 0) {
 		VSLb(bo->vsl, SLT_BackendClose, "%d %s", vbc->fd,
 		    bp->display_name);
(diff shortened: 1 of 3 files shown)


Dridi:
For reference, the test case from the original ticket (passes on my machine).

```vcl
varnishtest "honor first-byte timeout on a stolen connection"

server s1 {
        rxreq
        expect req.url == "/first"
        txresp -hdr "Connection: keep-alive"

        rxreq
        expect req.url == "/second"
        delay 2
        txresp
} -start

varnish v1 -vcl {
        backend s1 {
                .host = "${s1_addr}";
                .port = "${s1_port}";
                .first_byte_timeout = 1s;
        }
} -start

client c1 {
        txreq -url "/first"
        rxresp
        expect resp.status == 200

        txreq -url "/second"
        rxresp
        expect resp.status == 503
} -run
```

gquintard:
discuss this during next bugwash?

daghf:
I gave this a spin, and it turns out that `shutdown(vbc->fd, SHUT_WR)` does NOT trigger an epoll event. If the idea was that this is there to trigger an event for epoll to let us clean up the connection immediately, it doesn't work.

Playing around with it a bit, I found:

- SHUT_RD: This closes the reader-side of the socket. This also triggers an epoll event, since there is now an `EOF` to be found.
- SHUT_WR: Closes the writer side, transmits a FIN. Does not trigger an epoll.

daghf:
Fixed via #2504

Issue #2325, opened 2017-05-07
Enforce proper symbol names for storage back-ends


Dridi:
I was looking under the storage carpet and found two bugs. After I squashed the first one (#2321) the second turned into some kind of huge monster and suddenly I could no longer submit a fix.

Storage back-ends are exposed in VCL as `storage.<name>` symbols but the name can currently be anything:

```
varnishtest "nonsensical name"

varnish v1 -arg "-s ...=malloc" -cli storage.list
```

From the test report:

```
**** v1    0.1 CLI TX|storage.list
***  v1    0.2 CLI RX  200
**** v1    0.2 CLI RX|Storage devices:
**** v1    0.2 CLI RX|\tstorage.... = malloc           <= SEE HERE
**** v1    0.2 CLI RX|\tstorage.Transient = malloc
**   v1    0.2 CLI 200 <storage.list>
*    top   0.2 RESETTING after test.vtc
```

When I first reported this it turned into a discussion about unicode support in VCL symbols instead of allowing C-like symbol names. It was also suggested that `...` would be a proper symbol name, provided that we could use an escaped name that would be a proper symbol name. So on top of supporting characters outside of ascii, any character could become valid (extreme case IIUC).

I'm tracking this bug until we decide whether we kill or tame the unicode monster. Otherwise I can send a patch.

bsdphk:
Now would be a good time to decide definitively what is a legal "name" in VCL.

Xcir's feedback seems to indicate that UNICODE is not a hot-ticket item, but I would prefer to not cut of that avenue yet.

I think my proposal would be `[a-zA-Z][a-zA-Z0-9_-]*`

The inclusion of '-' means that we need a canonical way to quote this into C-identifiers, but this is entirely confined to VCC and therefore not an API issue, neither now or going forward.

We already have code which does this one place:

        VSB_printf(vsb, "&VGC_%s_", vh->rname);
        for (p = b, u = 1; p < e; p++, u++)
                if (vct_isalpha(*p) || vct_isdigit(*p))
                        VSB_putc(vsb, *p);
                else
                        VSB_printf(vsb, "_%02x_", *p);
        AZ(VSB_finish(vsb));

And I think that works OK for me, and it is UTF-8 robust, should we need it.

The trailing '_' is not necessary for undoing the quoting, but I feel it improves readability.

We should add a canonical function (in VCT ?) to validate a VCL compliant name, so MGT can bail out early and intelligently.

[...]

PR #2354, opened 2017-06-23
First series of "vcl 6.0;" changes

Dridi:
This is a partial implementation of what was reported here: https://github.com/varnishcache/varnish-cache/issues/2325#issuecomment-301457374

The changes to the VCL syntax are summed up in this test case:

https://github.com/Dridi/varnish-cache/blob/vsym/bin/varnishtest/tests/c00084.vtc

Basically, all kinds of VCL symbols can contain dashes (`-`) and VCL labels moved to a name space. This isn't done yet for VMODs, it is not possible to declare a function with this character yet. The other missing bit is symbol uniqueness for everything that is not in a name space.

I'm not continuing this effort yet, I will take advantage of the probable delay until it's reviewed and hopefully merged to work on some other assignment. However I will of course reply to comments as soon as I see them.

### DIFF
--- a/bin/varnishtest/tests/c00084.vtc
+++ b/bin/varnishtest/tests/c00084.vtc
@@ -1,0 +1,38 @@
+varnishtest "legal symbol names"
+
+varnish v1 -arg "-s my-store=malloc" -vcl {
+	import directors;
+
+	acl my-acl { "127.0.0.1"; }
+
+	probe my-pb { }
+	backend my-be {
+		.host = "${bad_backend}";
+		.probe = my-pb;
+	}
+
+	sub vcl_init {
+		new my-dir = directors.round_robin();
+		my-dir.add_backend(my-be);
+	}
+
+	sub vcl_recv {
+		call my-sub;
+	}
+
+	sub my-sub {
+		if (client.ip ~ my-acl) { }
+		set req.storage = storage.my-store;
+		set req.backend_hint = my-dir.backend();
+	}
+} -start
+
+varnish v1 -cli "vcl.label my-label vcl1"
+
+varnish v1 -vcl {
+	backend dummy { .host = "${bad_backend}"; }
+
+	sub vcl_recv {
+		return (vcl(my-label));
+	}
+}
--- a/bin/varnishtest/tests/d00006.vtc
+++ b/bin/varnishtest/tests/d00006.vtc
@@ -22,11 +22,11 @@ server s4 {
 	txresp -body "4444"
 } -start
 
-varnish v1 -errvcl {Name of VCL object, 'rr1-xx', contains illegal character '-'} {
+varnish v1 -errvcl {Name of VCL object, 'rr1.xx', contains illegal character '.'} {
 	import directors;
 	backend b1 { .host = "127.0.0.1"; .port = "8080";}
 	sub vcl_init {
-		new rr1-xx = directors.round_robin();
+		new rr1.xx = directors.round_robin();
 	}
 }
 
--- a/bin/varnishtest/tests/v00020.vtc
+++ b/bin/varnishtest/tests/v00020.vtc
@@ -204,12 +204,12 @@ varnish v1 -errvcl {IP + IP not possible.} {
 	}
 }
 
-varnish v1 -errvcl {Name of function, 'foo-bar', contains illegal character '-'} {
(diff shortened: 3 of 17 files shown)


bsdphk:
Re VCT_bad_name():

1. It should be VCT_invalid_name(), we don't pass value judgements :-)

2. We should implement a VCT_NAMEXTRA class to contain '-' and '_'

3. We should #define vct_isname(x) vct_is(x, VCT_ALPHHA|VCT_DIGIT|VCT_NAMEEXTRA)

4. I prefer the "if (e == NULL) e = strchr(b, '\0');" rather than scattered and mixed "if (b>e || *b=='\0') checks

bsdphk:
Re mcf_bad_vclname()

I wouldn't bother pointing out exactly which character was offensive, it just makes for more work writing coverage tests.  "Invalid name (%s)" is sufficient.

bsdphk:
Second patch adds symbol->cname, and fills it in without using it ?

->cname should just alias ->name if no escaping is necessary.

Avoiding the kind of two-pass string processing in vcc_c_symbol() which risks getting out of sync if somebody modifies the code later, is exactly why VSB exists.

bsdphk:
I'm to too thrilled about the overall direction of the rest of the patches, it doesn't feel right and adds too much code where things should get simpler.

Some of this is probably the weird/hybrid symbol-table in VCC and fixing that first may be a good idea.

Dridi:
> It should be VCT_invalid_name(), we don't pass value judgements :-)

Named after `mcf_bad_vclname()` for consistency, so we need to rename that one too ;-)

> We should implement a VCT_NAMEXTRA class to contain '-' and '_'
>
> We should #define vct_isname(x) vct_is(x, VCT_ALPHHA|VCT_DIGIT|VCT_NAMEEXTRA)

Yes but `vct_isname` wouldn't be accurate for the first character. Once again, patched that way because derived from existing code.

> I prefer the "if (e == NULL) e = strchr(b, '\0');" rather than scattered and mixed "if (b>e || *b=='\0') checks

Noted, I'd like to go through the all the comments before amending the PR.

> Re mcf_bad_vclname()
>
> I wouldn't bother pointing out exactly which character was offensive, it just makes for more work writing coverage tests. "Invalid name (%s)" is sufficient.

This is the current behavior, left untouched. I can change that, but I didn't want to change more than necessary to accommodate the new rules. Will amend once we agree on all your comments.

There is similar reporting somewhere else, I can simplify that too.

> Second patch adds symbol->cname, and fills it in without using it ?

Yes, I wanted to keep this change isolated/highlighted from the rest. I encountered other needed changes before actually starting to use the `cname`. I can probably move this patch before the first `cname` use if that helps.

> ->cname should just alias ->name if no escaping is necessary.
>
> Avoiding the kind of two-pass string processing in vcc_c_symbol() which risks getting out of sync if somebody modifies the code later, is exactly why VSB exists.

Good point, I will change that. It actually did go out of sync for a tiny short while but it was caught by the assert when I moved from `_%02x` to `_%02x_`. :)

> I'm to too thrilled about the overall direction of the rest of the patches, it doesn't feel right and adds too much code where things should get simpler.
>
> Some of this is probably the weird/hybrid symbol-table in VCC and fixing that first may be a good idea.

So enforcing symbol uniqueness first and then supporting the '-' character? I did go for the low-hanging fruits first, so I can try this angle if you confirm I'm understanding your comment right.

Dridi:
Following IRC discussions, I pushed the following commits:

745a6e0
e7ed02c
56f7ff4
3b2c6d3
ff4f115
1a14e5e

And for now I'll rebase this PR against master.

Dridi:
New set of commits to review once they pass the CI checks: a0b34bd...f6a66f9

Dridi:
Considering that August is already here, I think we should keep VCL to 4.0 for the next release, so keep all these patches but the last one for now. I don't think we have enough time at this point to bake in all the VCL breaking changes we discussed.

nigoroll:
+1 re @Dridi

Dridi:
I removed f6a66f9 so now this PR essentially allows any symbol to have dashes in their names. This was the consensus during today's bugwash.

hermunn:
Backport review:Since this is "breaking changes" they will not make it to 4.1.

PR #2356, opened 2017-06-27
Add extra locking to protect the pools list and refcounts

mbgrydeland:
Probes currently running on a worker thread at the time they are
deleted will delay the release of the refcount they hold on the TCP
pool. Since this call will not be from the CLI thread we need locking
to protect these datastructures.


bsdphk:
OK

PR #2357, opened 2017-06-28
Add the vtc feature ignore_unknown_macro.

slimhazard:
When this feature is switched on, strings of the form ``"${foo}"`` may be used in a vtc test; the test does not fail with the "Macro not found" error.

I needed it for a VMOD that binds pcre2, which has a new substitution function where the substitution strings may use syntax like ``${1}``.

``vtc(7)`` points out that this feature should be used very carefully, if ever, since misspelled macros will be silently ignored.

### DIFF
--- a/bin/varnishtest/vtc.c
+++ b/bin/varnishtest/vtc.c
@@ -58,6 +58,7 @@ volatile sig_atomic_t	vtc_error;	/* Error encountered */
 int			vtc_stop;	/* Stops current test without error */
 pthread_t		vtc_thread;
 static struct vtclog	*vltop;
+static int		ign_unknown_macro = 0;
 
 /**********************************************************************
  * Macro facility
@@ -236,13 +237,18 @@ macro_expand(struct vtclog *vl, const char *text)
 		p += 2;
 		m = macro_get(p, q);
 		if (m == NULL) {
-			VSB_destroy(&vsb);
-			vtc_fatal(vl, "Macro ${%.*s} not found", (int)(q - p),
-			    p);
-			NEEDLESS(return (NULL));
+			if (!ign_unknown_macro) {
+				VSB_destroy(&vsb);
+				vtc_fatal(vl, "Macro ${%.*s} not found",
+					  (int)(q - p), p);
+				NEEDLESS(return (NULL));
+			}
+			VSB_printf(vsb, "${%.*s}", (int)(q - p), p);
 		}
-		VSB_printf(vsb, "%s", m);
-		free(m);
+		else {
+			VSB_printf(vsb, "%s", m);
+			free(m);
+		}
 		text = q + 1;
 	}
 	AZ(VSB_finish(vsb));
@@ -662,8 +668,9 @@ cmd_delay(CMD_ARGS)
 
 /* SECTION: feature feature
  *
- * Test that the required feature(s) for a test are available, and skip the test
- * otherwise. feature takes any number of arguments from this list:
+ * Test that the required feature(s) for a test are available, and skip
+ * the test otherwise; or change the interpretation of the test, as
+ * documented below. feature takes any number of arguments from this list:
  *
  * SO_RCVTIMEO_WORKS
  *        The SO_RCVTIMEO socket option is working
@@ -685,6 +692,13 @@ cmd_delay(CMD_ARGS)
  *        The varnish group is present
  * cmd <command-line>
  *        A command line that should execute with a zero exit status
+ * ignore_unknown_macro
+ *        Do not fail the test if a string of the form ${...} is not
+ *        recognized as a macro.
+ *
+ * Be careful with the last feature, because it may cause a test with a
+ * misspelled macro to fail silently. You should only need if you must
+ * run a test with strings of the form "${...}".
  */
 
 static void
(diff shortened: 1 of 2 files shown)


bin/varnishtest/vtc.c
```diff
@@ -746,7 +760,8 @@ cmd_feature(CMD_ARGS)
 				vtc_stop = 2;
 			}
 #endif
-		} else if (!strcmp(*av, "cmd")) {
+		}
+		else if (!strcmp(*av, "cmd")) {
```
Dridi: Although I personally prefer have `else` branches on the next line, as pointed out to me several times this is not the FreeBSD/Varnish code style.

Dridi:
This is very error-prone, shouldn't we simply need to be able to escape (`\${...}`) macros?

Also with the decisions made for #2325 (being implemented in #2354) I was planning to also allow "Varnish identifiers" for `varnishtest` macros. So `${1}` wouldn't be a legal macro, and that instead could be ignored and left alone.

slimhazard:
Tbh I'm doing it this way because it was @bsdphk 's suggestion on IRC. %^)

I'd be open to other solutions, although it might not be worth it to open too many cans of worms for something that has never been a problem for years, and has now only come up for a very specific application.

The escape solution would work, although eventually someone will come along who needs a test with literal ``\${...}``, that is a string with backslash, dollar, brace ... So then we'd need ``\\`` to escape the escape.

Escapes would be inconsistent with the interpretation of strings everywhere else in Varnish -- ``\`` characters are not special in VCL or vtc, so things like ``\n`` or ``\$`` just mean backslash followed by ``n`` or ``$``. We'd have to decide, and make clear in the docs, whether backslashes affect anything else in vtc, or just the dollar.

Rejecting macros with non-legal identifiers wouldn't be enough for the pcre2 use case, because the substitution syntax can also have ``${name}`` for a named reference, and the ``name`` would pass as a VCL identifier.

@Dridi I can see your point, and maybe we could go with the escape suggestion. But I could also see the point in going with a simple solution that may be error-prone, but doesn't make us think about corner cases, since it's apparently been a problem only once in over ten years of Varnish.

That might, in fact, count as a reason not to do this at all. So far I've been able to go with a workaround -- if you include code with the ``${foo}`` syntax in another file, then varnishtest doesn't object and the tests can be run.

bin/varnishtest/vtc.c
slimhazard: Really? In the next hunk further down, that would look like this:
```
                        else
                                vtc_stop = 2;
                } else if (!strcmp(*av, "ignore_unknown_macro")) {
                        ign_unknown_macro = 1;
                        good = 1;
                }
```
Which looked too crowded and ugly to me -- you can hardly see the separation from the if block. So I put the else if on a separate line, and changed this line only so that the syntax would be consistent.

But all right then, if that's the house style, that's what we'll do.

(Not that it matters, since it's the house style, but I despise else on the same line as the closing brace for if.)

Dridi: I know, I would rather break the line after the brace...

Dridi:
> The escape solution would work, although eventually someone will come along who needs a test with literal \${...}, that is a string with backslash, dollar, brace ... So then we'd need \\ to escape the escape.

Not exactly, we'd do it right away. If we escape anything, the escape needs to be escapable ;)

> @Dridi I can see your point, and maybe we could go with the escape suggestion. But I could also see the point in going with a simple solution that may be error-prone, but doesn't make us think about corner cases, since it's apparently been a problem only once in over ten years of Varnish.

No, I've actually had the case with the `shell` and `process` commands where sometimes I would need the `${...}` syntax.

> That might, in fact, count as a reason not to do this at all. So far I've been able to go with a workaround -- if you include code with the ${foo} syntax in another file, then varnishtest doesn't object and the tests can be run.

I'd rather have an escape instead of needlessly moving things outside of my VTCs :-/

slimhazard:
@Dridi knowing how you and I like to discuss things down to the smallest nuance like world champion shedders of bikes (and I'm saying that as a friend), I suggest we keep it short and bring it up at the next bugwash. I've added the appropriate labels.

> Not exactly, we'd do it right away. If we escape anything, the escape needs to be escapable ;)

OK, but as I said, does this extend to everything else, like ``\n`` and so forth, which is never interpreted as an escape sequence in VCL? I fear that there may be madness at the end of that road.

> No, I've actually had the case with the shell and process commands where sometimes I would need the ``${...}`` syntax.

Aaaah, OK you've convinced me, there really is a need for a general solution here, it's not just a one-time quirk.

Dridi:
> OK, but as I said, does this extend to everything else, like \n and so forth, which is never interpreted as an escape sequence in VCL? I fear that there may be madness at the end of that road.

As long as it is properly documented, I don't have an opinion on what to escape. If we need to keep escaping to a minimum, so be it, `\\` and `\${`.

> OK you've convinced me

Cool! And for the record, all my comments are short in this thread! 😬

bsdphk:
The choice to not have \-escapes in VTC was very deliberate, and I'm not at all inclined to change it.

@slimhazard's suggestion for a feature for this very special corner case has my vote.

Dridi:
In that case I'll make sure that this feature will also ignore invalid names once macros can be named with `[a-zA-Z][a-zA-Z0-9-_]` like VCL symbols. That will do for conflicts with the shell syntax.

Dridi:
I looked at the failure on Travis CI and it wasn't interesting so I restarted the failing jobs and now it's passing.

bin/varnishtest/vtc.c
slimhazard: Note to self: the doc here should say "You should only need **it** if ...".

hermunn:
Back ported to 4.1 as 1e09758e0d5e5a27.

PR #2362, opened 2017-07-06
Introduce the vmod vtc for varnishtest

Dridi:
This is moving some of the generally useful functions of vmod debug to a
new one that is installed for use out of the Varnish source tree. Things
weren't just moved from one source file to another, some functions were
sligthly renamed, and the behavior of `barrier_sync` was changed to
adhere to the "universal fail" and no longer returns a boolean.

While at it, I added missing \_\_match_proto__ macros to vmod debug.

Closes #2276

### DIFF
--- a/bin/varnishtest/tests/c00070.vtc
+++ b/bin/varnishtest/tests/c00070.vtc
@@ -1,4 +1,4 @@
-varnishtest "Test workspace functions in vmod_debug"
+varnishtest "Test workspace functions in vmod_vtc"
 
 server s1 {
 	rxreq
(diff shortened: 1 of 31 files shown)


bsdphk:
This is probably a good idea.

We should not pull the rug on people using vmod_debug without notice.  Add the new vmod_vtc, and have the vmod_debug versions VSL a notice to switch to vmod vtc, until we remove them from vmod_debug at a later date (in a year/two releases ?)

hermunn:
Backport review: This is not going to 4.1.

PR #2372, opened 2017-07-19
Fix problem with purging and the n_obj_purged counter

hermunn:
The events leading to this patch started with me looking at different counters and their documentation. I realized that the counter `n_obj_purged` is not correctly implemented. Unfortunately, the bug is worse than this, as described in the commit message. The test on the `oc->ttl` is ugly, but it is there because the `oc->flags` is just 8 bits, and they are all taken. A "competing" patch is in https://github.com/hermunn/varnish-cache/tree/master-unpopular-fix-n_obj_purged, but, as you can see from the name of the branch, I do not think it will be popular. Please have a look at both before commenting.

The two fixes (this and the unpolular) causes different behavior in the following scenario:
* A "soft purge" sets the TTL of the object to 30 seconds.
* Shortly after, a "soft purge" sets the TTL to 60 seconds.
(keep and grace are set to the same value in the two "soft purge" calls).

In the unpopular branch the resulting TTL will be 60 seconds, while on this it will be 30 seconds, and `n_obj_purged` will be the number of times the TTL changed.

### DIFF
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -590,7 +590,7 @@ HSH_Purge(struct worker *wrk, struct objhead *oh, double ttl, double grace,
 double keep)
 {
 	struct objcore *oc, **ocp;
-	unsigned spc, ospc, nobj, n;
+	unsigned spc, ospc, nobj, n, n_tot = 0;
 	int more = 0;
 	double now;
 
@@ -598,6 +598,20 @@ double keep)
 	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
 	ospc = WS_Reserve(wrk->aws, 0);
 	assert(ospc >= sizeof *ocp);
+	/*
+	 * Because of "soft" purges, there might be oc's in the list that has
+	 * the OC_F_PURGED flag set. We do not want to let these slip through,
+	 * so we need to clear the flag before entering the do..while loop.
+	 */
+	Lck_Lock(&oh->mtx);
+	assert(oh->refcnt > 0);
+	VTAILQ_FOREACH(oc, &oh->objcs, hsh_list) {
+		CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
+		assert(oc->objhead == oh);
+		oc->flags &= ~OC_F_PURGED;
+	}
+	Lck_Unlock(&oh->mtx);
+
 	do {
 		more = 0;
 		spc = ospc;
@@ -621,6 +635,15 @@ double keep)
 			}
 			if (oc->flags & OC_F_DYING)
 				continue;
+			if (oc->flags & OC_F_PURGED) {
+				/*
+				 * We have already called EXP_Rearm on this
+				 * object, and we do not want to do it
+				 * again. Plus the space in the ocp array may
+				 * be limited.
+				 */
+				continue;
+			}
 			if (spc < sizeof *ocp) {
 				/* Iterate if aws is not big enough */
 				more = 1;
@@ -629,6 +652,7 @@ double keep)
 			oc->refcnt++;
 			spc -= sizeof *ocp;
 			ocp[nobj++] = oc;
+			oc->flags |= OC_F_PURGED;
 		}
 		Lck_Unlock(&oh->mtx);
 
@@ -638,9 +662,10 @@ double keep)
 			EXP_Rearm(oc, now, ttl, grace, keep);
 			(void)HSH_DerefObjCore(wrk, &oc, 0);
 		}
+		n_tot += nobj;
 	} while (more);
 	WS_Release(wrk->aws, 0);
-	Pool_PurgeStat(nobj);
+	Pool_PurgeStat(n_tot);
 }
 
 /*---------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


fgsch:
Can we just change the loop to be:

```
loop through the objects {
   if no space {
     purge
     reset space
   }
   add object
}
if anything remainng {
   purge
}
```

I appreciate we'll be calling purge in 2 different places but this seems tidier than the proposed approach. We could move purge into a helper function for dedup purposes.

fgsch:
Something like this:

```
diff --git a/bin/varnishd/cache/cache_hash.c b/bin/varnishd/cache/cache_hash.c
index 986af6f00..0280215a9 100644
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -585,60 +585,71 @@ hsh_rush2(struct worker *wrk, struct rush *r)
  * Purge an entire objhead
  */
 
+static void
+hsh_purge(struct worker *wrk, struct objcore **ocp, unsigned noc,
+    double now, double ttl, double grace, double keep)
+{
+	struct objcore *oc;
+	unsigned n;
+
+	for (n = 0; n < noc; n++) {
+		oc = ocp[n];
+		CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
+		EXP_Rearm(oc, now, ttl, grace, keep);
+		(void)HSH_DerefObjCore(wrk, &oc, 0);
+	}
+}
+
 void
 HSH_Purge(struct worker *wrk, struct objhead *oh, double ttl, double grace,
-double keep)
+    double keep)
 {
 	struct objcore *oc, **ocp;
-	unsigned spc, ospc, nobj, n;
-	int more = 0;
+	unsigned spc, ospc, nobj, noc;
 	double now;
 
 	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
 	ospc = WS_Reserve(wrk->aws, 0);
 	assert(ospc >= sizeof *ocp);
-	do {
-		more = 0;
-		spc = ospc;
-		nobj = 0;
-		ocp = (void*)wrk->aws->f;
-		Lck_Lock(&oh->mtx);
-		assert(oh->refcnt > 0);
-		now = VTIM_real();
-		VTAILQ_FOREACH(oc, &oh->objcs, hsh_list) {
-			CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
-			assert(oc->objhead == oh);
-			if (oc->flags & OC_F_BUSY) {
-				/*
-				 * We cannot purge busy objects here, because
-				 * their owners have special rights to them,
-				 * and may nuke them without concern for the
-				 * refcount, which by definition always must
-				 * be one, so they don't check.
-				 */
-				continue;
-			}
-			if (oc->flags & OC_F_DYING)
-				continue;
-			if (spc < sizeof *ocp) {
-				/* Iterate if aws is not big enough */
-				more = 1;
-				break;
-			}
-			oc->refcnt++;
-			spc -= sizeof *ocp;
-			ocp[nobj++] = oc;
+	spc = ospc;
+	noc = 0;
+	nobj = 0;
+	ocp = (void*)wrk->aws->f;
+	now = VTIM_real();
+	Lck_Lock(&oh->mtx);
+	assert(oh->refcnt > 0);
+	VTAILQ_FOREACH(oc, &oh->objcs, hsh_list) {
+		CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
+		assert(oc->objhead == oh);
+		if (oc->flags & OC_F_BUSY) {
+			/*
+			 * We cannot purge busy objects here, because
+			 * their owners have special rights to them,
+			 * and may nuke them without concern for the
+			 * refcount, which by definition always must
+			 * be one, so they don't check.
+			 */
+			continue;
 		}
-		Lck_Unlock(&oh->mtx);
-
-		for (n = 0; n < nobj; n++) {
-			oc = ocp[n];
-			CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
-			EXP_Rearm(oc, now, ttl, grace, keep);
-			(void)HSH_DerefObjCore(wrk, &oc, 0);
+		if (oc->flags & OC_F_DYING)
+			continue;
+		if (spc > sizeof *ocp) {
+			Lck_Unlock(&oh->mtx);
+			hsh_purge(wrk, ocp, noc, now, ttl, grace, keep);
+			Lck_Lock(&oh->mtx);
+			ocp = (void*)wrk->aws->f;
+			spc = ospc;
+			noc = 0;
 		}
-	} while (more);
+		oc->refcnt++;
+		spc -= sizeof *ocp;
+		ocp[noc++] = oc;
+		nobj++;
+	}
+	Lck_Unlock(&oh->mtx);
+	if (noc > 0)
+		hsh_purge(wrk, ocp, noc, now, ttl, grace, keep);
 	WS_Release(wrk->aws, 0);
 	Pool_PurgeStat(nobj);
 }
```

hermunn:
Is unlocking oh->mtx inside of the `VTAILQ_FOREACH` loop is safe? I think the existing code is like it is because we cannot trust the state of the for loop after releasing and reacquiring the lock.

fgsch:
Of course not, you are right 😞 

I prefer the approach in your unpopular branch.

bsdphk:
The unpopular branch just got much more popular because OC_F (1<<0) is available.

OK to that.

hermunn:
This was applied to master as c21fae9a2b, closing.

hermunn:
Backported to 4.1 as c2e69f4d with the two necessary follow-up commits 8625540bf7 and 4f1244464e6.

PR #2376, opened 2017-07-21
Expose to VCL whether or not a fetch is a background fetch

nigoroll:
The use case are cluster requests: Intra-cluster bgfetches should trigger a synchronous fetch on the peer-varnish in order to avoid additional short-lived / expired objects being created. Or, in other words, a bgfetch should actually get a fresh object and not one in grace from another cache.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -1020,19 +1020,27 @@ VBF_Fetch(struct worker *wrk, struct req *req, struct objcore *oc,
 	AN(oc->flags & OC_F_BUSY);
 	CHECK_OBJ_ORNULL(oldoc, OBJCORE_MAGIC);
 
-
-	switch (mode) {
-	case VBF_PASS:		how = "pass"; break;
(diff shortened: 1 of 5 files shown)


fgsch:
Can we follow the existing pattern and have the variable called `is_background`? 
In fact, can we name it `is_bgfetch`? I don't think `background' is distinctive enough.

rezan:
Much needed (+1). Have been spending a lot of time explaining this to people, the pitfalls of running multiple tiers and caching features like grace. [Some VCL](https://gist.github.com/rezan/91499d1588b4148612e99fda4f3f9ea8).

nigoroll:
@fgsch the `is_*` were all `beresp` properties, but I'm fine with `is_bgfetch`.
Force-pushed the changed patch

fgsch:
+1 for the idea and code.

nigoroll:
travis failure is in r01490.vtc, which is unrelated.

bsdphk:
Looks OK.

fgsch:
Applied.

hermunn:
This was just back ported as bf167ec4c.

PR #2381, opened 2017-08-01
retire s_req - it is identical to client_req

nigoroll:
do we need any more words?


bsdphk:
No objection.

Not sure if fallout in any important tools.

nigoroll:
it should be easy change or alias in tools. Will add a changelog entry

fgsch:
Applied.

hermunn:
Backport review: We will not do this in 4.1, as it the gain of a back port is tiny compared to the potential problems this can cause.

PR #2382, opened 2017-08-02
Move the cli buffer to malloc (from stack)

hermunn:
If the cli buffer is allocated on the stack, the mgt process will crash
hard for big values of this parameter. Instead we allocate on the heap.

A new error is introduced, "CLI buffer overrun", when the buffer is too
small for the payload.

(I will change the file name for the test as soon as a number is assigned)

### DIFF
--- a/bin/varnishd/mgt/mgt_cli.c
+++ b/bin/varnishd/mgt/mgt_cli.c
@@ -66,6 +66,8 @@ static int		cli_i = -1, cli_o = -1;
 struct VCLS		*mgt_cls;
 static const char	*secret_file;
 
+struct vsb		*cli_buf = NULL;
+
 /*--------------------------------------------------------------------*/
 
 static void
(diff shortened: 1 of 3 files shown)


hermunn:
A small note: The alternative is to introduce a maximum size for the `cli_buffer` parameter, and have it low enough for the default stack size (mgt process) to easily contain the buffer. However, I see no reason not to allocate the buffer through `malloc`.

bsdphk:
Why not simply make it a VSB ?

hermunn:
> Why not simply make it a VSB ?

I have no good explanation for this. :) It is just me not being used to the varnish objects. Do you want me to update the pull request to a VSB solution?

bsdphk:
Yes, I think a VSB buffer will be much simpler and less code.

I think the reason it isn't a VSB is simply that code were written before we imported sbufs from FreeBSD.

hermunn:
An updated commit is 188574a695d957 , but I put it in another branch. If this looks OK (please discuss), then I can update this PR before merging.

The problem with the VSB approach is the `write` call, that might fail for very big inline VCLs (if I am not mistaken).

hermunn:
Since we are releasing 5.2 very soon, we should also consider removing the parameter `cli_buffer` completely without going through a *deprecation* period. If we go for this, the test case will need to go away from the patch 188574a.

fgsch:
Related to this, should we impose some upper limit to the size?

hermunn:
The question about upper limit that @fgsch brings up is crucial. It needs to be settled before this goes in.

Apart from that I think the PR should be very straightforward.

We can also discuss if we want to remove the parameter entirely for 5.2 instead of having a deprecation period, but this is secondary.

bsdphk:
Looks pretty good to me.

I think we should just remove the parameter ?

hermunn:
@bsdphk, I agree to remove the parameter and make a note of this in the changelog. I do not see any problems with the `VSB_auto` (unbounded size), but I wanted you to agree before pushing, especially since @fgsch brought it up.

fgsch:
I'm OK with removing the parameter but it will break installations of people using it, that's why I suggested to keep it (with the deprecated message) until next release.

hermunn:
After pondering this a bit, I think that this should go in as it is now. Well, one exception; I will make a note in the changelog about this being deprecated.

hermunn:
Back ported to 4.1 as 3f6230961612599f.

PR #2383, opened 2017-08-08
Fix sourcing of varnish.m4 when building on FreeBSD

feld:
Noticed when switching to Github for fetching source on FreeBSD and using autotools it never correctly sources varnish.m4 so configure fails with obtuse errors about missing functions which are defined in varnish.m4.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,4 +1,4 @@
-ACLOCAL_AMFLAGS = -I m4
+ACLOCAL_AMFLAGS = -I m4 -I .
 
 SUBDIRS = include lib bin etc doc man
 


Dridi:
Thanks!

hermunn:
@feld: Is this patch needed on 4.1? If so I will back port it (it applies cleanly, but I do not know if it makes sense). If not, I can remove the backport-review label.

Dridi:
@hermunn, yes, we should back-port it.

hermunn:
@Dridi Thanks for confirming. Back ported as 5e84d03.

Issue #2367, opened 2017-07-11
Varnish crashes with separate VCL files and POST requests


ernestojpg:
Hello!

I'm getting this segmentation fault in Varnish 5.1.2 when receiving POST requests in a simple configuration with two VCL files:

```
Error: Child (96827) died signal=6
Error: Child (96827) Panic at: Mon, 10 Jul 2017 21:38:34 GMT
Assert error in child_sigsegv_handler(), mgt/mgt_child.c line 271:
  Condition(Segmentation fault by instruction at 0x0) not true.
version = varnish-5.1.2 revision 6ece695, vrt api = 6.0
ident = Darwin,16.6.0,x86_64,-jnone,-smalloc,-smalloc,-hcritbit,kqueue
now = 3282252.606878 (mono), 1499722714.474813 (real)
Backtrace:
  0x102ca1a61: 0   varnishd                            0x0000000102ca1a61 pan_ic + 437
  0x102cd08f3: 0   varnishd                            0x0000000102cd08f3 child_sigsegv_handler + 102
  0x7fff8b1cfb3a: 0   libsystem_platform.dylib            0x00007fff8b1cfb3a _sigtramp + 26
  0x102d4f4c8: 0   varnishd                            0x0000000102d4f4c8 waiter + 5544
  0x102ca4b83: 0   varnishd                            0x0000000102ca4b83 vrb_pull + 199
  0x102ca51a7: 0   varnishd                            0x0000000102ca51a7 VRB_Ignore + 53
  0x102ca62da: 0   varnishd                            0x0000000102ca62da CNT_Request + 3754
  0x102cc504c: 0   varnishd                            0x0000000102cc504c http1_req + 491
  0x102cbd532: 0   varnishd                            0x0000000102cbd532 WRK_Thread + 1655
  0x102cbce5d: 0   varnishd                            0x0000000102cbce5d pool_thread + 51
thread = (cache-worker)
thr.req = 0x7f8401025420 {
  vxid = 2, transport = HTTP/1 {
    state = HTTP1::Proc
  }
  step = R_STP_FETCH,
  req_body = R_BODY_TAKEN,
  restarts = 0, esi_level = 0,
  sp = 0x7f8400e01480 {
    fd = 16, vxid = 1,
    t_open = 1499722714.474208,
    t_idle = 1499722714.474208,
    transport = HTTP/1 {
      state = HTTP1::Proc
    }
    client = 127.0.0.1 49537,
[...]

PR #2385, opened 2017-08-14
Move req->ws_req snapshotting until after VFP_Setup

daghf:
This lets the request's vfp stack survive a VCL label change.

Fixes: #2367


bsdphk:
This needs to be revisited when we add the "list of VFP's" facility to vcl_recv{}, but for now: OK.

Issue #2390, opened 2017-08-18
varnish 4.1.8 / Assert error in VCL_Get(), cache/cache_vcl.c line 168


n4n0r:
Hi,

Following the quick discussion the chat, I am openning an issue for this one. 

Running 4.1.8 official debian precise package with vmod std and directors.
Unfortunately I don't have more traces to reproduce the issue.

```
Last panic at: Fri, 18 Aug 2017 08:27:41 GMT
"Assert error in VCL_Get(), cache/cache_vcl.c line 168:
  Condition((pthread_rwlock_rdlock(&vcl_active->temp_rwl)) == 0) not true.
thread = (cache-worker)
version = varnish-4.1.8 revision d266ac5c6
ident = Linux,3.2.0-113-generic,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
now = 25113631.416082 (mono), 1503044858.623294 (real)
Backtrace:
  0x4349db: pan_ic+0x17b
  0x4414a2: varnishd() [0x4414a2]
  0x450ea3: HTTP1_Session+0x2a3
  0x43b01d: SES_Proto_Req+0x5d
  0x44b3aa: WRK_Thread+0x4da
  0x44b80b: pool_thread+0x2b
  0x7f8ffe23fe9a: libpthread.so.0(+0x7e9a) [0x7f8ffe23fe9a]
  0x7f8ffdf6d36d: libc.so.6(clone+0x6d) [0x7f8ffdf6d36d]
req = 0x7f8d29b7f020 {
  vxid = 18958891, step = R_STP_RESTART,
  req_body = R_BODY_INIT,
  restarts = 0, esi_level = 0,
  sp = 0x7f8ee60e2e20 {
    fd = 87, vxid = 19634664,
    client = xxx.xxx.xxx.xxx 7328,
    t_open = 1503044858.361968,
    t_idle = 1503044858.551950,
    step = S_STP_H1NEWREQ,
  },
  ws = 0x7f8d29b7f200 {
    id = \"req\",
    {s,f,r,e} = {0x7f8d29b81000,+976,(nil),+122872},
  },
  http_conn = 0x7f8d29b7f128 {
    fd = 87,
    doclose = NULL,
    ws = 0x7f8d29b7f200,
    {rxbuf_b, rxbuf_e} = {0x7f8d29b81000, 0x7f8d29b813cb},
    {pipeline_b, pipeline_e} = {(nil), (nil)},
    content_length = 0,
[...]

PR #2391, opened 2017-08-22
Straighten locking wrt vcl_active

mbgrydeland:
The locking around the use of vcl_active does not include the checking
of its magic value or the temperature asserts, leading to a race when
changing vcl_active.

Fixes: #2390


hermunn:
Backported by merging #2392.

PR #2392, opened 2017-08-22
Straighten locking wrt vcl_active

mbgrydeland:
The locking around the use of vcl_active does not include the checking
of its magic value or the temperature asserts, leading to a race when
changing vcl_active.

Fixes: #2390


mbgrydeland:
@hermunn This is for merging if/when #2391 goes into master

hermunn:
Thanks to @mbgrydeland for doing the back porting to 4.1!

Issue #2311, opened 2017-04-16
Missing errorhandling code in h2_end_headers(), http2/cache_http2_proto.c line 535


simonvik:
```
Panic at: Thu, 13 Apr 2017 14:36:01 GMT
Missing errorhandling code in h2_end_headers(), http2/cache_http2_proto.c line 535:
  Condition((Pool_Task(wrk->pool, &req->task, TASK_QUEUE_REQ)) == 0) not true.
version = varnish-5.1.1 revision de38712, vrt api = 6.0
ident = Linux,4.4.30-20161130-1423-0a2a05d,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
now = 2003538.537251 (mono), 1492094144.795931 (real)
Backtrace:
  0x438c25: /usr/sbin/varnishd() [0x438c25]
  0x4616db: /usr/sbin/varnishd() [0x4616db]
  0x46285a: /usr/sbin/varnishd(h2_rxframe+0x5aa) [0x46285a]
  0x462eee: /usr/sbin/varnishd() [0x462eee]
  0x4520a2: /usr/sbin/varnishd() [0x4520a2]
  0x45254b: /usr/sbin/varnishd() [0x45254b]
  0x7fa4ae6096ba: /lib/x86_64-linux-gnu/libpthread.so.0(+0x76ba) [0x7fa4ae6096ba]
  0x7fa4ae33f82d: /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7fa4ae33f82d]
thread = (cache-worker)
thr.req = 0x7f852e21a020 {
  vxid = 100213221, transport = H2
  step = 0x0,
  req_body = R_BODY_INIT,
  err_code = 1, err_reason = (null),
  restarts = 0, esi_level = 0,
  sp = 0x7f8d2b3e1020 {
    fd = 82533, vxid = 100213221,
    t_open = 1492092973.394044,
    t_idle = 1492094142.870537,
    transport = H2 {
      streams {
        0x00000000 idle  
        0x00000289 c-rem 
      }
    }
    client = 94.224.111.74 55237,
    privs = 0x7f8d2b3e1088 {
    },
  },
  ws = 0x7f852e21a208 {
    id = \"req\",
    {s, f, r, e} = {0x7f852e21c008, +440, (nil), +122864},
  },
  http_conn = 0x7f852e21a130 {
    fd = 82533 (@0x7f8d2b3e1038),
    doclose = NULL,
[...]

PR #2394, opened 2017-08-22
Refuse h2 streams that can't be scheduled

Dridi:
The initial plan was a one-liner to get rid of an `XXXAZ` assert (2nd patch) but testing this case proved to be hard because of a bug in the thread pools (fixed in 1st patch) and while I was playing a bit with h2 resets I realized that the `sess_drop` counters includes streams so I introduced a new `req_drop` counter (3rd patch).

I expect the 3rd patch to be potentially controversial, but I tried to keep the changes to a minimum.

As a side note, since we are breaking libvarnishapi, it may be a good occasion to sort the counters. For example `sess_drop` and `sess_dropped` could be next to each other.

### DIFF
--- a/bin/varnishd/main.vsc
+++ b/bin/varnishd/main.vsc
@@ -240,8 +240,14 @@
 .. varnish_vsc:: sess_dropped
 	:oneliner:	Sessions dropped for thread
 
-	Number of times session was dropped because the queue were too long
-	already. See also parameter thread_queue_limit.
+	Number of times an HTTP/1 session was dropped because the queue was
+	too long already. See also parameter thread_queue_limit.
+
+.. varnish_vsc:: req_dropped
+	:oneliner:	Requests dropped
+
+	Number of times an HTTP/2 stream was refused because the queue was
+	too long already. See also parameter thread_queue_limit.
 
 .. varnish_vsc:: n_object
 	:type:	gauge
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -294,9 +294,13 @@ enum task_prio {
 	TASK_QUEUE_BO,
 #define TASK_QUEUE_RESERVE	TASK_QUEUE_BO
 	TASK_QUEUE_REQ,
+	TASK_QUEUE_STR,
 	TASK_QUEUE_VCA,
 	TASK_QUEUE_END
 };
+
+#define TASK_QUEUE_CLIENT(prio) \
+	(prio == TASK_QUEUE_REQ || prio == TASK_QUEUE_STR)
 
 /*--------------------------------------------------------------------*/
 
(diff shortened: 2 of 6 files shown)


Dridi:
Fixes #2311

mbgrydeland:
For the counter patch, we have two option as I see it:

1. Docfix the sess_drop / rename it, to be the sum of scheduling errors (which could be H1 sessions, and H2 sessions or streams). Then we don't need to special case the scheduling type, and just increase ndropped also for these scheduling errors.

2. Keep the H1 and H2 errors counted separately. I am not sure I see the value of keeping these errors separate though, as they both draw from the same resource pool (the threads), and increasing the pool size is the possible action in both cases. If we go this way, the separate counter should be included in the pools locking scheme like ndropped is today, instead of adding a new lock type.

mbgrydeland:
I agree that the REFUSED_STREAM is the right response to send to the client on these types of errors. It is correct to have done the HPACK processing even though we are not replying on the stream, to keep up with the clients HPACK state.

Dridi:
> For the counter patch, we have two option as I see it:

I went for solution 2 because `sess_drop` isn't accurate for h2 streams. Depending on the consensus I will update the 3rd patch accordingly. Now that you mention it, the separate counter for `req_drop` is so obvious that it's embarrassing :)

Dridi:
I just realized that `sess_drop` isn't the correct homologous counter, so I just pushed a patch going for solution 2 again, taking @mbgrydeland's remarks into account, and introducing `req_dropped` instead.

Dridi:
I just realized that `sess_drop` isn't the correct homologous counter, so I just pushed a patch going for solution 2 again, taking @mbgrydeland's remarks into account, and introducing `req_dropped` instead.

bin/varnishd/main.vsc
```diff
@@ -240,8 +240,14 @@
...
-	Number of times session was dropped because the queue were too long
-	already. See also parameter thread_queue_limit.
+	Number of times an HTTP/1 session was dropped because the queues were
+	too long already. See also parameter thread_queue_limit.
+
+.. varnish_vsc:: req_dropped
+	:oneliner:	Requests dropped
+
+	Number of times an HTTP/2 stream was refused because the queues were
+	too long already. See also parameter thread_queue_limit.
```
fgsch: This sentence sounds weird. The "queue was full" or something similar would be better.

Dridi:
@fgsch thanks, now updated. It wasn't a good idea to make this change that late :)

bsdphk:
Tested on FreeBSD, looks OK to me.

daghf:
@Dridi, would you mind also having a look at the `XXXAZ(Pool_Task(..))` in `h2_ou_session` ?

PR #2396, opened 2017-08-23
Sigaltstack

nigoroll:
Some progress on handling stack overflows.

### DIFF
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -352,9 +352,23 @@ mgt_launch_child(struct cli *cli)
 			memset(&sa, 0, sizeof sa);
 			sa.sa_sigaction = child_signal_handler;
 			sa.sa_flags = SA_SIGINFO;
-			(void)sigaction(SIGSEGV, &sa, NULL);
 			(void)sigaction(SIGBUS, &sa, NULL);
 			(void)sigaction(SIGABRT, &sa, NULL);
+
+#ifdef HAVE_SIGALTSTACK
+			stack_t ss;
+			size_t sz = SIGSTKSZ + 4096;
+			if (sz < mgt_param.wthread_stacksize)
+				sz = mgt_param.wthread_stacksize;
+			ss.ss_sp = malloc(sz);
+			AN(ss.ss_sp);
+			ss.ss_size = sz;
+			ss.ss_flags = 0;
+			AZ(sigaltstack(&ss, NULL));
+			sa.sa_flags |= SA_ONSTACK;
+#endif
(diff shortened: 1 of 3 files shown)


Dridi:
The first commit should definitely land in master and 4.1 regardless of the outcome for the `sigaltstack` change.

bin/varnishd/mgt/mgt_child.c
```diff
@@ -352,9 +352,23 @@ mgt_launch_child(struct cli *cli)
...
 			sa.sa_flags = SA_SIGINFO;
-			(void)sigaction(SIGSEGV, &sa, NULL);
 			(void)sigaction(SIGBUS, &sa, NULL);
 			(void)sigaction(SIGABRT, &sa, NULL);
+
+#ifdef HAVE_SIGALTSTACK
+			stack_t ss;
+			size_t sz = SIGSTKSZ + 4096;
+			if (sz < mgt_param.wthread_stacksize)
+				sz = mgt_param.wthread_stacksize;
```
Dridi: What if `thread_pool_stack` is later increased? Won't this will leave the alternate stack with a smaller allocation than expected by an end-user?

nigoroll: Correct, but as the stack is only used for the signal handler, I would not consider this an issue.

Dridi: So maybe a `thread_pool_altstack` parameter with experimental+needs_restart flags would be a better choice? It could have the same default value as `thread_pool_stack` and a minimum value of `SIGSTKSZ + 4096`.

nigoroll: I'd consider this overkill, but if there is a good argument for it, not a problem.

Dridi:
The test's inline C doesn't compile on all Travis CI jobs:

https://travis-ci.org/varnishcache/varnish-cache/jobs/267529977#L2346

nigoroll:
thx @Dridi

hermunn:
Back ported to 4.1 as e44e54d and 5efbf4c.

nigoroll:
@hermunn I had pushed a couple more commits as vtest results were coming in:
* 6e6e7e738d991ba5baf4d39b9b9f22836215622b
* b7d42dd543688436a3a594878380d92680a8f7f0
* 421ebab3589fde462f551b1ff0103a646afbc25a

hermunn:
@nigoroll : Thank you for pointing at these commits. They have been ported to 4.1 as 0ef00ea1d e05e9f85d 772e716dc (oldest to newest). I will leave the backport review tag on for a bit, in case you want to have a look at it.

PR #2404, opened 2017-08-29
Introduce vmod-purge

Dridi:
This module offers a finer-grained control over purges, in the old
pre-vcl 4.0 style. For that, the signatures of VRT_purge and HSH_Purge
were updated to inform about the number of rearmed objects.

Initially suggested in #2339, it was agreed on IRC to review such a VMOD.

### DIFF
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -586,7 +586,7 @@ hsh_rush2(struct worker *wrk, struct rush *r)
  * Purge an entire objhead
  */
 
-void
+unsigned
 HSH_Purge(struct worker *wrk, struct objhead *oh, double ttl, double grace,
 double keep)
 {
@@ -667,6 +667,7 @@ double keep)
 	} while (more);
 	WS_Release(wrk->aws, 0);
 	Pool_PurgeStat(n_tot);
+	return (n_tot);
 }
 
 /*---------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


bsdphk:
Why cant vmod_purge::soft() set ttl to non-zero values ?

Dridi:
That is the behavior of vmod-softpurge, neuter TTL but keep other periods. In my initial proposal there was a `purge.rearm` function that could update all three of them. I can either add that function, or add a `ttl` parameter to `purge.soft` with a default value of 0. I have no strong opinion on the matter.

bsdphk:
I think you should add the ttl=0 param, I really don't see any reason not to.

What about error handling, anything which should cause VRT_fail() ?

Dridi:
Error handling is already in `VRT_purge`, this VMOD is merely a wrapper around it, I will update the code.

Dridi:
Patch updated.

fgsch:
I wonder, could we expose this in VCL, e.g. `return(purge(ttl=0s))` and expose the number of objects purged via some variable (e.g. `obj.purged`)?
We could do this for all the purge (and soft purge) variants.

bsdphk:
Test m00031 fails about one time out of 20 on my machine, here are some of the trigger lines:

  ---- c1    1.1 EXPECT resp.status (200) == "404" failed
  ---- s1   31.3 HTTP rx timeout (fd:4 30000 ms)
  ---- c1   31.3 HTTP rx timeout (fd:18 30000 ms)
  ---- s1   31.1 HTTP rx timeout (fd:4 30000 ms)
  ---- c1   31.1 HTTP rx timeout (fd:18 30000 ms)
  ---- s1   31.0 HTTP rx timeout (fd:4 30000 ms)
  ---- c1   31.0 HTTP rx timeout (fd:18 30000 ms)

Dridi:
Test case updated, no longer timing-sensitive on my system.

Dridi:
OK'd during today's bugwash.

Issue #2402, opened 2017-08-29
(be)req.hash return type


nigoroll:
reminder issue, so we don't forget to finish with the discussion we had on IRC:
With cd44549432ec809c508b5966c80347e6ff47b582 and 7bfd13b4bf5f1d1307495705278e643b00410d7b, access to the hash value was implemented as a hex-serialized string.
IMHO, the correct data type would be BLOB. The conversion to whatever encoding should be left to a vmod like https://code.uplex.de/uplex-varnish/libvmod-blobcode

fgsch:
Unless there is a specific usecase I don't see the point of returning a BLOB that will require a VMOD call (which we don't have in tree). This change was based on RL input and I consider it (to return a formatted string) to be the norm, not the exception.

nigoroll:
To explain my POV also:
* the native data type of a sha256 hash is BLOB, which has been introduced to do away with (arbitrary) choices of string encodings in the first place.
* Different users will require different encodings, which is why we want to abstract from particular use cases
* for any users not requiring hex encoding, the current decision will imply unnecessary overhead
* The fact that certain functionality is not in tree should not keep us from making the right decision, but rather motivate us to pull stuff into tree

fgsch:
All your points above assume that the hash either will be part of something else or needs to be represented in a different way (likely to be consumed by something else). 
This is not the input I got and the reason I wanted to expose the hash.

I'm all for changing this to a BLOB _if_ there is any need. I found that your proposal comes from the need to generalise rather than something concrete and will punish RL users that don't need this generalisation.

nigoroll:
So what _is_ your use case then?
Other than that, good interface design does need to generalize.

fgsch:
I believe most people expressed the desire to send the hash in a header without having to (re)compute it. As simple as that. I think @rezan might have some other use cases.

Feature creep, code rotting and security issues are also related to unneeded generalisation.

rezan:
The use case I have for this is using the hash as a string and hex encoding works well. The only concern I had was automatically converting all blob types to hex encoding when referenced as a string. I think encoding should be done case by case, and sometimes not at all. However, this could become a non issue with the right code abstractions. Either way, I do not feel very strongly here and as long as we can get the hash as a string (ie header), im good.

nigoroll:
@fgsch totally ok to have the hash in the header, my concerns are simply about the encoding
@rezan can you please explain "right code abstractios"?
[...]

PR #2407, opened 2017-08-30
Add VMOD blob

slimhazard:
Add a VMOD that supports binary-to-text encodings for the BLOB data type to the standard distribution.

This is a possible solution for the issue under discussion in #2402 .

This is nearly identical to the VMOD that has been maintained at:

https://code.uplex.de/uplex-varnish/libvmod-blobcode

The main difference is that the version of the VMOD in the PR invokes VCL failure (``VRT_Fail()``) for its error handling (mostly for illegal encodings, or when there is insufficient workspace for a string or blob to be generated).


slimhazard:
@bsdphk suggested naming this VMOD just "blob" (on IRC and in varnish-dev). The name can be easily changed. I prefer "blobcode", first of all because the VMOD is not "all about BLOBs", but strictly about binary-to-text encodings. Also, we already have a VMOD named "blob" in productive use, and it would be a bit of a PITA to have to go change the name where it's used:

https://code.uplex.de/uplex-varnish/libvmod-blob

VMOD blob provides these functions:
```
BOOL blob.same(BLOB, BLOB)
BOOL blob.equal(BLOB, BLOB)
INT blob.length(BLOB)
INT blob.integer(BLOB [, ENUM padding])
BLOB blob.subblob(BLOB, BYTES length [, BYTES offset])
```
With fairly obvious meanings, I think, except possibly for ``blob.integer()``, which returns the first bytes of the BLOB as an INT.

What we could do is add these functions to the present VMOD as well. Then it really is "all about BLOBs", and IMO "blob" becomes a better name.

bsdphk:
If we add a BLOB related vmod to the project, it should be the (not too big) kitchen-sink for all stuff related to BLOBs.

For of the five functions you propose to add looks OK to me, but blob.integer() sounds too specific for my taste.  Maybe generalise it something like:

INT blob.integer(BLOB, INT offset=0, ENUM {vbe8,vbe16,vbe32, vbe64,vle8,vle16,vle32,vle64}, ...)

(Of course that forces the issue of how big an INT actually is...)

slimhazard:
@bsdphk what about this:
```
INT blob.integer(BLOB, BYTES offset=0, ENUM {BIG, LITTLE} endian=0, BYTES bytes=0)
```
If ``endian`` is NULL, return whatever INT results from the endianness of the platform, otherwise endianness as specified.

If ``bytes`` is 0, return whatever you get from ``sizeof(INT)``, otherwise fill the INT with the number of bytes specified, but no more than ``sizeof(INT)``.

We can bikeshed about whether to silently ignore ``bytes > sizeof(INT)``, or invoke ``VRT_fail()`` in that case.

Dridi:
> If `endian` is NULL, return whatever INT results from the endianness of the platform, otherwise endianness as specified.

What about `ENUM {HOST, BIG, LITTLE} endian = "HOST"` ?

slimhazard:
@Dridi good idea, yes.

slimhazard:
I'll squash all of the commits up to this point, so that the VMOD will only ever have been named 'blob'.

And will now start adding the other four functions.

slimhazard:
Added the four functions discussed above, will squash the earlier commits this afternoon.

bsdphk:
I tried this on FreeBSD-current and got::

      CC       libvmod_blob_la-vmod_blob.lo
    vmod_blob.c:276:22: error: cast from 'const char *' to 'void *' drops const qualifier [-Werror,-Wcast-qual]
                                                            enc, (void *) s, len,
                                                                          ^
    vmod_blob.c:281:21: error: cast from 'const char *' to 'void *' drops const qualifier [-Werror,-Wcast-qual]
                                                    free((void *) s);
                                                                  ^
    vmod_blob.c:285:17: error: cast from 'const char *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
                                                    *((char *)s + len) = '\0';
                                                              ^
    vmod_blob.c:311:18: error: cast from 'const char *' to 'void *' drops const qualifier [-Werror,-Wcast-qual]
                            free((void *) b->encoding[i]);
                                          ^
    vmod_blob.c:363:21: error: cast from 'const char *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
                    b->priv = (char *)p;
                                      ^
    vmod_blob.c:374:20: error: cast from 'const char *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
            b->priv = (char *)p;
                              ^
    6 errors generated.
      CC       libvmod_blob_la-base64.lo
    base64.c:89:33: error: cast from 'const char *' to 'unsigned char *' drops const qualifier [-Werror,-Wcast-qual]
            const uint8_t *in = (uint8_t *)inbuf;

slimhazard:
Squashes and fixes are in.

nigoroll:
bugwash: rebase, squash into fewer commits, merge

Issue #1799, opened 2016-03-04
Backend request coalescing fail / vcl_hit{} returns miss without busy object.  Doing pass.


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1799

nigoroll:
Here's another workaround:
* mark `return(miss)` cases in `vcl_hit` with a req header
* in `vcl_pass`, check for this header and, if present, `return(restart)`
* in `vcl_miss`, clear this header
* in `vcl_recv`, check for restart and the presence of the marker header and and set `hash_always_miss`
```
sub vcl_hit {
    // ...
    set req.http.workaround-1799 = "fixup";
    return (miss);
}
sub vcl_pass {
    if (req.http.workaround-1799 && req.restarts < 2) {
        return (restart);
    }
}
sub vcl_miss {
    unset req.http.workaround-1799;
}
sub vcl_recv {
    if (req.restarts >= 1 && req.http.workaround-1799) {
        set req.hash_always_miss = true;
    }
}
```

nigoroll:
saving irc discussion for easy reference

> (13:46:48) martin: was this alleviated by the hfp->hfm change?
> (13:47:01) slink: So I was thinking about a couple of options with this in mind and the hit-for-pass vs. hit-for-miss: with the current hit-for-miss, we insert new objects all the time, which might also turn out to be problematic for some cases. So the one most drastic question is: should we have a late cache insert after fetch?
> (13:47:56) slink: martin its just something i thought about in the new context
> (13:48:23) slink: and I dont think 1799 has got any better
> (13:48:43) slink: miss from hit is still a pass
> (13:48:51) martin: yeah, i think you're right. that one is still open
> (13:49:13) martin: a late cache insert would make the request coalescing fail wouldn't it?
> (13:49:37) slink: yes, but only for the case where you have an object and decide to miss anyway
> (13:51:29) slink: martin: hm but yes, I wonder of there is a better option...
> (13:52:04) martin: this is tricky stuff, really touches on several areas that need to be just right for things work out
> (13:52:15) slink: martin: absolutely
> (13:53:38) slink: But I think there are some basic questions: for coalescing we need a busy object. So we have the options of inserting one which we may not need or inserting it later and accept the risk that we may miss out for some time
> (13:53:56) slink: "always" inserting one was an option phk really had concerns about
> (13:54:09) slink: but basically hfm is not that far away from it
[...]

PR #2422, opened 2017-09-11
Ignore req.ttl when keeping track of epired objects

Dridi:
The goal of `req.ttl` is to allow lower requirements for known
transactions (since this has to be done in VCL) but objects with
a TTL lower than `req.ttl` would always be considered expired
during lookup even if their actual TTL (obj.ttl) is still positive.

Ignoring `req.ttl` gives a better control over explicit refreshes
made to optimize latency and greatly reduces the risk of running
into #1799 (for a subset of use cases).


Dridi:
Pushed with a slightly improved commit message in aa4298c.

Dridi:
Ported to 4.1 in ad92c56.

PR #2425, opened 2017-09-14
VMOD blob: use a DEFAULT value for the case ENUM.

slimhazard:
DEFAULT is interpreted as LOWER for the HEX and URL encodings, and is the only legal value for all other encodings.

This responds to objections from @nigoroll and @Dridi to the API changes implemented as a result of the bugwash on Sep 11.

One of those changes was to remove the *UC and *LC encodings (HEXUC, HEXLC, URLUC and URLLC), and instead use a ``case`` ENUM in the encode method and functions with values ``LOWER`` and ``UPPER``, default ``LOWER``. That way, the enum values ``HEX`` and ``URL`` can be used for both encodings and decodings.

The ``case`` ENUM is not relevant for the ``BASE64*`` and ``IDENTITY`` encodings, and is ignored for those. In fact it may be misleading. Setting the case to ``UPPER`` with ``IDENTITY`` might appear as if the string encoding is upper-cased, but it is just ignored. And of course changing the case of the ``BASE64*`` encodings would be just wrong.

The documentation in the master version points all of this out, and advises that you should use ``std.toupper`` and ``std.tolower`` to force case. But this just became clear to @nigoroll and @Dridi today, and they nevertheless object.

One of the reasons: with named arguments, you can leave the ``case`` ENUM out, and the result looks reasonable enough:
```
blob.encode(BASE64, blob=req.hash)
```
But with positional arguments, one of the ``case`` values must appear in the call, making it look as if the result might be upper- or lower-cased:
```
blob.encode(BASE64, LOWER, req.hash)
```
So, this PR changes the VMOD so that:
* ``case`` always defaults to ``DEFAULT``.
* ``DEFAULT`` is interpreted as ``LOWER`` for the ``HEX`` and ``URL`` encodings.
* ``DEFAULT`` is the only legal value for the other encodings (``IDENTITY`` and ``BASE64*``). If any other value is given for ``case`` with those encodings, then VCL failure is invoked.

Then the invocation with positional args becomes:
```
blob.encode(BASE64, DEFAULT, req.hash)
```
And no other value in the second position is allowed.

### DIFF
--- a/bin/varnishtest/tests/m00033.vtc
+++ b/bin/varnishtest/tests/m00033.vtc
@@ -54,13 +54,9 @@ varnish v1 -vcl {
 	      blob.encode(IDENTITY,
 			  blob=blob.decode(HEX, encoded="666f6f00626172"));
 
-	    set resp.http.lc =
-	      blob.encode(IDENTITY, LOWER, blob.decode(IDENTITY,
-						       encoded="Don't care"));
-
-	    set resp.http.uc =
-	      blob.encode(IDENTITY, UPPER, blob.decode(IDENTITY,
-						       encoded="Don't care"));
+	    set resp.http.pos =
+	      blob.encode(IDENTITY, DEFAULT, blob.decode(IDENTITY,
+						         encoded="foobar"));
 	}
 } -start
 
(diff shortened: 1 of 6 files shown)


slimhazard:
Merged into master as d1799b8.

PR #2439, opened 2017-09-25
Expose the varnishd optstring via -x

Dridi:
This helps write scripts like this:

    optstring=$(varnishd -x optstring)

    while getopts $optstring opt
    do
        case $opt in
        n)
            # handle $OPTARG
            ;;
        # handle other options
        *)
            # ignore unneeded options
            ;;
        esac
    done

    varnishd "$@"

Otherwise if optstring is not kept in sync, getopts will stop processing
options if it encounters one that is not specified.

### DIFF
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -74,6 +74,8 @@ static struct vpf_fh *pfh2 = NULL;
 static struct vfil_path *vcl_path = NULL;
 static VTAILQ_HEAD(,f_arg) f_args = VTAILQ_HEAD_INITIALIZER(f_args);
 
+static const char opt_spec[] = "a:b:Cdf:Fh:i:I:j:l:M:n:P:p:r:S:s:T:t:VW:x:";
+
 int optreset;	// Some has it, some doesn't.  Cheaper than auto*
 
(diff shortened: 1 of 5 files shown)


bsdphk:
As mentioned in bugwash:  I think all our tools in the tree should be able to output the optstring, and I suggest we find a common flag for it.

One possibility is to put the optstring in the first line of the usage() output and make sure -? is implemented in all programs.

Dridi:
Anothernother solution would be to support an `--optstring` long option. VUTs already support long options so it wouldn't be hard to add one, and `varnishd` could either use the same option or stick with `-x optstring`.

Which would you prefer?

bsdphk:
I can live with varnishd being the odd man out.

Dridi:
I just added `--optstring` support to VUTs. That leaves `varnishadm` and `varnishtest` out, but this was initially meant for `varnishd` to intercept interesting options before it would daemonize, and all the daemons are covered now.

bsdphk:
You should put the example on top of this page in the docs somewhere...

Dridi:
Added documentation somewhere*s*, ready for review.

Dridi:
Ported to 4.1 as ff5085527 and 67d56d172, skipping the VUT part since they don't support long options on that branch.

PR #2442, opened 2017-09-28
H2: Respect flow control limits when sending

daghf:
This PR addresses one of the key H/2 spec violations in the current implementation by implementing flow control adherence for sending bits to clients.

This is achieved by making the streams wait in `H2_Send` when they are out of window credits. To allow for transmission of stream0 control messages and a to have level of concurrency (not block other streams on the same connection that still have credits available), we drop sender privileges before waiting. To accomplish this I've introduced "AssertHeld" versions of `H2_Send_Get`/`H2_Send_Rel` that permits calling those functions while already holding the lock.

I was not able to come up with a way to make a VTC that verifies that the flow control limits are respected (input welcome if anyone have any ideas on that). I have however subjected it to load testing, and in my testing I haven't seen it crash.

It's currently missing timeout handling, but I was hoping to get some feedback on whether this is headed in the right direction before continuing.

### DIFF
--- a/bin/varnishd/http2/cache_http2.h
+++ b/bin/varnishd/http2/cache_http2.h
@@ -144,9 +144,6 @@ struct h2_sess {
 	struct h2_req			*mailcall;
 	pthread_cond_t			*cond;
 
-	int64_t				r_window;
-	int64_t				t_window;
-
 	struct sess			*sess;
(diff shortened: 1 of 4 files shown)


daghf:
Merged

Issue #2387, opened 2017-08-15
Assert error in WS_ReleaseP


daghf:
Spotted via http2fuzz,

```
Error: Child (7585) Panic at: Tue, 15 Aug 2017 09:36:08 GMT
Assert error in WS_ReleaseP(), cache/cache_ws.c line 283:
  Condition(ptr >= ws->f) not true.
version = varnish-trunk revision c41ab0150, vrt api = 6.1
ident = Linux,4.9.0-3-amd64,x86_64,-jnone,-smalloc,-smalloc,-hcritbit,epoll
now = 304082.847244 (mono), 1502789767.932500 (real)
Backtrace:
  0x454250: bin/varnishd/varnishd() [0x454250]
  0x45404b: bin/varnishd/varnishd(pan_ic + 811) [0x45404b]
  0x4856f3: bin/varnishd/varnishd(WS_ReleaseP+0xf3) [0x4856f3]
  0x49b687: bin/varnishd/varnishd(h2h_decode_fini+0x207) [0x49b687]
  0x49d607: bin/varnishd/varnishd(h2_kill_req+0x1d7) [0x49d607]
  0x4a1a75: bin/varnishd/varnishd(h2_new_session + 1765) [0x4a1a75]
  0x483f7c: bin/varnishd/varnishd(Pool_Wrk_Thread + 1612) [0x483f7c]
  0x4837f1: bin/varnishd/varnishd(WRK_Thread + 833) [0x4837f1]
  0x48347a: bin/varnishd/varnishd() [0x48347a]
  0x7f9b33b34494: /lib/x86_64-linux-gnu/libpthread.so.0(+0x7494) [0x7f9b33b34494]
thread = (cache-worker)
thr.req = 0x7f9b26582020 {
  vxid = 561644, transport = H2
  step = 0x0,
  req_body = R_BODY_INIT,
  err_code = 1, err_reason = (null),
  restarts = 0, esi_level = 0,
  sp = 0x7f9b25417e20 {
    fd = 29, vxid = 561644,
    t_open = 1502789767.693881,
    t_idle = 1502789767.812064,
    transport = H2 {
      streams {
        0x00000000 idle  
        0x1ad3c95b open  
        0x533606d3 open  
      }
    }
    client = ::1 45108,
    privs = 0x7f9b25417e88 {
    },
  },
  ws = 0x7f9b26582208 {
    id = \"req\",
[...]

PR #2443, opened 2017-09-29
Enforce sequence requirement of header block frames

daghf:
This commit enforces that a HEADERS frame without an END_HEADERS flag
must be followed by a CONTINUATION frame on the same stream.

Sending interleaved/out of sequence header or continuation frames will
now be treated as a PROTOCOL_ERROR, as per rfc7540.

This also fixes the panics seen in #2387.

Fixes: #2387


daghf:
Merged in 053c93cd12f79a00d9801a5d0f9e67ea3aa2fcbd

PR #2446, opened 2017-10-02
Do not possibly underflow rlen

### LINKED ISSUE #2444: vbp_poke rlen handling
in the read loop, we overwrite our buffer with every read after we filled it once. While I see a case for drain-reading probe responses, such a probe request will almost inevitably fail because the status line is long overwritten.
Also, we probably shouldn't have rlen += i because this will overflow rlen for i <0, just to be on the safe side (rlen is not used outside the loop, so we're safe at the moment, but we might want to avoid future confusion)

nigoroll:
for i < 0, rlen could underflow. We are safe because of the check for i < 0 further down, so this change is just a minor cleanup.
    
Fixes #2444

### DIFF
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -340,8 +340,10 @@ vbp_poke(struct vbp_target *vt)
 			    sizeof vt->resp_buf - rlen);
 		else
 			i = read(s, buf, sizeof buf);
+		if (i <= 0)
+			break;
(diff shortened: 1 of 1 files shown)


bsdphk:
Turn it into a regular while (1) { ...} and then OK

nigoroll:
merged with bedaceb355a8522f71e3b1b4b963f59e239a9b1f

hermunn:
Backported to 4.1 as 50bfa24ea117ec23.

Issue #2405, opened 2017-08-29
keep req.backend_hint across restarts?


nigoroll:
We clear req.backend_hint when restarting https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_req_fsm.c#L727
This complicates the #1799 workaround (the workaround vcl as given only works with `bereq.backend` being set on the backend side) and may have other drawbacks.
On the other hand, this would be a breaking change, so if we decide to make the change, we'd want it in a major release.

nigoroll:
Sorry for the wrong reference in the commit message.

nigoroll:
bugwash notes:

The proposal is that restart and retry should keep all of the `req`/`bereq` state except for `req.restarts` and `bereq.retries`.

Please review this proposal before next bugwash.

side issue `std.rollback()` (_not_ the scope of this ticket):
* documentation needs to be improved
* rollback does not play with vmods keeping references to objects on the workspace
* known, but not yet resolved - probably to be solved with the vcl variables/objects/scope rework

Dridi:
I'd like to see a `VRT_Reset` with `VRT_Rollback`'s current behavior and have the latter only roll back `req`/`bereq`, leaving the workspace alone. Or even a new VRT symbol to avoid confusion, eg. `VRT_RollbackHttp`.

nigoroll:
bugwash conclusion: Yes, implement restart and retry should keep all of the `req`/`bereq` state except for `req.restarts`, `bereq.retries` and `*.xid`.
More work on rollback/reset to follow later, maybe by @Dridi @slimhazard or myself

nigoroll:
#2447 addresses `restart` only, not yet `retry`

PR #2447, opened 2017-10-02
Restarts cleanup & consistent vcl_recv state with vcl switching

nigoroll:
* first patch is just to get access to all req properties in deliver - I don't see how this could cause any harm (famous last words(tm))
* fix the client side of #2405 (keep req attributes across restarts)
* side issue: notices that vcl switching will call the label vcl's vcl_recv with the other init from cnt_recv missing, so moved that to a common cnt_recv_prep


nigoroll:
updated to current master

nigoroll:
bugwash result: OK, needs release note (anyway)

nigoroll:
Merged

PR #2450, opened 2017-10-06
Pass delivery abandoned does not qualify as an error

nigoroll:
... so log it under the Debug tag.

FetchErrors should be actual errors which can be addressed. In this case,
nothing is wrong in any way, the fact that we abort a fetch if we don't
need the body is a varnish internal optimization (which makes sense, but
comes at the cost of closing a connection).

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -451,8 +451,8 @@ vbf_stp_fetchbody(struct worker *wrk, struct busyobj *bo)
 			 * objects to be created.
 			 */
 			AN(vfc->oc->flags & OC_F_PASS);
-			VSLb(wrk->vsl, SLT_FetchError,
-			    "Pass delivery abandoned");
+			VSLb(wrk->vsl, SLT_Debug,
+			    "Fetch: Pass delivery abandoned");
 			bo->htc->doclose = SC_RX_BODY;
 			break;
 		}


nigoroll:
force-pushed rebase

nigoroll:
rebased

nigoroll:
rebased

bsdphk:
Strictly speaking no VSL is necessary at all, because there is already a ``VCL_return      b abandon`` VSL which says the exact same thing.

I think the thinking behind FetchError was that all fetches which do not complete should get one of those.

We should probably poll the bugwash...

bmancone:
It looks like this issue was fixed in 4.1.10, but not in the 5.x branch. Then, in the 6.x branch, the issue is fixed again. @nigoroll, am I missing something here ?

- https://github.com/varnishcache/varnish-cache/blob/varnish-4.1.9/bin/varnishd/cache/cache_fetch.c#L494 - 4.1.9 : **SLT_FetchError**

- https://github.com/varnishcache/varnish-cache/blob/varnish-4.1.10/bin/varnishd/cache/cache_fetch.c#L498 - 4.1.10 : **SLT_Debug**

- https://github.com/varnishcache/varnish-cache/blob/varnish-5.0.0/bin/varnishd/cache/cache_fetch.c#L502 - 5.0.0 : **SLT_FetchError**

- https://github.com/varnishcache/varnish-cache/blob/varnish-6.0.0/bin/varnishd/cache/cache_fetch.c#L456 - 6.0.0 : **SLT_Debug**

nigoroll:
this patch came after 5.1.3, and I am not sure about the support status of 5.1 at the time.
And, at any rate, I have never been the maintainer of any LTS version.
@bmancone what specifically are you asking here?

bmancone:
@nigoroll I wanted to understand if the errors I saw after upgrading from 4.1.11 to 5.1.3 were actual errors. I did not have in mind that this patch came after 5.1.3, everything is clear now. Thank you for your quick answer !

nigoroll:
all right. I would recommend to move on to 6.0.4 or 6.2.1

bmancone:
@nigoroll Thank you !

PR #2455, opened 2017-10-11
Add cache_grace_hit counter

hermunn:
The new `cache_grace_hit` counter keeps track of the number of grace hits. To be precise, it counts the number of times lookup returns an expired object, `vcl_hit` is called and decides to `return(deliver)`.

The test case has gotten a temporary name (bin/varnishtest/tests/b00099.vtc) to avoid potential crashes with other incoming patches. I will rename this before pushing if I get an OK.

Another note: Maybe lines like `varnish v1 -expect cache_grace_hit == 1` should be added to other test cases instead of adding this test case. However, to keep the discussion on track, I include the test case to demonstrate the functionality. I hope that we first can discuss the code (is this the correct interpretation of a *grace hit*?), and then try to reduce the number of test cases without testing less.

### DIFF
--- a/bin/varnishd/VSC_main.vsc
+++ b/bin/varnishd/VSC_main.vsc
@@ -56,6 +56,13 @@
 	Count of cache hits.  A cache hit indicates that an object has been
 	delivered to a client without fetching it from a backend server.
 
+.. varnish_vsc:: cache_hit_grace
+	:oneliner:	Cache grace hits
+
+	Count of cache hits with grace. A cache hit with grace is a cache
+	hit where the object is expired. Note that such hits are also
+	included in the cache_hit counter.
+
 .. varnish_vsc:: cache_hitpass
 	:oneliner:	Cache hits for pass.
 
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -523,6 +523,8 @@ cnt_lookup(struct worker *wrk, struct req *req)
 		}
 		wrk->stats->cache_hit++;
 		req->is_hit = 1;
+		if (lr == HSH_EXP || lr == HSH_EXPBUSY)
+			wrk->stats->cache_hit_grace++;
 		req->req_step = R_STP_DELIVER;
 		return (REQ_FSM_MORE);
 	case VCL_RET_MISS:
--- a/bin/varnishtest/tests/b00099.vtc
+++ b/bin/varnishtest/tests/b00099.vtc
@@ -1,0 +1,52 @@
+varnishtest "The grace_hit_counter"
+
+server s1 {
+	# normal fetch
+	rxreq
+	expect req.url == "/1"
+	txresp -hdr "Age: 1" -hdr "Cache-Control: max-age=2" -body "1"
+
+	# background fetch:
+	rxreq
+	expect req.url == "/1"
+	txresp -body "2"
+
+	# normal fetch
+	rxreq
+	expect req.url == "/2"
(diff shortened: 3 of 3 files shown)


hermunn:
During bugwash there seemed to be no objections to the code, but the name was not immediately accepted. Suggestions included `cache_graced_hit`, `cache_hitgrace`, but there might be better options.

It was also noted that we probably want more counters that are similar to this one. Feel free to specify here if you have any wishes (with names!), and I might create a new PR with these. I am mixing this in so that it will be easier to find a good naming convension for all the potential new counters.

nigoroll:
I'm fine with this change, but regarding "any wished" I'd point to dynamic (vcc defined through vmods) counters

gquintard:
there's already hitmiss, hitpass, so I'd go for hitgraced to keep things more or less consistent.

hermunn:
Naming is hard. We probably want a counter for the message *vcl_hit{} returns miss without busy object. Doing pass.* as well, but it will not be called `hitmiss`. Lets call it `hit_became_miss` for now.

Note that `hit_became_miss` and `cache_grace_hit` (proposed in this PR) count the number of times `vcl_hit` was called *and* something specific happened (the outcome of the function was `return(miss)` or `return(fetch)` with grace).

On the other hand, `hitmiss` and `hitpass` counts things that happen earlier, and both prevent `vcl_hit` being called.

Maybe this is a reason not to change the name from `cache_grace_hit` to `hitgrace`?

rezan:
Can you add a counter for ESI requests (ved_include)?

hermunn:
@rezan, good idea.

However, first I need an OK on this. I know people care a lot about naming, and this is why I have not merged it yet.

I will create pull request on other counters requested by team members at Varnish Software, but I do not want to do those in the same PR. (I am trying to add some pressure here.)

My suggestion is:

* `cache_graced_hit`: Number of cache hits that served graced (expired) objects.
* `hit_became_miss_pass` (or some other name, name to be discussed in the *next* pull request): Number of times `vcl_hit` became a `return(miss)` and we had to do a `pass` because of #1799.

fgsch:
My vote is for `cache_hit_grace`, with or without the last underscore.

hermunn:
I am OK with `cache_hit_grace`.

hermunn:
Bugwash conclusion: `cache_hit_grace` got the popular vote, but I need to get the description approved by @Dridi before pushing. I am a bit busy right now, but will got this done *soon*.

hermunn:
@Dridi Rebased and updated the documentation. As you can see, the text on `cache_hit` is not 100% accurate, since it does not take into account the case where `vcl_delvier` stops the delivery to the client. I would like to merge this and let you have a go at improving the text a bit.

Dridi:
LGTM

hermunn:
Then I will change the name of the test case and push.

hermunn:
This was pushed as 1d62f5da43a245.

llach-gog:
It's closed already, but I'm just wondering - if we will have such counter for "grace" shouldn't we also have it for "keep"? Like "cache_hit_keep"?

hermunn:
@llach-gog Yeah, there probably should be more counters in the area, but I do not (yet) have a complete list. This history for this counter, `cache_hit_grace`, is short and simple:  I got a request from a colleague.

If there are specific counters that you need, list them here, and with some luck I (or someone else) will find the time to implement them before the next major release.

hugocruz:
@hermunn good stuff !

hermunn:
Back ported as a33bf527ee50f46.

PR #2456, opened 2017-10-11
Add unit of "Varnish:time_firstbyte" in varnishncsa man page

jgarrouste:
Simple change to add the unit (seconds) of "Varnish:time_firstbyte" in varnishncsa documentation


fgsch:
Thanks for the PR. 
I've fixed it in a slightly different way following similar constructions in the rest of that document.

Issue #2268, opened 2017-03-20
Assert error in VRTPRIV_dynamic_kill(), cache/cache_vrt_priv.c line 111:


xcir:
```
Panic at: Mon, 20 Mar 2017 08:39:07 GMT
Assert error in VRTPRIV_dynamic_kill(), cache/cache_vrt_priv.c line 111:
  Condition((vp)->magic == 0x24157a52) not true.
version = varnish-5.1.1 revision de38712, vrt api = 6.0
ident = Linux,4.4.0-24-generic,x86_64,-junix,-smalloc,-sfile,-sfile,-smalloc,-hcritbit,epoll
now = 1956981.218037 (mono), 1489999146.497627 (real)
Backtrace:
  0x438b37: /usr/sbin/varnishd() [0x438b37]
  0x44b290: /usr/sbin/varnishd(VRTPRIV_dynamic_kill+0x160) [0x44b290]
  0x43ae9e: /usr/sbin/varnishd(Req_Cleanup+0xce) [0x43ae9e]
  0x461100: /usr/sbin/varnishd(h2_del_req+0xb0) [0x461100]
  0x461289: /usr/sbin/varnishd(h2_do_req+0x79) [0x461289]
  0x451b69: /usr/sbin/varnishd() [0x451b69]
  0x451feb: /usr/sbin/varnishd() [0x451feb]
  0x7fad689e06ba: /lib/x86_64-linux-gnu/libpthread.so.0(+0x76ba) [0x7fad689e06ba]
  0x7fad6871682d: /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7fad6871682d]
thread = (cache-worker)
thr.req = 0x7fa3549c0020 {
  vxid = 3899727, transport = H2
  step = R_STP_TRANSMIT,
  req_body = R_BODY_NONE,
  restarts = 0, esi_level = 0,
  sp = 0x7fad5c679c20 {
    fd = 218, vxid = 3899722,
    t_open = 1489999146.309901,
    t_idle = 1489999146.497617,
    transport = H2 {
      streams {
        0x00000000 idle
      }
    }
    client = *** ***,
  },
  ws = 0x7fa3549c0208 {
    id = \"req\",
    {s, f, r, e} = {0x7fa3549c2008, +1816, (nil), +122864},
  },
  http_conn = 0x7fa3549c0130 {
    doclose = NULL,
    ws = (nil) {
    },
[...]

PR #2457, opened 2017-10-12
Add locking around client-side PRIV handling

daghf:
With H/2 we now have multiple requests in flight concurrently for the same session, so we need to synchronize access to the `sp->privs` list.

Another option to this patch would be to move the `PRIV_{TASK,TOP}` priv pointers to `struct req`, which would avoid the locking introduced here.

Fixes: #2268

### DIFF
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -202,7 +202,9 @@ ved_include(struct req *preq, const char *src, const char *host,
 		AZ(req->wrk);
 	}
 
+	Lck_Lock(&sp->mtx);
 	VRTPRIV_dynamic_kill(sp->privs, (uintptr_t)req);
+	Lck_Unlock(&sp->mtx);
 
 	AZ(preq->vcl);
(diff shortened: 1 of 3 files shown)


Dridi:
With the uncertain future of `PRIV_*` straightening the locking is probably good enough.

nigoroll:
my vote goes to moving & avoid locking. Doing so would also reduce the lookup time for privs also because less privs would sit on each list.
@Dridi why would the `PRIV_*` future be uncertain?

Dridi:
@nigoroll I was under the impression that trunk will move away from `PRIV_*` in favor of VMOD objects and considering that the 5.x series only has experimental h2 support it seemed enough. If that is not the case, then we should definitely avoid the locking altogether.

nigoroll:
@Dridi `PRIV_CALL` is probably going away, but my guess is that the VRT priv interface might be used even see more use with per-task objects.

daghf:
Please see #2458 for the alternative approach.

daghf:
Closing in favor of #2458

PR #2458, opened 2017-10-12
Move the req-specific PRIV pointers to struct req

daghf:
NB: This is an alternative solution to PR #2457 

This PR kills `sp->privs` and moves it over to `struct req`.

Keeping the privs list in struct sess would necessitate extra locking for h/2 where we have multiple reqs in flight concurrently.

Previously we used the `id` argument to tell apart pointers for different requests on the same session, this is still required for the very specific case where a VMOD would use both PRIV_TOP and PRIV_TASK.

Fixes: #2268

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -574,6 +574,8 @@ struct req {
 
 	/* Temporary accounting */
 	struct acct_req		acct;
+
+	struct vrt_privs	privs[1];
 };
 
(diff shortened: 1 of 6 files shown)


nigoroll:
the diff lgtm. Have not conducted other checks

Dridi:
Positive feedback from https://github.com/varnishcache/varnish-cache/issues/2268#issuecomment-338532833 too.

PR #2459, opened 2017-10-12
VIP12 vmod function/method polymorphism

nigoroll:
Ref: https://github.com/varnishcache/varnish-cache/wiki/VIP12:-vmod-polymorphism-(for-type-conversions)

We now allow definition of vmod function/methods by the same name but
different signatures.

VCC file:

For $Function/$Method names, we add a "poly suffix" {C_SUFFIX} which
is appended to the function names for the C-Interface, but not for the
VCL interface.

vmodtool:

The change to extract the poly suffix is straight forward. The
RST generation is unchanged for now. All information is contained,
but we might want to improve the documentation format for polymorphic
functions later.

vcc:

We hang polymorphic variants of the same symbol name as a chain of
children under the first symbol found. While using just one layer
below a primary symbol would be a bit more efficient for symbol
insertion, this way we do not need to break the rule of symbol names
being unique at each level.

In vcc_Eval_SymFunc(), the first matching variant is now looked up by
successively trying vcc_func() until success or symbols exhausted.

This implementation differs from the VIP in some regards:

* "for polymorphic arguments, STRING always needs to be supported"

I cannot see any more why we would want to be that strict. For string
arguments, a polymorphic function (as any function/method) effectively
takes any argument type through string folding, and a vmod might want
to not allow that.

* "at vcl compile time, vcc finds the best match ** arbitration
   left->right best match (e.g. stringification of rightmost arguments
   preferred)"

Such heuristic appears pretty arbitrary. Instead, the current
implementation just uses the first match found, in the order of
definition in the vcc file. This should give vmod authors maximum
flexibility - including varying argument counts.

### DIFF
--- a/bin/varnishtest/tests/m00000.vtc
+++ b/bin/varnishtest/tests/m00000.vtc
@@ -33,6 +33,12 @@ varnish v1 -vcl+backend {
 		std.log("VCL" + " initiated " + "log");
 		std.syslog(8 + 7, "Somebody runs varnishtest");
 		debug.rot52(resp);
+		set resp.http.f_poly_int = debug.f_poly(42);
+		set resp.http.f_poly_real = debug.f_poly(30.12);
+		set resp.http.f_poly_string = debug.f_poly(33s + 3m);
+		set resp.http.m_poly_int = obj.m_poly(42);
+		set resp.http.m_poly_real = obj.m_poly(30.12);
+		set resp.http.m_poly_string = obj.m_poly(33s + 3m);
 	}
 } -start
 
@@ -62,6 +68,18 @@ logexpect l1 -v v1 -g raw -d 1 {
 	expect 0 =    RespHeader	{^not: -1}
 	expect 0 =    VCL_Log		{^VCL initiated log}
 	expect 0 =    RespHeader	{^Encrypted: ROT52}
+	expect 0 =    Debug		{^f_poly_int\(42\)}
+	expect 0 =    RespHeader	{^f_poly_int: 42$}
+	expect 0 =    Debug		{^f_poly_real\(30.120000\)}
+	expect 0 =    RespHeader	{^f_poly_real: 30$}
+	expect 0 =    Debug		{^f_poly_string\(213.000\)}
+	expect 0 =    RespHeader	{^f_poly_string: 213$}
+	expect 0 =    Debug		{^m_poly_int\(42\)}
+	expect 0 =    RespHeader	{^m_poly_int: 42$}
+	expect 0 =    Debug		{^m_poly_real\(30.120000\)}
+	expect 0 =    RespHeader	{^m_poly_real: 30$}
+	expect 0 =    Debug		{^m_poly_string\(213.000\)}
+	expect 0 =    RespHeader	{^m_poly_string: 213$}
 	expect 0 =    VCL_return	{^deliver}
 } -start
 
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -598,22 +598,37 @@ vcc_Eval_Func(struct vcc *tl, const char *spec,
  */
 
 void v_matchproto_(sym_expr_t)
-vcc_Eval_SymFunc(struct vcc *tl, struct expr **e, const struct symbol *sym,
+vcc_Eval_SymFunc(struct vcc *tl, struct expr **ee, const struct symbol *sym,
     vcc_type_t fmt)
 {
+	struct expr *e = NULL;
+	struct token *t1;
 
 	(void)fmt;
+
 	assert(sym->kind == SYM_FUNC);
 	AN(sym->eval_priv);
 
 	SkipToken(tl, ID);
+	t1 = tl->t;
 	assert(sym->fmt == VCC_Type(sym->eval_priv));
-	vcc_func(tl, e, sym->eval_priv, sym->extra, sym);
+	do {
+		if (e != NULL)
+			vcc_delete_expr(e);
+		e = NULL;
+		tl->t = t1;
+		tl->err = 0;
+		VSB_clear(tl->sb);
+		vcc_func(tl, &e, sym->eval_priv, sym->extra, sym);
+		if (! tl->err)
+			break;
+	} while ((sym = VTAILQ_FIRST(&sym->children)));
 	ERRCHK(tl);
-	if ((*e)->fmt == STRING) {
-		(*e)->fmt = STRINGS;
-		(*e)->nstr = 1;
+	if (e->fmt == STRING) {
+		e->fmt = STRINGS;
+		e->nstr = 1;
 	}
+	*ee = e;
 }
 
 /*--------------------------------------------------------------------
--- a/lib/libvcc/vcc_symb.c
+++ b/lib/libvcc/vcc_symb.c
@@ -153,6 +153,8 @@ VCC_Symbol(struct vcc *tl, struct symbol *parent,
 			continue;
 		break;
 	}
+	if (sym != NULL && create == 2 && q == e)
+		return (VCC_Symbol(tl, sym, b, e, kind, create));
 	if (sym == NULL && create == 0 && parent->wildcard != NULL) {
 		AN(parent->wildcard);
 		parent->wildcard(tl, parent, b, e);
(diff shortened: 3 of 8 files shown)


Dridi:
I feel this needs more testing because it opens interesting edge cases:

```
$Module foo
$Function INT poly{int}(INT val=0, REAL other)
$Function INT poly{real}(REAL val)
$Function INT poly{string}(STRING val)
```

Which do you pick for `foo.poly(1.0)`? What would pola say? I know very well this is not a great example but we should expect polymorphism for more complex signatures.

Example:

```
$Function PROBE probe{url}(STRING url, INT expected_response = 200, ...)
$Function PROBE probe{request}(STRING request, INT expected_response = 200, ...)
```

The only way to differentiate those would be to look at parameter names. That last one could be argued as pebkac on the vmod author's side, but arguably vmodtool could also reject two suffixes with the same signature.

Speaking of vmodtool, it doesn't seem to enforce that a function or method should either be unique, or all variants should have a prefix. Currently you can do this:

```
$Function INT func()
$Function INT func{poly}()
```

Dridi:
References in the generated RST don't include the polymorphic suffix.

nigoroll:
@Dridi:
* Yes, we need testing, and I hope that coming up with this early in the release cycle helps that.
* Yes, I've chosen the simple approach to declare precedence by order in the vcc file. At this point I think this is appropriate for our use case and as pola as can be. But, also yes, we leave it with the vmod author to come up with a sensible declaration. So while your example illustrates the basic problem with polymorphism, I do not think that trying to solve it in VCC is the right way.
* I would argue that your second example (`poly`) is just plain wrong. This is not a sensible use case for polymorphism, you'd probably want named arguments with null defaults, two functions or some switch argument.
* Yes, vmodtool.py currently lacks a lot of checks - this is not related to this PR, but existed before. Though we could make improvements at this point, I am not sure if it's worth it. Your C compiler will barf at you anyway.
* I'll fix the RST references, thank you

bin/varnishtest/tests/m00000.vtc
```diff
@@ -29,6 +29,12 @@ varnish v1 -vcl+backend {
 		std.log("VCL" + " initiated " + "log");
 		std.syslog(8 + 7, "Somebody runs varnishtest");
 		debug.rot52(resp);
+		set resp.http.f_poly_int = debug.f_poly(42);
+		set resp.http.f_poly_real = debug.f_poly(30.12);
+		set resp.http.f_poly_string = debug.f_poly(33s + 3m);
```
fgsch: This is duration

Dridi:
> - Yes, I've chosen the simple approach to declare precedence by order in the vcc file. At this point I think this is appropriate for our use case and as pola as can be. But, also yes, we leave it with the vmod author to come up with a sensible declaration. So while your example illustrates the basic problem with polymorphism, I do not think that trying to solve it in VCC is the right way.

I'm fine with such an approach, but this needs to be addressed (for example as documentation, or warnings).

> - I would argue that your second example (probe) is just plain wrong. This is not a sensible use case for polymorphism, you'd probably want named arguments with null defaults, two functions or some switch argument.

Is it though? I'm trying to think VCL here, not C.

Because we have named parameters, possibly with default values, it means that in VCL we can omit or reorder parameters as we please when names are provided. So I'd consider named parameters as part of the signature because of how we can write VCL.

- Yes, vmodtool.py currently lacks a lot of checks - this is not related to this PR, but existed before. Though we could make improvements at this point, I am not sure if it's worth it. Your C compiler will barf at you anyway.

Examples above compile fine, I tested them (not entirely as-is, but tested the underlying intent) and with the patch I can currently build and use a VMOD for all three cases.

nigoroll:
@Dridi 
* Document how this works: yes - will do after merge (if any)
* `probe` example: My point here is that the decision problem you get regarding the polymorphism stems from the fact that the declaration does not make sense in the first place. If I failed to get my point across, let's discuss it on IRC somewhen
* Ah yes, you are right, `func` / `func{poly}` will actually compile and it should even work as if you had a different _poly_ suffix for the first one too.

nigoroll:
quick notes on a 1:1 irc disucssion with @Dridi 
* we agree that this needs to be documented properly (yes, still)
* the "accept the first signature which vcc would accept otherwise" approach seems simple yet powerful
** for the probe example, the argument names would differentiate
* we both like the fact that also return values can vary

slimhazard:
Overloaded return types (when they're not distinguished by differing method signatures) can bring up some thorny issues, and hence are famously not possible in languages like Java and C++ (except for covariant return types in recent versions of Java). I don't know if VCL is subject to the same issues, or if @nigoroll and @Dridi have already discussed them, so I'd like to bring some of it out in the open.

For the examples assume:
```
INT foo()
REAL foo()
VOID bar(REAL)
```
What it boils down to is that, under certain conditions, it may be impossible to know which version to select.

* Calls in void context -- say we could do:
```
sub vcl_recv {
    # just want the side effects of the call, don't care about the return value
    foo();
}
```
Actually I believe that the last time I tried something like that, VCC didn't allow it (haven't tried it recently). Even if it isn't possible at present, I don't think it should foreclosed for the future. I for one have wanted to call functions in VCL that discard the return value, and it would be good to make that possible if it isn't already.

At any rate, the difficulty is obvious -- there's nothing about the calling context that determines which version of ``foo()`` to choose.

* Implicit type conversions (such as numeric promotion)

Again, I don't know off the top of my head if we have something like implicit INT-to-REAL conversion in VCL (I think that we do, but if we don't, again I would be against precluding the possibility). Assuming that we do, then we have an ambiguity in:
```
bar(foo())
```
Do we call the first version, take the INT result and convert it to REAL? Or call the second version that returns REAL?

Similar examples could be shown for the VCL types that have implicit STRING conversions.

The solution advocated above is: in an ambiguous context, always choose the first version declared. That would mean that in the potentially ambiguous examples, it would be impossible to choose the REAL version. This would be taken as a failing of the VMOD interface design.

Still, I think it raises the legit question as to whether this is really a good idea. To solve the ambiguities, the VMOD author would have to change the interface to something like:
```
INT foo()
REAL foo_real()
```
Now there's no ambiguity, but there's also no return type overloading, and we're shifting the name of the return type into the name of the function. So what was the point of having overloaded return types at all?

I'm not saying this is conclusively the wrong idea. The fact that the two (for better or worse) most widely known programming languages with method overloading have chosen *not* to allow overloaded return types in general makes me cautious. They may have avoided going down certain roads to madness, and maybe we should think carefully about heading that way ourselves.

nigoroll:
@slimhazard: If I understand correctly, the reason why the two languages you are referring to require such huge complication of polymorphism is that they implement a type inference system and thus actually need to make a decision.
VCCs type system is really simple (for built-in operators, first type determines overall type, just minimal coercion). I thus think that an equally simple resolution for polymorphic functions/methods (_first match_) is a good choice.
In other words: We don't need to solve a problem we have not created in the first place.

> The solution advocated above is: in an ambiguous context, always choose the first version declared.

This is exactly what I am proposing - except that we _always_ choose the first version matching - no matter how ambiguous the context or declaration may be.

I still fail to see the issue here: It is up to the vmod implementor to come up with a suitable interface and it is up to her to think about ambiguities.

Our documentation will contain general advise regarding the recommended priority of polymorphic variants with respect to the VCC coercion, such as "REAL first", "STRING last".

nigoroll:
and just a quick reminder: Polymorphism was suggested to avoid complicating the type system to begin with, see #2089

nigoroll:
force-push updated to master

nigoroll:
force-pushed to reflect recent intentation change

nigoroll:
FTR before resolving the conflicts of this patch due to the vcc changes introduced since I opened the PR, I would appreciate feedback from @bsdphk on concept & implementation.

bsdphk:
I am going to nix this implementation in light of subsequent events.

First, as to the original problem in #2089:

    set beresp.ttl = std.random(beresp.ttl, * 0.9, beresp.ttl);

Why not ?

    set beresp.ttl *= std.random(0.9, 1.0):

Second, we have implemented optional arguments to vmod functions.

I find `mod.something(dur=beresp.ttl)` vs. `mod.something(real=355./113.)` perfectly acceptable, even if it turns "you didn't pass any input" into a run-time error from the vmod.

Third, as @slimhazard points out, opening the party for variant return-types gives even CS types nightmares.

See https://github.com/varnishcache/varnish-cache/wiki/VIP12:-vmod-polymorphism-(for-type-conversions) for further progress.

PR #2462, opened 2017-10-13
Add bans on obj.ttl, obj.age, obj.grace and obj.keep

nigoroll:
Also add a duration argument type and the operators >, >=, < and <= for use with it (besides the existing == and !=).
    
obj.ttl and obj.age are compared relative to the time the ban was issued, as required by the fact that bans are evaluated some arbitrary time after they are created. For this reason, the ban_dups parameter has no effect on obj.ttl and obj.age bans, duplicates of bans using these fields are never removed.
    
obj.grace and obj.keep are compared as absolute values.

In an effort to support The Proprietary Stevedore (tm) [or any other storage engine supporting persistent bans - if any?], the existing ban serialisation format is preserved and upgrades _should_ work seamlessly. But downgrades won't.
    
Using the BANS_HAS_* macros, we try to make the relationship between the ban serialisation in cache_ban_build.c and access in cache_ban.c more obvious.
    
We now generate all fields and operators from table files as well as the definition of allowed operators for fields. The user help text about allowed operators is generated at init time.


nigoroll:
Last commit message should have read "thank you, travis", but no big deal anyway

nigoroll:
force-pushed a rebase to current master - only 8ef1c56 is relevant

nigoroll:
rebased

nigoroll:
rebased

nigoroll:
rebased

nigoroll:
FTR; discussed with @bsdphk and @daghf on irc:
* explained motivation to phk
* briefly discussed the option to have bans call special vcl subs for advanced logic

nigoroll:
FTR:

> briefly discussed the option to have bans call special vcl subs for advanced logic

reflecting on this again and again, I noticed that while this would be a good option to have, it won't be a replacement: Interpreted bans are an important ac hoc tool and are used by many applications.

So I continue to request this patch to be merged, as it complements a vital feature.

nigoroll:
FTR, I messed up (some of) the force-pushes between  4e911a6 and c4344ad yesterday, https://github.com/nigoroll/varnish-cache/commit/233520da75c1d577c531f91cfa37dba39042d17e is good again

nigoroll:
FTW I still owe a vtc

nigoroll:
I have polished the patch and force-pushed:
* better vtc coverage added
* fixed a bug in `ban_evaluate` for `BANS_OPER_EQ` with durations
worked remaining TODOs:
* there is now a flag preventing de-duplication of `obj.ttl` and `obj.age` bans: Because their reference point is the ban time, they should not be de-duplicated, unless we want to bloat the dedup code with boolean and temporal logic (iow "does obj.ttl > 1h at time x include obj.ttl > 2m at time y ?")
* duration string rendering added

bsdphk:
This is probably the hairiest macro-magic we have yet, and we should probably not try to beat that record again :-)

OK

Issue #2305, opened 2017-04-08
Assert error in h2_do_req(), http2/cache_http2_proto.c line 493


simonvik:
```
Panic at: Sat, 08 Apr 2017 16:06:41 GMT
Assert error in h2_do_req(), http2/cache_http2_proto.c line 493:
  Condition(req->req_body_status == REQ_BODY_NONE) not true.
version = varnish-5.1.1 revision de38712, vrt api = 6.0
ident = Linux,4.4.30-20161130-1423-0a2a05d,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
now = 1648807.363910 (mono), 1491667599.027457 (real)
Backtrace:
  0x438c25: /usr/sbin/varnishd() [0x438c25]
  0x460839: /usr/sbin/varnishd() [0x460839]
  0x4520a2: /usr/sbin/varnishd() [0x4520a2]
  0x45254b: /usr/sbin/varnishd() [0x45254b]
  0x7fc79bf006ba: /lib/x86_64-linux-gnu/libpthread.so.0(+0x76ba) [0x7fc79bf006ba]
  0x7fc79bc3682d: /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7fc79bc3682d]
thread = (cache-worker)
thr.req = 0x7fc204c06020 {
  vxid = 105614567, transport = H2
  step = R_STP_LOOKUP,
  req_body = R_BODY_WITH_LEN,
  restarts = 0, esi_level = 0,
  sp = 0x7fc4dfc1ba20 {
    fd = 323, vxid = 105614565,
    t_open = 1491667598.302321,
    t_idle = 1491667598.347864,
    transport = H2 {
      streams {
        0x00000000 idle  
        0x00000001 c-rem 
        0x00000003 c-rem 
        0x00000005 c-rem 
        0x00000007 c-rem 
      }
    }
    client = to: ********* 53192,
    privs = 0x7fc4dfc1ba88 {
    },
  },
  ws = 0x7fc204c06208 {
    id = \"req\",
    {s, f, r, e} = {0x7fc204c08008, +1616, +122864, +122864},
  },
  http_conn = 0x7fc204c06130 {
    doclose = NULL,
    ws = (nil) {
    },
    {rxbuf_b, rxbuf_e} = {(nil), (nil)},
    {pipeline_b, pipeline_e} = {(nil), (nil)},
    content_length = 0,
    body_status = none,
[...]

PR #2466, opened 2017-10-19
Don't panic if we reembark with a request body

daghf:
This moves the body status decision to h2_end_headers, to make sure that
it is only done once per request.

We also move the Cookie header concatenation to h2_end_headers, to avoid
stepping on any changes that may have taken place in VCL.

Fixes: #2305

### DIFF
--- a/bin/varnishd/http2/cache_http2_proto.c
+++ b/bin/varnishd/http2/cache_http2_proto.c
@@ -480,25 +480,10 @@ h2_do_req(struct worker *wrk, void *priv)
 {
 	struct req *req;
 	struct h2_req *r2;
-	const char *b;
 
 	CAST_OBJ_NOTNULL(req, priv, REQ_MAGIC);
 	CAST_OBJ_NOTNULL(r2, req->transport_priv, H2_REQ_MAGIC);
 	THR_SetRequest(req);
-
-	// XXX: Smarter to do this already at HPACK time into tail end of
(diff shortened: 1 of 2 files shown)


daghf:
Added another commit to address the second issue brought up in #2305.

If the END_STREAM flag is set  we will with this treat it as a PROTOCOL_ERROR if a content-length header is spotted.

bsdphk:
Re second commit:  Should we not make an exception for C-L: 0 ?

daghf:
@bsdphk agree - I'll update with a fix

daghf:
@bsdphk Updated now - I changed the `http_GetHdr(..)` into an `http_GetContentLength(..)`:

```
--- a/bin/varnishd/http2/cache_http2_proto.c
+++ b/bin/varnishd/http2/cache_http2_proto.c
@@ -531,7 +531,7 @@ h2_end_headers(struct worker *wrk, struct h2_sess *h2,
                        req->req_body_status = REQ_BODY_WITH_LEN;
        } else {
                assert (req->req_body_status == REQ_BODY_NONE);
-               if (http_GetHdr(req->http, H_Content_Length, &b))
+               if (http_GetContentLength(req->http) > 0)
                        return (H2CE_PROTOCOL_ERROR); //rfc7540,l,1838,1840
        }
 ```

bsdphk:
OK

PR #2474, opened 2017-10-30
Get rid of m4/ax_with_curses.m4

gquintard:
Tackle the same issue as #2331, hopefully in a cleaner way.


Dridi:
Reviewed, tweaked and tested successfully on my system. Does not approve the current indentation, otherwise in favor if it passes on other platforms we support ;)

gquintard:
won't work on systems lacking proper pc support

PR #2475, opened 2017-10-31
Use pkg-config for curses, if possible

gquintard:
I hoped to get rid of m4/ax_with_curses.m4, but it needs to stays for platforms not shipping .pc files.
For for those that do however, it solves the issue of split libtinfo.


bsdphk:
No objection.

nigoroll:
@gquintard approved again during bugwash

PR #2476, opened 2017-10-31
chunked processing fixes and reduce v1f syscalls

nigoroll:
* v1f performance:
v1f called reads even on one-byte buffers. As we normally have a bit of pipeline buffer space from session init, we can re-use it for some readahead. As the readahead must not block, we need to switch blocking on and off more frequently, so we move blocking to the HTC layer and cache the current blocking state.
* chunked:
  * fix parsing of chunk extensions (as required by httpbis)
  * fix handing of trailers (by ignoring them for now, but I want to add full trailer support)

I have *intentionally* left in two disabled Debug VSLs to make life easier for reviewers.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -192,6 +192,8 @@ struct http_conn {
 	char			*rxbuf_e;
 	char			*pipeline_b;
 	char			*pipeline_e;
+	char			*rxra_b;
+	char			*rxra_e;
 	ssize_t			content_length;
 	void			*priv;
 
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -173,6 +173,12 @@ struct http {
  *
  */
 
+enum htc_blocking {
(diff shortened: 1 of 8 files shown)


nigoroll:
FTR I've force-pushed a minor fix to the recursion in `v1f_read()` and stabilized the test. The test failure previously reported by Travis was caused by the fact that I had several sends for the trailers so vtc could complain about the close from varnishd (timing sensitive)
the current code survives `./varnishtest -n 200 -j 40 -t 10 -i tests/b00007.vtc` on my box

bsdphk:
I don't understand the need for the rxra fields, how are they different from the pipeline fields ?

nigoroll:
@bsdphk once the pipeline is consumed, the pipeline fields are nulled. the rxra fields just point to the pipeline region for re-use

nigoroll:
rebased and added dae7d5aeaedb345a8e8db8e104d93f3c4de370f3 - this would be squeezed with the previous commit for merge, but keeping it separate for reviewers

nigoroll:
continued in #2477 ...

PR #2477, opened 2017-11-01
http1 chunked encoding trailers support & other improvements

nigoroll:
Overall documentation from 10d24f445bac7a520b0aa47385ddc4c35299f1c2:

## Usage / VCL
    
Trailer processing is enabled with ``set bereq.http.TE = "trailers";`` in ``vcl_backend_fetch{}``. Only then may the backend send a chunked response with trailers. Any ``TE`` client header remains filtered, so trailer processing is neither implicitly enabled nor would trailers ever be sent downstream (except in pipe mode).
    
The above only allows the backend to send trailers, whether or not it actually makes use of that capability (or if it even chooses chunked encoding) is up to the backend.
    
For processing of the backend response, we enable trailer handling only if a ``Trailers:`` backend response header is present - either from the backend or set in ``vcl_backend_response{}``. ``Trailers:`` must contain header names to be accepted in the chunked trailer. In addition, ``set beresp.http.Trailers = "*";`` can be used in vcl to instruct varnish to accept any header in the chunked trailer. ``Trailers: *`` is never accepted from a backend.
    
Any accepted trailer headers are copied into the object without additional VCL control. In other words, any headers allowed by the ``Trailers`` header are accepted as-are. Additional filtering is possible on the client side.

Trailer processing implicitly disables streaming. The backend response is read into cache and trailer headers are delivered as normal response headers.
    
## Implementation notes

The object API requires that the size of the attribute space is known at object creation time, so we need to guess the amount of space to reserve for trailers. For this initial implementation, the reserve is hardcoded as 1KB and we'll probably want some VCL control later.

For trailer processing, the setting OA_HEADERS now needs happen after reading the body, obviously. It is unclear if The Proprietary Stevedore (tm) supports this.

To avoid code duplication for header dissection, http1_dissect_hdrs() is now generalised and declared in cache.h as HTTP1_DissectHdrs()

For logging new/deleted headers from vbf, http_VSLH() and http_VSLH_del() are now declared in cache_http.h

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -192,6 +192,8 @@ struct http_conn {
 	char			*rxbuf_e;
 	char			*pipeline_b;
 	char			*pipeline_e;
+	char			*rxra_b;
+	char			*rxra_e;
 	ssize_t			content_length;
 	void			*priv;
 
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -173,6 +173,12 @@ struct http {
  *
  */
 
+enum htc_blocking {
+	_HTC_BLK_UNKNOWN = 0,
+	HTC_NONBLOCKING,
+	HTC_BLOCKING
+};
+
 struct http_conn {
 	unsigned		magic;
 #define HTTP_CONN_MAGIC		0x3e19edd1
@@ -180,6 +186,7 @@ struct http_conn {
 	int			*rfd;
 	enum sess_close		doclose;
 	enum body_status	body_status;
+	enum htc_blocking	blocking;
 	struct ws		*ws;
 	char			*rxbuf_b;
 	char			*rxbuf_e;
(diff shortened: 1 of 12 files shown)


fgsch:
Out of curiosity, is anyone using trailers actively?

nigoroll:
I will be very actively once this is finished.

fgsch:
I was talking about browsers or servers. 

For browsers it looks like is mostly not supported, e.g.: 
http://www.browserscope.org/?category=network&v=top

nigoroll:
Yes, for my use case, both client and server will be varnish

nigoroll:
because additional changes to fully support Trailers (including VRT/VCL) depend on #2478, I have created incremental pull requests
1. #2478 basis for 2
2. V1F / backend response trailer support https://github.com/nigoroll/varnish-cache/pull/1
3. Delivery / client response trailer support and VRT/VCL https://github.com/nigoroll/varnish-cache/pull/2

nigoroll:
bugwash result: evaluate a 'trigger fetch from vcl' model, then come back

PR #2478, opened 2017-11-02
v1l should live on the session workspace

nigoroll:
The motivation is to remove the reservation from req->ws during delivery, but actually line delivery is a function of the session, thus memory required for it should also come from there:
    
* We avoid requiring to have an obscure surplus of workspace_client for delivery
  * which is also added for every subrequest though not required there
* We get predictable performance as the number of IO-vectors available is now only a function of workspace_session rather than the amount of memory which happens to be available on the request workspace.

As a sensible side effect, we now also fail with an internal 500 for workspace_session overflows (which we should have already, but as the session workspace was not used for much, this was not relevant).
    
We raise the default to 2k to accomodate the ~120 iovs to send 60 HTTP1 headers in a single writev() call.


nigoroll:
bugwash discussion: We should use `workspace_thread`.
Will change the patch and re-open new PR

PR #2481, opened 2017-11-05
Remove cli buffer

hermunn:
The pull request #2382 was not in fact completely removing `cli_buffer` parameter, and this should fix it. The addition to VSB might be controversial, and I would like to get feedback on what the new function should be called (if the functionality is accepted), and what parameters should be (pointers vs indices, start + length vs. start + end).

### DIFF
--- a/lib/libvarnish/vlu.c
+++ b/lib/libvarnish/vlu.c
@@ -36,32 +36,29 @@
 
 #include "vas.h"	// XXX Flexelint "not used" - but req'ed for assert()
 #include "miniobj.h"
+#include "vdef.h"
+#include "vsb.h"
 
 #include "vlu.h"
 
 struct vlu {
 	unsigned	magic;
 #define LINEUP_MAGIC	0x8286661
-	char		*buf;
-	unsigned	bufl;
-	unsigned	bufp;
+	struct vsb	*buf;
 	void		*priv;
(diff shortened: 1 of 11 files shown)


lib/libvarnish/vlu.c
```diff
@@ -102,23 +99,22 @@ LineUpProcess(struct vlu *l)
...
-		l->bufp = 0;
+		VSB_clear(l->buf);
 	return (0);
 }
 
 int
 VLU_Fd(int fd, struct vlu *l)
 {
 	int i;
+	char buf[BUFSIZ];
```
hermunn: Actually, I think that `BUFSIZ` is too low here, and that we can afford more than this in each `read` call.

bsdphk:
I did some code spelunking, and some renovation seems overdue anyway, so with Hermunns permission I'm taking over this issue.

bsdphk:
With 6cb21b47a16be6b97b8197dd3cb79cef42999273 this is handled.

hermunn:
Backport review: The solution for 4.1 is to accept that `cli_buffer` is not removed. See 96534906297fb8440b for the revert of the documentation change. In other words, `cli_buffer` is still in 4.1, but we keep the fix #2382 so that a high value will not cause a stack overflow.

PR #2483, opened 2017-11-07
thread ws from heap and v1l from thread ws

nigoroll:
* Allocate thread_workspace from the heap rather than from the stack
* take v1l memory from the thread workspace
  * for ESIs, switch to a new thread ws when needed

### DIFF
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -100,6 +100,8 @@ ved_include(struct req *preq, const char *src, const char *host,
 	struct worker *wrk;
 	struct sess *sp;
 	struct req *req;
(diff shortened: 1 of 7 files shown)


bsdphk:
Reducing default threadstack from 48 to 46k helps nobody, it is page-aligned.

The ESI switcheroo is a no-go.

nigoroll:
* stack: right, it should be 44k or remain at 48, yes
* regarding the dual purpose of thread ws let's discuss options please

nigoroll:
Outcome of a very constructive IRC discussion with @bsdphk 
In real life, we need most iovs for headers (roughly number of headers x 2, so commonly in the order of 30-50). Body delivery uses only a comparably small number of IOVs:

PR #2485, opened 2017-11-07
sketch/poc for changes to enable backend body fetch from vcl

nigoroll:
suggestion based of the concept of advancing the fetch fsm from within an fsm state (STP_STARTFETCH) to enable calling VBF_FetchBody from VCL. The actual call to FetchBody is not included


nigoroll:
bugwash disucssion outcome: prepare another poc/sketch without advancing the fsm, just pulling in the body keeping the fsm in STP_STARTFETCH

nigoroll:
Continued in #2491

PR #2487, opened 2017-11-12
http1 delivery to thread workspace take 2

nigoroll:
continuing #2483 with the following differences:
* do not set up a new thread workspace for esi requests, but rather allocate a fixed small number of io vectors (parameter esi_iovs defaulting to 10) on the thread workspace for esi body delivery
* no changes for header delivery or non-esi

As shown in #2483, we should not need more than 5 io vectors during esi processing (because we try to push down ESI chunks ASAP), yet still the default leaves some headroom and as a last resort the parameter can be tuned.

---
commit message d5999c769b95584b2890779f04a38bc921a2f6e4 : 

The motivation is to remove the reservation from req->ws during delivery, but actually line delivery memory should not come from the request space - as originally designed:
    
* We avoid requiring to have an obscure surplus of workspace_client for delivery
** which is also added for every subrequest though not required there
* We get predictable performance as the number of IO-vectors available is now only a function of workspace_thread or esi_iovs (see below) rather than the amount of memory which happens to be available on the request workspace.
    
As a sensible side effect, we now also fail with an internal 500 error for workspace_session and workspace_thread overflows in addition to the existing check on workspace_client for completeness.
    
For ESI requests, we run all of the client side processing, which uses the thread workspace, with V1L set up. Thus, V1L now needs its control structure together with a small amount of io vectors as an allocation on the workspace.

Real world observation has shown that no more than five io vectors are normally in use during ESI, yet still we make this number configurable and have a default with some safety margin.

For non-ESI requests and headers, we use all of the thread_workspace for io vectors, as before.
    
As V1L does not necessarily reserve workspace any more, functions have been renamed to better reflect the purpose:
    
V1L_Reserve -> V1L_Open
V1L_FlushRelease -> V1L_Close

### DIFF
--- a/bin/varnishd/cache/cache_req.c
+++ b/bin/varnishd/cache/cache_req.c
@@ -226,7 +226,6 @@ Req_Cleanup(struct sess *sp, struct worker *wrk, struct req *req)
 	req->is_hit = 0;
 
 	WS_Reset(req->ws, 0);
-	WS_Reset(wrk->aws, 0);
 }
 
 /*----------------------------------------------------------------------
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -1001,14 +1001,6 @@ CNT_Request(struct worker *wrk, struct req *req)
 		CHECK_OBJ_ORNULL(wrk->nobjhead, OBJHEAD_MAGIC);
 		CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 
-		/*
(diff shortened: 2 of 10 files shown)


bsdphk:
OK.

nigoroll:
Merged:
* 9c330b6dd76f11ddbdab30a2c0b62c45176c47c9
* beeaa19cced3fe1ab79381b2b1b7b0b5594cbb18
* changelog 2c71cabe13e6345f22eb44c047218c147680d6f0
waving at @hermunn

PR #2490, opened 2017-11-13
Backend "extrachance" retry fixes

daghf:
This PR addresses #2135 

We permit a single retry and force a fresh backend connection the second time around.

First commit adds a `force_fresh` parameter to `VTP_Get`. Second commit addresses the retry logic.

Failure will land us in `vcl_backend_error`, where the VCL user can choose to do further attempts via `return (retry);`.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -112,7 +112,7 @@ vbe_dir_getfd(struct worker *wrk, struct backend *bp, struct busyobj *bo)
 	bo->htc->doclose = SC_NULL;
 
 	FIND_TMO(connect_timeout, tmod, bo, bp);
-	vtp = VTP_Get(bp->tcp_pool, tmod, wrk);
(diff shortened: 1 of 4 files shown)


daghf:
Also I should mention this PR only affects H/1 backends. Potential future H/2 backends and file based backends implement their own set of `struct director` callbacks, and are thus not affected.

bsdphk:
Just on general principles, this is one of the kind of changes I'd either like to see comprehensively tested in VTC or failing that, run a bit in production before we commit it ?

daghf:
@bsdphk 

I've added a test case now.

With #1772 not being fixed it was kind of tricky to do a VTC. I was able to work around it by sending a whitespace-only response to trigger an epoll event, which wakes up the fetch thread from its `VTP_Wait` call and triggers a retry.

We've also had a customer running a build of this in a production environment for the last week without any issues.

bsdphk:
OK'ed.

I assume this also means we can close #2135 ?

daghf:
Thanks for the OK.

Closing 2135 should happen automagically with the "Fixes" tag.

PR #2491, opened 2017-11-13
POC for early backend body fetches (from VCL)

nigoroll:
Take 2 on how support for early body fetches could look like, this time inclusive of a VCL demo.

`VBF_Fetchbody()` contains all the steps necessary to trigger a (non streaming) Fetch from VCL. While I have not used an fsm inside out approach this time, I havekept the fsm functions: we simply need to do the work in `vbf_stp_fetch()` and `vbf_stp_fetchbody()` somewhere and it does not seem to make much sense to just wrap the functions to get rid of the `fetch_step`. Neither would we want code duplication.

So now `VBF_Fetchbody()` is a simplistic minimal fsm within the outer fsm, taking two steps (fetch and fetchbody) at a time and only once.

`bo->fetch_step_next` is used to pass the state `VBF_Fetchbody()` has left off to the fsm.

Another issue is that we can now add headers (in vcl) after the fetch is complete, so

* we need to postpone copying headers to the objects until later and
* we need a guestimate of free space
  * which we remember in `bo->hdr_spc`

The final code should have a fall back in case `hdr_spc` is not sufficient. In this case, we need to copy the object (on the stevedore level, not on the boc level or above), which should be fine because:

* this only can happen for the early body fetch case
* and only for do_stream = 0

For the inefficient copy case, we'd increase a statistic and log a warning. For the efficient "guestimate was sufficient" case we should create a log entry to get statistics about wasted space.

`vbf_fetch_prep` consolidates all work to be done before the stevedore object can be created based on a good `OA_HEADERS` size estimate.

some esi tests failed due to different log order, these would need adjustment before merge.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -410,6 +410,8 @@ struct busyobj {
 	 * is recycled.
 	 */
 	int			retries;
+	enum fetch_step		fetch_step_next;
+	unsigned		hdr_spc;
 	struct req		*req;
 	struct sess		*sp;
 	struct worker		*wrk;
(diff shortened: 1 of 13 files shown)


nigoroll:
FYI https://github.com/nigoroll/varnish-cache/pull/3 is not yet based upon this PR, I'm waiting for feedback first.

nigoroll:
rebased, force-pushed

nigoroll:
continued in #2501

PR #2499, opened 2017-11-21
RFC: bless the fact that our booleans are uints

nigoroll:
... to allow for defaults which are neither true nor false


nigoroll:
buried after 2/2 votes on irc against

PR #2500, opened 2017-11-21
RFC: early peek at shard config rework specs

nigoroll:
No code here, just .vcc interface specs and some comments.

The plan is to make the shard director more useful and facilitate complex configurations where
* shard parameters are determined independent of the selection of the director (and probably before director selection)
* shards are to be layered below other directors and thus need to be configured by some other means than parameters to the `backend` method.

The goal is to enable lazy resolution as with the other directors.

Reviews/comments of the specs welcome.


nigoroll:
TODO from bugwash (only marginally related to this RFC):
remove hash choice from .key

PR #2501, opened 2017-11-23
POC for early backend body fetches (from VCL)

nigoroll:
Take 2 1/2 on how support for early body fetches could look like, inclusive of a VCL demo.

`VBF_Fetchbody()` contains all the steps necessary to trigger a (non streaming) fetch from VCL. While I have not used an fsm inside out approach this time, I have still kept the fsm functions: we need to do the work in `vbf_stp_fetch()` and `vbf_stp_fetchbody()` somewhere and it does not seem to make much sense to just wrap the functions to get rid of the fetch_step enum.

So now `VBF_Fetchbody()` is a simplistic minimal fsm within the outer fsm, taking two steps (fetch and fetchbody) at a time.

`bo->fetch_step_next` is used to pass the state `VBF_Fetchbody()` has left off to the outer fsm.

Another issue is that we can now add headers (in vcl) after the fetch
is complete, so

* we need to postpone copying headers to the objects until later (latest possible time for streaming is earlier than for buffered)
* we need a guestimate of free space
* which we remember in `bo->hdr_spc`

If the reserved space does not suffice, we fail after returning from `vcl_backend_fetch()` *1)

`vbf_fetch_prep` consolidates all work to be done before the stevedore
object can be created based on a good `OA_HEADERS` size estimate.

some esi tests failed due to different log order, these would need adjustment before merge.

*1)

A theoretical alternative would be to recreate the object, but the object layer is interwoven with the busy object in a way which, IMHO after having attempted an implementation, prevents clean and simple solutions at the moment:

We would need to signal `BOS_FINISHED,` which, in the obj layer, triggers the bos condvar while in fact the busy object is not ready

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -410,6 +410,8 @@ struct busyobj {
 	 * is recycled.
 	 */
 	int			retries;
+	enum fetch_step		fetch_step_next;
+	unsigned		hdr_spc;
 	struct req		*req;
 	struct sess		*sp;
 	struct worker		*wrk;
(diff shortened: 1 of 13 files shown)


nigoroll:
irc diskussion, mostly with @bsdphk and @mbgrydeland : 
* read body into additional temp object in Transient
* handle like 304

Issue #2502, opened 2017-11-24
objcore reference count leak


mbgrydeland:
We leak an objcore reference if we in cnt_deliver switch to synth because of error while unpacking the object's headers.

hermunn:
Back ported as 23dcf55bdf6a24cd long ago.

Issue #1807, opened 2016-03-04
Assert error in cnt_deliver()


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1807

lkarsten:
backport review: commited to 4.1, but not in changelog.

PR #2503, opened 2017-11-24
Deref the objcore before switching to synth on error.

mbgrydeland:
Also add asserts for the references held in req->objcore and
req->stale_oc.

The test case for #1807 catches this bug after adding the asserts.

Fixes: #2502


bsdphk:
OK.

Good Catch.

hermunn:
Backport review: Back ported to 4.1 as 23dcf55bdf6a24cd98.

PR #2504, opened 2017-11-26
First byte timeout fixes

daghf:
This PR fixes the case where `first_byte_timeout` is not respected for recycled backend connections. 

- First commit teaches `VPT_Close` to also handle connections in state `VTP_STATE_STOLEN`. 

- Second commit removes a `VTP_Wait` that was only used for piped connections that are about to be closed. This is no longer necessary, since `VTP_Close` can now deal with also STOLEN connections.

- Third commit adds a timeout in `VTP_Wait` and bails out to let `vbe_dir_finish` do the necessary `VTP_Close`. Failure to respond within `first_byte_timeout` lands you in `vcl_backend_error`.

Fixes: #1772 
Supersedes: #2347


bsdphk:
Looks OK.

The test case may be a bit to aggressive with timing, but we can relax that later.

PR #2507, opened 2017-11-28
Avoid premature H/2 session close

daghf:
Timing out an H/2 session needs to take into account any active streams being processed.

This fix should also hopefully contribute to t02003.vtc showing up less frequently in VTEST.


bsdphk:
OK

Looks like it stabilizes t020003 here.

Issue #2438, opened 2017-09-25
2018-03 Release Coordination


bsdphk:
This ticket is for coordination of the March 2018 release.

# Desirable content of this release:
* Clear separation of VMOD API levels in .h files (phk)
* VCL object model nailed down (phk)
* ~~VCL-configurable VFP/VDP stacks (phk)~~
* UNIX domain sockets (geoff)
* VMOD VSC counters (dridi/phk)
* Better everything (everybody)

# Things people want to see improved, without putting their name on the line:
* H2

# Things for discussion:

## VCL identifiers case-sensitive (... 5.1) or not (-trunk)
  * HTTP headers are not case sensitive, so req.http.foo and req.http.FoO are the same anyway.

## Horrible director/backend naming in C name-space
  * Time to change backend → http_backend and director → backend ?

# TODO list for release

## Document #2521 issue in release note

nigoroll:
bugwash:
* do we want to retain strict 4.0 vcl semantics? See for example #2509 or c2406f4ff586b7f00efb86d06a250ec249f6f53d

nigoroll:
during last bugwash we agreed that we wanted to ponder this. Here's my bit:

general thoughts:
* if we aim to keep semantics for existing vcl, we might still fail by introducing new bugs (for instance in wrapper code)
* we would not want to maintain current semantics indefinitely, so the only benefit for users would the option to delay migration of their vcl
* this endeavor diverts resources from making actual progress
* how many vtcs would we need to double?
* would we need to duplicate all vmods?
* Do bundled vmods count towards vcl semantic compatibility? If we tell users "you will be able to keep your vcl", they would expect this to include vmods, right?

code:
* gc beresp.storage #2509 
  * could keep old code or
  * implement wrapper to retain old behavior
  * is this relevant at all? Worth the effort?
* change restart handling c2406f4
  * would need to keep two code paths in core
* backend connection recycle/retry #2135
  * would not make sense to optionally switch back to old behavior, IMHO
* doesn't VRT_fail break existing semantics?

my opinion:
I understand that we have upset some users in the past by requiring work on VCL changes, but my impression is that this was mainly caused by the client/backend separation and a number of renames. I do not see much value in diverting developer resources to developing and maintaining a compatibility layer for something like "the last two major version's VCL semantics": The benefit appears rather minor and there is a risk of failing on details.
[...]

PR #2516, opened 2017-12-11
Shard key alg retire

nigoroll:
Ref #2500: As suggested by @bsdphk : Simplify the shard `.key()` method

Holding back due to the ongoing discussion regarding vcl 4.0 compatibility - see #2438


nigoroll:
rebased and added missing tbl files

nigoroll:
merged as 15cc069bdc4ec7ff1925848b8cd8bfb2e9edeece

Issue #2351, opened 2017-06-22
Assert error in http_linkh


daghf:
Also via http2fuzz. Was still able to trigger this with vastly elevated WS sizes, so it does not look like a regular WS overflow.

    Error: Child (19155) Panic at: Wed, 21 Jun 2017 11:39:12 GMT
    Assert error in http_linkh(), cache/cache_http.c line 1120:
      Condition(((fm->hd[n]).b) != 0) not true.
    version = varnish-trunk revision b2623920d, vrt api = 6.1
    ident = Linux,4.9.0-3-amd64,x86_64,-jnone,-smalloc,-smalloc,-hcritbit,epoll
    now = 92434.758327 (mono), 1498045152.781766 (real)
    Backtrace:
      0x453ed0: bin/varnishd/varnishd() [0x453ed0]
      0x453ccb: bin/varnishd/varnishd() [0x453ccb]
      0x44a34c: bin/varnishd/varnishd() [0x44a34c]
      0x44a24f: bin/varnishd/varnishd(http_FilterReq+0x14f) [0x44a24f]
      0x435508: bin/varnishd/varnishd() [0x435508]
      0x434a53: bin/varnishd/varnishd() [0x434a53]
      0x4839cc: bin/varnishd/varnishd() [0x4839cc]
      0x483241: bin/varnishd/varnishd() [0x483241]
      0x482eca: bin/varnishd/varnishd() [0x482eca]
      0x7fc4d2061494: /lib/x86_64-linux-gnu/libpthread.so.0(+0x7494) [0x7fc4d2061494]
    thread = (cache-worker)
    thr.req = 0x7fc4c142f020 {
      MAGIC 0x00000000 (Should:REQ_MAGIC/0x2751aaa1)
      vxid = 0, transport = NULL
      step = 0x0,
      req_body = R_BODY_INIT,
      restarts = 0, esi_level = 0,
      ws = 0x7fc4c142f208 {
        MAGIC 0x00000000 (Should:WS_MAGIC/0x35fac554)
        id = \"\",
        {s, f, r, e} = {(nil), +0, +0, +0},
      },
      http[req] = (nil) {
      },

daghf:
Slightly more detailed panic, with symbol names via gdb.

```
Info: Child (7905) said Child starts
Error: Child (7905) died signal=6
Error: Child (7905) Panic at: Wed, 12 Jul 2017 15:10:57 GMT
[...]

PR #2517, opened 2017-12-12
H/2: Fail streams with missing :method or :path

daghf:
Fixes: #2351


daghf:
Merged in 1d4149f26e96a866a179ab9

PR #2520, opened 2017-12-13
Make H/2 VSL less chatty by default

daghf:
Logging every single bit transferred is a bit excessive, so we add a debug flag for enabling this.

This commit adds (`-p debug=+h2_raw`) that enables logging of the raw H/2 frame bytes (H2TxHdr, H2TxBody, H2RxHdr, H2RxBody).


daghf:
@Dridi comments that an alternative approach here would be to instead add H2TxHdr, H2TxBody, H2RxHdr and H2RxBody to the default `vsl_mask`

bsdphk:
The original idea here was somewhat hazy, and the plan was to make it up along the way.

In particular, the hexdumping should be happening on the API side, not in varnishd, but I never got around to that.

It may be prudent to have a middle level where we dump sufficient bytes of all frames to still have an idea what's going on.

daghf:
Closing this to go for a different approach

Issue #2511, opened 2017-12-08
The option of naming a storage could be more explicit in the man page


gquintard:
while there, also explain implicit naming.

PR #2523, opened 2017-12-14
More info about the -s argument

gquintard:
fix: #2511


hermunn:
This documentation fix is not completely accurate for 4.1. The committer can back port to 4.1 if he so desires.

Issue #2515, opened 2017-12-11
Regression: VMOD ENUMs cannot default to NULL


slimhazard:
## Expected Behavior
Existing VMODs that set NULL (literal 0) as the default value for an ENUM should not be broken.

Since ``VCL_ENUM`` is ``typedef const char *``, NULL is a valid value, and may be the best choice as a default, depending on the requirements of the VMOD -- by default "none of the above". (Of course the VMOD must be implemented to handle NULL correctly, but that's the VMOD's problem.)

Until the change mentioned below, it was possible to set NULL as the default with this syntax in the VCC declaration:
```
... foo(ENUM {VAL1, VAL2} bar=0)
```
Examples of VMODs that do this:
https://code.uplex.de/uplex-varnish/libvmod-re2
https://code.uplex.de/uplex-varnish/libvmod-pcre2
https://code.uplex.de/uplex-varnish/libvmod-gcrypt

## Current Behavior
Since the change in c91a6ce, the syntax shown above is rejected by the vmodtool as an error. This is because the values listed in ``{...}`` are appended to a ``.spec`` field in the vmodtool (a list of strings), and the default value is now checked if it appears in the list. If it doesn't, the VCC declaration is rejected as an error.

## Possible Solution
* Recognize the ``=0`` case and allow it, setting NULL as the default; or
* (Maybe better) Permit a distinguished value like ``__NONE__`` as the default, and set NULL as the default in that case. That would preclude VMODs from using the ``__NONE__`` literal (or whatever it turns out to be), but a choice like that with the double underscore, or some other "magic" spelling, is unlikely to break anything.

## Steps to Reproduce (for bugs)
Build any of the VMODs mentioned above.

## Context
Trying to keep VMODs compatible with master (and get a head start on compatibility with the March 2018 release), without having to chase down ... *that* many breaking changes ... (?)

## Your Environment

* Version used: master since c91a6ce (any platform)

nigoroll:
The obvious simple solution is to designate one ENUM (e.g. `undef`) for this purpose, but this would duplicate the logic for each vmod and I do agree that having a standard value for `undef` semantics would make sense.
I'd even propose to widen the scope and define `undef` for as many VCL types as possible:
* `NULL` for all the pointers (`VCL_ACL`, `VCL_BACKEND`, `VCL_BLOB` etc.)
* `USHRT_MAX` for `BOOL`
[...]

PR #2524, opened 2017-12-17
Add ARGMASK for vmod information on which arguments were given explicitly

nigoroll:
Ref: #2515

### DIFF
--- a/bin/varnishtest/tests/m00000.vtc
+++ b/bin/varnishtest/tests/m00000.vtc
@@ -10,6 +10,7 @@ varnish v1 -vcl+backend {
 	import std;
 	import debug;
 	import vtc;
+	import blob;
 
 	sub vcl_init {
 		new obj = debug.obj();
@@ -33,6 +34,10 @@ varnish v1 -vcl+backend {
 		std.log("VCL" + " initiated " + "log");
 		std.syslog(8 + 7, "Somebody runs varnishtest");
 		debug.rot52(resp);
+		set resp.http.a0 = blob.encode(HEX, blob=debug.args());
+		set resp.http.a2 = blob.encode(HEX, blob=debug.args(b=1,ag=32));
+		set resp.http.a5 = blob.encode(HEX, blob=debug.args(a=0,c=2));
+		set resp.http.aX = blob.encode(HEX, blob=debug.args(af=31));
 	}
 } -start
 
@@ -48,6 +53,10 @@ client c1 {
 	expect resp.http.encrypted == "ROT52"
 	expect resp.http.what >= 16
 	expect resp.http.not == -1
+	expect resp.http.a0 == "00000000"
+	expect resp.http.a2 == "00000002"
+	expect resp.http.a5 == "00000005"
+	expect resp.http.aX == "80000000"
 } -run
 
 logexpect l1 -v v1 -g raw -d 1 {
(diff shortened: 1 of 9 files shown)


slimhazard:
What happens if there are more than 32 args? (I'll bet you can guess which VMOD I could show as an example where there are more than 32 args for a method.)

Wouldn't a vbm make for a general solution? It could be defaulted to 32 bits, although it could actually default to one byte, since VMOD methods & functions with more than 8 args are rare.

nigoroll:
> What happens if there are more than 32 args?

They are ignored for the argmask (the vtc I have actually tests for this here https://github.com/varnishcache/varnish-cache/pull/2524/commits/7a7efa719289a4f4ffc608a4d64dd7e2fcb70f80#diff-bdd5299dfe537695c7a6f8336ccec344R38 )

My take on the bitfield width was that having the information only available for the first _n_ arguments (going to _n_=64 would be trivial).

> Wouldn't a vbm make for a general solution? It could be defaulted to 32 bits, although it could actually default to one byte, since VMOD methods & functions with more than 8 args are rare.

I think this should be doable, though using a simple integer arguments is easier and I'd consider a vbm overkill for the reason given above.

As the number of arguments is known at vcc time, the bitfield width would always be fixed and sized just right.

nigoroll:
@Dridi, good point! I'll go ahead with this.

Famous last words: _32 `$Undef` arguments ought to be enough for everyone._

slimhazard:
Are we talking about putting a fixed limit on the number of params that a VMOD method or function can have? Or on how many of them can have a default value? Seriously?!

The ``regex.sub`` function in [this VMOD](https://code.uplex.de/uplex-varnish/libvmod-pcre2) has 48 params. I know that seems like a lot, but it works fine, because two of them are ``PRIV_*``, two others are required, and all of the rest have default values.

A vbm for the purpose of ARGMASK is not "overkill". An arbitrary-length bitmap interface is *just exactly what the feature calls for*.

Dridi:
> The `regex.sub` function in this VMOD has 48 params.

Do they all need an `$Undef` default?

slimhazard:
> Do they all need an $Undef default?

The VMOD is fine so far without it. But what if someone comes along with a VMOD that does need more than 32 $Undefs?

Can someone help me understand the attraction to a solution that imposes a fixed limit, although we have a perfectly good interface baked right into Varnish that avoids exactly that? This is blowing my mind.

nigoroll:
> Can someone help me understand the attraction to a solution that  imposes a fixed limit

simplicity.

slimhazard:
> simplicity.
```
if (ctx->argmask & (1U << n))
```
has no such advantage over:
```
if (vbit_test(ctx->argmask, n))
```
Whereas the latter has the advantage of imposing no fixed maximum at all. Looking at it now, I'd say that the latter is actually clearer as well.

I remain utterly flabbergasted. I know that I, in part, got this started with the ENUM=0 issue, but I am very much -1 on this PR if it creates an unnecessary limitation like this.

nigoroll:
@slimhazard I was taking about implementation simplicity on the vcc side of things.

Anyway, I will go ahead now and implement this with a vbm

nigoroll:
FTR, I think there is a much simpler and better, yet still fully generic way than vbm. Stay tuned.

nigoroll:
next iteration PoC in #2526

PR #2525, opened 2017-12-20
shard director: LAZY mode (vdi resolve function), parameter objects

nigoroll:
*XXX* needs more changelog tlc / cleanup before final merge

suspected phk concerns upfront:

* to pass parameters from vcl to the resolve function, we need some state which we can lookup from both sides. We use a vrt task priv for this, looked up via the vmod _object_ id. phk recently said that this use case was not the original design intention, but it is the safest/cleanest I can currently see for this use case.

* the vdi resolve function would need a vrt_ctx argument. As we don't have it at the moment, we construct one for the task priv lookup.

requirements:

* #2524 ARGMASK
* #2516 shard key alg retire

on the actual change:

We introduce a shard_param object to hold the shard director lookup parameters which until now could only be passed to the .backend() method.

By associating a parameter object with a shard director, we enable LAZY lookups as with the other directors. Parameter objects are defined with VCL scope (normal vmod objects), but can be overridden per backend request using a task priv.

We use the same concept to carry shard.backend() parameters to vdi resolve for LAZY mode: They get saved in a per-director task scope parameter object.

Each object points to another object providing defaults for values which are not defined.

Actual resolution of the various parameter objects does not happen before they are used to allow changing them independently (ie, shard .backend() parameters have precedence over an associated
parameter object, which by itself can be overridden).

Overview of parameter objects (pointers are alternatives)

```
shard() director        shard_param() object    default praram

             --------------------------------->   vmod static
  VCL obj   /                                ->
  .param  -+--------->    VCL obj           /  _
                          .default  --------   /|
                                              /
                             ^               /
                             |              /
                                           /
                          .default        /
        ------------->    TASK priv      /
       /                                /
  .default -----------------------------
  TASK priv

```


doc/changes.rst
```diff
@@ -47,6 +47,52 @@ VCL and bundled VMODs
...
+  * for ``alg=RS``, replace::
+
+      <dir>.backend(by=KEY, key=<dir>.key(<string>, RS))
+
+    with::
+
+      <dir>.backend(by=BLOB, key_blob=blobdigest.hash(RS,
+	blob.decode(encoded=<string>)))
+
+
```
slimhazard: The custom thus far in the project has been that changes.rst has terse summaries of the changes (most of them one-liners), and longer texts like this that motivate the change are in the What's New/Upgrading release docs. The entry in changes.rst can link to the longer text. Not that this against the law or something, but I do think that users are more likely to read up about this in W'sN/U than in changes.rst.

doc/sphinx/reference/vmod.rst
```diff
@@ -149,14 +158,21 @@ VCL data types are targeted at the job, so for instance, we have data
...
 
 	A type for named ACLs declared in VCL.
 
+ARGMASK
+	C typedef: ``VCC_ARGMASK``
+
+	C-type: ``uint32_t``
+
+	See :ref:`ref-vmod-named-args` above.
+
```
slimhazard: Shouldn't the docs about ARGMASK be a part of the PR for ARGMASK?

doc/changes.rst
nigoroll: Absolutely, this belongs to the next release notes. I just wanted to park the text in changes.rst until we start an `upgrading-x.x.rst` guide
If that's a concern, we can start `upgrading.rst` now.

nigoroll:
rebased onto 111a42f745cf188ec13a0be10b6e15b9d1fd3f15 to avoid #2535

nigoroll:
force-pushed again: rebased onto 0773342679f5cf4e696162ce1e20069d2a3c84b6 and added that fix

nigoroll:
@bsdphk and myself have discussed the blob-wrapping of the shard param object to pass it to the {{backend}} and {{associate}} methods. Agreement:
* It's a transitional interface
* we're not too fond of it and want to replace it with VCL support to pass object references
* phk can live with it because it's contained in a vmod

nigoroll:
merged as 9b8f6e21809bf867eb5bd792331484d0e70eb367

PR #2526, opened 2017-12-22
PoC: vmod function $Undef feature: argmask with dynamic struct decl

nigoroll:
really an unfinished PoC, loose ends:
* cleanup needed
* vmodtool: not yet implemented for vmod objects (but that's easy)

showing how an argmask to indicate used vmod function arguments could look like which fulfils the requirements uttered by these three folks:

* @slimhazard fully dynamic in size, no "_n_ bits ought..."
* @Dridi `$Undef` rules it all
* @rezan symbolic access by argument name


lib/libvcc/vmodtool.py
```diff
@@ -401,8 +414,55 @@ def c_fn(self, args=[], h=False):
...
+            s += 'struct '
+        else:
+            s += 'td_'
+        s += struct_args_pfx
+        if not h:
+            s += self.vcc.modname + '_'
+        s += self.cname(h)
+        return s
+
+        s = 'struct '
```
nigoroll: yes, I know, needs cleanup

bsdphk:
I have gotten this to work in a way which provides much more leverage going forward.

Please see 6515f17ed61ee979a170612a3467d87fedd757b1 for example.

PR #2529, opened 2018-01-08
Add parameters for various H/2 settings

daghf:
This adds varnish parameters for all H2 settings except for `SETTINGS_ENABLE_PUSH` (that setting does not make sense from a server POV).

The next step here is to actually start enforcing the configured settings (max concurrent streams, max frame size, etc.). Currently the only one that is actually used is `SETTINGS_INITIAL_WINDOW_SIZE`.

### DIFF
--- a/bin/varnishd/http2/cache_http2_session.c
+++ b/bin/varnishd/http2/cache_http2_session.c
@@ -62,6 +62,16 @@ static const struct h2_settings H2_proto_settings = {
 #include "tbl/h2_settings.h"
 };
 
+static void
+h2_local_settings(struct h2_settings *h2s)
+{
+	*h2s = H2_proto_settings;
+#define H2_SETTINGS_PARAM_ONLY
+#define H2_SETTING(U, l, ...)			\
+	h2s->l = cache_param->h2_##l;
+#include "tbl/h2_settings.h"
+#undef H2_SETTINGS_PARAM_ONLY
+}
 
 /**********************************************************************
  * The h2_sess struct needs many of the same things as a request,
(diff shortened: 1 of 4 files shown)


nigoroll:
we'll probably want something like `sub vcl_transport {}` or `sub vcl_protocol {}` with the option of setting these per connection at some point, but +1 for at least adding the parameters as a first step.

bsdphk:
I don't like the "assert(len >= 36);", it should be inside the H2_SETTINGS macro so adding settings updates it automatically.

daghf:
@bsdphk something like this?

```
diff --git a/bin/varnishd/http2/cache_http2_session.c b/bin/varnishd/http2/cache_http2_session.c
index f7ed018cb..73bab0f78 100644
--- a/bin/varnishd/http2/cache_http2_session.c
+++ b/bin/varnishd/http2/cache_http2_session.c
@@ -55,11 +55,13 @@ static size_t
 h2_enc_settings(const struct h2_settings *h2s, uint8_t *buf, size_t len)
 {
        uint8_t *b;
+       size_t l = 0;
 
-       assert(len >= 36);
        b = buf;
 #define H2_SETTING(U,l,v,d,...)                                \
        if (h2s->l != d) {                              \
+               l += 6;                                 \
+               assert(l <= len);                       \
                vbe16enc(b, v);                         \
                b += 2;                                 \
                vbe32enc(b, h2s->l);                    \
```

daghf:
PR updated

bsdphk:
Yeah, that looks better.  OK when CI-checks complete

PR #2534, opened 2018-01-10
bytearray needs an encoding in python3

gquintard:
tested with both pythons


bsdphk:
Is ascii the right thing, or should we use utf8 ?

Does it make any difference ?

gquintard:
Files are identical, and while utf8 is the native string encoding for python, I'd argue that ascii is the one for C, which is what we are producing here.

PR #2536, opened 2018-01-11
Get accurate byte counters

dmatetelki:
The future of the patch can either be:

- Keeping the old ones for compatibility.
- Replacing the old counters with the new counters as the old ones are only useful for debugging mostly varnish itself, maybe the site.

------------
There was a regression from Varnish 4.0 to 4.1, where the response
bytes was counted as the number of bytes fed to the outgoing write
vector, rather than the bytes that was actually handed off to the OS'
socket buffer. This would cause for many cases the complete object
size counted as transmitted bytes, even though the client hung up the
connection early.

To fix this, two new fields has been added to the ReqAcct and
BereqAcct log lines. These fields show the number of protocol bytes
that were read or written to the socket. These bytes will include any
extra protocol overhead (e.g. chunked encoding).

varnishncsa has been changed to use these new counters in its byte
count output. For the %b formatter (body bytes sent), the output will
be the smallest of the response body bytes, or the protocol response
bytes. This is to give the expected output for known object sizes
without inflation on connection hangups.


Dridi:
> The future of the patch can either be:
>
> - Keeping the old ones for compatibility.
> - Replacing the old counters with the new counters as the old ones are only useful for debugging mostly varnish itself, maybe the site.

If they bear the same semantics but do a better job at it, the new counters should replace the old ones. As long as we maintain the fields as before it shouldn't break anything for users (and yes, I know, famous last words and all that jazz =).

gquintard:
I'd vote to remove the delta. Either we remove that from VCP or we push
that upstream.

-- 
Guillaume Quintard

bsdphk:
I think the consensus at last weeks bugwash was to have only one set of counters and have a feature flag for the old (ie: present) and arguably wrong way of counting.

dmatetelki:
@bsdphk : I've re-read the irc discussion and my impression is that the feature-flag idea was opposed (@fgsch, @Dridi,...). You wanted someone to come up with a proposal, as H2 counters needs to be taken consideration. I can do that with @daghf.

n4n0r:
Hi,
I can confirm we encounter this issue (in varnish 4.1.8). This is quite impacting for measuring bandwidth usage, in particular when video streaming is involved. 
What release are you planning to release this fix to ?

dmatetelki:
@n4n0r : It is on the top of my list and i hope it will be in the next release.

dmatetelki:
Abandoning for @mbgrydeland new #2558

PR #2540, opened 2018-01-15
Introduce obj.storage for VCL

### LINKED ISSUE #2533: Wrong beresp.storage on pass transactions
I haven't checked to what extent this is broken, and I've only checked trunk:

```
varnishtest "wrong beresp.storage"

server s1 {
	rxreq
	txresp
} -start

varnish v1 -vcl+backend {
	sub vcl_recv {
		return (pass);
	}
	sub vcl_backend_response {
		set beresp.http.Storage = beresp.storage;
	}
} -start

client c1 {
	txreq
	rxresp
	#expect resp.http.storage ~ Transient
} -run

varnish v1 -expect SMA.s0.c_req == 0
varnish v1 -expect SMA.Transient.c_req > 0
```

If you enable `c1`'s `expect` you'll get this:

```
---- c1    0.4 EXPECT resp.http.storage (storage.s0) ~ "Transient" failed
```

Haven't looked further yet in case there are volunteers.

Dridi:
In addition, instead of always initializing a transaction's storage
backend using the round-robin selection, pick Transient when we know
in advance that beresp will be uncacheable.

Closes #2533

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -261,8 +261,7 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 	CHECK_OBJ_NOTNULL(bo, BUSYOBJ_MAGIC);
 
 	AZ(bo->storage);
-
-	bo->storage = STV_next();
+	bo->storage = bo->do_pass ? stv_transient : STV_next();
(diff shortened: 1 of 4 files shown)


fgsch:
LGTM

bsdphk:
OK

PR #2543, opened 2018-01-17
Enforce h2_max_concurrent_streams

daghf:
With this patch we fail new streams with a `REFUSED_STREAM` if h2_max_concurrent_streams is exceeded. New streams may be opened again after the count drops below the limit.

I'm using the value of `h2->refcnt` for the counting. If at some point we decide other bits will need to hold `h2_sess` refs, we will need to revisit this.

For the test case we need to keep N streams in flight while also making sure they are started in order, so I ended up using quite a few barriers to stabilize that.


bsdphk:
Looks OK

PR #2547, opened 2018-01-21
Changes in timestamps, prepare for 2519

hermunn:
These patches prepares for #2519. The first patch, "Update t_req on restart" is simple: `return (restart)` should reset the time for the request.

The second patch is a bug fix. Without it, the `ttl` variable will change during VCL processing, and taking a timestamp can change the result of VCL processing, which is just plain wrong. The test case demonstrates this - remove the `std.timestamp()`, and the test case passes (on master).

The bug was disvovered while working on #2519, and it is required to make it water tight. Unfortunately, the second patch + #2519 without the first patch will break `c00041.vtc` (test purging from vcl). For this reason I have updated #2519 to include the patches in this PR.

In other words, the first patch in this PR should be accepted because it is necessary when a PURGE request invalidates the object, and then uses `return (restart)` to fetch a new object from the backend.

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -747,8 +747,10 @@ cnt_restart(struct worker *wrk, struct req *req)
 		req->err_code = 503;
 		req->req_step = R_STP_SYNTH;
 	} else {
+		double now = W_TIM_real(wrk);
+		req->t_req = now;
 		// XXX: ReqEnd + ReqAcct ?
-		VSLb_ts_req(req, "Restart", W_TIM_real(wrk));
+		VSLb_ts_req(req, "Restart", now);
 		VSL_ChgId(req->vsl, "req", "restart",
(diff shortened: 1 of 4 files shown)


bsdphk:
I disagree on this bit: return(restart) should *not* reset t_req.

t_req is the honest time we received the request, how much we fumble with it in VCL should not affect that.

hermunn:
I suspected this, or, to be fair, a colleague predicted this. Actually, since submitted this I realized the text is wrong. The problem is that leaving out the first patch makes c00041.vtc brittle, even without additional changes.

Proposition: In `cnt_purge` we call `HSH_Purge` which takes a timestamp before calling `EXP_Rearm`. Perhaps we should use `t_req` here instread. I will create a branch end test.

hermunn:
I made a new pull request (as you can see) that should fix c00041 in a different way. I guess it is better than this one, but please dicuss pros and cons of both.

hermunn:
Closing if favor of #2555.

PR #2548, opened 2018-01-22
Changes in timestamps, purges get better at killing objects

hermunn:
This is slightly different than #2547, and I think I found the right way to not cause problems with `c00041.vtc` like #2547 does.

The only question is if a call to `EXP_Rearm` now should use `t_req` for `now` in all cases. In core varnish all rearms are purges, but there are VMODs that use `EXP_Rearm` for *soft purges*.

### DIFF
--- a/bin/varnishd/cache/cache_expire.c
+++ b/bin/varnishd/cache/cache_expire.c
@@ -164,7 +164,9 @@ EXP_Rearm(struct objcore *oc, double now, double ttl, double grace, double keep)
 	VSL(SLT_ExpKill, 0, "EXP_Rearm p=%p E=%.9f e=%.9f f=0x%x", oc,
 	    oc->timer_when, when, oc->flags);
 
-	if (when < oc->t_origin || when < oc->timer_when)
+	if (ttl == 0. && grace == 0. && keep == 0.)
+		HSH_Kill(oc);
+	else if (when < oc->t_origin || when < oc->timer_when)
 		exp_mail_it(oc, OC_EF_MOVE);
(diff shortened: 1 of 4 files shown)


Dridi:
We now have a `vmod-purge` for soft purges, and the `VRT_purge` function can now rearm all the things without tying VMODs to `EXP_Rearm`. In `varnish-modules` we no longer build `vmod-softpurge` for the 5.x series.

I think your check in 5b84e9b should look at `oc->*` instead of the functions parameters to be more accurate. `grace` and `keep` may be NaN on the function call and fail the check even though their `oc->` counterparts are both zero. Same for `ttl`, it may be left alone (NaN) but arithmetic on `oc->ttl` may tell us it's stale.

hermunn:
> I think your check in 5b84e9b should look at oc->* instead of the functions parameters to be more accurate

I had to change  5b84e9b to 77fe035 to avoid a bad race. But I think you are right, the test should probably be

```
if (ttl == 0.0 && oc->grace == 0.0 && oc->keep == 0.0)
```

hermunn:
I updated the PR according to @Dridi's suggestion, but that broke `r02339.vtc` and `m00031.vtc`. I am not sure if it is the code or the test that is in error, thogh.

Input appreciated.

hermunn:
Actually, it looks like `r02339.vtc` and `m00031.vtc` breaks for all variants of my proposal. Hmmmm, not sure what is correct.

hermunn:
Closing if favor of #2555

PR #2553, opened 2018-01-31
Introduce optional numeric arg to std.ip

gquintard:


Dridi:
In order to maintain current behavior, `BOOL numeric = 1` should actually be `BOOL numeric = 0`. Thank @daghf for this pointing out, I got this one backwards...

gquintard:
duh, I knew that. Thanks for the keen eye. Re-running the test before amending

bsdphk:
If your resolver library needs this, it is badly broken.

A well-formed numerical address should never cause the resolver library to do network traffic...

Dridi:
@bsdphk on the other hand if we get said IP addresses from an untrusted third party, we may want to prevent spurious lookups.

fgsch:
I wonder if we should make numeric the default and add another function that resolves.

bsdphk:
Bugwash consensus:

Make the new optional arg "resolve = true" and don't change default.

Issue #2530, opened 2018-01-09
Varnish shouldn't test gunzip for range responses


Dridi:
This is a placeholder, I will close this issue once fixed in both master and 4.1 branches. Confirmed for any version since at least 4.0, and I have a patch for trunk. I will submit the first patch once I have a more comprehensive test case.

Dridi:
I'm using this issue also as a placeholder for a collateral bug found while working on this, regarding ESI and partial responses.

The reasons why it took so long to send the patches despite having them ready at the time are 1) I was sick that week 2) I found the ESI bug and 3) it got OBE. Sorry about the latter.

bsdphk:
Go see #2554 instead

PR #2554, opened 2018-01-31
VFP and VDPs for partial responses

Dridi:
Otherwise all bets are off :)

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -514,13 +514,19 @@ vbf_figure_out_vfp(struct busyobj *bo)
 	 *
 	 */
 
-	/* No body or no GZIP supprt -> done */
+	/* No body or no GZIP support -> done */
 	if (bo->htc->body_status == BS_NONE ||
 	    bo->htc->content_length == 0 ||
 	    !cache_param->http_gzip_support) {
 		http_Unset(bo->beresp, H_Content_Encoding);
 		bo->do_gzip = bo->do_gunzip = 0;
 		bo->do_stream = 0;
+		return (0);
+	}
+
+	/* Don't mess with a partial body */
+	if (http_GetStatus(bo->beresp) == 206) {
+		AN(bo->uncacheable);
 		return (0);
 	}
 
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -527,6 +527,11 @@ vbf_figure_out_vfp(struct busyobj *bo)
 	/* Don't mess with a partial body */
 	if (http_GetStatus(bo->beresp) == 206) {
 		AN(bo->uncacheable);
+		if (bo->do_esi) {
+			VSLb(bo->vsl, SLT_FetchError,
+			    "ESI enabled on partial response");
+			return (-1);
+		}
 		return (0);
 	}
 
--- a/bin/varnishtest/tests/r02530.vtc
+++ b/bin/varnishtest/tests/r02530.vtc
@@ -1,0 +1,74 @@
+varnishtest "Don't test gunzip for partial responses"
+
+# The use of ETags is only here to ensure they aren't accidentally weakened.
+
+server s1 {
+	# pass'ed range request
+	rxreq
+	expect req.url == "/pass"
+	expect req.http.Accept-Encoding == gzip
+	expect req.http.Range == bytes=0-1
+	txresp -status 206 -nolen \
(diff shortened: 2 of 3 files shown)


Dridi:
I decided not to wait for the 4.1 patch to be ready to submit this one.

fgsch:
Why failing? Can we ignore it instead?
I think it's pretty common that people blindly sets do_esi.

Dridi:
> Why failing? Can we ignore it instead?

Because we can't guarantee the ESI expansion on a partial backend response, so we need to either fail or retry a new `bereq` without the range.

>  I think it's pretty common that people blindly sets do_esi.

Then maybe it shouldn't be ported to 4.1 in this case (only the testGunzip fix).

fgsch:
> Because we can't guarantee the ESI expansion on a partial backend response, so we need to either fail or retry a new bereq without the range.

We cannot guarantee it, ever. 
`do_esi` is just an indicator - we might not find any ESI tags even for non-partial requests.

Since this can be workaround in VCL with a few lines I don't think we should go for this.
There are too many assumptions and potential fallout to make it worth it.

If anything I'd document it.

Dridi:
> We cannot guarantee it, ever.
> do_esi is just an indicator - we might not find any ESI tags even for non-partial requests.

This case is fine, if there are no ESI tags then so be it. It's different than missing ESI tags because they are out of range or truncated.

>  There are too many assumptions and potential fallout to make it worth it.

The one case we briefly discussed is having a cacheable 206, but if it's cacheable the Range header is unset in `bereq` so we always cache the whole object. And if we were on the cacheable track and get an unattended partial response (backend misbehaving, covered by the test case) then `beresp` is marked as uncacheable and I don't think we can revert that decision.

fgsch:
Except you could insert the Range header in v_b_f{}.

I still fail to understand why we want to add this as C code. Why not VCL?

Dridi:
> Except you could insert the Range header in v_b_f{}.

At this point you should know what you are doing.

> I still fail to understand why we want to add this as C code. Why not VCL?

I wouldn't mind a VCL fix, so that would be a check in the built-in's v_b_r? But in this case you could still bypass the check.

fgsch:
> At this point you should know what you are doing.

We could say the same about ESI since it's disabled by default.

Anyhow, if we add it in vcl_backend_fetch you could still bypass it, correct.

Dridi:
I think we have finally reached mutual understanding, I hope we'll get to decide on a solution for next bugwash.

bsdphk:
The bigger picture here is that doing any VFP/VDP processing on range responses is iffy.

Presently we only have "our own" filters (gzip/gunzip/esi) but VMOD filters are in our future, and it seems to me that filters will have to be told that a response is partial at ->init() time, and decide for themselves what to do.

This mechanism needs to ensure that you cannot bypass mandatory filtering by sending a range for the entire object with Cookie or Authentication header.

(I'm closing #2530 so we only have one issue for this)

Dridi:
@bsdphk I failed to capture the final decision:

- The first patch disables VFPs on partial responses, so far so good?
- The second patch fails the fetch transaction for partial responses with `do_esi == true` (explicitly requested vs header-negotiated). Do we ultimately want this behavior? I think we vetoed the idea of a feature flag to control it.
- The second patch has both a `FetchError` that could be considered documentation, and actual documentation. Is that enough, or should we have more bells and whistles in `vcl(7)`?
- Do we also want to fail a `do_g[un]zip == true` on a partial backend response?
- Should I update this pull request to neuter VDPs on partial responses as well?

rezan:
I am a little bit nervous reading some of the language here myself. The solution is to disable VFP/VDPs that we *know* about and which break in this configuration, in particular, gzip and esi. I just want to make sure... but we cannot assume all processors need to be skipped when a partial is encountered. This is the exact challenge processors bring up, how do we know what any processor actually does and when they need to be used? In particular, if I do a range of 0-99999999, I should not be able to bypass security, encoding, and storage processors.

Dridi:
I fixed the testGunzip bug in trunk (5384df74abd77a54d03727e847e8610c7af068a8) and 4.1 (https://github.com/varnishcache/varnish-cache/commit/94a7f427c21d8bd0a91541b29141471c8cab220a) branches. I'll be happy to revisit this pull request once we reach a decision for ESI and processors in general.

hermunn:
About back porting this: The fix by @Dridi is already in the 4.1 branch, and as far as I can see, 08bf3c8 is not necessary for backporting to 4.1. Yell if I am wrong.

PR #2555, opened 2018-02-04
New way of calculating TTLs in VCL.

hermunn:
This PR supersedes #2548 and #2547 and is compatible with 71b93143d from #2519.

After looking closely at the available timestamps in varnish, I came to the conclusion that we need a new variable in `struct ctx` as a base for ttl calculations. I have named it `ttl_now`, but maybe `now_for_ttl`, `now_ttl` or `now_for_ttl_calculations` would be better. The reason is that the current use of `now` is problematic for reasons explained in #2548 and #2547 and demonstrated by the test cases.

If this PR is accepted, then, in this repository, only the shard director will use `now`. There are probably external vmods that also use it, and some of these will probably have to switch to `ttl_now` to work correctly.

As far as I can see it, the main discussion point for this PR will be what `obj.ttl` should be in `sub vcl_deliver`. For consistency maybe it should maybe be calculated from `t_req` or `t_restart` instead of `now`, but it is unclear. I hope the test cases covers the cases that we need to discuss.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -510,6 +510,7 @@ struct req {
 	double			t_first;	/* First timestamp logged */
 	double			t_prev;		/* Previous timestamp logged */
 	double			t_req;		/* Headers complete */
+	double			t_restart;	/* Time of last restart */
 
 	struct http_conn	htc[1];
 	struct vfp_ctx		*vfc;
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -167,6 +167,7 @@ ved_include(struct req *preq, const char *src, const char *host,
 
 	req->req_step = R_STP_RECV;
 	req->t_req = preq->t_req;
+	req->t_restart = preq->t_restart;
 	assert(isnan(req->t_first));
 	assert(isnan(req->t_prev));
 
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -590,13 +590,12 @@ hsh_rush2(struct worker *wrk, struct rush *r)
  */
 
 unsigned
-HSH_Purge(struct worker *wrk, struct objhead *oh, double ttl, double grace,
-double keep)
+HSH_Purge(struct worker *wrk, struct objhead *oh, double ttl_now, double ttl,
+double grace, double keep)
 {
 	struct objcore *oc, **ocp;
 	unsigned spc, ospc, nobj, n, n_tot = 0;
 	int more = 0;
-	double now;
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
@@ -623,7 +622,6 @@ double keep)
 		ocp = (void*)wrk->aws->f;
 		Lck_Lock(&oh->mtx);
 		assert(oh->refcnt > 0);
-		now = VTIM_real();
 		VTAILQ_FOREACH(oc, &oh->objcs, hsh_list) {
 			CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
 			assert(oc->objhead == oh);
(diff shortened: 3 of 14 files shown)


hermunn:
After talking to @mbgrydeland I have written 5a2c2c10097 on top of these two patches, and it is up for discussion. The commit partly reverts 34be747 by removing the `t_restart` variable, which might be desirable in itself, but can only be done if restart should not affect `ctx->now` in VCL.

hermunn:
So, why did I conclude that a new variable in `struct vrt_ctx` is the way to go? First of all, we need a consistent way of calculating `ttl` for some meaning of *consistent*.

In `struct vrt_ctx` the only time variable is `now`, which cannot be used for TTLs due to its nature - it will often change between VCL `sub`s. We need to keep `now` for uses that are not TTL calculations, and thus the choice is between introducing a new variable in `struct vrt_ctx` or to always use a value from a request or a busy object to calculate TTLs. The second option gives us less flexibility of what *now* for TTLs should be, and, in my opinion, harder code to read and write.

hermunn:
After a long discussion on IRC I am summing up what we agree on, and what is still open. Feel free to comment on any of the bullet points:

* We agree that `r02519.vtc` is correct - when an object expires after the request was received but before lookup AND the object was found (due to *keep* or a lagging expiry thread), then we want VLC to see this as a *hit*. This is the center assumption of this PR, and it fixes a known problem uncovered after  #2519 was discussed and tested.
* We agree that `t_restart` is unnecessary and that `t_req` should be the basis for TTL on the client side. For this reason I needed to change a `>=` to a `>` in `HSH_Lookup`. I have squashed the original commits with togheter with 5a2c2c1, and the result is a commit that also changes the purge mechanism.
* There was some talk about calculating TTL relative to when hashing is completed, but it did not seem to catch on. Please comment if you want to argue for this, but remember that restarts come into the picture here.
* We agree that `c00086.vtc`, now renamed to `s00007.vtc` in 30ee9da9, is correct.
* There is disagreement on `c000087.vtc`, now renamed to `s00008.vtc` in 30ee9da9, where we either expect approximately 10s or 12s as `obj.ttl` in `sub vcl_deliver`. I can see good arguments for both, but right now I am leaning to 12s. Case for 12s: It is up to the RFC code to create headers that accurately describes the expiry of the object, and this data should stay in the object. The role of `obj.ttl` on the client side is to keep the TTL relative to `t_req`, with no exceptions. Case for 10s: Switching to 12s is a change of behavior. Also, if the client is another varnish, you might put `obj.ttl` in a header. There it should be a high resolution time to live that is as accurate as possible, so then it is natural to make it relative to delivery of the object.
* The test case `c000088.vtc`, now renamed to `s00009.vtc` in 30ee9da9, was not discussed. It is very closely related to the previous test case, and can probably be removed.
* @bsdphk would prefer to avoid adding `ttl_now`, but I find it difficult to do in a clean way, if at all. I will listen to suggestions, of course, but maybe we can postpone this to after the PR is in?

What I would like now, is for the following to happen *soon*:
* We agree on the correct interpretation of `ttl_now` for the next March 15th release.
* I update this PR to reflect what we have agreed.
* The PR gets a final review and is OKed.
* I update #2519 and get it OKed.

When all of this is done we can close #1799 with a documentation on how to use keep to deliver old objects when a backend is marked as unhealthy (do `return (deliver)` in `sub vcl_hit`).

And I can tell you what I want, what I really really want: To close #1799.

bsdphk:
This was discussed at bugwash today and agreement apart from some finer points of Age header was reached.

Conclusion is that we will always have either a ctx->req or ctx->bereq to provide the relevant timestamp for TTL purposes, so we do not need a ttl_now field.

hermunn:
I put another commit on top of the original one, which simply removes the `ttl_now` variable, and introduces a static function that calculates it.

In the post *bugwash* discussion consensus was reached: The interpretation of `obj.ttl` in `sub vcl_recv` stays as it is, and it is consistent with the `age` calculation in the same `sub`.

I see no reason not to merge this now, but I give you a final opportunity to nit pick.

Oh, a final question: When I rebase to current master, should I squash my to commits? I do not have a strong opinion, but if no one chimes in, I will squash. I will also rename the *regression* test case to r02555.vtc.

bsdphk:
OK

hermunn:
Squashed and wrote a better commit message. Will push after verifying that the tests pass.

hermunn:
This was back ported to 4.1 as a02e4f277c3ff12c.

PR #2558, opened 2018-02-06
Accurate byte counters

mbgrydeland:
There was a regression from Varnish 4.0 to 4.1, where the response
bytes was counted as the number of bytes fed to the outgoing write
vector, rather than the bytes that was actually handed off to the OS'
socket buffer. This would cause for many cases the complete object
size counted as transmitted bytes, even though the client hung up the
connection early.

This patch changes the counters to show the amount of bytes sent as
reported from the write() system calls rather than the bytes we planned
and prepared to send. The counters will include any protocol overhead (ie
chunked encoding in HTTP/1 and the frame headers in HTTP/2).

ESI subrequests will as before in their log transactions report the number
of bytes it (and any subrequests below it) contributed to the total body
bytes produced.

Some test cases have been adjusted to account for the new counter behaviour.

One new test case added that attempts to catch this regression.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -233,7 +233,8 @@ vbe_dir_gethdrs(const struct director *d, struct worker *wrk,
 		if (vtp->state != VTP_STATE_STOLEN)
 			extrachance = 0;
 
-		i = V1F_SendReq(wrk, bo, &bo->acct.bereq_hdrbytes, 0);
+		i = V1F_SendReq(wrk, bo, &bo->acct.bereq_hdrbytes,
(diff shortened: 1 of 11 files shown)


mbgrydeland:
Yes, this supersedes #2536 . 

I see that my enterprising test case for this failed on travis. It seemed to work fine on my computer, but I was worried that the test case for this would be brittle. We might want to drop the test case, or come up with some better way of testing the interrupted writes.

bsdphk:
Ditch the test-case and commit, testing partial writes to buffers is not realistic.

mbgrydeland:
Merged without the test-case in 51176640603b123a38b49ae271cb9c9de9df1d28

PR #2561, opened 2018-02-08
Support proxy-protocol-v2 TLV

ehocdet:
I have some bits to support proxy-protocol-v2 TLV as described in  https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt.

TLV like crc32c for checksum does not required storing, others like ALPN do.

Patches series include support of CRC32c, ALPN and SSL version.
. alpn can be used in VCL via `proxy.alpn`
. ssl version can be used in VCL via `proxy.ssl`

More TLV can be easily added but in increased of `uint16_t    sattr[SA_LAST]` in `struct sess`

What do you think?


bsdphk:
This ties into the Unix-domai-sockets in oblique ways.

The answer to the patch is no, because I do not want the VCL built-in variables to sprout in all sorts of directions this way.

The way to do this is to add vmod_proxy in the tree, which uses private APIs to talk to cache_proxy_proto.c to extract the necessary information.

This private API should go in a new "noinst" include file varnishd/proxy/cache_proxy.h

PR #2564, opened 2018-02-15
Support Unix domain sockets

slimhazard:
This supersedes PR #2371, in light of @bsdphk's decisions in his mail to varnish-dev on Feb 13th 2018. See #2371 and the mail for details.

I'll be adding commits to this branch/PR incrementally as we go along.

### DIFF
--- a/bin/varnishtest/tests/a00002.vtc
+++ b/bin/varnishtest/tests/a00002.vtc
@@ -21,3 +21,24 @@ client c1 -connect ${s1_sock} {
 client c1 -run
 
 server s1 -wait
+
+# The same tests with unix domain sockets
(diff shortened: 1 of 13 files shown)


bsdphk:
Looks good.

I wonder if it makes sense `server` also gets a shortand for `-listen "${tmpdir}/${instance}.sock"` ?

OK

slimhazard:
Rebased the PR/feature branch to master and pushed dcce66a to master. Keeping the PR open for ongoing commits (and removed the OK label so that new commits can be OK'd again).

slimhazard:
@bsdphk, a shorthand like `server -listen_uds` should be easy enough, but I'd prefer to get back to that after we've gone through the harder parts.

PR #2571, opened 2018-02-16
Support Unix domain sockets for -a listen addresses

slimhazard:
Sorry about the multiple PR, github apparently closes a PR once it's been merged, and then it can't be re-opened.

So, this once again supersedes PR #2371, in light of @bsdphk's decisions in his mail to varnish-dev on Feb 13th 2018, and now follows #2564.

### DIFF
--- a/lib/libvarnish/vsa.c
+++ b/lib/libvarnish/vsa.c
@@ -36,7 +36,9 @@
 #include <string.h>
 #include <stdlib.h>
 #include <sys/socket.h>
+#include <sys/types.h>
 #include <netinet/in.h>
+#include <netdb.h>
 
 #include "vdef.h"
 #include "vas.h"
@@ -173,6 +175,27 @@ struct suckaddr {
 };
 
 const int vsa_suckaddr_len = sizeof(struct suckaddr);
+
+/*
+ * Bogus IPv4 address 0.0.0.0:0 to be used for VCL *.ip variables when the
+ * "real" address is not IP (such as UDS addresses).
+ */
+static struct suckaddr bogo_ip_vsa;
+const struct suckaddr *bogo_ip = &bogo_ip_vsa;
+
+void
+VSA_Init()
+{
+	struct addrinfo hints, *res = NULL;
+
+	memset(&hints, 0, sizeof(hints));
+	hints.ai_family = AF_INET;
+	hints.ai_flags = AI_NUMERICHOST | AI_NUMERICSERV;
+	AZ(getaddrinfo("0.0.0.0", "0", &hints, &res));
+	AN(VSA_Build(&bogo_ip_vsa, res->ai_addr, res->ai_addrlen));
+	assert(VSA_Sane(bogo_ip));
+	freeaddrinfo(res);
+}
 
 /*
  * This VRT interface is for the VCC generated ACL code, which needs
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -99,6 +99,7 @@ struct poolparam;
 struct sess;
 struct transport;
 struct worker;
+struct listen_sock;
 
 #define DIGEST_LEN		32
 
@@ -571,6 +572,7 @@ struct sess {
 #define SESS_MAGIC		0x2c2f9c5a
 
 	uint16_t		sattr[SA_LAST];
+	struct listen_sock	*listen_sock;
 	int			refcnt;
 	int			fd;
 	uint32_t		vxid;
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -673,8 +673,12 @@ ccf_listen_address(struct cli *cli, const char * const *av, void *priv)
 
 	AZ(pthread_mutex_lock(&shut_mtx));
 	VTAILQ_FOREACH(ls, &heritage.socks, list) {
-		VTCP_myname(ls->sock, h, sizeof h, p, sizeof p);
-		VCLI_Out(cli, "%s %s\n", h, p);
+		if (!ls->uds) {
+			VTCP_myname(ls->sock, h, sizeof h, p, sizeof p);
+			VCLI_Out(cli, "%s %s\n", h, p);
+		}
+		else
+			VCLI_Out(cli, "%s -\n", ls->endpoint);
 	}
 	AZ(pthread_mutex_unlock(&shut_mtx));
 }
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -37,6 +37,8 @@
 #include <stdlib.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
+#include <sys/socket.h>
+#include <sys/un.h>
 
 #include "cache_varnishd.h"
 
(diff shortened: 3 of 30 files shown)


slimhazard:
Rationale for not changing X-Forwarded-For when the request was received via UDS: the obvious choices that I can think of -- the UDS path, the bogo-IP 0.0.0.0, or "localhost" -- are either inaccurate or potentially insecure. 0.0.0.0 is inaccurate (deliberately bogus); "localhost" implies TCP loopback, which is also inaccurate; and passing along the UDS path risks exposing info about the filesystem layout to backends that may or may not be trustworthy.

Neither of haproxy or nginx update X-F-F automatically the way Varnish does. In both cases, X-F-F is changed if you have a config rule to do so, and it isn't if you don't, and that goes for both IP and UDS connections. So I don't think there are any precedents to consider here.

I think it's best not to update X-F-F implicitly, document that fact, and advise users to change X-F-F in VCL if they prefer.

slimhazard:
Rationale for ReqStart output with UDS (`/path -`): we could have the endpoint name at the first position (-a arg name, by default `a0`, etc). But that position is used for the varnishncsa `%h` formatter ("remote host"), and an endpoint name likely doesn't make sense as a remote host.

On the other hand, a UDS path doesn't designate a "remote host" either. `-` as the output for varnishncsa `%h` might make more sense in this case, but we'd have to figure out how to make varnishncsa smart enough to recognize the situation.

bsdphk:
The static initialization of bogo_ip fails to init the sockaddr length field on platforms which have it.

We have avoided this problem everywhere else by never constructing a sockaddr, leaving that to library functions which presumably know what the local platform wants.

That implies runtime initialization, but we can do that during master-process startup, so its no big deal.

bsdphk:
I'm not fond of the many `if (*ls->endpoint == '/')` tests all over the place.

Make the detemination one place, and put a flag in struct listen_sock which is tested everywhere else.

bsdphk:
The test in v00053 doesn't actually depend on the fields being IP typed, it wouldn't notice if any of them were STRING.

I would probably have written it to match an ACL, that way the type mattered.

bsdphk:
I don't agree with logging the endpoint in ReqStart, its a long constant string which conveys no information.  I would have logged listen_sock->name instead, it's shorter and users can put useful information in it.

bsdphk:
I don't approve of the change to for instance client.identity and ip_tos

The entire point about the bogo_ip is to *not* force the users to special-case everything in their VCL.

If you use UDS without PROXY, then client.identity is 0.0.0.0 - live with it, and setting ip_tos on a UDS should just be ignored.

bsdphk:
... Likewise about not setting sess-attr CLIENT_IP:  It should be set for UDS, to bogo_ip, so that we don't need to special-case everything everywhere.

Panic code for instance should log IP+PORT+ENDPOINT for all connections.

slimhazard:
> The static initialization of bogo_ip fails to init the sockaddr length field on platforms which have it.
> That implies runtime initialization, but we can do that during master-process startup, so its no big deal.

Ok, then add a VSA_Init() to be called by mgt_main()? The child process uses it as well, wouldn't we need to hand it off via heritage?

> I'm not fond of the many if (*ls->endpoint == '/') tests all over the place.
> Make the detemination one place, and put a flag in struct listen_sock which is tested everywhere else.

ACK

> The test in v00053 doesn't actually depend on the fields being IP typed, it wouldn't notice if any of them were STRING.
> I would probably have written it to match an ACL, that way the type mattered.

c00087.vtc matches the bogo-IPs against ACLs, looks like it does what you have in mind.

I'd still be in favor of having a test that explicitly demonstrates getting 0.0.0.0 for the *.ip variables. This will be something we'll want to emphasize in the docs, and I know of productive VCL that does just that (assign a *.ip to a header), and suspect that it's very widespread.

> I don't agree with logging the endpoint in ReqStart, its a long constant string which conveys no information. I would have logged listen_sock->name instead, it's shorter and users can put useful information in it.

ACK, was considering the same thing.

I've been concerned that something meaningful needs to go into ReqStart and BackendStart, because varnishncsa reads them for the `%h` formatter. A symbolic name could probably go there just as well, but we better be sure to document that, for people who might be expecting varnishncsa to produce something parseable as a network address.

While we're holding back on the backends, I'd like to point out that BackendStart currently doesn't work, comes out as "Conversion failed", since it calls VTCP_myname() which assumes VSAs. I don't see anything in V1F_SendReq(), where it's logged, with which we could navigate back to a name, not even the backend name.

And also about the backends -- I also have BackendOpen logging the path, but that could just be '-' (the backend name is already in the log entry).

> I don't approve of the change to for instance client.identity and ip_tos

ACK on ip_tos doing nothing for a UDS. Then I'd at least log a message with VCL_Error.

client.identity worries me much more, because it's commonly used for the hash director, and someone could find the hash director going drastically wrong, if all at once it directs all traffic to one backend and leaves all of the other backends idle. Because of that, I thought that we'd be helping users a lot more by failing early and often.

We could add a VCL_Error message in that case, saying that "this is probably not distinguishing client identities the way you thought it would", and hope that people think to read the log.

But this is case where I see potential for "OMG you broke my site" if we *don't* make it fail.

> ... Likewise about not setting sess-attr CLIENT_IP: It should be set for UDS, to bogo_ip, so that we don't need to special-case everything everywhere.

Well, CLIENT_IP and CLIENT_PORT are string attributes. But sure, we can put "0.0.0.0" and "0" in there. CLIENT_ADDR is being set to bogo-IP.

> Panic code for instance should log IP+PORT+ENDPOINT for all connections.

ACK

slimhazard:
Edited the commits and force-pushed to address @bsdphk's comments:
- added `VSA_Init()`, called by `mgt_main`, to build the bogo_ip VSA via `getaddrinfo()` and `VSA_Build()`
- added flag `uds` to `struct listen_sock`, so as not to check `*ls->endpoint == '/'` all the time
- session string attributes `CLIENT_IP` and `CLIENT_PORT` filled with "0.0.0.0" and "0" for UDS -a addresses
- `std.set_ip_tos()` is silently ignored for a UDS
- `ReqStart` emits "0.0.0.0 0" if the listen address is UDS
- `client.identity` is "0.0.0.0" if unset and the listen address is UDS
- panic output for the client address is: `<addr> <port> <endpoint name>` in all cases, where `<addr> <port>` should be "0.0.0.0 0" for UDS

We should think about the logs -- now that `ReqStart` emits "0.0.0.0" for any UDS listener, then unless I'm mistaken we have no client VSL records that distinguish the listener endpoint. So if there's more than one UDS listener, there's currently no way to tell them apart by reading varnishlog or with a log reader like varnishncsa.

slimhazard:
About UDS backends: I believe that it won't be possible to implement them in a VMOD while re-using backend core code unless the TCP pools can manage UDS addresses in addition to IPv4 and IPv6. Since the backend and probe code makes use of the TCP pools, we'd end up having to re-implement all of the backend code in the VMOD.

What I think is the minimum necessary change for the backend core code to make it re-usable for a UDS backend VMOD:
- tcp pools manage UDS addresses as well as IPv4 and IPv6
- `struct vrt_backend` has the `path` field
- the backend code just hands off the `vrt_backend.path` to the TCP pool

The dynamic UDS backends from vmod_debug in commit 53a4122 do just that, so we could make use of what's already here.

Still ...

If UDS backends are moved to a VMOD, then we couldn't have `-b /path` on the command line or `-vcl+backend` in vtc scripts to refer to a UDS backend, unless we're willing to have the code generated for those two functions import the VMOD.

So all in all, I'm still in favor of UDS backends in native VCL. But if we decide to go with a UDS VMOD, I believe it can be done very easily with the core changes described above.

lib/libvarnish/vsa.c
```diff
@@ -174,6 +176,25 @@ struct suckaddr {
...
+static struct suckaddr bogo_ip_vsa;
+const struct suckaddr *bogo_ip = &bogo_ip_vsa;
+
+void
+VSA_Init()
+{
+	struct addrinfo hints, *res = NULL;
+
+	memset(&hints, 0, sizeof(hints));
+	hints.ai_family = AF_INET;
```
Dridi: Add `AI_NUMERICHOST` and `AI_NUMERICSERV` flags?

slimhazard: Yes, thought about that after the commit, we should have it.

slimhazard:
Another force push, only changes the first commit:
- build the bogo-IP with `AI_NUMERICHOST | AI_NUMERICSERV`, as pointed out by @Dridi 
- include `<sys/socket.h>` in `mgt_main.c`, required for `vsa.h` (tripped up the Travis "sanitizer" build)

slimhazard:
Yet another force-push:
- `SessOpen` has `0.0.0.0 0` at the local and remote IP/port positions for UDS listeners.
- X-F-F is not a special case -- `0.0.0.0` is added if the listener is UDS and `client.ip` has not been set differently by PROXY.
- VTC-test-only commits have been squashed into the "big commit"
- Commits about UDS backends have been moved into #2579

bsdphk:
There's a few nits of stylistic order, but we can polish that later.

Thanks for your patience.

OK!

slimhazard:
Merged to master and pushed, PR for UDS backend support is #2579.

PR #2577, opened 2018-02-21
Symbolic status for HTC FetchError

nigoroll:
Right now, we output a user-unfriendly integer for fetch errors caused by HTC.
This patch moves the enum definition to a table file which also holds a short and long description.
The long description is not currently used but intended for auto-generated documentation (tbc if the PR gets accepted)


nigoroll:
rebased

nigoroll:
rebased

nigoroll:
rebased

nigoroll:
rebased, force-pushed

PR #2585, opened 2018-02-26
Add the listener endpoint name as a third field to ReqStart.

slimhazard:
Makes it possible to distinguish listeners in a VSL client transaction. Especially if there is more than one UDS listen address, since all of them are logged as the bogo-IP 0.0.0.0:0 where addresses are logged.

We have the endpoint name in SessOpen, but that is only seen with `-g session`, not with any of `-g vxid`, `-g request` or `-c`. `-g session` is not possible for varnishncsa, and rarely used for any of the other log readers.


slimhazard:
Closed via ddc624d

PR #2588, opened 2018-02-27
Add Obj* records to the default vsl_mask

Dridi:
They add up to a considerable chunk of the VSL traffic, especially when cookies are involved, while providing little value when troubleshooting. In an effort to reduce VSL cycles on heavy workloads I found these to be good candidates for several reasons:

- `obj.*` fields are immutable in VCL
- after a hit, we initially get the same `Resp*` records in `vcl_deliver`
- before `vcl_backend_response` returns, we can infer them from `Beresp*`

The main difference would be `ObjHeader:X-Varnish`, but it would be the same as `BerespHeader:X-Varnish`, and be part of the `RespHeader` counterpart in the case of a hit. That, and cache-related headers adjusted to the current time.

One case where they wouldn't appear is for example on a hit followed by a restart. I have yet to find a use case where I'd need to query them with `varnishlog` or some other VUT.


hermunn:
Backport review: This will not be back ported to 4.1 as it changes the defaults. The workaround is to disable these records during runtime.

Issue #2551, opened 2018-01-30
Partial availability of listen sockets yields user unfriendly error message.


lkarsten:
Issue: On Varnish Cache 5.0.0 on Ubuntu, when attempting to use "-a :80", and another process (here: apache2) is bound to [::]:80, Varnish asserts on startup.

Expected: an error message that Varnish was unable to bind to the interfaces specified.

```
Jan 30 14:39:16 hostname varnishd[6576]: Manager got SIGINT
Jan 30 14:39:16 hostname varnishd[6576]: Debug: Stopping Child
Jan 30 14:39:16 hostname varnishd[6576]: Stopping Child
Jan 30 14:39:17 hostname varnishd[6576]: Error: Child (9998) died signal=6 (core dumped)
Jan 30 14:39:17 hostname varnishd[6576]: Child (9998) died signal=6 (core dumped)
Jan 30 14:39:17 hostname varnishd[6576]: Error: Child (9998) Panic at: Tue, 30 Jan 2018 13:39:17 GMT
Jan 30 14:39:17 hostname varnishd[6576]: "Assert error in vca_acct(), cache/cache_acceptor.c line 507:
Jan 30 14:39:17 hostname varnishd[6576]:   Condition(ls->sock > 0) not true.
Jan 30 14:39:17 hostname varnishd[6576]: thread = (cache-acceptor)
Jan 30 14:39:17 hostname varnishd[6576]: version = varnish-5.0.0 revision 99d036f
Jan 30 14:39:17 hostname varnishd[6576]: ident = Linux,4.13.0-32-generic,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
Jan 30 14:39:17 hostname varnishd[6576]: Backtrace:
Jan 30 14:39:17 hostname varnishd[6576]:   0x55bfa6a796df: varnishd(+0x426df) [0x55bfa6a796df]
Jan 30 14:39:17 hostname varnishd[6576]:   0x55bfa6a54962: varnishd(+0x1d962) [0x55bfa6a54962]
Jan 30 14:39:17 hostname varnishd[6576]:   0x7fadd567f7fc: libpthread.so.0(+0x77fc) [0x7fadd567f7fc]
Jan 30 14:39:17 hostname varnishd[6576]:   0x7fadd53acb5f: libc.so.6(clone+0x3f) [0x7fadd53acb5f]
Jan 30 14:39:17 hostname varnishd[6576]: "
[...]

PR #2590, opened 2018-02-28
Changes to better report child socket initialization errors

mbgrydeland:
This patch set enables the child process to report down to management (over CLI) if it fails to do its part of the socket initialization. Errors are reported over the CLI. If the child is auto-started on management startup, the daemonize code will notice and report error exit on failures.

The tcp_fastopen and cache_filter parameters will now cause error messages and fail if they can't be applied (instead of silently ignored).

Failures to execute listen() will cause failure (fixing #2551).

### DIFF
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -631,7 +631,12 @@ ccf_start(struct cli *cli, const char * const *av, void *priv)
 			    ls->endpoint, strerror(errno));
 			return;
 		}
-		AZ(listen(ls->sock, cache_param->listen_depth));
+		if (listen(ls->sock, cache_param->listen_depth)) {
+			VCLI_SetResult(cli, CLIS_CANT);
+			VCLI_Out(cli, "Listen failed on socket '%s': %s",
+			    ls->endpoint, strerror(errno));
+			return;
+		}
 		vca_tcp_opt_set(ls, 1);
 		if (cache_param->accept_filter && VTCP_filter_http(ls->sock)) {
 			VCLI_SetResult(cli, CLIS_CANT);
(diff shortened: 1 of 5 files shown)


bsdphk:
I'm not sure failing hard on accept-filters is a good idea, we make no attempt to find out if they are actually loaded in the kernel or not?

mbgrydeland:
But they are not enabled by default are they? The varnish runtime parameter that is? Isn't OK to fail it hard if the user tries to enable it and the setting fails for whatever reason?

mbgrydeland:
Oh, I see now that accept-filters are indeed on by default. Then it would be bad to suddenly fail hard on not being able to enable it. Maybe we should have it print a warning instead? Or change and document the default of accept-filters?

slimhazard:
We're relying on the current treatment of accept_filters and tcp_open (if they fail, log an Error in VSL but proceed anyway) as a way of minimizing the impact of introducing UDS -- "you won't have to change that, those two params just won't do anything if you listen at UDS". There's already documentation to that effect in "Upgrading to 6.0".

I've verified that both of them fail on FreeBSD when applied to a UDS, and I think we can safely assume that they will fail on any platform for which they're available. I think it would be better if Varnish doesn't stop working if someone wants to use -a UDS, if they have a start script that turns on one or both params -- the problem is not likely to be immediately obvious.

> Maybe we should have it print a warning instead?

We emit a VSL message with tag Error on vxid 0 if either one of them fails (so you need -g raw), but I believe that's not documented anywhere (I did mention it in the "Upgrading" addition). We could add that fact to the params docs.

mbgrydeland:
Attempting to set options known to not work seems a bit strange to me, maybe the code should test the proto of the socket against the intended protocols before running setting the options? Seems a small impact.

But anyways the error reporting part on those options (tcp_fastopen and accept_filters) are not important to me. I just noticed that the other code changes allowed for easy error reporting also on those. I'm fine with having them mostly silently ignore failures as before.

My goal is simply to be able to report and fail properly when the child fails to call the listen() on the sockets.

bsdphk:
OK with footnotes as agreed in bugwash

mbgrydeland:
This has been merged, without the changing of behaviour if TCP fastopen or kernel filter should fail.

PR #2593, opened 2018-03-01
Document grace and keep, and how to deal with problematic servers

hermunn:
This PR is rather gloomy, as I realize that finding a workaround for #1799 is even harder than I thought.

Unfortunately we cannot close issue #1799 for the reasons described in the documentation. I hope that there is a workaround that I have overlooked, and that the PR can be updated accordingly.

Feedback appreciated.

### DIFF
--- a/doc/sphinx/users-guide/vcl-grace.rst
+++ b/doc/sphinx/users-guide/vcl-grace.rst
@@ -1,11 +1,15 @@
 .. _users-guide-handling_misbehaving_servers:
 
-Misbehaving servers
+Grace mode and keep
 -------------------
 
-A key feature of Varnish is its ability to shield you from misbehaving
-web- and application servers.
+Sometimes you want Varnish to serve content that is somewhat stale
+instead of waiting for a fresh object from the backend. For example,
+if you run a news site, serving a main page that is a few seconds old
+is not a problem if this gives your site faster load times.
 
+In Varnish this is achieved by using `grace mode`. A related idea
+is `keep`, which is also explained here.
 
 Grace mode
 ~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


steven-j-wojcik:
@hermunn Should [line 210](https://github.com/varnishcache/varnish-cache/pull/2593/files#diff-e4e78f2d087eea382cb784fadfeda063R210) be:

> Grace mode allows Varnish **to deliver** slightly stale content to clients while getting a fresh version from the backend. 

instead of: 

> Grace mode allows Varnish **delivered** slightly stale content to clients while getting a fresh version from the backend.

slimhazard:
There's some overlap with the chapters "Cache misses" and "Uncacheable Content" in the User's Guide page "Achieving a high hitrate". But not as much as I thought there might be at first -- "Cache misses" describes `beresp.keep`, as if it's the first introduction, that's the only thing I see that obviously comes up twice.

The point of those chapters is to describe request coalescing, pass, hit-for-miss, hit-for-pass, and how all of those topics are related to conditional requests and the 304 response. I see the overall point of the docs in this PR as something different, so I'm not saying that it's already covered.

But if we do add this part, then maybe we should look for where we could edit the one or the other down a bit. Not necessarily a lot -- I don't think it's always wrong to mention something more than once, if it's not excessively redundant. Sometimes it's not so bad to hear about something twice when you're learning something new.

hermunn:
@steven-j-wojcik You are correct. I will fix this.

@slimhazard The main issue with the old version is that it is wrong. :) I have no strong feelings about what the right way is, but I needed to write something that is correct, and that explains the remaining challenges with the use case.

We can accept this and then fix the redundancy later, or we can reject it and explain everything in one place. My main concern is not disclosing that `return(miss)` in `vcl_hit` in the way that the old version suggested, can cause serious problems.

bsdphk:
To be honest, not quite sure what bugwash concluded.

Doc improvements always welcome of course, and don't need review by me.

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -19,52 +23,194 @@ If you are serving thousands of hits per second the queue of waiting
...
+Keep
+~~~~
+
+Setting an object's `keep` tells Varnish that it should keep an object
+in the cache for some additional time. There are two reasons to do this:
 
-So, in order to serve stale content we must first have some content to
-serve. So to make Varnish keep all objects for 2 minutes beyond their
-TTL use the following VCL::
+* To use the object as a candidate for ``304 NOT MODIFIED`` from the server.
```
nigoroll: more precise:
* To use the object to construct a conditional ``GET`` backend request (with ``If-Modified-Since:`` and/or ``Ìf-None-Match:`` headers), allowing the backend to reply with a ``304 Not Modified`` response, which may be more efficient on the backend and saves re-transmitting the unchanged body.

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -19,52 +23,194 @@ If you are serving thousands of hits per second the queue of waiting
...
+The effect of grace and keep
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For most users setting the default grace and/or a suitable grace for
+each object is enough. The default VCL will do the right thing and
+behave as described above. However, if you want to customize how varnish
+behaves by changing ``sub vcl_hit``, then you should know some of the
+details on how this works. 
+
+When a request is made for a resource where an object is found, but TTL
```
nigoroll: better maybe?
When encountering expired objects upon a cache hit, Varnish considers the following:

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -19,52 +23,194 @@ If you are serving thousands of hits per second the queue of waiting
...
+wish to use when the backend is sick, for example::
+
+  sub vcl_backend_response {
+       set beresp.grace = 24h;
+       // no keep
+  }
+
+Then you can use the following code as your ``sub vcl_hit``::
+
+   if (std.healthy(req.backend_hint)) {
```
nigoroll: general note: we should have VTCs for all the snippets in the docs to make sure that we don't break them.

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -19,52 +23,194 @@ If you are serving thousands of hits per second the queue of waiting
...
+
+This method has one potentially serious problem when more than one
+client asks for an object that has expired its TTL. If the second of
+these requests arrives after the effective grace, but before the first
+request has completed, then the second request will be turned into a
+`pass`.
+
+In practice this method works well in most cases, but if you
+experience excessive `pass` behavior, this translates to a reduced the
+hit rate and higher load on the backend. When this happens you will
```
nigoroll: s:the::

doc/sphinx/users-guide/vcl-grace.rst
hermunn: I wrote something longer in the last commit (just pushed). I hope it is clear.

doc/sphinx/users-guide/vcl-grace.rst
hermunn: Excellent, thanks.

doc/sphinx/users-guide/vcl-grace.rst
hermunn: Thanks.

PR #2595, opened 2018-03-02
Fix "TypeError: err() takes at most 2 arguments (4 given)"

simonvik:
Err takes two arguments, 4 was sent.

```
TypeError: err() takes at most 2 arguments (4 given)
```


bsdphk:
Already fixed as part of other commit

Issue #2468, opened 2017-10-23
Incompatible VMOD - hashes do not match


paulborile:
Hi,
we release a varnish module in binary form and we do make them for almost all varnish versions in order to not have ABI mismatch problems. Our procedure builds the module .so by compiling it against the released versions that we download from git with git checkout <release_hash>. Release hashes are found at the releases link in github (https://github.com/varnishcache/varnish-cache/releases) and this as always worked fine (only one strange hash in 5.0 time ago) up to 5.2.0. We get this message while loading our binary module, built with 5.2.0 sources, commit 4c4875c , against 5.2.0 binary version we download from : https://packagecloud.io/varnishcache/varnish5/packages/ubuntu/xenial/varnish_5.2.0-1~xenial_amd64.deb/download.deb

```
Incompatible VMOD wurfl
	File name: /usr/lib64/varnish/vmods/libvmod_wurfl.so
	ABI mismatch, expected <Varnish 5.2.0 4c4875cbf>, got <Varnish 5.2.0 4c4875c>
```
Looks like the git hash short SHA is long 9 characters instead of 7 you used so far. Is this correct ?

Dridi:
My comment from varnishcache/pkg-varnish-cache#91:

> It looks like it depends on your git version :(
>
> I noticed that short hashes were longer when I upgraded git to 2.13 but didn't pay attention to it.
>
> What you should do instead is to install the `varnish-dev` package from packagecloud to make sure to build against the same ABI anyway. Technically, this is "working" because you build against sources but run against packages, different targets... But of course we don't want to consider them different if it's the same commit.
>
> In `generate.py` we should probably replace `git show -s --pretty=format:%h` with `git rev-parse --short=7` to get a consistent short hash. And we should probably add a check to the release process from now on. To me, this is a problem in varnish-cache, not the packaging.

Dridi:
We probably want this in 4.1 too, please advise :)

hermunn:
Dridi, if you think this is not breaking people's setups, then I agree. I have not looked at this closely, so if you think it is safe to port to 4.1, please do.

Dridi:
Added to 4.1 too.

paulborile:
Are we sure that we will be able to build binary packages with the new generate.py that will ABI match with old releases of varnish ? If I understand correctly you are bringing the untruncated hashes to 4.1 : assuming only for versions > 4.1.8, correct ? Same applies to 5.1 ?

[...]

PR #2598, opened 2018-03-06
Run include/generate.py only when needed

### LINKED ISSUE #2597: 'vcs_version.h' file not found
```
$ wget https://github.com/varnishcache/varnish-cache/archive/master.tar.gz
[...]
$ tar zxf master.tar.gz
$ cd varnish-cache-master/
$ ./autogen.des
[...]
$ make -s
Making all in include
  GEN      vcs_version.h
  CC       vbm_test.o
  CCLD     vbm_test
  GEN      vcs_version.h                  <= supposedly generated here
Making all in lib
Making all in libvarnish
  CC       libvarnish_a-binary_heap.o
  CC       libvarnish_a-vas.o
  CC       libvarnish_a-vav.o
  CC       libvarnish_a-vcli_proto.o
  CC       libvarnish_a-vcli_serve.o
  CC       libvarnish_a-vct.o
  CC       libvarnish_a-version.o
version.c:37:10: fatal error: 'vcs_version.h' file not found
#include "vcs_version.h"
         ^~~~~~~~~~~~~~~
1 error generated.
make[3]: *** [Makefile:859: libvarnish_a-version.o] Error 1
make[2]: *** [Makefile:419: all-recursive] Error 1
make[1]: *** [Makefile:559: all-recursive] Error 1
make: *** [Makefile:445: all] Error 2
$ find . -name vcs_version.h
```

Dridi:
In order to do that reliably, it needs to be figured out at configure time. We should also not assume that .git is a directory, it can also be a regular file.

When REFRESH_VCS is true, the logic to see whether the Git HEAD changed moved to the Makefile and could be removed from the python script. We may still want to keep it on both sides to stay extra careful.

We need a .PHONY target to always check the Git HEAD but on the other hand we don't want generate.py to also always run. Otherwise we'd always see this in the output even when it doesn't really happen:

    GEN     vcs_version.h

To solve that, we make generate.py depend on the .PHONY target and just touch it when the Git HEAD changed. This forces us to add generate.py as a source for vcs_version.h, which is actually the case.

When building outside of a Git clone, this is either a dist/release archive and it already contains a vcs_version.h that won't be updated, or it contains none and generate.py will make a NOGIT one.

Finally, instead of trying to be smart about the NOGIT case, it's a trivial fallback to the git rev-parse command. As a side effect, a full SHA1 hash is always used to describe the commit and aligns with the decision made in #2468 for VMODs metadata.

Fixes #2597

### DIFF
--- a/include/generate.py
+++ b/include/generate.py
@@ -50,21 +50,17 @@ def file_header(fo):
 	fo.write("""/*
  * NB:  This file is machine generated, DO NOT EDIT!
  *
- * Edit and run lib/libvcc/generate.py instead.
+ * Edit and run include/generate.py instead.
  */
 
 """)
 
 #######################################################################
 
-if os.path.isdir(os.path.join(srcroot, ".git")):
-	v = subprocess.check_output([
-		"git --git-dir=" + os.path.join(srcroot, ".git") +
-		" show -s --pretty=format:%H"
-	], shell=True, universal_newlines=True)
-	v = v.strip()
-else:
-	v = "NOGIT"
+
+v = subprocess.check_output([
+    "git rev-parse HEAD 2>/dev/null || echo NOGIT"
+    ], shell=True, universal_newlines=True).strip()
 
 vcsfn = os.path.join(srcroot, "include", "vcs_version.h")
 
--- a/configure.ac
+++ b/configure.ac
@@ -729,6 +729,9 @@ AC_SUBST(VSC_GEN_H)
 # Make sure this include dir exists
 AC_CONFIG_COMMANDS([mkdir], [$MKDIR_P doc/sphinx/include])
 
+# In order to reliably generate vcs_version.h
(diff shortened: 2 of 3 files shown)


Dridi:
Works on my machine, didn't go the extra mile to try it on other systems. Please test.

include/generate.py
```diff
@@ -50,21 +50,17 @@ def file_header(fo):
...
-		"git --git-dir=" + os.path.join(srcroot, ".git") +
-		" show -s --pretty=format:%H"
-	], shell=True, universal_newlines=True)
-	v = v.strip()
-else:
-	v = "NOGIT"
+
+v = subprocess.check_output([
+    "git rev-parse HEAD 2>/dev/null || echo NOGIT"
+    ], shell=True, universal_newlines=True).strip()
```
nigoroll: please merge your simplification of `generate.py` :)

PR #2599, opened 2018-03-06
Log cacheability

gquintard:
Ham-fisted attempt at logging cacheability


bsdphk:
Apart from wondering if "pass" is the right thing to log for "uncacheable" I'm OK with this.

lkarsten:
Is it wise to add another meaning of "pass" in Varnish?

cacheable | uncacheable? too long?

gquintard:
(un)cacheable seemed a bit long, but I have no problem with changing it.

nigoroll:
I had a hard time with the `uncacheable` attribute in VCL, I would have preferred if we only used `pass` instead. But I understand that `uncacheable` combines all the different cases (`pass`, `hitpass`, `hitmiss`, make uncacheable) so I think that, as we once decided to go with `uncacheable`, the right output choce would be `uncacheable` and `cacheable`

bin/varnishd/cache/cache_rfc2616.c
```diff
@@ -198,9 +198,10 @@ RFC2616_Ttl(struct busyobj *bo, double now, double *t_origin,
 	}
 
 	VSLb(bo->vsl, SLT_TTL,
-	    "RFC %.0f %.0f %.0f %.0f %.0f %.0f %.0f %u",
+	    "RFC %.0f %.0f %.0f %.0f %.0f %.0f %.0f %u %s",
 	    *ttl, *grace, *keep, now,
-	    *t_origin, h_date, h_expires, max_age);
+	    *t_origin, h_date, h_expires, max_age,
+	    bo->uncacheable ? "pass" : "cache");
```
nigoroll: `bo->uncacheable ? "uncacheable" : "cacheable");`

gquintard:
Will commit with cacheable/uncacheable on Thursday if no one yells before then

gquintard:
closed via 9e174d4c1d38b752e029d25c28127f9eeace39c9

PR #2602, opened 2018-03-07
Implement VCL vars local.endpoint and .socket (VCL >= 4.1)

slimhazard:


slimhazard:
The result at the end of `v00025.vtc` came as a surprise: 
- there are two `-a` addresses
- a client instigates a backend fetch over the one address, then another fetch over the other one

The second time through, `local.endpoint` and `.socket` as accessed on the backend side are not the same as when they are accessed on the client side, in the same transaction that brought about the fetch.

On the second fetch:
- Client-side `local.endpoint` and `.socket` correspond to the Varnish `-a` address to which the client connected.
- On the backend side, they correspond to the `-a` address used to instigate the _first_ fetch.
- My expectation was that they would be identical to the corresponding client-side values in every case.

I take this as meaning that the busyobj used in the subsequent fetch does not necessarily point to the same session as `req->sp` on the client side for the same client/backend transaction.

If this is too confusing for the common VCL author (and I think it might be), then maybe the two new variables should only be readable on the client side after all. If you want to carry those values into the backend side, then assign them to a `req.http.*` header and retrieve them from `bereq.http.*`.

We'd lose symmetry with the `*.ip` variables (which are readable from both the client and backend side). But "fixing" this is probably more trouble than it's worth, and if we leave it this way, then explaining why it can happen probably doesn't make it any less confusing.

Dridi:
I wouldn't worry about confusion and simply be consistent with `local.ip` since it's already readable from both sides.

slimhazard:
@Dridi well to put it more bluntly, to me it feels like this is _wrong_.

A use case for `local.socket` and `local.endpoint` on the backend side is access control -- only allow connections to backend B if the client connected over endpoint A. This wouldn't work for that, it might even be a security risk.

Think about how we point this out in documentation. "They might not be what you expect on the backend side, because a busy object might be re-used that ...". How do we explain it without confronting users with internal details that would leave them understanding even less than they did before?

Seems to me that this is addressed much more easily by advising users to send the values over bereq headers.

Dridi:
If you're using this for access control then `vcl_recv` is the place. I don't really care about `local.*` being accessed on the backend side, I'm merely pointing out that we already have `local.ip` from which we can reuse the same semantics.

If you feel so strongly against the new variables being readable on the client side maybe `vcl 4.1` should retire `local.ip` as well? And the new variables would only be available on the client side? Isn't there an overlap between `ip` and `socket` ?

Either way, if 1) we want to make them readable from both sides and 2) `local.ip` remains in `vcl 4.1` then they should have the same semantics on the backend side.

slimhazard:
@Dridi I'm fine if @bsdphk wants to make a decision, or if we take a vote at bugwash.

slimhazard:
closed via b38c3d32bc060c213e380ca553dfa927822e863d and 29c0c5711049ac2d4cde84ad0eecabd1cbc25a3e

PR #2604, opened 2018-03-09
Add the $Synopsis directive to vmod.vcc sources.

slimhazard:
Vaild values are 'auto' or 'man', default 'auto'. With 'auto' you
get the auto-generated SYNOPSIS in the VMOD docs. With 'man', the
SYNOPSIS is just skipped, so the VMOD author has to write it.

### DIFF
--- a/lib/libvcc/vmodtool.py
+++ b/lib/libvcc/vmodtool.py
@@ -602,15 +602,16 @@ class s_module(stanza):
         fo.write("\n")
         fo.write(":Manual section: " + self.vcc.mansection + "\n")
 
-        fo.write("\n")
-        write_rst_hdr(fo, "SYNOPSIS", "=")
-        fo.write("\n")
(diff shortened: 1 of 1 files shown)


Dridi:
Nitpick: please pick `auto` and `manual` as I find `man` confusing (because `vmodool` generates both `.rst` and `.man.rst` files). Or if anything I would even go for `enabled` and `disabled` with the former being the default.

slimhazard:
@Dridi done (manual)

Dridi:
After some more thought, you may want the synopsis in order to generate HTML docs but not in your manual.

```
$Synopsis all|doc|man|none
```

I would definitely keep a ToC for sphinx builds and not have the generated synopsis for manual pages.

slimhazard:
closed via 8337f6ffc66c68c0b5f8265eb58e36112d06a507

PR #2609, opened 2018-03-14
Fix memory leak of vary string on stevedore alloc fail

mbgrydeland:
If the stevedore failed the object creation, we would leak the temporary
VSB holding the computed vary string. This patch frees it.

Problem exists in 4.1 and later.


hermunn:
Back ported as 26d5261500473.

PR #2611, opened 2018-03-15
Go back to trunk after release of 6.0.0

hermunn:
This is a quite trivial commit, and it is a PR so that we can discuss it.

The question is: How long until we push things that might be good for a 6.0.1 release? I am hoping that we can have some time where we do only bug fixes. My suggestion then becomes that merging this commit marks the point where we start doing changes that are not bug fixes (or other similar fixes).


slimhazard:
OK'd at Apr 2nd bugwash

hermunn:
Pushed as c78555b62eb200634282f.

Issue #2573, opened 2018-02-19
varnish 4.1 vcc accepts $thing.http.$thing.http.$Header


nigoroll:
This test does not trigger an error as it should (ie the -errvcl does not succeed)

```
diff --git a/bin/varnishtest/tests/v00019.vtc b/bin/varnishtest/tests/v00019.vtc
index c4bd725..dd08480 100644
--- a/bin/varnishtest/tests/v00019.vtc
+++ b/bin/varnishtest/tests/v00019.vtc
@@ -63,3 +63,10 @@ varnish v1 -errvcl {Comparison of different types: STRING '==' INT} {
                }
        }
 }
+
+varnish v1 -errvcl {foo} {
+       backend b { .host = "127.0.0.1"; }
+       sub vcl_recv {
+               set req.http.req.http.foo = "bar";
+       }
+}

```

nigoroll:
FTR: 4.1 has no prio for me, so I'd be happy if someone else looked after the fix if it is at all required

gquintard:
Indeed, and I know of at least one client that will get burned when upgrading to 6.

Dridi:
Regarding `resp.http.Cache-Control[max-age]` I like the idea but we need clear semantics since we can't really express the difference between `field=value` or just `field`.

This for example would feel like magic to me:

```
set resp.http.Cache-Control[public] = BOOL; # or just set/unset without a value?
```

And if we take that from the client side instead, the one we don't trust, how do we deal with duplicates? For example multiple entries with the same name/value in an `Accept*` header.

I would rather write this by the way:

```
set resp.http[Cache-Control].max-age = 30s;
```

Using double quotes would be consistent with VSL where the prefix may contain spaces:

```
Timestamp:"my custom timestamp"
```

Again, not a big fan of double quotes for header names.

fgsch:
We could use `:`, e.g.:

`set resp.http.Cache-Control:public`

ftr, this is how Fastly does it.

nigoroll:
does fastly handle the additional characters?

Dridi:
I don't like the syntax, but I'd rather use `:` and bridge the gap. That's not incompatible with square brackets or double quotes if we want to support the full range of header names.

fgsch:
@nigoroll, please elaborate on additional characters.

`if (req.http.foo:bar) { ... }` will be true if field is present, regardless of its value (or lack of).

slimhazard:
[...]

PR #2632, opened 2018-03-29
Return 400 for invalid header names (or lack of)

fgsch:
Related to #2573


fgsch:
OBE: Only headers lacking the name will return 400. Invalid names will be handled elsewhere.

PR #2633, opened 2018-03-29
For HTTP/1.1 requests, Host is mandatory

### LINKED ISSUE #2631: Enforce Host requirement for HTTP/1.1 requests
We currently allow HTTP/1.1 requests without the Host header, however this is mandatory.

As per RFC 7230:

```
   A client MUST send a Host header field in all HTTP/1.1 request messages.
```

Also:

```
   A client MUST send a Host header field in an HTTP/1.1 request even if
   the request-target is in the absolute-form, since this allows the
   Host information to be forwarded through ancient HTTP/1.0 proxies
   that might not have implemented Host.
```

And finally:

```
   A server MUST respond with a 400 (Bad Request) status code to any
   HTTP/1.1 request message that lacks a Host header field and to any
   request message that contains more than one Host header field or a
   Host header field with an invalid field-value.
```

Adding this ticket to discuss it during bugwash.

fgsch:
Fixes #2631

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -36,8 +36,14 @@ vcl 4.0;
 
 sub vcl_recv {
     if (req.method == "PRI") {
-	/* This will never happen in properly formed traffic (see: RFC7540) */
-	return (synth(405));
+        /* This will never happen in properly formed traffic (see: RFC7540) */
+        return (synth(405));
+    }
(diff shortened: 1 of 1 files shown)


nigoroll:
lgtm

bsdphk:
OK

fgsch:
Merged in ff86ca7e1eb8bee3c34c7cf5be5e352780add1d3.

PR #2634, opened 2018-04-02
Use libumem only as a storage allocator, not to override malloc etc.

slimhazard:
We do this by not linking with libumem, but rather using dlopen/dlsym
to load the slab allocator interface, which is used by the stevedore.

### DIFF
--- a/bin/varnishd/Makefile.am
+++ b/bin/varnishd/Makefile.am
@@ -169,7 +169,7 @@ varnishd_LDADD = \
 	@SAN_LDFLAGS@ \
 	@JEMALLOC_LDADD@ \
 	@PCRE_LIBS@ \
-	${DL_LIBS} ${PTHREAD_LIBS} ${NET_LIBS} ${RT_LIBS} ${LIBM} ${UMEM_LIBS}
+	${DL_LIBS} ${PTHREAD_LIBS} ${NET_LIBS} ${RT_LIBS} ${LIBM}
 
 noinst_PROGRAMS = vhp_gen_hufdec
 vhp_gen_hufdec_SOURCES = hpack/vhp_gen_hufdec.c
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -136,7 +136,7 @@ usage(void)
 
 	printf(FMT, "-s [name=]kind[,options]", "Storage specification");
 	printf(FMT, "", "Can be specified multiple times.");
-#ifdef HAVE_LIBUMEM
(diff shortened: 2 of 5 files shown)


slimhazard:
Background: when libumem is linked, it overrides `malloc(3)` and friends, so umem takes on _all_ memory allocation.

@nigoroll  has said that he has wanted to make this change for a while. We also suspect that it was one of the reasons why VMOD unix was failing in the `getpeerucred(3)` case. `getpeerucred(3)` implicitly allocates an opaque `ucred_t` object for you, which then has to be freed with `ucred_free(3)`. In the tests of VMOD unix, we saw crashes with a stack that included `umem_free(3)`, which is the umem substitute for `free(3)`.

@nigoroll, I've put this up for your review, but I still don't think I'm getting it right -- it passes `make distcheck`, but on the joyent host I still see `libumem.so.1` in `ldd varnishd`, which makes me think that libumem is still there overriding the rest of memory allocation. Maybe you can check if you see that as well, and if so maybe you know how we can definitively prevent the link to libumem.

Dridi:
Travis CI is choking on `python-docutils`'s absence, probably a consequence of switching to `make distcheck`.

slimhazard:
@Dridi, yes, there's apparently a bug in yesterday's modification to `travis.yml`:
```
The command "if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then make -j3 distcheck; fi" exited with 2.
```
In: https://travis-ci.org/varnishcache/varnish-cache/jobs/361338847

That's OK, this is the first PR that's putting the new Travis run to the test. %^)

slimhazard:
Rebased to get the `travis.yml` fix and force-pushed, that worked.

nigoroll:
the ld dependency originated from libpcre, libedit and libncurses. I've rebuilt all on varnishdev-il.ham1.v0.uplex.de and now ldd shows no dependency on libumem.
Will test in production.

nigoroll:
Ah no, not yet. I see failing tests, checking

nigoroll:
I am having trouble getting ahead on this path for now, so please hold

nigoroll:
@Dridi jemalloc and umem are xor for all practical purposes

nigoroll:
meged.
* @slimhazard s commit: 2c4e481ed1e7ddfc8cc04233cea53b0942340ab2
* additional details:
** 11a50e7944fe3ae69e649204b9b6683b92d2e463
** e241de074fd3b47437bba0ddb8449fb755102f51
** 929dd59c27383ee344f6b8654afa2c496aadd5e3

PR #2635, opened 2018-04-03
add debug bit to count contended locks

nigoroll:
As the only way to find relevant contended locks is to get statistics from production systems, we should have a runtime switch to enable gathering of more statistics.

### DIFF
--- a/bin/varnishd/VSC_lck.vsc
+++ b/bin/varnishd/VSC_lck.vsc
@@ -30,5 +30,15 @@
 	:oneliner:	Lock Operations
 
 
+.. varnish_vsc:: busy
+	:type:	counter
+	:level:	debug
+	:oneliner:	Contended locks
+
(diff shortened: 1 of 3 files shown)


nigoroll:
rebased due to travis error caused outside this patch

bsdphk:
We actually had something like this earlier and it wasn't used much, and tended to skew things when used, becuase congestion on contested locks increased as a result.

That said I'm not against it the idea, but I think you are counting the wrong thing, because this way you also involve "legitimate" TryLock (if any for that Lck) in your stats.

I would increment the new counter in Lck__Lock() when TryLock fails and I would give it a suitable cryptic name (ie: not "contested") so people don't get the wrong idea when the debugging is not enabled.  Maybe name it "try_fail" ?

nigoroll:
will prep a new patch

Issue #2622, opened 2018-03-20
proxy backend -> EADDRNOTAVAIL -> probes failing -> all backends down -> site down


nigoroll:
I've been bitten badly by a series of issues which I think we need to address. Each of them may not be that bad in isolation but I think the whole picture is relevant for understanding their relevance.

Seen on ee1d34dea8ecb5576d0f2740e807ad4d9e67b3c6 Linux 3.14.53+ Debian 8.10 - note that the connection pool code has changed since then but I do not see semantic differences

### setup

* cache clustering/sharing is used in a two-varnish setup: requests which are not identified as passes are forwarded to the other varnish based on the backend selection of the shard director.
* The other varnish is configured as a backend with `proxy_header`
* The rate of miss+hitmiss is ~1k/s, so the expected rate of connections to the other varnish is ~.5k/s
* 1M max open files
* <100k client connections

### symptoms

* connections to "the other varnish" are starting to fail (`FetchError backend ...: fail`)
* backend probes are starting to fail, until almost all backends are down (`FetchError no backend`)
* **site goes down**

### lack of visibility

analyzing this issue was complicated by the facts that
* `FetchError` does not give information about _why_ the connection failed
* probes do not log  _why_ a connection failed

### assured facts

* For proxy backend requests, the connection is closed after a single request: https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_backend.c#L184 and will end up in `TIME_WAIT` because it is closed by the client
* Thus, with default Linux tcp tuning (connections staying in `TIME_WAIT` for 120s and `ip_local_port_range` ~28k), connections will fail when a request rate of ~235/s is exceeded

### open question

I suspect a causal relation between the failing proxy connections and the fact that backend probes failed, but I cannot prove it:

The hypothesis is that failing backend requests use up file descriptors, for example via a delayed close() in the pool code, but I failed to come up with a plausible explanation.

So I am still looking for either support of the hypothesis or an alternative hypothesis which would explain why backend probes are failing in the situation documented under _assured facts_

### TODOs

* backend proxy connections are problematic as-are due to the `TIME_WAIT` issue and, in particular, future scenarios like connecting to an ssl-onloader require connection reuse/pooling.
* our error messages should point to the exact root-cause and should include the relevant `errno`

Dridi:
> The hypothesis is that failing backend requests use up file descriptors, for example via a delayed close() in the pool code, but I failed to come up with a plausible explanation.

[...]

PR #2636, opened 2018-04-03
sess_fail_* and vcp stats, holddown for vcp

nigoroll:
This should address all the issues documented in #2622 
* visibility for accept and vcp errors
* holddown for vcp errors

### DIFF
--- a/bin/varnishd/VSC_vcp.vsc
+++ b/bin/varnishd/VSC_vcp.vsc
@@ -1,0 +1,120 @@
+..
+	This is *NOT* a RST file but the syntax has been chosen so
+	that it may become an RST file at some later date.
+
+.. varnish_vsc_begin::	vcp
+	:oneliner:	Connection Pool Counters
+	:order:		65
+
+
+	Counter for operations on connection pools.
+
+.. varnish_vsc:: n_conn
+	:type:	counter
+	:level:	diag
+	:oneliner:	Number of pooled connections
+
+
+.. varnish_vsc:: n_kill
+	:type:	counter
+	:level:	diag
+	:oneliner:	Number of connections still to be closed
+
+
+.. varnish_vsc:: n_used
+	:type:	counter
+	:level:	diag
+	:oneliner:	Number of open connections
+
+
+.. varnish_vsc:: opened
+	:type:	counter
+	:level:	diag
+	:oneliner:	Connections opened
+
+	Connections opened successfully
+
+	This counter may be slightly inaccurate for performance reasons.
+
+
+.. varnish_vsc:: closed
+	:type:	counter
+	:level:	diag
+	:oneliner:	Connections closed
+
+	Connections closed
+
+
+.. varnish_vsc:: recycle
+	:type:	counter
+	:level:	diag
+	:oneliner:	Connections recycled
+
+	Connections kept open (cached) for later reuse
+
+
+.. varnish_vsc:: recycle_fail
+	:type:	counter
+	:level:	diag
+	:oneliner:	Connection recycles (Wait_Enter) failed
+
+
+.. varnish_vsc:: helddown
+	:type:	counter
+	:level:	diag
+	:oneliner:	Connection not attempted due to holddown
+
+	See parameters backend_local_error_holddown and
+	backend_remote_error_holddown.
+
+
+.. varnish_vsc:: fail
+	:type:	counter
+	:level:	diag
+	:oneliner:	Connections failed
+
+
+	Counter of failed opens. Detailed reasons are given in the
+	fail_* counters and in Debug VSL.
+
+	This counter is the sum of all detailled fail_* counters.
+
+	All fail_* counters may be slightly inaccurate for efficiency.
+
+.. varnish_vsc:: fail_eacces
+	:type:	counter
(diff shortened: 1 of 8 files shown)


bin/varnishd/cache/cache_tcp_pool.c
```diff
@@ -240,8 +247,26 @@ VCP_New(struct conn_pool *cp, const void *id, void *priv,
...
+	else {
+		if (! tp->ip4 || (cache_param->prefer_ipv6 && tp->ip6))
+			addr = tp->ip6;
+		else
+			addr = tp->ip4;
+
+		AN(addr);
+		VTCP_name(addr, abuf, sizeof abuf, pbuf, sizeof pbuf);
+		cp->stats = VSC_vcp_New(NULL, &cp->vsc, "%p.%s:%s", id,
+					abuf, pbuf);
```
Dridi: Until Varnish 4.0 backend counters used to include the inet4/inet6/port tuple in the name, for example:

```
VBE.default(127.0.0.1,::1,8080).*
```

I think we should inherit that behavior for the new counters.

bin/varnishd/VSC_vcp.vsc
```diff
@@ -0,0 +1,120 @@
...
+	:oneliner:	Connection Pool Counters
+	:order:		65
+
+
+	Counter for operations on connection pools.
+
+.. varnish_vsc:: n_conn
+	:type:	counter
+	:level:	diag
+	:oneliner:	Number of cached connections
```
Dridi: s/cache/pooled/ ?

bin/varnishd/cache/cache_tcp_pool.c
```diff
@@ -347,27 +350,52 @@ VCP_Open(struct conn_pool *cp, double tmo, const void **privp)
 
 	CHECK_OBJ_NOTNULL(cp, CONN_POOL_MAGIC);
 
+	while (cp->holddown > 0) {
+		Lck_Lock(&cp->mtx);
+		if (cp->holddown == 0) {
```
Dridi: How can it be `== 0` if you need `> 0` to enter the loop?

Should it be this instead? `while (cp->holddown >= 0) {`

nigoroll:
@Dridi  thank you for your feedback, see last two added commits

bin/varnishd/cache/cache_tcp_pool.c
```diff
@@ -279,32 +307,121 @@ VCP_Rel(struct conn_pool *cp)
...
+VCP_Open(struct conn_pool *cp, double tmo, const void **privp)
+{
+	int r;
+	double h;
+
+	CHECK_OBJ_NOTNULL(cp, CONN_POOL_MAGIC);
+
+	while (cp->holddown > 0) {
+		Lck_Lock(&cp->mtx);
+		if (cp->holddown == 0) {
```
Dridi: This `if` statement is still inconsistent with the `while` above.

nigoroll: github had shown me your initial comment on this as outdated, so I thought you had taken it back.

So what is happening here is that holddown can be modified concurrently to a thread in open. We want to avoid taking the lock each time, so we accept to miss a holddown in favor of somehow less locking overhead. But for the reverse case that the holddown has been cleared concurrently, we want to ensure that we actually see it cleared.

We we check unprotected first but then, under the lock, we make sure that holddown has not been cleared concurrently.

nigoroll: BTW, the main motivation for the lock is to get a consistent view on holddown and holddown_errno. While we could juggle membars, a simple mutex is easier and safe and with the above optimization the whole thing has close to no overhead for the base case (no connection errors).

Dridi: I cleaned my glasses and finally was able to see that glaring lock in-between. Thank you for your patience :)

nigoroll:
rebased & force pushed

nigoroll:
the first patch `sess_fail_*` was accepted by phk and is now merged as 3f4cd9297d53b1e18e594c83d9fd068d4dd4a666

I've squashed and rebased the other patches, leaving us with three commits.

To summarize today's irc discussion:
* I have implemented the backend connection errors on the VCP layer for these reasons:
  * this is where the connection `open()` happens
  * we already had other statistics about pools which we did not expose yet could be useful
  * connections are shared across backends and vcls
* yet an argument can be made to expose the open error counters on the backend (VBE) layer, per backend (with semantics _number of errors this backend experienced when it happend not to be able to reuse a recycled connection_), as this avoids involving users in another layer of indirection.
  * if we went this route, some counters would remain on the pool side which I think should still be exposed
* Another option we discussed are _VSC symlinks_ to link backends to pools (won't work when we start using different VCP keys)

@bsdphk please make a decision and I will be happy to change the patches accordingly

bsdphk:
Decision:  Counters go with other backen counters.

The postulated clarity of having them on the pool is too speculative to warrant the complexity required at this point in time.

nigoroll:
I will update the PR

nigoroll:
Continued in #2699

PR #2640, opened 2018-04-07
Include headers to compile on musl

gquintard:


bsdphk:
Can I get you to annotate these #includes with "// for MUSL" so that automatic scans don't flag them for removal later ?

OK with that change.

Dridi:
Aren't automatic scans misled by platform-specific includes? I often notice with glibc that you can get away with less includes than what the manual says.

gquintard:
rebased, merging.

PR #2642, opened 2018-04-09
add debug bit to count contended locks

nigoroll:
Follow-Up to #2635: separate stats for the lock and trylock case

I cannot quite follow phk's obscurity argument as this will only trigger additional questions. So instead, make clear in the name and documentation that the counters are only significant if the `lck` debug bit is set.

### DIFF
--- a/bin/varnishd/VSC_lck.vsc
+++ b/bin/varnishd/VSC_lck.vsc
@@ -30,5 +30,27 @@
 	:oneliner:	Lock Operations
 
 
+.. varnish_vsc:: dbg_busy
+	:type:	counter
+	:level:	debug
+	:oneliner:	Contended lock operations
+
+	If the ``lck`` debug bit is set: Lock operations which
+	returned EBUSY on the first locking attempt.
+
+	If the ``lck`` debug bit is unset, this counter will never be
+	incremented even if lock operations are contended.
+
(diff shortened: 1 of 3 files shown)


nigoroll:
rebased, force-pushed

bsdphk:
We should check if the pthread_mutex implementations have undocumented/non-standard stats available ?

nigoroll:
> We should check if the pthread_mutex implementations have undocumented/non-standard stats available ?

I've checked glibc/linux and can't find any.

Refs:

* https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_mutex_lock.c;h=29cc143e6cbf2421e853267c2a26c0f0e6918198;hb=HEAD#l312

* https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/lowlevellock-futex.h;h=fc834ed16e15b831d388cc4519a64fbd9995a211;hb=HEAD#l71

It can be done with systemtap, but that's probably not that practical for identifying single contented varnish logs: https://sourceware.org/systemtap/SystemTap_Beginners_Guide/futexcontentionsect.html

nigoroll:
backporters: this also needs f7d1c8bf39f348c6b214760b9bcf46877a3b5cb1 - sorry

PR #2644, opened 2018-04-10
vcc: vmod function named argument clarifications

nigoroll:
minor cleanup and additional assertions to vcc code generation for vmod function arguments:

* We remove setting function argument names to the empty string such that there are only two cases left: either the function argument name is NULL or points to a proper name
* We assert that only named arguments can be possibly be speficied multiple times at this place in the code
* We assert that only named arguments can be optional
* We add handling for positional arguments being missing

### DIFF
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -524,7 +524,6 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 		if (!memcmp(vvp->value, "PRIV_", 5)) {
 			fa->result = vcc_priv_arg(tl, vvp->value,
 			    sym->name, sym->vmod);
-			fa->name = "";
 			continue;
 		}
 		fa->type = VCC_Type(vvp->value);
@@ -579,6 +578,7 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 			return;
 		}
 		if (fa->result != NULL) {
+			AN(fa->name);
 			VSB_printf(tl->sb, "Argument '%s' already used\n",
 			    fa->name);
 			vcc_ErrWhere(tl, tl->t);
@@ -602,6 +602,7 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 	VTAILQ_FOREACH_SAFE(fa, &head, list, fa2) {
 		n++;
 		if (fa->optional) {
+			AN(fa->name);
 			bprintf(ssa, "\v1.valid_%s = %d,\n",
 				fa->name, fa->avail);
 			e1 = vcc_expr_edit(tl, e1->fmt, ssa, e1, NULL);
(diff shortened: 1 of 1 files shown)


lib/libvcc/vcc_expr.c
```diff
@@ -497,7 +497,6 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 		if (!memcmp(vvp->value, "PRIV_", 5)) {
 			fa->result = vcc_priv_arg(tl, vvp->value,
 			    sym->name, sym->vmod);
-			fa->name = "";
```
nigoroll: * either the function argument is NULL or properly named, we remove the case where it also can be the empty string in order to avoid additional tests and ambiguity

lib/libvcc/vcc_expr.c
```diff
@@ -552,6 +551,7 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 			return;
 		}
 		if (fa->result != NULL) {
+			AN(fa->name);
```
nigoroll: * only named arguments can be possibly be speficied multiple times

lib/libvcc/vcc_expr.c
```diff
@@ -575,6 +575,7 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 	VTAILQ_FOREACH_SAFE(fa, &head, list, fa2) {
 		n++;
 		if (fa->optional) {
+			AN(fa->name);
```
nigoroll: * only named arguments can be optional

lib/libvcc/vcc_expr.c
```diff
@@ -584,7 +585,7 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 		if (fa->result == NULL && fa->val != NULL)
 			fa->result = vcc_mk_expr(fa->type, "%s", fa->val);
 		if (fa->result != NULL && sa != NULL) {
-			if (fa->name && *fa->name != '\0')
+			if (fa->name)
```
nigoroll: see above, names can't be the empty string

lib/libvcc/vcc_expr.c
```diff
@@ -593,8 +594,11 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 			e1 = vcc_expr_edit(tl, e1->fmt, "\v1,\n\v2",
 			    e1, fa->result);
 		} else if (!fa->optional) {
-			VSB_printf(tl->sb, "Argument '%s' missing\n",
-			    fa->name);
+			if (fa->name)
+				VSB_printf(tl->sb, "Argument '%s' missing\n",
+					   fa->name);
+			else
+				VSB_printf(tl->sb, "Argument %d missing\n", n);
```
nigoroll: positional arguments could be missing (I think?!)

bsdphk:
Looks good.

PR #2650, opened 2018-04-13
VUT apps can be stopped cleanly via vtc process -stop.

### LINKED ISSUE #2649: assert failure stopping a VUT app via vtc process -stop
This came up while I was trying to construct a test case for #2646.

When a VUT app is started with `process -start` (I found this with varnishncsa), and then stopped with `process -stop`, this assertion failure is invoked:
```
Assert error in vut_sighandler(), varnishncsa.c line 1127:
  Condition((vut) != NULL) not true.
```
It's happening because `VUT_Fini()` starts with `TAKE_OBJ_NOTNULL`, which NULLs the `vut` object, and the signal handler gets the object after that.

Apparently the only way we run varnishncsa in the current tests is with the `shell` command, and then it can be stopped with `shell "kill"`, apparently without any problems:
```
varnishtest "assert error stopping varnishncsa via vtc process -stop"

varnish v1 -vcl { backend b { .host="${bad_ip}"; } } -start

# No problem starting varnishncsa -D via the shell command, but then
# we can't use the process command to check the error status, stderr
# output, etc.
shell {
	varnishncsa -n ${v1_name} -D -P ${tmpdir}/ncsa.pid -w ${tmpdir}/ncsa.log
}

delay 1

# Killing the process with the shell command doesn't seem to cause any
# errors.
shell "kill -TERM `cat ${tmpdir}/ncsa.pid`"

# Now we use process -start, to use some of its checks in a test.
process p1 {
	varnishncsa -n ${v1_name} -P ${tmpdir}/ncsa.pid -w ${tmpdir}/ncsa.log
} -start

delay 1

# This results in assertion failure and coredump:
# Assert error in vut_sighandler(), varnishncsa.c line 1127:
#   Condition((vut) != NULL) not true.
#   errno = 4 (Interrupted system call)
process p1 -stop
```
But for the bugfix I need to use `process` for the checks it provides, which can't be done when it's started as a demon with `shell`.

I'll have a PR to fix this shortly, and another PR based on the first one as a fix for #2646.

slimhazard:
Don't NULL the vut handle before a signal handler needs it.

Fixes #2649

### DIFF
--- a/bin/varnishtest/tests/r02649.vtc
+++ b/bin/varnishtest/tests/r02649.vtc
@@ -1,0 +1,14 @@
+varnishtest "Cleanly stop a VUT app via vtc process -stop"
+
+varnish v1 -vcl { backend b { .host="${bad_ip}"; } } -start
+
+process p1 {
+	varnishncsa -n ${v1_name} -P ${tmpdir}/ncsa.pid -w ${tmpdir}/ncsa.log
+} -start
+
+delay 1
+
+process p1 -expect-exit 0 -stop -wait
+
+# Expect empty stderr output
+shell -match {^0\b} "wc -c ${tmpdir}/p1/stderr"
--- a/lib/libvarnishapi/vut.c
+++ b/lib/libvarnishapi/vut.c
@@ -315,7 +315,8 @@ VUT_Fini(struct VUT **vutp)
 {
 	struct VUT *vut;
 
-	TAKE_OBJ_NOTNULL(vut, vutp, VUT_MAGIC);
+	vut = *vutp;
+	CHECK_OBJ_NOTNULL(vut, VUT_MAGIC);
 	AN(vut->progname);
 
 	free(vut->n_arg);
@@ -336,6 +337,7 @@ VUT_Fini(struct VUT **vutp)
 
 	memset(vut, 0, sizeof *vut);
 	FREE_OBJ(vut);
+	*vutp = NULL;
 }
 
 int


Dridi:
Or maybe we should simply not assert that the VUT is not NULL since the signal handler can be shared by multiple VUTs in one process, and that a signal may land during/after the fini operation.

https://github.com/varnishcache/varnish-cache/blob/5d026a310b04e6a4a51cadbf69479e7011e98002/bin/varnishncsa/varnishncsa.c#L1124-L1129

We should probably `CHECK_OBJ_ORNULL` instead since we have no control over when this function is called.

slimhazard:
@Dridi if the VUT can be shared then we should do that; glad I asked you. But then `VUT_Signaled()` also has `CHECK_OBJ_NOTNULL`, and uses some of its fields -- bail out of that one if VUT==NULL?

I still think we should NULL the object as late as possible, since there's evidently some need for it for a while. The fix in #2651 needs it at least a few lines further down in `VUT_Fini()`.

Dridi:
I said it's the signal handler that may be shared by multiple VUTs, not VUTs that can be shared. Our VUTs are single-threaded so just checking ORNULL in the signal handler is enough to assert, and then `VUT_Signaled` should only be called when we have a VUT (so ignore the signal if we already called `VUT_Fini` and started tearing things down).

Now that you mention it, the only reason why you need a VUT in #2651 is because you want to use `VUT_Error` and `vut->P_Arg`. I didn't catch that when I first looked, I will comment there.

slimhazard:
I was about to change this as @Dridi requested, but on taking a closer look I don't think that just changing the signal handler in varnishncsa is the right solution:
```
varnish-cache/bin$ git grep -B1 -A4 ^vut_sighandler
varnishhist/varnishhist.c-static void
varnishhist/varnishhist.c:vut_sighandler(int sig)
varnishhist/varnishhist.c-{
varnishhist/varnishhist.c-      VUT_Signaled(vut, sig);
varnishhist/varnishhist.c-}
varnishhist/varnishhist.c-
--
varnishlog/varnishlog.c-static void
varnishlog/varnishlog.c:vut_sighandler(int sig)
varnishlog/varnishlog.c-{
varnishlog/varnishlog.c-        CHECK_OBJ_NOTNULL(vut, VUT_MAGIC);
varnishlog/varnishlog.c-        VUT_Signaled(vut, sig);
varnishlog/varnishlog.c-}
--
varnishncsa/varnishncsa.c-static void
varnishncsa/varnishncsa.c:vut_sighandler(int sig)
varnishncsa/varnishncsa.c-{
varnishncsa/varnishncsa.c-      CHECK_OBJ_NOTNULL(vut, VUT_MAGIC);
varnishncsa/varnishncsa.c-      VUT_Signaled(vut, sig);
varnishncsa/varnishncsa.c-}
--
varnishtop/varnishtop.c-static void
varnishtop/varnishtop.c:vut_sighandler(int sig)
varnishtop/varnishtop.c-{
varnishtop/varnishtop.c-        CHECK_OBJ_NOTNULL(vut, VUT_MAGIC);
varnishtop/varnishtop.c-        VUT_Signaled(vut, sig);
varnishtop/varnishtop.c-}
```
This problem is all over the place, which IMO is a clear sign that it should be fixed in the VUT API and not in the apps that use it.

This feels like a heretic question, but: what exactly is the importance of NULLing the VUT object as early as possible in `VUT_Fini()`? Why, for that matter, is it absolutely necessary to NULL it at all?

One answer to that might be the problem we're looking at here -- the race between `VUT_Fini()` and the signal handlers. Maybe `struct VUT` should have a state field, to distinguish between "running" and "tearing down". `VUT_Fini()` could set the state "tearing down" (followed by a membar), and the signal handlers need to know that they can't safely run in that state.

@mbgrydeland I think I need your input on this after all.

Issue #2646, opened 2018-04-11
varnishnsca segfaults when shutting down


michbsd:
On FreeBSD, when restarting varnishncsa (via rc.d) it segfaults 

```
root@nsXXXXX:~ # /usr/local/etc/rc.d/varnishncsa restart
Stopping varnishncsa.
Waiting for PIDS: 20855.
Starting varnishncsa.

```
in /var/log/messages:
`Apr 11 11:20:22 nsXXXXX kernel: pid 20855 (varnishncsa), uid 430: exited on signal 6 (core dumped)`

Here's the backtrace : 

```
#0  0x00000008015f339a in thr_kill () from /lib/libc.so.7
[New Thread 802406400 (LWP 100671/<unknown>)]
(gdb) bt
#0  0x00000008015f339a in thr_kill () from /lib/libc.so.7
#1  0x00000008015f3386 in raise () from /lib/libc.so.7
#2  0x00000008015f3309 in abort () from /lib/libc.so.7
#3  0x000000080082e160 in VAS_Fail (func=0x80085e7ba "vut_vpf_remove", file=0x80085e36a "vut.c", line=77, cond=0x80085e7c9 "(VPF_Remove(pfh)) == 0", kind=VAS_ASSERT)
    at ../libvarnish/vas.c:77
#4  0x000000080085360b in vut_vpf_remove () at vut.c:77
#5  0x00000008008537b0 in VUT_Fini (vutp=0x60c280) at vut.c:326
#6  0x00000000004022c8 in main (argc=11, argv=0x7fffffffea28) at varnishncsa.c:1254
(gdb) #
```

* Version used: varnish-6.0.0 revision NOGIT
* Operating System and version: FreeBSD 10.3-REL-p19

slimhazard:
This looks like an assertion failure (`VPF_Remove(pfh) == 0` was not true) and abort, not a segfault.

On Linux at any rate, signal 6 is SIGABRT, not SIGSEGV. @michbsd can you check if SIGABRT==6 on your platform?

`VPF_Remove()` closes and removes the PID file. It fails if `unlink(2)` or `close(2)` fails, or if, before the close is attempted, `fstat(2)` returns a dev and/or ino that are different from dev/ino obtained by `fstat()` just after the file was opened.

@michbsd, I assume you're using varnishncsa with `-P` for a pid file? And is there reason to believe that the pid file was changed or became inaccessible since the time it was created?

michbsd:
[...]

PR #2651, opened 2018-04-13
VUT apps fail gracefully when removing a PID file fails

slimhazard:
References #2646

### DIFF
--- a/lib/libvarnishapi/vut.c
+++ b/lib/libvarnishapi/vut.c
@@ -315,7 +315,8 @@ VUT_Fini(struct VUT **vutp)
 {
 	struct VUT *vut;
 
-	TAKE_OBJ_NOTNULL(vut, vutp, VUT_MAGIC);
+	vut = *vutp;
+	CHECK_OBJ_NOTNULL(vut, VUT_MAGIC);
 	AN(vut->progname);
 
 	free(vut->n_arg);
@@ -336,6 +337,7 @@ VUT_Fini(struct VUT **vutp)
 
 	memset(vut, 0, sizeof *vut);
 	FREE_OBJ(vut);
+	*vutp = NULL;
 }
 
 int
--- a/lib/libvarnishapi/vut.c
+++ b/lib/libvarnishapi/vut.c
@@ -75,8 +75,12 @@ static void
 vut_vpf_remove(void)
 {
 	if (pfh != NULL) {
-		AZ(VPF_Remove(pfh));
-		pfh = NULL;
+		errno = 0;
+		if (VPF_Remove(pfh) != 0)
+			fprintf(stderr, "Cannot remove pid file: %s\n",
+				strerror(errno));
+		else
+			pfh = NULL;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


slimhazard:
Based on PR #2650, fix for #2649.

lib/libvarnishapi/vut.c
```diff
@@ -336,6 +337,7 @@ VUT_Fini(struct VUT **vutp)
 
 	memset(vut, 0, sizeof *vut);
 	FREE_OBJ(vut);
+	*vutp = NULL;
```
Dridi: NULLing here doesn't close the race with the signal handler. Once it's NULL we may still receive a signal and crash in the `varnishncsa` code (and possibly the other VUTs). We should keep the "take" semantics here and not assert in the signal handler since we have no control over signal delivery in `varnishncsa`'s current state. (We could, but we don't need to.)

lib/libvarnishapi/vut.c
```diff
@@ -75,8 +75,12 @@ static void
...
 {
 	if (pfh != NULL) {
-		AZ(VPF_Remove(pfh));
-		pfh = NULL;
+		errno = 0;
+		if (VPF_Remove(pfh) != 0)
+			fprintf(stderr, "Cannot remove pid file: %s\n",
+				strerror(errno));
+		else
+			pfh = NULL;
```
Dridi: Always set `pfh` to NULL, if we failed to remove it once we will likely not succeed when the function is called again (because of `atexit`).

lib/libvarnishapi/vut.c
```diff
@@ -319,15 +323,18 @@ VUT_Fini(struct VUT **vutp)
 	CHECK_OBJ_NOTNULL(vut, VUT_MAGIC);
 	AN(vut->progname);
 
+	errno = 0;
+	if (pfh != NULL && VPF_Remove(pfh) != 0)
+		VUT_Error(vut, 1, "Cannot remove pid file %s: %s", vut->P_arg,
```
Dridi: `-P` is a global option, it isn't tied to a single VUT.

lib/libvarnishapi/vut.c
```diff
@@ -319,15 +323,18 @@ VUT_Fini(struct VUT **vutp)
...
+	pfh = NULL;
+
 	free(vut->n_arg);
 	free(vut->P_arg);
 	free(vut->q_arg);
 	free(vut->r_arg);
 	free(vut->t_arg);
 
-	vut_vpf_remove();
-	AZ(pfh);
```
Dridi: This code should be restored, and `pfh` should be NULL even if we failed to remove the PID file. At the very least an error message will have been printed already.

lib/libvarnishapi/vut.c
slimhazard: Um, yes it is tied to the VUT object? It's a field in `struct VUT`? And all of the VUT apps in the Varnish distribution except for varnishhist hand off the `-P` arg to `VUT_Arg()`, which sets the field? (varnishhist uses -P for something else, and doesn't have an arg for a PID file.)

slimhazard:
Most of @Dridi's requests for changes here are actually about #2650 -- see the conversation over there.

PR #2669, opened 2018-05-03
allocate dynamic privs on the appropriate workspace

nigoroll:
A show stopper used to be that the top request's workspace
was reserved for delivery during ESI processing, but since
beeaa19cced3fe1ab79381b2b1b7b0b5594cbb18 the path is clear.


nigoroll:
bugwash: run in production, then merge

nigoroll:
will merge after testing with no issues on a pre-live installation with automated unit tests (issuing ~1M real world requests checking for many corner cases). Will also take this into production soon, so the dog will eat his own food.

nigoroll:
merged as de2b431086a13b243dc8b3e71cd8697db1df2c7f

PR #2670, opened 2018-05-03
vmodtool --boilerplate improvements

Dridi:
This is a series of patches that gradually improve the boilerplate generator. Because we bundle out-of-tree VMODs in our Varnish distribution, adding this helps importing VMODs following the `varnish.m4` conventions without tweaks or renames and makes it easier to keep bundled VMODs updated (and less error-prone over time). All that while retaining the benefits of offloading the boilerplate.

With that even vmod-debug can now have its boilerplate file, and it improves readability regarding where VCC stuff ends and VSC setup starts.

With the increase of in-tree VMODs I also opted to rename files to avoid collisions as it's taking more time in my editor to cycle through `*.vcc` files when I want to just open one.


nigoroll:
bugwash
> (13:35:35) phk: Ok, slink reviews, and when both slink&dridi are happy it goes in

Dridi:
@nigoroll I would like to postpone your review some time after our initial planning. Taking care of the libvgz CFLAGS item from the last VDD will help meet your condition that this patch set should only be included once out-of-tree usage becomes possible. I will open a pull request for the CFLAGS stuff and rebase this pull request on top of it.

nigoroll:
@Dridi should we maybe close this one until you had the chance to get back to it?

PR #2673, opened 2018-05-09
VDP API changes

mbgrydeland:
This PR changes the VDP API to enable some optimizations for VDPs that need to act differently when dealing with the very last data to be sent out (e.g. set the last frame marker in H/2, or an on-the-fly compressor adding its trailer data as part of the last data chunk).

The H/2 parts of the optimizations isn't finished yet, @daghf will be looking into that part. There are some comments in the code explaining those parts.

### DIFF
--- a/bin/varnishd/cache/cache_deliver_proc.c
+++ b/bin/varnishd/cache/cache_deliver_proc.c
@@ -49,28 +49,34 @@
  * r > 0:  Stop, breaks out early without error condition
  */
 int
-VDP_bytes(struct req *req, enum vdp_action act, const void *ptr, ssize_t len)
+VDP_bytes(struct req *req, enum vdp_flush flush, const void *ptr, ssize_t len)
 {
-	int retval;
+	enum vdp_status status;
 	struct vdp_entry *vdpe;
 	struct vdp_ctx *vdc;
 
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+
 	vdc = req->vdc;
-	assert(act == VDP_NULL || act == VDP_FLUSH);
-	if (vdc->retval)
-		return (vdc->retval);
+	if (vdc->status != VDP_OK)
+		return (vdc->status);
 	vdpe = vdc->nxt;
 	CHECK_OBJ_NOTNULL(vdpe, VDP_ENTRY_MAGIC);
-	vdc->nxt = VTAILQ_NEXT(vdpe, list);
 
-	assert(act > VDP_NULL || len > 0);
+	/* We should be called only once with the VDP_LAST flag */
+	AZ(vdpe->f_seen_last);
+
 	/* Call the present layer, while pointing to the next layer down */
-	retval = vdpe->vdp->func(req, act, &vdpe->priv, ptr, len);
-	if (retval && (vdc->retval == 0 || retval < vdc->retval))
-		vdc->retval = retval; /* Latch error value */
+	vdc->nxt = VTAILQ_NEXT(vdpe, list);
+	CHECK_OBJ_ORNULL(vdc->nxt, VDP_ENTRY_MAGIC);
+	status = vdpe->vdp->bytes(req, flush, &vdpe->priv, ptr, len);
 	vdc->nxt = vdpe;
-	return (vdc->retval);
+	if (status != VDP_OK && (vdc->status == VDP_OK || status < vdc->status))
+		vdc->status = status; /* Latch error status */
+	if (flush == VDP_LAST)
+		vdpe->f_seen_last = 1;
+
+	return (vdc->status);
 }
 
 int
@@ -83,19 +89,20 @@ VDP_push(struct req *req, const struct vdp *vdp, void *priv, int bottom)
 	vdc = req->vdc;
 	AN(vdp);
 	AN(vdp->name);
-	AN(vdp->func);
+	AN(vdp->bytes);
 
-	if (vdc->retval)
-		return (vdc->retval);
+	if (vdc->status != VDP_OK)
+		return (vdc->status);
 
 	if (DO_DEBUG(DBG_PROCESSORS))
 		VSLb(req->vsl, SLT_Debug, "VDP_push(%s)", vdp->name);
 
 	vdpe = WS_Alloc(req->ws, sizeof *vdpe);
 	if (vdpe == NULL) {
-		AZ(vdc->retval);
-		vdc->retval = -1;
-		return (vdc->retval);
+		VSLb(req->vsl, SLT_Error,
+		    "VDP_push(%s) workspace_client overflow", vdp->name);
+		vdc->status = VDP_ERROR;
+		return (vdc->status);
 	}
 	INIT_OBJ(vdpe, VDP_ENTRY_MAGIC);
 	vdpe->vdp = vdp;
@@ -106,9 +113,9 @@ VDP_push(struct req *req, const struct vdp *vdp, void *priv, int bottom)
 		VTAILQ_INSERT_HEAD(&vdc->vdp, vdpe, list);
 	vdc->nxt = VTAILQ_FIRST(&vdc->vdp);
 
-	AZ(vdc->retval);
-	vdc->retval = vdpe->vdp->func(req, VDP_INIT, &vdpe->priv, NULL, 0);
-	return (vdc->retval);
+	if (vdpe->vdp->init)
+		vdc->status = vdpe->vdp->init(req, &vdpe->priv);
+	return (vdc->status);
 }
 
 void
@@ -121,13 +128,10 @@ VDP_close(struct req *req)
 	vdc = req->vdc;
 	while (!VTAILQ_EMPTY(&vdc->vdp)) {
 		vdpe = VTAILQ_FIRST(&vdc->vdp);
-		if (vdc->retval >= 0)
-			AN(vdpe);
-		if (vdpe != NULL) {
-			CHECK_OBJ_NOTNULL(vdpe, VDP_ENTRY_MAGIC);
-			VTAILQ_REMOVE(&vdc->vdp, vdpe, list);
-			AZ(vdpe->vdp->func(req, VDP_FINI, &vdpe->priv,
-			    NULL, 0));
+		CHECK_OBJ_NOTNULL(vdpe, VDP_ENTRY_MAGIC);
+		VTAILQ_REMOVE(&vdc->vdp, vdpe, list);
+		if (vdpe->vdp->fini) {
+			vdpe->vdp->fini(req, &vdpe->priv);
 			AZ(vdpe->priv);
 		}
 		vdc->nxt = VTAILQ_FIRST(&vdc->vdp);
@@ -139,9 +143,8 @@ VDP_close(struct req *req)
 static int v_matchproto_(objiterate_f)
 vdp_objiterator(void *priv, int flush, int last, const void *ptr, ssize_t len)
 {
-	if (last && len == 0)
-		return (0);	/* XXX: For now do not send empty last
-				 * events down the chain. */
+	if (last)
+		return (VDP_bytes(priv, VDP_LAST, ptr, len));
 	return (VDP_bytes(priv, flush ? VDP_FLUSH : VDP_NULL, ptr, len));
 }
 
@@ -152,9 +155,12 @@ VDP_DeliverObj(struct req *req)
 	int r;
 
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	if (req->vdc->status != VDP_OK)
+		return (req->vdc->status);
 	r = ObjIterate(req->wrk, req->objcore, req, vdp_objiterator,
 	    req->objcore->flags & OC_F_PRIVATE ? 1 : 0);
-	if (r < 0)
-		return (r);
-	return (0);
+	if (r < 0 && req->vdc->status == VDP_OK)
+		/* Streaming fetch error */
+		req->vdc->status = VDP_ERROR_FETCH;
+	return (req->vdc->status);
 }
--- a/bin/varnishd/cache/cache_deliver_proc.c
+++ b/bin/varnishd/cache/cache_deliver_proc.c
@@ -137,8 +137,11 @@ VDP_close(struct req *req)
 /*--------------------------------------------------------------------*/
 
 static int v_matchproto_(objiterate_f)
-vdp_objiterator(void *priv, int flush, const void *ptr, ssize_t len)
+vdp_objiterator(void *priv, int flush, int last, const void *ptr, ssize_t len)
 {
+	if (last && len == 0)
+		return (0);	/* XXX: For now do not send empty last
+				 * events down the chain. */
 	return (VDP_bytes(priv, flush ? VDP_FLUSH : VDP_NULL, ptr, len));
 }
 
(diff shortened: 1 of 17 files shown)


bin/varnishd/cache/cache_deliver_proc.c
```diff
@@ -49,30 +49,56 @@
  * r > 0:  Stop, breaks out early without error condition
  */
 int
```
nigoroll: not just here, but in most of the touched functions: shouldn't the return value be changed into `enum vdp_status` also?

bin/varnishd/cache/cache_deliver_proc.c
```diff
@@ -83,19 +109,20 @@ VDP_push(struct req *req, const struct vdp *vdp, void *priv, int bottom)
...
 
 	if (DO_DEBUG(DBG_PROCESSORS))
 		VSLb(req->vsl, SLT_Debug, "VDP_push(%s)", vdp->name);
 
 	vdpe = WS_Alloc(req->ws, sizeof *vdpe);
 	if (vdpe == NULL) {
-		AZ(vdc->retval);
-		vdc->retval = -1;
-		return (vdc->retval);
+		VSLb(req->vsl, SLT_Error,
```
nigoroll: why pull in the error reporting? We got more error cases in the caller, so keeping the reporting in the caller would appear to be the natural choice?

bin/varnishd/cache/cache_deliver_proc.c
```diff
@@ -121,13 +148,10 @@ VDP_close(struct req *req)
...
 	while (!VTAILQ_EMPTY(&vdc->vdp)) {
 		vdpe = VTAILQ_FIRST(&vdc->vdp);
-		if (vdc->retval >= 0)
-			AN(vdpe);
-		if (vdpe != NULL) {
-			CHECK_OBJ_NOTNULL(vdpe, VDP_ENTRY_MAGIC);
-			VTAILQ_REMOVE(&vdc->vdp, vdpe, list);
-			AZ(vdpe->vdp->func(req, VDP_FINI, &vdpe->priv,
-			    NULL, 0));
+		CHECK_OBJ_NOTNULL(vdpe, VDP_ENTRY_MAGIC);
```
nigoroll: this partly reverts 1e82f502a027b89caa94be88496b20aaf3970c8d and I don't see why this should be a good idea:
shouldn't VDP_close work also on a failed init?

bin/varnishd/cache/cache_esi_deliver.c
```diff
@@ -695,6 +724,12 @@ ved_objiterate(void *priv, int flush, const void *ptr, ssize_t len)
 		}
 	}
 	assert(len == 0);
+
+	/* Flush always because foo->dbits will be reused on the next
+	 * call */
+	if (ved_bytes(foo->req, foo->preq, VDP_FLUSH, NULL, 0))
+		return (-1);
```
nigoroll: VDP_ERROR?

bin/varnishd/cache/cache_esi_deliver.c
```diff
@@ -695,6 +724,12 @@ ved_objiterate(void *priv, int flush, const void *ptr, ssize_t len)
...
+
 	return (0);
```
nigoroll: VDP_OK?

bin/varnishd/cache/cache_obj.c
```diff
@@ -238,22 +238,21 @@ ObjExtend(struct worker *wrk, struct objcore *oc, ssize_t l)
...
 	CHECK_OBJ_NOTNULL(oc->boc, BOC_MAGIC);
-	Lck_Lock(&oc->boc->mtx);
+
+	Lck_AssertHeld(&oc->boc->mtx);
+
 	while (1) {
-		rv = oc->boc->len_so_far;
-		assert(l <= rv || oc->boc->state == BOS_FAILED);
-		if (rv > l || oc->boc->state >= BOS_FINISHED)
+		assert(l <= oc->boc->len_so_far ||
```
nigoroll: I had reviewed the original code just recently and one of the results I remember was that the consistent view through the rv local variable was important for correctness.
So here I'd like to turn the argumentation around: What would be the good reason to change it?

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -407,12 +404,10 @@ cnt_transmit(struct worker *wrk, struct req *req)
 			    "Content-Length: %jd", req->resp_len);
 	}
 
-	if (err == 0)
```
nigoroll: see previous comment, I don't think we should push error reporting down into VDP

bin/varnishd/http1/cache_http1_deliver.c
```diff
@@ -116,21 +114,25 @@ V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
...
-	if (sendbody && VDP_push(req, &v1d_vdp, NULL, 1)) {
-		v1d_error(req, "workspace_thread overflow");
-		AZ(req->wrk->v1l);
+	if (sendbody)
+		VDP_push(req, &v1d_vdp, NULL, 1);
+
+	/* Check if any of the VDP initialization calls failed */
+	assert(req->vdc->status <= VDP_OK);
+	if (req->vdc->status < VDP_OK) {
+		v1d_error(req, "VDP initialization error", SC_VDP_ERROR);
```
nigoroll: to me it looks like we'd report an error twice, so, again, ....

bin/varnishd/http1/cache_http1_deliver.c
```diff
@@ -163,11 +165,11 @@ V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
 
 	if (!sendbody) {
 		AZ(req->wrk->v1l);
-		VDP_close(req);
```
nigoroll: so where is this happening?

bin/varnishd/storage/storage_simple.c
```diff
@@ -255,77 +249,115 @@ sml_iterator(struct worker *wrk, struct objcore *oc,
...
-		st = VTAILQ_NEXT(st, list);
-		if (st != NULL && st->len == 0)
-			st = NULL;
-		Lck_Unlock(&boc->mtx);
-		assert(l > 0 || boc->state == BOS_FINISHED);
-		ret = func(priv, st != NULL ? final : 1, p, l);
-		if (ret)
-			break;
 	}
+	Lck_Unlock(&boc->mtx);
```
nigoroll: I have not reviewed this in full, but IIUC I had just recently discussed this code with @bsdphk and IIUC he had deliberately reduced lock contention here?
If this change survives, I'd want to review again and in more detail, for now, the diff has overwhelmed me

rezan:
> Do we really need to differentiate between FLUSH and LAST

I would say yes, VFPs already have an end state and there are many usecases where a VDP needs to know that its at the end of the stream. I think its too late to re-purpose FLUSH as the end.

daghf:
In my opinion the key point regarding keeping FLUSH and LAST separate is that it lets us also have separate callbacks for the cleanup code and the byte-shuffling code, which makes for cleaner code.

bin/varnishd/cache/cache_deliver_proc.c
```diff
@@ -137,8 +161,10 @@ VDP_close(struct req *req)
 /*--------------------------------------------------------------------*/
 
 static int v_matchproto_(objiterate_f)
-vdp_objiterator(void *priv, int flush, const void *ptr, ssize_t len)
+vdp_objiterator(void *priv, int flush, int last, const void *ptr, ssize_t len)
```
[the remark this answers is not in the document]
mbgrydeland: From the object iterator POV, `flush` means the content is ephemereal, or that the pointers are only valid up until the function returns. `last` is the very last call. So they are not really the same, though for the consumer (VDP), `last` would also imply `flush` as we would then be finished delivering the response and a flush is executed anyways.

So yes, this could be combined to a bit field, or even an enum like in the VDP interface with last implying flush. But since the two args are not really the same, and this isn't a very critical code path (only called once per delivery), I think it's fine to use two unsigned args.

bin/varnishd/http1/cache_http1_deliver.c
mbgrydeland: The VDP_close() call was moved up one stack frame to cache_req_fsm.c, instead of relying on the individual transports to do it before returning. This makes it easier to know that it will always be called, and at the same time elliminating memory leaks from failure to call it for each of the H/1 canned responses in reaction to WS exhaustion (they failed to call VDP_close()).

bin/varnishd/cache/cache_deliver_proc.c
mbgrydeland: My thought on the error reporting was to allow the one noticing the error to do the reporting, as it is best equiped to do a specific error text rather than just a generic "VDP initialization error". This particular line is the VDP_push reporting that it failed to set up the VDP stack because of WS exhaustion.

In general, a VDP that wants to fail the delivery during its initialization should log appropriately before returning error. But since we can't trust a random VMOD to adhere to that rule, we also log a generic  "VDP initialization error" if things went south. This will result in multiple SLT_Error log lines instead of a single just to be sure. If we don't like this, we maybe should adopt a something like in VFP_Error instead, where we record if an error message has already been printed, and not do it twice.

bin/varnishd/cache/cache_deliver_proc.c
mbgrydeland: VDP_close() will call fini() on VDPs in the stack only when VDP_push() reported VDP_OK when it was inserted. This behaviour is needed because we do not know whether the VDP_push() is going to succeed or not, and if the init() function of the VDP to be inserted will even be called. This can happen because we latch the error status and skip on initialization when an error is recorded, or we skip it because of WS exhaustion in VDP_push. If the code calling VDP_push is handing off a resource through the priv arg, it needs to free that resource in case of failure instead of relying on the fini() function to do it for it.

bin/varnishd/cache/cache_esi_deliver.c
mbgrydeland: Yeah

bin/varnishd/cache/cache_req_fsm.c
mbgrydeland: See comments on the error handling above.

bin/varnishd/storage/storage_simple.c
mbgrydeland: It is my understanding that the new scheme results in a few less attempts to lock the mutex, thus reducing the contention.

bin/varnishd/cache/cache_deliver_proc.c
nigoroll: Yeah something like VDP_Error sounds like a sensible thing to do. This way we can achieve your sensible goal to ensure there always is an error emitted while still keeping it concise inside the VDP.

nigoroll:
@mbgrydeland thank you for your responses. I will probably need to spend some more time with the code then. As you are implying that the lock changes are to stay, that's where I'd want to focus on. It would be helpful if we could reduce the number of LOCs changed, but I'll try anyway.
Regarding the contention and amount if synchronization required, I am pretty sure that the previous code was quite good, but for my review I will assume that your suggestion is even better.

nigoroll:
@mbgrydeland I've finally got back to this and will re-start my review with the iterator changes.
FTR and as a reference, here's the IRC log (edited out irrelevant bits) when I last discussed this piece of code with @bsdphk:

Conversation with #varnish-hacking at Wed 11 Apr 2018 10:00:47 CEST on slink@irc.linpro.no (irc)
...
(11:37:36) ehocdet: phk: https://github.com/varnishcache/varnish-cache/commit/8a34320c3b5d10638658125943158af3fcd2d5f7#diff-8c8c6a8320a60ec4cca4098f6458fb8eL149
(11:37:47) ehocdet: pthread_cond_broadcast is out of the lock, a typo?
(12:09:27) phk: ehocdet, no, that is deliberate.
(12:09:39) slink: phk can you elaborate?
(12:09:55) phk: ehocdet, usually the threads you wake with pthread_cond will try to grab the protecting mutex as the first thing they do.
(12:10:19) phk: so if you have the broadcast inside the mutex, you end up with an unecessary context-switch or three
(12:10:39) phk: Since this is not a hard lock, but merely a notification, we can avoid that by releasing the mutex first.
(12:10:58) slink: but is it ok that the bcast can be missed?
(12:11:01) phk: If the thread happens to wake on its own between our letting go of the mutex and the broadcast, nothing bad happens.
(12:11:17) phk: it can only be missed if the thread already woke up
(12:12:03) ehocdet: interesting, thanks
...
(13:04:52) slink: phk re condvars, I am not convinced. My understanding is this:
(13:04:52) slink: cond_wait, internally, registers itself on the cv, then releases the lock
(13:04:52) slink: so if we use cond_broadcast outside the lock, it can happen that the broadcast happens before cond_wait could register and thus the signal gets lost.
(13:07:52) phk: but the condwait is registerd holding the lock, so we couldn't get here in that case.
(13:37:59) slink: phk so just reflecting on the condvar
(13:37:59) slink: thr a ObjExtend gets up to Unlock (inclusive)
(13:37:59) slink: thr b ObjWaitExtend gets up to CondWait because it rv is not sufficient
(13:37:59) slink: now before b registers, a signals -> signal is lost
(13:37:59) slink: BUT in this case we're waiting for another Extend, so the second one will be caught
(13:37:59) slink: so for this case it would be ok
(13:37:59) slink: I am not entirely convinced, though
(13:40:18) phk: slink: Thread a holds the mutex.
(13:40:51) slink: phk I constructed the case where it has just released it
(13:40:54) phk: If it holds the mutex before thread be takes the mutex, it has the old value and sleeps on the condvar, thread b gets mutex, updates var, releases mutex, signals, a wakes up, sees new value, all is fine.
(13:41:18) phk: If a grabs the mutex right after b releases it, but before b signals, a sees the new value and all is fine.
(13:41:50) slink: phk I was thinking about the case that the new value is not sufficient
(13:42:14) slink: I think in this case it will miss the first signal, but a second one must be coming later in this case
(13:42:25) phk: what case is that ?
(13:43:18) phk: (ie: what do you mean "new value is not sufficient" ?
(13:43:29) slink: I might suffer from not having the big picture, just staring at the code as is: if rv == l in WaitExtend
(13:44:39) phk: slink, that is the case where you have all there is to be had at this point, and you should sleep until next call to ObjExtend() or ObjSetState()
(13:44:58) phk: and note how SetState calls broadcast inside the mutex ?
(13:45:09) slink: thats the point I got to: there must be another signal coming up in this case
(13:45:14) slink: yes I did
(13:55:54) slink: phk thank you for your explanations. I had learned in the past that broadcast-outside-mtx is usually wrong, but I feel convinced that it is ok in this case.
(13:57:05) phk: slink, as long as you get the final one right, it's an OK optimization.
(13:57:29) slink: phk full ack

nigoroll:
@mbgrydeland I'd like to think out loud while reviewing:
You are proposing to change `sml_iterator()` to basically always hold the busyobject lock except when sleeping on the CV in `ObjWaitExtend` or during the callback func (usually writing to the client). The advantage is halving the number of lock ops in `sml_iterator()`, so the question is if we would now be holding the lock during longer operations, and my answer would be plain no. However, in the old code we could have a short path avoiding the lock in `ObjWaitExtend` `if (oc->boc->len_so_far > l || oc->boc->state >= BOS_FINISHED)`
Staring at the code, I wonder if we can determine the `last` flag safely like this: Wouldn't we need to always stay one chunk behind? In other words, couldn't it happen that we have already processed all bytes when the state changes to `BOS_FINISHED` without any new bytes, for example with `BS_EOF`?
will continue after your response.

nigoroll:
discussed during VDD: `last` is just an opportunistic optimization

nigoroll:
@mbgrydeland close for OBE?

mbgrydeland:
OBE from recent work on master. Closing the PR.

PR #2680, opened 2018-05-17
move logic of VDI_Resolve() to VRT as VRT_DirectorResolve()

nigoroll:
VDI_Resolve() remains for error handling for calls from VDI_GetHdr() and VDI_Http1Pipe()

### DIFF
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -288,12 +288,12 @@ VRT_LookupDirector(VRT_CTX, VCL_STRING name)
 VCL_BACKEND
 VRT_DirectorResolve(VRT_CTX, VCL_BACKEND d)
 {
-	VCL_BACKEND d2;
-
-	for (; d != NULL && d->vdir->methods->resolve != NULL; d = d2) {
+	while (d != NULL) {
 		CHECK_OBJ_NOTNULL(d, DIRECTOR_MAGIC);
 		AN(d->vdir);
-		d2 = d->vdir->methods->resolve(ctx, d);
+		if (d->vdir->methods->resolve == NULL)
+			break;
+		d = d->vdir->methods->resolve(ctx, d);
 	}
 	CHECK_OBJ_ORNULL(d, DIRECTOR_MAGIC);
 	if (d != NULL)
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -286,6 +286,24 @@ VRT_LookupDirector(VRT_CTX, VCL_STRING name)
 /*--------------------------------------------------------------------*/
 
 VCL_BACKEND
+VRT_DirectorResolve(VRT_CTX, VCL_BACKEND d)
+{
+	VCL_BACKEND d2;
+
+	for (; d != NULL && d->vdir->methods->resolve != NULL; d = d2) {
+		CHECK_OBJ_NOTNULL(d, DIRECTOR_MAGIC);
+		AN(d->vdir);
+		d2 = d->vdir->methods->resolve(ctx, d);
+	}
+	CHECK_OBJ_ORNULL(d, DIRECTOR_MAGIC);
+	if (d != NULL)
+		AN(d->vdir);
+	return (d);
+}
+
+/*--------------------------------------------------------------------*/
+
+VCL_BACKEND
 VCL_DefaultDirector(const struct vcl *vcl)
 {
 
--- a/bin/varnishd/cache/cache_director.c
+++ b/bin/varnishd/cache/cache_director.c
@@ -83,28 +83,23 @@ VDI_Ahealth(const struct director *d)
 static VCL_BACKEND
 VDI_Resolve(VRT_CTX)
 {
-	const struct director *d;
-	const struct director *d2;
+	VCL_BACKEND d;
 	struct busyobj *bo;
 
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	bo = ctx->bo;
 	CHECK_OBJ_NOTNULL(bo, BUSYOBJ_MAGIC);
-	CHECK_OBJ_ORNULL(bo->director_req, DIRECTOR_MAGIC);
 
-	for (d = bo->director_req; d != NULL &&
-	    d->vdir->methods->resolve != NULL; d = d2) {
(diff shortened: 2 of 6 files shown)


nigoroll:
FTR, I had accidentally pushed a bad commit in the meantime, now force-pushed the branch as it should be

nigoroll:
FTR, had to copy this into the vmod to get ahead: https://code.uplex.de/uplex-varnish/libvmod-cluster/commit/f24e54036ea36508c2cf429df948b585abf1ef9c

nigoroll:
as this is a prerequisite for #2850 which appears to take on momentum, can we please try to get this one ticked off?

bsdphk:
My problem wit

bsdphk:
Sorry, hit wrong button by accident.  Comment still stands.

nigoroll:
First of all, a side note on the director refcounting issue: During a VDD, we developed the idea of a director _dough_ to be baked when it is time to make the decision. I had actually pondered this a bit, but it appeared to me that this would actually be less powerful than the parameter object we already got https://github.com/varnishcache/varnish-cache/blob/master/lib/libvmod_directors/vmod.vcc#L584 because the parameter object can be shared between directors. I do not want to rule out this option, but so far I do not (yet?) see any great advantage.
Looking at the issue from the opposite angle, the killer argument against #2725 was that we need to take a mutex once per request. It was only much later that I realized that it should be possible to combine taking the cool list reference with taking the VCL reference such that no additional serialization operation per request was required. I have not got back to this yet, but I still find the idea perused by the PR attractive and hope that if the (only?) concern against it could be ruled out, it could become a viable option (again).

Now regarding this PR: There is absolutely nothing which would prevent vmods from not doing the right thing. There is no difference between directly referencing a backend and referencing one after resolution through a director. The function of this proposed interface is similar to https://github.com/varnishcache/varnish-cache/blob/master/lib/libvmod_directors/vmod.vcc#L430 but generalized over directors.

So I agree that we got an issue to solve and while I would want avoid it getting worse, I can not follow the argument that this proposal would lead us into the wrong direction.

nigoroll:
rebased onto #2939

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -274,6 +274,24 @@ VRT_LookupDirector(VRT_CTX, VCL_STRING name)
...
+VCL_BACKEND
+VRT_DirectorResolve(VRT_CTX, VCL_BACKEND d)
+{
+	VCL_BACKEND d2;
+
+	for (; d != NULL && d->vdir->methods->resolve != NULL; d = d2) {
+		CHECK_OBJ_NOTNULL(d, DIRECTOR_MAGIC);
+		AN(d->vdir);
+		d2 = d->vdir->methods->resolve(ctx, d);
+	}
```
Dridi: I'm a bit worried that we dereference `d->vdir->methods` before we check `d`. I'm also worried that we are painting ourselves in a corner if we expose this function with this signature. I will try to discuss this with you before adding more noise to this review.

Dridi:
Before we merge this I would like to see an improved commit log explaining at least why we want something like `VRT_DirectorResolve` in the first place. Spoilers: allow eager backend resolution using VMODs. Compared to the lazy backend resolution that is done outside VCL once a fetch is initiated for real.

I discussed the function signature with @nigoroll and here is a summary of how we want to follow up this change. It should be possible to keep the new `VRT_DirectorResolve` signature as-is, but we should add a note mentioning potential changes before it's part of a release.

#### VRT_DirectorHealth

Introduce a `VCL_REAL VRT_DirectorHealth(...)` function that returns a value between 0 and 1. This allows directors to provide a non-binary health meter that opens the ability to have more built-in directors.

For VBE backends, this is the ratio of available connections. This would fix a bug where today a backend can be selected by a director even when it has no available slot to open a connection.

For a backend with `max_connections = 0` (the default) we can use an arbitrary maximum like `UINT32_MAX` to allow unbounded backends to have a relative health meter when they are part of the same cluster.

A new director with a selection algorithm based on this criteria would effectively enable a "least connections" strategy for regular backends. Bikeshed available with enough paint for everyone to give it name.

Built-in directors also need to implement the relative health method since they can be stacked together.

#### Backend/Director reservation ticket

Even though a director would no longer resolve a backend with no available connection, there is still a TOCTOU race between the moment we select a backend and the moment we try to use it to send a request.

Our suggestion is to use the PRIV_TASK facility for backends that support it to acknowledge that they were selected and guarantee that by the time their `gethdrs()` method is called this is still the case. Then they are free to fail for other reasons.

The problem with PRIV_TASK is ownership. If a backend is resolved on the client side (`req.backend_hint`) then this priv needs to be moved to the backend transaction if a fetch or bgfetch is triggered. So that no more than one transaction owns the reservation ticket. For that we may want a new PRIV_DIRECTOR that is essentially the same, but with move capabilities.

Things to consider:

- eager resolution (the object of this pull request) on the client side before going to the waitinglist
- multiple eager resolutions in the same transaction, because once we have this, it will likely happen
- pipe transactions and when to release the privs

nigoroll:
discussion with @Dridi and @bsdphk : update director refcount PR, then get back to this one

nigoroll:
FYI a quick report on progress with director refcounting
* I have an updated version of #2725 for which all checks pass
* Currently working on a version which does not require an additional lock and integrated with VCL Refcounting
* latest spin off is #3198

bsdphk:
From what contexts is it OK to call VRT_DirectorResolve ?

As far as I can tell, vmod_directors::shard would panic on the lack of ctx->bo if one called it from vcl_recv{} ?

nigoroll:
> From what contexts is it OK to call VRT_DirectorResolve ?

Any, IMHO. If directors do not support certain contexts, they should return `NULL` or `VRT_fail()`.

> As far as I can tell, vmod_directors::shard would panic on the lack of ctx->bo if one called it from vcl_recv{} ?

Can you direct me to the respective LOC?

No, this should not happen, and I have just pushed to this PR a patch to add `directors.resolve()` with a respective test.

bsdphk:
You're right, I overlooked the check for ctx-Bo != NULL in get_key.

But the question still stands: Do we want this to be usable everywhere ?

nigoroll:
> But the question still stands: Do we want this to be usable everywhere ?

Yes, my answer was: If directors do not support certain contexts, they should return `NULL` or `VRT_fail()`.

bsdphk:
Please see #3311 for an alternative implementation using type-specific methods.

nigoroll:
merged #3311 instead

PR #2688, opened 2018-05-22
varnishtop -1, -d and -f

### LINKED ISSUE #2686: varnishtop -1 does not work as documented
The manual says:

> Instead of a continuously updated display, print the statistics once and exit. Implies -d.

However it also seems to filter out anything but the top results. By that I mean any record having the highest ranking value.

Consider this test case:

```
varnishtest "varnishtop -1"

server s1 {
        rxreq
        txresp
} -start

varnish v1 -vcl+backend { } -start

client c1 {
        txreq
        rxresp
} -run

process p1 -dump {varnishtop -n ${v1_name} -d -i ReqMethod} -start
process p1 -winsz 30 80
delay 2
process p1 -expect-text 3 6 {1.00 ReqMethod      GET}
process p1 -screen_dump -write q -wait

process p2 -dump {varnishtop -n ${v1_name} -1 -i ReqMethod} -start
process p2 -winsz 30 80
delay 2
process p2 -expect-text 1 6 {1.00 ReqMethod GET}
process p2 -screen_dump -wait
```

When `-d` is used we get `ReqMethod` as the first and only result, as expected. When `-1` is used the output is empty. If we remove the `-i` option in the `-1` process we get the following results:

```
     2.00 VCL_return fetch
     2.00 VCL_return deliver
```

Instead of only waiting to display statistics it also removes anything that is not in the "top 1" from the output.

It leads me to believe that when we filter the output we still needlessly count fields that aren't going to be displayed. I can open a separate issue if we want to enhance the `-i` option to save cycles.

By the way, the manual is quite cryptic for `-i` and `-I`:

> If a tag include option is the first of any tag selection options, all tags are first marked excluded.

Spotted while reproducing: 9938f01fe8a2b148013423c310c221273b8297b4

I haven't tested on older releases yet, but I believe it goes back to at least 4.1 (and probably 4.0).

Dridi:
Fixes #2686, bonus docfix and coverage for `-f`.

Please see individual commits for details.

### DIFF
--- a/bin/varnishtop/varnishtop_options.h
+++ b/bin/varnishtop/varnishtop_options.h
@@ -40,7 +40,9 @@
 #define TOP_OPT_f							\
 	VOPT("f", "[-f]", "First field only",				\
 	    "Sort and group only on the first field of each log entry."	\
-	    " This is useful when displaying e.g. stataddr entries,"	\
+	    " For log entries in the form ``prefix: value`` it is the"	\
+	    " prefix without the colon that is sorted and grouped."	\
+	    " This is useful when displaying e.g. ReqStart entries,"	\
 	    " where the first field is the client IP address."		\
 	)
 
--- a/bin/varnishtest/tests/r02686.vtc
+++ b/bin/varnishtest/tests/r02686.vtc
@@ -1,0 +1,27 @@
+varnishtest "varnishtop -1/-d"
+
+server s1 {
+        rxreq
+        txresp
+} -start
+
+varnish v1 -vcl+backend { } -start
+
+client c1 {
+        txreq
+        rxresp
+} -run
+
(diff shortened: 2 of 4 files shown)


bin/varnishtop/varnishtop_options.h
```diff
@@ -40,7 +40,9 @@
 #define TOP_OPT_f							\
 	VOPT("f", "[-f]", "First field only",				\
 	    "Sort and group only on the first field of each log entry."	\
-	    " This is useful when displaying e.g. stataddr entries,"	\
+	    " For log entries in the form ``prefix: value`` it is the"	\
+	    " prefix without the colon that is sorted and groupped."	\
```
fgsch: grouped.

PR #2689, opened 2018-05-23
Add JSON ping output

guardalben:
This adds -j support to the command

varnishadm ping -j

Tested on RFC4627/RFC7159/ECMA404 standards (all valid JSON)

### DIFF
--- a/lib/libvarnish/vcli_serve.c
+++ b/lib/libvarnish/vcli_serve.c
@@ -103,6 +103,16 @@ VCLS_func_ping(struct cli *cli, const char * const *av, void *priv)
 	VCLI_Out(cli, "PONG %jd 1.0", (intmax_t)t);
 }
 
+void v_matchproto_(cli_func_t)
+VCLS_func_ping_json(struct cli *cli, const char * const *av, void *priv)
+{
+	(void)av;
+	(void)priv;
+	VCLI_JSON_begin(cli, 2, av);
+	VCLI_Out(cli, ", \"PONG\"\n");
+	VCLI_JSON_end(cli);
+}
+
 /*--------------------------------------------------------------------*/
 
 void v_matchproto_(cli_func_t)
--- a/bin/varnishd/cache/cache_cli.c
+++ b/bin/varnishd/cache/cache_cli.c
@@ -111,7 +111,7 @@ CLI_Run(void)
 /*--------------------------------------------------------------------*/
 
 static struct cli_proto cli_cmds[] = {
-	{ CLICMD_PING,	"i", VCLS_func_ping },
+	{ CLICMD_PING,	"i", VCLS_func_ping, VCLS_func_ping_json },
 	{ CLICMD_HELP,	"i", VCLS_func_help, VCLS_func_help_json },
 	{ NULL }
 };
--- a/bin/varnishd/mgt/mgt_cli.c
+++ b/bin/varnishd/mgt/mgt_cli.c
@@ -327,7 +327,7 @@ mcf_help_json(struct cli *cli, const char * const *av, void *priv)
 
 static struct cli_proto cli_auth[] = {
 	{ CLICMD_HELP,		"", mcf_help, mcf_help_json },
-	{ CLICMD_PING,		"", VCLS_func_ping },
+	{ CLICMD_PING,		"", VCLS_func_ping, VCLS_func_ping_json },
 	{ CLICMD_AUTH,		"", mcf_auth },
 	{ CLICMD_QUIT,		"", VCLS_func_close },
 	{ NULL }
--- a/bin/varnishtest/tests/b00008.vtc
+++ b/bin/varnishtest/tests/b00008.vtc
@@ -1,4 +1,4 @@
-varnishtest "Test CLI help and parameter functions"
+varnishtest "Test CLI commands and parameter functions"
 
 varnish v1 -arg "-b ${bad_ip}:9080"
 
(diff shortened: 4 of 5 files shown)


bin/varnishtest/tests/b00099.vtc
```diff
@@ -0,0 +1,19 @@
+varnishtest "Test varnishadm -j option"
+
+feature ignore_unknown_macro
+
+feature cmd {python --version -m json.tool 2>&1 | grep -q 'Python [0-9].[0-9]'}
+
+varnish v1 -vcl { backend foo { .host = "127.0.0.1"; } } -start
+
+shell -expect "VALID" {
+	${topbuild}/bin/varnishadm/varnishadm -n ${v1_name} help -j | python -m json.tool >> /dev/null && echo "VALID" || echo "NOT"
```
fgsch: You can drop the topbuild from all these.

bin/varnishtest/tests/b00099.vtc
```diff
@@ -0,0 +1,19 @@
...
+}
+
+shell -expect "VALID" {
+	${topbuild}/bin/varnishadm/varnishadm -n ${v1_name} backend.list -j | python -m json.tool >> /dev/null && echo "VALID" || echo "NOT"
+}
+
+shell -expect "VALID" {
+	${topbuild}/bin/varnishadm/varnishadm -n ${v1_name} ping -j | python -m json.tool >> /dev/null && echo "VALID" || echo "NOT"
+}
```
fgsch: You can drop the topbuild from all these.

guardalben:
Thanks, have dropped as you suggested.

bin/varnishtest/tests/b00099.vtc
guardalben: Thanks, have dropped as you suggested.

bsdphk:
Varnishtest supports testing CLI output for JSON wellformedness now, we should use that instead of running extra python processes.  See `-cli-json`

guardalben:
Agreed, I've pushed a simple varnishtest using -cli-json and will implement and validate new commands.

bsdphk:
One other thing, we're trying to stem the explosion of test-cases, because they have an overhead of several seconds and we're already at 700.

Can I get you to merge the test-case into b00008 instead ?

guardalben:
done!

lib/libvarnish/vcli_serve.c
```diff
@@ -102,6 +102,21 @@ VCLS_func_ping(struct cli *cli, const char * const *av, void *priv)
...
+{
+	time_t t;
+
+	(void)av;
+	(void)priv;
+	t = time(NULL);
+	VCLI_JSON_begin(cli, 1, av);
+	VSB_indent(cli->sb, 2);
+	VCLI_Out(cli, ",\n {\n \"PONG\": %jd \n }", (intmax_t)t);
+	VSB_indent(cli->sb, -2);
```
fgsch: These VSB_indent calls don't really follow the output from `help -j`, which if I'm reading right should be:

```
  {
     "PONG": 1234
  }
```

guardalben: Thanks! you're absolutely right. I've pushed a new version following the same format as in cache_director.c, lines [373-381].

bsdphk:
So final thing:  Why the dictionary in the output ?

I suggest we just emit the "PONG" and the timestamp as part of the top-level list:

    [ 1, ["ping", "-j"], "PONG", 528094150 ]

gquintard:
maybe we should have a timestamp in all responses, and so have it before
the variable element?
```

[ 1, ["ping", "-j"], 528094150, "PONG" ]

```


This will be my last bikeshedding contribution.

-- 
Guillaume Quintard

bsdphk:
I don't see the point of timestamps in all responses, but can be convinced by good arguments.

Dridi:
Rather than a timestamp in all responses, having the ability to pass a nonce in a command to check it back in the response? `varnishadm`/`libvarnishapi` could do that transparently to make users' lives simpler and help us not run into the likes of #2010 in the future. Alas I'm afraid in its current form the CLI protocol makes it somewhat hard to incorporate it as an optional feature.

slimhazard:
I believe it's fairly common to send the output of ``backend.list`` to a monitoring tool in order to keep an eye on the health state of backends. JSON makes that easier, and a timestamp is useful info -- "these were the health states at time T".

Maybe it would be generalizing from just two examples, but it seems to me that "CLI told me that this info was true at time T" generally makes sense. ``vcl.list`` and ``param.show`` come to mind as commands for which a timestamp seems relevant.

hermunn:
> Maybe it would be generalizing from just two examples, but it seems to me that "CLI told me that this info was true at time T" generally makes sense. vcl.list and param.show come to mind as commands for which a timestamp seems relevant.

I fully agree, but I don't think removing the timestamp has been suggested. For the text version we get

```
# varnishadm ping
PONG 1528101934 1.0
```
In my opinion, removing the timestamp in the JSON when it is present in the text version is not OK. More general questions are
1. Should we always return the command with the reply?
2. Should the timestamp always be the third item when it makes sense to have a timestamp (almost always)? Or maybe it should always be the last item?

If we can agree on these, then I think we have the answer to what the output should be for the `ping` command.

guardalben:
IMHO, to have useful information to feed third-party monitoring tools (e.g., for time series analysis) the timestamp is mandatory to track back problems and/or to create patterns to further predict problems based on time analysis. Perhaps, the granularity of the information may need to be considered as well (e.g., info returned every 1sec, 5sec, 10sec...) which depends on the use case. This is a trade-off between overhead to provide info vs useful info at exact moment (should we provide them this possibility?)

bsdphk:
Well put a timestamp in the common JSON header for all JSON replies. (I'll do that)

bsdphk:
I have updated the generic code so that all CLI JSON outputs have a timestamp.

That means the output from the ping -j can be reduced to::

        VCLI_JSON_begin(cli, 2, av);
        VCLI_Out(cli, ", \"PONG\"\n");
        VCLI_JSON_end(cli);

OK'ed with that change.

hermunn:
About back porting: JSON versions of command output will not be back ported by me to 4.1. Protest here if you really need it, and I will look into it.

Issue #2624, opened 2018-03-20
Assert error in HTC_RxStuff(), cache/cache_session.c line 296


fgsch:
```
Assert error in HTC_RxStuff(), cache/cache_session.c line 296:
  Condition(z >= 0) not true.
version = varnish-6.0.0 revision 5e2b0d8abda18e807b73ed2f6c0746688001e85b, vrt api = 7.0
ident = Darwin,17.4.0,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,kqueue
now = 1032608.898370 (mono), 1521572311.301582 (real)
Backtrace:
  0x105d02f6c: 0   varnishd                            0x0000000105d02f6c pan_backtrace + 348
  0x105d02b46: 0   varnishd                            0x0000000105d02b46 pan_ic + 1446
  0x1060dc549: 0   varnishd                            0x00000001060dc549 VAS_Fail + 361
  0x105d5520e: 0   varnishd                            0x0000000105d5520e HTC_RxStuff + 10254
  0x105ec182f: 0   varnishd                            0x0000000105ec182f h2_rxframe + 2335
  0x105edf20f: 0   varnishd                            0x0000000105edf20f h2_new_session + 9807
  0x105e0a662: 0   varnishd                            0x0000000105e0a662 Pool_Work_Thread + 15858
  0x105e0649f: 0   varnishd                            0x0000000105e0649f WRK_Thread + 1903
  0x105e05cab: 0   varnishd                            0x0000000105e05cab pool_thread + 891
  0x7fff64cf16c1: 0   libsystem_pthread.dylib             0x00007fff64cf16c1 _pthread_body + 340
thread = (cache-worker)
thr.req = 0x631000050820 {
  vxid = 1000, transport = H2
  step = 0x0,
  req_body = R_BODY_INIT,
  err_code = 1, err_reason = (null),
  restarts = 0, esi_level = 0,
  sp = 0x61500000ffa0 {
    fd = 25, vxid = 1000,
    t_open = 1521572311.300687,
    t_idle = 1521572311.301546,
    ws = 0x61500000ffe0 {
      id = "ses",
      {s, f, r, e} = {0x615000010018, +104, 0x0, +352},
    },
    transport = H2 {
      streams {
        0x00000000 idle  
      }
    }
    client = 127.0.0.1 58497 127.0.0.1:58492,
  },
  ws = 0x631000050968 {
    id = "req",
[...]

Issue #2285, opened 2017-03-28
Assert error in HTC_RxStuff(), cache/cache_session.c line 294


simonvik:
```
Panic at: Tue, 28 Mar 2017 09:53:14 GMT
Assert error in HTC_RxStuff(), cache/cache_session.c line 294:
  Condition(i >= 0) not true.
version = varnish-5.1.1 revision de38712, vrt api = 6.0
ident = Linux,4.4.30-20161130-1423-0a2a05d,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
now = 5510903.711436 (mono), 1490694793.346442 (real)
Backtrace:
  0x438975: /usr/sbin/varnishd() [0x438975]
  0x440c92: /usr/sbin/varnishd(HTC_RxStuff+0x352) [0x440c92]
  0x461a71: /usr/sbin/varnishd(h2_rxframe+0x81) [0x461a71]
  0x4624b5: /usr/sbin/varnishd() [0x4624b5]
  0x451de2: /usr/sbin/varnishd() [0x451de2]
  0x45228b: /usr/sbin/varnishd() [0x45228b]
  0x7f7d5a1316ba: /lib/x86_64-linux-gnu/libpthread.so.0(+0x76ba) [0x7f7d5a1316ba]
  0x7f7d59e6782d: /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7f7d59e6782d]
thread = (cache-worker)
thr.req = 0x7f7d38817020 {
  vxid = 425985, transport = H2
  step = 0x0,
  req_body = R_BODY_INIT,
  err_code = 1, err_reason = (null),
  restarts = 0, esi_level = 0,
  sp = 0x7f7d3bc0e220 {
    fd = 21, vxid = 425985,
    t_open = 1490694793.208348,
    t_idle = 1490694793.346433,
    transport = H2 {
      streams {
        0x00000000 idle  
        0x00000001 c-rem 
        0x00000003 c-rem 
      }
    }
    client = ********* 64787,
    privs = 0x7f7d3bc0e288 {
    },
  },
  ws = 0x7f7d38817208 {
    id = \"req\",
    {s, f, r, e} = {0x7f7d38819008, +16648, +122864, +122864},
  },
  http_conn = 0x7f7d38817130 {
    fd = 21 (@0x7f7d3bc0e238),
    doclose = NULL,
    ws = 0x7f7d38817208 {
      [Already dumped, see above]
    },
[...]

Issue #2219, opened 2017-02-17
Crash in 4.1.5 with "Assert error in SES_RxReInit(), cache/cache_session.c line 179"


akuktin:
We just had two crashes with varnish-4.1.5 installed from packages from the repo at http://repo.varnish-cache.org/ubuntu . Syslog reports:

```
Feb 17 15:00:19 lbs01-pa varnishd[26949]: Child (26953) Last panic at: Fri, 17 Feb 2017 14:00:19 GMT
"Assert error in SES_RxReInit(), cache/cache_session.c line 179:
  Condition(l <= htc->ws->r - htc->rxbuf_b) not true.
thread = (cache-worker)
version = varnish-4.1.5 revision 2c82b1c
ident = Linux,3.13.0-63-generic,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
now = 275547.556234 (mono), 1487340017.329933 (real)
Backtrace:
  0x433f42: varnishd() [0x433f42]
  0x43adca: varnishd(SES_RxReInit+0xba) [0x43adca]
  0x45dc74: varnishd(VPX_Proto_Sess+0x1c4) [0x45dc74]
  0x44a419: varnishd() [0x44a419]
  0x44a88b: varnishd() [0x44a88b]
  0x7fe514af6184: libpthread.so.0(+0x8184) [0x7fe514af6184]
  0x7fe51482337d: libc.so.6(clone+0x6d) [0x7fe51482337d]

"
```
```
Feb 17 15:20:21 lbs01-pa varnishd[26949]: Child (29560) Last panic at: Fri, 17 Feb 2017 14:20:21 GMT
"Assert error in SES_RxReInit(), cache/cache_session.c line 179:
  Condition(l <= htc->ws->r - htc->rxbuf_b) not true.
thread = (cache-worker)
version = varnish-4.1.5 revision 2c82b1c
ident = Linux,3.13.0-63-generic,x86_64,-junix,-smalloc,-smalloc,-hcritbit,epoll
now = 276660.009976 (mono), 1487341129.783674 (real)
Backtrace:
  0x433f42: varnishd() [0x433f42]
  0x43adca: varnishd(SES_RxReInit+0xba) [0x43adca]
  0x45dc74: varnishd(VPX_Proto_Sess+0x1c4) [0x45dc74]
  0x44a419: varnishd() [0x44a419]
  0x44a88b: varnishd() [0x44a88b]
[...]

PR #2701, opened 2018-06-07
Make HTC_RxStuff reserve a byte for NUL character again and fix maxbytes

mbgrydeland:
handling

This partly reverses the fix for #2285, as that fix was missdiagnosed. The
original fix for #2285 also introduced a bug similar to #2219 where the
proxy code would not find space to insert a NUL character.

The real problem in #2285 (and the duplicate #2624) appears during H/2
prior knowledge upgrade, and the difference in maxbytes argument to
HTC_RxStuff between the H/1 and H/2 code paths. If the pipelined bytes
exceeded the maxbytes input on the second run we would run into an assert.

(#2624 has received a workaround in the H/2 code that perhaps should be
reverted).

Changes in this patch:

HTC_RxStuff will now reserve a space to insert a NUL character for those
completion functions that needs that (HTTP/1 and Proxy_V1).

HTC_RxStuff will no longer assert on seeing pipelined data exceeding the
maxbytes input.


bsdphk:
I belive I have fixed the PROXY and H1 code to not require the NUL insert.

PR #2705, opened 2018-06-10
A handfull of steps towards fixing #1799

hermunn:
These five commits are hopefully on the right path to finally closing #1799, but we are still not there.

Only one, the last, of these five commits is not eligible for back porting. This commit changes VCL/the final state machine, and can conceivably break some users' VCL logic. Actually, this can be discussed, as the only change is that we go to miss with a 304 candidate for "keep" objects, while other cases stay the same.

Not included is the patch where `return (miss)` in `sub vcl_hit` gets disallowed.


bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -507,14 +507,23 @@ cnt_lookup(struct worker *wrk, struct req *req)
...
-		if (busy != NULL) {
+		/* Found nothing or an object that was out of grace */
+		if (oc == NULL) {
+			/* Nothing */
+			if (busy != NULL) {
+				AN(busy->flags & OC_F_BUSY);
+				req->objcore = busy;
+				req->req_step = R_STP_MISS;
+			} else {
+				req->req_step = R_STP_PASS;
```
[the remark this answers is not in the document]
hermunn: Indeed it is, but we cannot remove this until we have disallowed `return (miss)` from `sub vcl_hit`. I hope that will happen soon.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -507,14 +507,23 @@ cnt_lookup(struct worker *wrk, struct req *req)
...
+			}
+		} else {
+			/* Very expired object */
```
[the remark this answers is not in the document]
hermunn: It means that even grace has expired. I can fix the comment.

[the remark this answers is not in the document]
hermunn: Alternatively we can add a possible return value for `Lookup found a value, but grace had expired`, and thus avoid the `if (oc == NULL)` test. It is a matter of taste.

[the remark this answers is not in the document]
nigoroll: I was really just wondering about the language. Sounds a bit like "most optimal" (I like to call it a _hyperlative_ as an exaggeration of the superlative)

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
nigoroll: thank you for clarifying, ok

bin/varnishtest/tests/r02219.vtc
```diff
@@ -14,7 +14,7 @@ varnish v1 -arg "-p workspace_client=9k" -proto PROXY -vcl+backend {
...
+	send "PROXY TCP4 127.0.0.1 127.0.0.1 1111 2222\r\nGET /AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA HTTP/1.1\r\n\r\n"
```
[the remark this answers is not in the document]
hermunn: If #2701 is merged or resolved, then this change can go away. It was necessary because `struct req` increased with 8 bytes (the new `d_grace` member), and that made the test case fail (with a panic), and I just wanted a passing test suite.

bin/varnishtest/tests/b00099.vtc
```diff
@@ -0,0 +1,116 @@
...
+	expect resp.body == "2"
+	expect resp.http.X-grace == "0.000"
+} -start
+
+delay .1
+
+# third client does not disable grace, and should get the grace object
+# even though c3 is waiting on the background thread to deliver a new
+# version.
+
```
daghf: instead of a delay in server{} we should use a barrier to ensure the fetch is still ongoing when c4 executes

bin/varnishd/cache/cache_hash.c
```diff
@@ -478,36 +478,50 @@ HSH_Lookup(struct req *req, struct objcore **ocp, struct objcore **bocp,
 		busy_found = 0;
 	}
 
+	if (exp_oc != NULL && EXP_Ttl_grace(req, exp_oc) < req->t_req) {
+		/* the newest object is out of grace */
+		if (!busy_found) {
+			/*
+			 * here we get to insert the busy object. This
+			 * transkltes to a MISS with a 304 candidate.
```
daghf: spelling

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -627,7 +629,11 @@ VRT_r_##which##_##fld(VRT_CTX)					\
 
 VRT_DO_EXP_R(obj, ctx->req->objcore, ttl,
     ttl_now(ctx) - ctx->req->objcore->t_origin)
-VRT_DO_EXP_R(obj, ctx->req->objcore, grace, 0)
+VRT_DO_EXP_R(obj, ctx->req->objcore, grace,
+    ctx->req->d_grace >= 0
+    && ctx->req->d_grace < ctx->req->objcore->grace
+    ? ctx->req->objcore->grace - ctx->req->d_grace
+    : 0)
```
bsdphk: Two questions:

1. Should obj.grace really take req.grace into account ?

2. If yes, why not call EXP_Ttl_grace() ?

doc/sphinx/reference/vcl_var.rst
```diff
@@ -304,6 +304,21 @@ req.ttl
...
+
+	Writable from: client
+
+
+	Upper limit on the object grace.
+
+	This affects the object's grace during lookup and also the
+	value of obj.grace in vcl_hit.
+
+
```
bsdphk: This needs better text.

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -121,8 +117,9 @@ you can check if the backend is sick and modify the behavior when it
 comes to grace. There are essentially two ways of doing this. You can
 explicitly deliver kept object (that is not within grace) when you see
 that the backend is sick, or you can explicitly `not` serve an expired
-object when you know that the backend is healthy. The two methods have
-slightly different characteristics, as we shall see.
+object when you know that the backend is healthy. At the current time,
+we only recommend the first strategy, as the behavior of `keep` will
+change with the next major release of Varnish Cache.
```
bsdphk: Have I heard about how keep changes ?

hermunn: Good point. Of course, the change is that *keep* objects cannot be delivered any more. If the last commit in the PR is pushed, then only the first method will work. I can move the documentation patch last if I get the other ones OK'ed, and update the text such that only the first strategy is left.

hermunn:
I have squashed some commits (the new one, *Reintroduce the req.grace variable, change keep behavior*, contains most of the work), changed the documentation substantially and added a test case that I wrote to make sure the documentation was correct.

A fresh review is in order.

I should rename the test cases before pushing. I forgot the last time, and now the naming is a bit off.

hermunn:
If @daghf can have a look, too, then I think we should be ready to make a decision. But @nigoroll, does the documentation update(s) look OK to you?

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -78,13 +74,18 @@ will consider the following:
...
+* In the following rules we know that the object is within grace, and
+  we expect the object to be delivered.
 * If there is no backend request for the object, one is scheduled and
   ``sub vcl_hit`` is called immediately.
-* If there is a backend request going on, but the object is under grace,
-  ``sub vcl_hit`` is called immediately.
-* If there is a backend request going on, but the grace has expired,
-  processing is halted until the backend request has finished and a
-  fresh object is available.
+* If there is a backend request going on, ``sub vcl_hit`` is called
```
nigoroll: the last two items are the same?

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -118,104 +127,43 @@ web- and application servers.
...
+  sub vcl_recv {
+       if (std.healthy(req.backend_hint)) {
+            // change the behavior for healthy backends: Cap grace to 10s
+            set req.grace = 10s;
+       }
+  }
 
-The effect of this is that, when the backend is healthy, objects with
+In the example above, the special variable ``req.grace`` is set.  The
+effect of this is that, when the backend is healthy, objects with
```
nigoroll: maybe remove the "of this"? ("of this is that" sounds a bit bumpy)

doc/sphinx/users-guide/vcl-grace.rst
```diff
@@ -118,104 +127,43 @@ web- and application servers.
...
-
 Summary
 ~~~~~~~
 
-Grace mode allows Varnish to deliver slightly stale content to clients while
-getting a fresh version from the backend. The result is faster load times
-with a low cost.
+Grace mode allows Varnish to deliver slightly stale content to clients
+while getting a fresh version from the backend. The result is faster
+load times with a low cost.
```
nigoroll: "at lower cost" maybe?

daghf:
+1 from me, both to @hermunn's patches and @nigoroll's edits. I don't have much to add.

I wonder if the req_fsm side of it could be made more readable if we gave `HSH_Lookup` specific return values for `HSH_HITMISS` / `HSH_HITPASS`

hermunn:
> I wonder if the req_fsm side of it could be made more readable if we gave HSH_Lookup specific return values for HSH_HITMISS / HSH_HITPASS

I agree with @daghf, but I think should be merged with @nigoroll's patch on top. I will fix the docs as proposed, amend, rebase and force push with @nigoroll's rewrite patch on top. When everything looks good I will add a comment here.

nigoroll:
@dag-erling +1 . I actually thought the same en passant, but did now want to suggest even more changes

hermunn:
Force pushed new rebased version. No changes in the actual code since the reviews from @daghf and @nigoroll , but I have changes some test case names, updated the documentation as requested by Nils, and removed the now unnecessary changes to `r02219.vtc`.

I consider this PR essentially approved, but I will let it ferment here until the bugwash on Monday. After this I want to make the PR that disallowes `return (miss)` from `sub vcl_hit`, and we should discuss if this will be `vcl 4.2;` or something else.

The goal is to back port this to Varnish Cache 4.1 and 6.0 very soon.

hermunn:
I have back ported all but the last commit to 4.1: d2d09318 721448ef a8dc8780 62163f4b

The last commit is a pure rewrite commit, and I do not expect this code to change any more in 4.1, so I leave it as it is.

PR #2717, opened 2018-06-20
Accurate ban statistics except for a few remaining corner cases

### LINKED ISSUE #2716: varnishstat MAIN.bans counter is wrong
## Expected Behavior
`MAIN.bans` counter should be around 1M (approximately)

## Current Behavior
Counter goes up to a very big value:
```
MAIN.bans                  18446744073709551579          .   Count of bans
MAIN.bans_completed                           0          .   Number of bans marked 'completed'
MAIN.bans_obj              18446744073709551572          .   Number of bans using obj.*
MAIN.bans_req                                 0          .   Number of bans using req.*
MAIN.bans_added                          509494         0.74 Bans added
MAIN.bans_deleted                        509536         0.74 Bans deleted
MAIN.bans_tested                        1268737         1.84 Bans tested against objects (lookup)
MAIN.bans_obj_killed                       6377         0.01 Objects killed by bans (lookup)
MAIN.bans_lurker_tested             19320875202     27991.41 Bans tested against objects (lurker)
MAIN.bans_tests_tested                303786326       440.12 Ban tests tested against objects (lookup)
MAIN.bans_lurker_tests_tested       19320879749     27991.42 Ban tests tested against objects (lurker)
MAIN.bans_lurker_obj_killed              143565         0.21 Objects killed by bans (lurker)
MAIN.bans_lurker_obj_killed_cutoff            0         0.00 Objects killed by bans for cutoff (lurker)
MAIN.bans_dups                           414859         0.60 Bans superseded by other bans
MAIN.bans_lurker_contention                   0         0.00 Lurker gave way for lookup
MAIN.bans_persisted_bytes              81466732          .   Bytes used by the persisted ban lists
MAIN.bans_persisted_fragmentation      81466716          .   Extra bytes in persisted ban lists due to fragmentation
```

But the counter is still working after the big jump: next second the value went from *1579 to *1580.

## Possible Solution
No idea, sorry.

I clearly don't know why it's doing this. Just observed this on a few servers, after a while (at least 2 days)

## Context
I'm monitoring bans to see if there's no congestion, so the alert goes off and is never removed. I'm using the following test:
`MAIN.bans - MAIN.bans_completed > 1000`

## Your Environment
* Version used: 6.0.0-1~stretch
* Operating System and version: Debian Stretch (stable)
* Source of binary packages used: packagecloud

nigoroll:
Fixes #2716 for most cases.

Open questions:

* make `wstat` lock visible to ban code and hold it during cli ban?
* do the same for `vcl_init` / `vcl_fini`  or will we add a worker?

From the commit message:

For ban statistics, we updated VSC_C_main directly, so if we raced
Pool_Sumstat(), that could undo our changes.

This patch fixes statistics by using the per-worker statistics
cache except for the following remaining corner cases:

* bans_persisted_* counters receive absolut updates, which does
  not seem to fit the incremental updates via the per-worker stats.

  I've kept these cases untouched and marked with comments. Worst
  that should happen here are temporary inconsistencies until the
  next absolute update.

* For BAN_Reload(), my understanding is that it should only
  happen during init, so we continue to update VSC_C_main
  directly.

* For bans via the cli, we would need to grab the wstat lock,
  which, at the moment, is private to the worker implementation.

  Until we make a change here, we could miss a ban increment
  from the cli.

* for VCL bans from vcl_init / fini, we do not have access
  to the worker struct at the moment, so for now we also
  accept an inconsistency here.

Fixes #2716 for relevant cases

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -571,7 +571,7 @@ struct sess {
 struct ban_proto *BAN_Build(void);
 const char *BAN_AddTest(struct ban_proto *,
     const char *, const char *, const char *);
-const char *BAN_Commit(struct ban_proto *b);
+const char *BAN_Commit(struct ban_proto *b, struct VSC_main *stats);
 void BAN_Abandon(struct ban_proto *b);
 
 /* cache_cli.c [CLI] */
--- a/bin/varnishd/cache/cache_ban.c
+++ b/bin/varnishd/cache/cache_ban.c
@@ -152,7 +152,7 @@ ban_equal(const uint8_t *bs1, const uint8_t *bs2)
 }
 
 void
-ban_mark_completed(struct ban *b)
+ban_mark_completed(struct ban *b, struct VSC_main *stats)
 {
 	unsigned ln;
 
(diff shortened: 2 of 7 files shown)


bsdphk:
Bugwash conclusion:

The real trouble is that CLI and VCL both frobble counters and CLI doesn't use stats-summing.

Solution: give CLI a worker struct (and in VRT_CTX)

nigoroll:
clarified with @bsdphk after bugwash 1:1 : we'll put this in and then pursue the next steps

Issue #2710, opened 2018-06-15
libvarnishapi needs soname bump


bzed:
## Current Behavior

Programs linking /usr/lib/libvarnishapi.so.1 expect LIBVARNISHAPI_1.0, but since 6.0 libvarnisharpi.so ships 
```
root@host:~# objdump -x /usr/lib/libvarnishapi.so.1 | grep API
2 0x00 0x000d1570 LIBVARNISHAPI_2.0
0000000000000000 g     O *ABS*	0000000000000000              LIBVARNISHAPI_2.0
```
## Possible Solution

Bump the soname of libvarnishapi to avoid linking issues.

## Steps to Reproduce (for bugs)
Install varnish 5, run objdump as shown above. Install varnish 6, do the same and compare the results.

## Context
Programs like collectd link libvarnishapi and fail badly due to missing symbols.

## Your Environment
Debian stretch with varnish from the official varnish repository.

bsdphk:
Merge this with #2718

PR #2718, opened 2018-06-21
Increasing libvarnishapi.so version from 1.0.6 to 2.0.0

dmatetelki:
Fixes: libvarnishapi needs soname bump #2710

Should be backported to branches: 5.0, 5.1, 5.2, 6.0

### DIFF
--- a/lib/libvarnishapi/Makefile.am
+++ b/lib/libvarnishapi/Makefile.am
@@ -9,7 +9,7 @@ AM_CPPFLAGS = \
 
 lib_LTLIBRARIES = libvarnishapi.la
 
-libvarnishapi_la_LDFLAGS = $(AM_LDFLAGS) -version-info 1:6:0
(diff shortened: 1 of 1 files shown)


bsdphk:
Bugwash conclusion: Merge in -trunk, no backporting

bzed:
@bsdphk please backport to 6.0. Backporting to 5.x is wrong imho - but 6.0 is broken as it is at the moment. Thanks!

bzed:
@bsdphk According to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=879471 5.2 is broken, too

dmatetelki:
> martin | do we agree that 5.x and 6.0 will remain broken with regard to the so-name of libvarnishapi?                              martin | i believe that is the right move, as to not confure anyone with regard to actual updates on those branches 

the answer was yes

PR #2723, opened 2018-06-28
vmod references and handles

nigoroll:
* dlopen vmods only once and handle multiple _Inits via the refcount
* add a simple interface for taking references to vmods
* optionally hand vmods their own handle (opaque to them)

### DIFF
--- a/bin/varnishd/cache/cache_vrt_vmod.c
+++ b/bin/varnishd/cache/cache_vrt_vmod.c
@@ -64,7 +64,11 @@ struct vmod {
 	unsigned		vrt_minor;
 };
 
+/* the vmods list is owned by the cli thread */
 static VTAILQ_HEAD(,vmod)	vmods = VTAILQ_HEAD_INITIALIZER(vmods);
+
+/* protects all (struct vmod).ref */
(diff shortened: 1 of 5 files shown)


bsdphk:
Can you explain why you need to refcount vmods separate from the VCLs which load them ?

nigoroll:
I have also reflected on this for some time:
* the current model is that the VCL owns directors and vmods, so if we finalize directors before vmods when the VCL goes cold, we should be fine.
* For one thing, we seem not to do this in the right order at the moment. I have not checked why exactly, because I thought we needed a broader solution
* ... which would cater for the option of a (director) vmod keeping references to directors across vmods: If vcl1 loads director vmod a and b, with b referencing a, then if we load vcl2 which only loads vmod b, we would need to keep vmod a

But I am not so sure any more if the latter would be a good idea, so I tend to think that understanding the fini order issue and fixing it specifically would be a better idea?
directors with a wider scope than vmods would then (continue to) be off limits.

Feedback welcome

nigoroll:
this pr is dead, the premise that directors could outlive a vcl is wrong.

PR #2726, opened 2018-07-01
Send PROXY UNKNOWN for PROXYv1 in probes to non-IP backends.

### LINKED ISSUE #2702: Wrong turn at cache/cache_backend_probe.c:255: Unknown family
See also #2151 and #2244.

```vcl
varnishtest "unix proxy"

server s1 "" -start

varnish v1 -proto PROXY -arg "-a ${tmpdir}/v1.sock" -vcl+backend {
        sub vcl_recv {
                # XXX: needs a relevant condition, this is broken logic
                if (client.ip != remote.ip || server.ip != local.ip) {
                        return (synth(400));
                }
                return (synth(200));
        }
} -start

varnish v2 -vcl {
        probe default {
                .window = 1;
                .initial = 1;
                .threshold = 1;
                .interval =0.5s;
        }

        backend v1 {
                .path = "${tmpdir}/v1.sock";
                .proxy_header = 1;
        }
} -start

client c1 -connect ${v2_sock} {
        txreq
        rxresp
        expect resp.status == 200
} -run

delay 1

varnish v2 -cliexpect "vcl1.v1[ ]+probe[ ]+1/1[ ]+good" backend.list
```

Setting `.proxy_header = 2` fails in different interesting ways.

slimhazard:
That is, if the backend socket's address family is neither of AF_INET or AF_INET6.

This means that UNKNOWN is sent in probes to UDS backends when ``.proxy_header=1`` is set.

Also verify that a UDS backend receives PROXY LOCAL in a probe when ``.proxy_header=2`` is set (this was already implemented, but not yet tested for UDS backends/probes).

Fixes #2702

### DIFF
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -269,8 +269,9 @@ vbp_write_proxy_v1(struct vbp_target *vt, int *sock)
 	else if (ss.ss_family == AF_INET)
 		VSB_printf(&vsb, " TCP4 ");
 	else
-		WRONG("Unknown family");
-	VSB_printf(&vsb, "%s %s %s %s\r\n", addr, addr, port, port);
+		VSB_printf(&vsb, " UNKNOWN\r\n");
+	if (ss.ss_family == AF_INET6 || ss.ss_family == AF_INET)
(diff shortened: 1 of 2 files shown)


bsdphk:
LGTM2

PR #2728, opened 2018-07-06
New %{Varnish:date}x varnishncsa format

Dridi:
This format may help spot latency problems caused by other proxies sitting in front of Varnish, as long as they leave the original Date header unmodified. With this format it becomes more convenient to compare the Date header with the `Timestamp:Start[1]` entry of an rxreq transaction.

This feature was accepted in principle during the last VDD and this is my first time in the varnishncsa code base. I think I could polish a couple things here and there if this passes the review.

In particular, if there is no Date header this format falls back to a plain `-`. But looking at how other formats `strdup` the fallback string and check for its presence I think we have a (negligible) waste of resources and a bit of dead code.

### DIFF
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -89,6 +89,7 @@ enum e_frag {
 	F_tend,			/* Time end */
 	F_ttfb,			/* %{Varnish:time_firstbyte}x */
 	F_host,			/* Host header */
+	F_date,			/* Date header */
 	F_auth,			/* Authorization header */
 	F__MAX,
 };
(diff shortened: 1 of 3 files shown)


Dridi:
@slimhazard suggested to generalize this to any header containing a date, I will update the pull request accordingly.

The proposed syntax is `%{Varnish:Timestamp:<header>}x`. I suggested `%{Timestamp:<header>}x` on IRC but that wouldn't make it Varnish-specific so here goes the longer name.

Examples:

- `%{Varnish:Timestamp:Date}x`
- `%{Varnish:Timestamp:Last-Modified}x`
- `%{Varnish:Timestamp:If-Modified-Since}x`

PR #2732, opened 2018-07-17
Accommodate musl libc that has a smaller stack

gquintard:
reviving this: https://git.alpinelinux.org/cgit/aports/tree/main/varnish/musl-fix-stack-overflow.patch

Alpine uses musl libc that has a smaller stack, so Varnish compiles but can't run with that big a number of events (8192). phk offered that maybe the NEEV number could need to be revised and we should have a look at what others do.

- nginx has it configurable, set to 512 by default, and I found no recommendation about changing those.
- haproxy also has it configurable with a 200 default, and says this:

```
tune.maxpollevents <number>
  Sets the maximum amount of events that can be processed at once in a call to
  the polling system. The default value is adapted to the operating system. It
  has been noticed that reducing it below 200 tends to slightly decrease
  latency at the expense of network bandwidth, and increasing it above 200
  tends to trade latency for slightly increased bandwidth.
```

We are certainly a bit too ambitious with 8192, but I think that whatever the value is, we could do with a malloc approach to make sure those kind of issues don't break "obscure" platforms when we tweak it.

### DIFF
--- a/bin/varnishd/waiter/cache_waiter_epoll.c
+++ b/bin/varnishd/waiter/cache_waiter_epoll.c
@@ -33,6 +33,8 @@
 //lint -e{766}
 #include "config.h"
 
+#include <stdlib.h>
+
 #if defined(HAVE_EPOLL_CTL)
 
 #include <sys/epoll.h>
@@ -69,7 +71,7 @@ struct vwe {
 static void *
 vwe_thread(void *priv)
 {
-	struct epoll_event ev[NEEV], *ep;
+	struct epoll_event *ev, *ep;
 	struct waited *wp;
 	struct waiter *w;
 	double now, then;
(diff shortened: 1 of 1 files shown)


bin/varnishd/waiter/cache_waiter_epoll.c
```diff
@@ -30,6 +30,8 @@
  * - thank you, Marc Alexander Lehmann
  */
 
+#include <stdlib.h>
```
Dridi: Should move after `config.h`, did you put it here because it would otherwise fail to build?

gquintard:
Nope, because my OCD compels me put system headers on top, but you are right.

Shouldn't we just add "-I config.h" to the CCPFLAGS and stop including that file manually?

bin/varnishd/waiter/cache_waiter_epoll.c
gquintard: fixed and force pushed

fgsch:
LGTM.

Explicit is always better than implicit, so re "-I config.h", 👎 .

rezan:
Why not make `NEEV` tunable here?

gquintard:
From what I saw, this parameter is really changed on other software, and we already have a big number of tunables that are never used, so I'd vote against that.

But if we make it tunable, all the more reason to apply that patch.

Dridi:
Unofficial bugwash consensus: merge as-is and turn into a tunable later if needed.

PR #2733, opened 2018-07-19
Add VRT functions for STRANDS.

slimhazard:
Re-implementations of the current functions for STRING_LIST.

Also removes VRT_STRANDS_string() from vrt.h, which was never implemented.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -339,6 +339,114 @@ VRT_CollectString(VRT_CTX, const char *p, ...)
 	return (b);
 }
 
+/*--------------------------------------------------------------------
+ * Collapse STRANDS into the space provided, or return NULL
+ */
+
+char *
+VRT_Strands(char *d, unsigned dl, VCL_STRANDS s)
+{
+	char *b;
+	const char *e;
+	unsigned x;
+
+	b = d;
+	e = b + dl;
+	for (int i = 0; i < s->n && b < e; i++)
+		if (s->p[i] != NULL && *s->p[i] != '\0') {
+			x = strlen(s->p[i]);
+			if (b + x < e)
+				memcpy(b, s->p[i], x);
+			b += x;
+		}
+	if (b >= e)
+		return (NULL);
+	*b++ = '\0';
+	return (b);
+}
+
+/*--------------------------------------------------------------------
+ * Copy and merge STRANDS into a workspace.
+ */
+
+VCL_STRING
+VRT_StrandsWS(struct ws *ws, const char *h, VCL_STRANDS s)
+{
+	char *b;
+	const char *q = NULL, *e;
+	VCL_STRING r;
+	unsigned u, x;
+	int i;
+
+	u = WS_Reserve(ws, 0);
+
+	for (i = 0; i < s->n; i++)
+		if (s->p[i] != NULL && *s->p[i] != '\0') {
+			q = s->p[i];
+			break;
+		}
+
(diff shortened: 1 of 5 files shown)


slimhazard:
I would also suggest that in the user-visible docs generated as RST from VCC, we show STRANDS as STRING, just as we do with STRING_LIST.

As far as a VCL author is concerned, a concatenated string is just a string. Trying to explain why a STRING is sometimes STRANDS is likely to confuse them more than it enlightens them.

bin/varnishd/cache/cache_vrt.c
```diff
@@ -339,6 +339,119 @@ VRT_CollectString(VRT_CTX, const char *p, ...)
...
+			q = s->p[i];
+			break;
+		}
+
+	if (h != NULL && q == NULL && WS_Inside(ws, h, NULL)) {
+		WS_Release(ws, 0);
+		return (h);
+	}
+
+	if (h == NULL && q != NULL && s->n == 1 && WS_Inside(ws, q, NULL)) {
```
nigoroll: what `if s->n == 2 && s->p[0] == NULL`
I think we should basically have the same logic as `VRT_String()`.

Skipped reviewing the rest of this function

nigoroll:
Yes, I'd really like getting off the ground with strands

bin/varnishd/cache/cache_vrt.c
```diff
@@ -339,6 +339,119 @@ VRT_CollectString(VRT_CTX, const char *p, ...)
...
 }
 
+/*--------------------------------------------------------------------
+ * Collapse STRANDS into the space provided, or return NULL
+ */
+
+char *
+VRT_Strands(char *d, unsigned dl, VCL_STRANDS s)
+{
+	char *b, *e;
```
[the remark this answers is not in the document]
slimhazard: Can be done

bin/varnishd/cache/cache_vrt.c
slimhazard: That case is covered in lines 403-411 further down. The idea in L393 is that this is likely to be a very common special case -- there was only one string in the STRANDS object.

And I would have said that the part in lines 393 to 412 are very similar to the same part of ``VRT_String()``, after changing the varags and ``vrt_magic_string`` stuff for STRANDS.

Maybe more so than it needs to be in fact, because this part could be simplified, and that might involve moving the L393 case further down. I'll try a simplification.

slimhazard:
I have some code stashed that would use STRANDS for vmod_blob instead of STRING_LIST, and I think it would be a welcome simplification. But it would need ``VRT_CollectStrands()`` from this patch.

An interesting question is whether we'd want to move more STRING_LISTs to STRANDS throughout Varnish, especially for something as central as ``VRT_SetHdr()``. The ``sethdr`` function in vmod_debug for this patch is meant to give that a look.

If we decide *not* to use STRANDS to set headers, then we probably don't need the ``h`` parameter in ``VRT_StrandsWS()`` (and that function could then be simplified further).

I "like" STRANDS better than the varargs stuff, but it's hard to justify that with more than a subjective aesthetic. The code just feels more straightforward than fooling around with ``va_list`` and ``va_start`` and ``vrt_magic_string_end`` and all that. (When I look at the machine code generated for varargs as opposed to STRANDS, the STRANDS code looks simpler at that level as well, but whatever that may be worth.)

bsdphk:
VRT_Strands should use size_t for the length

All these functions should AN() their pointer arguments.

Otherwise: OK

slimhazard:
We should consider the same changes (size_t and AN()) for the VRT STRING_LIST functions, since the new ones just imitate the old ones.

PR #2737, opened 2018-07-27
Add std.fnmatch()

slimhazard:


fgsch:
I'm not necessarily against this but I wonder why we need to have both regexp and this in core (as opposed to have this in an external vmod).

Can you share your thoughts?

slimhazard:
@fgsch that's a fair question, and you may be right. Some thoughts:

This would become an alternative for VCL authors who are more comfortable with shell/wildcard syntax than with regex. (I'd like for everyone to be fluent with regexen, but experience and reality turn out differently.)

I think that this might make a common error less common -- writing ``/foo/.*/bar`` with the intention of having only one path component in the middle, although it matches a string like ``/foo/baz/quux/bar``. With the default params, ``/foo/*/bar`` will get that right. It's deliberate that the docs and examples in the patch bring that up.

Generally, I think ``fnmatch(3)`` is a good fit for matching URLs.

And I expect that it's universally available, which is one of the reasons why we include things in VMOD std. If it required an external library, then I'd agree with you, it should be in a 3rd party VMOD and not in the Varnish tree at all. Of course these things are surprising sometimes, but I doubt that Varnish will ever be built for a platform that doesn't support ``fnmatch(3)``.

slimhazard:
Another thing about the regex ``/foo/.*/bar`` when what you really wanted was ``/foo/[^/]+/bar`` -- probably most of the time, that would fail a match in cases for which you'd want it to fail, coincidentally as it were. But because of ``.*``, the regex matcher is obliged to search the URL all the way to the end before it determines no match. If that's happening thousands of times per second, a lot of work is being done for no good purpose.

As I said, I wish we could show everyone how to get the regex right.

But this way, they could also say ``/foo/*/bar`` with ``fnmatch``, and the search stops as soon as it's clear that the URL won't match.

nigoroll:
I also have a use case where some token contains a glob pattern, which, as the token is dynamic, is also dynamic per se. To avoid custom code, the only other option would be to transform the glob pattern to a dynamic regex (see vmod_re `match_dyn` https://code.uplex.de/uplex-varnish/libvmod-re/blob/master/src/vmod_re.vcc#L179) which, unless `fnmatch` was terribly inefficient, does not sound terribly attractive

bsdphk:
OK with me, though I wonder if it should be called "glob" instead ?

slimhazard:
Naming is not entirely easy for this.

"glob" has connotations of generating a set of matching filenames -- there are functions named "glob" in Python, Ruby, PHP and Perl that do that.

"match" is not distinguished well from regex matching. Something like "shellmatch" or "globmatch" or "wildcardmatch" gets to be a mouthful.

I'm not exactly crazy about "fnmatch", since VCL authors who don't know from POSIX functions won't be familiar with it, and it's not really about filenames. But it feels like a better description than the alternatives, and functions named "fnmatch" with this purpose exist in Python, Ruby and PHP.

We already have "strstr" in VMOD std, and I guess using the library names is probably best for clarity.

Issue #2668, opened 2018-05-03
u00008.vtc systematic timeout


Dridi:
For some reason u8 always hangs and eventually times out after a minute. This should be the last of my "just upgraded to Fedora 28" series (with #2666 and #2667). Here is a command (pun intended) that shows it in action:

```
$ bin/varnishtest/varnishtest -i bin/varnishtest/tests/u00008.vtc -t10 -n32 -j4 -kq                                                              
#    top  TEST bin/varnishtest/tests/u00008.vtc TIMED OUT (kill -9)
#    top  TEST bin/varnishtest/tests/u00008.vtc FAILED (10.003) signal=9
#    top  TEST bin/varnishtest/tests/u00008.vtc TIMED OUT (kill -9)
#    top  TEST bin/varnishtest/tests/u00008.vtc FAILED (10.002) signal=9
#    top  TEST bin/varnishtest/tests/u00008.vtc TIMED OUT (kill -9)
#    top  TEST bin/varnishtest/tests/u00008.vtc FAILED (10.002) signal=9
[...]
32 tests failed, 0 tests skipped, 0 tests passed
$ grep ^CURSES_LIBS= config.log
CURSES_LIBS='-lncursesw -ltinfo '
$ rpm -qf /usr/lib64/lib{ncursesw,tinfo}.so
ncurses-devel-6.1-4.20180224.fc28.x86_64
ncurses-devel-6.1-4.20180224.fc28.x86_64
```

bsdphk:
Try to find out which -expect-text fails and why ?

bsdphk:
Is this still a problem?

Dridi:
Still happening on an up to date Fedora 28 for current master. I'll try to figure what exactly is timing out.

bsdphk:
Couldn't reproduce on FreeBSD-12-CURRENT

bsdphk:
The problem is the lack of support from "CSI # b" "repeat previous graphic # times" in teken

bsdphk:
Note to self: Update teken from FreeBSD source tree

PR #2741, opened 2018-08-16
Make it possible to change ncurses update rate.

lkarsten:
At times the 1.0s update rate makes counters increase in uneven chunks. In period increasing by 3, the next period by 200. 

This makes it hard to see what is going on. Making it possible to increase the refresh rate leads to longer averaging period and more stable numbers.

This commit introduces ncurses keys + and - (ala bwm-ng), adjusting the refresh interval.


lkarsten:
Yeah, this breaks u00008.vtc and possibly other tests.

Not immediately obvious what is wrong. Help fixing the test case would be appreciated.

Dridi:
u8 is already broken on recent linux systems, see #2668.

bsdphk:
This sounds like a good idea to me.

Make sure to cover the code in a test-case.

bin/varnishstat/varnishstat_curses.c
```diff
@@ -899,6 +901,13 @@ handle_keypress(int ch)
 		current = n_ptarray - 1;
 		page_start = (current - l_points) + 1;
 		break;
+	case '+':
+		interval = interval + 0.1;
```
nigoroll: just picking nits: `interval += 0.1;`

bin/varnishstat/varnishstat_curses.c
```diff
@@ -899,6 +901,13 @@ handle_keypress(int ch)
...
+		break;
+	case '-':
+		if (interval - 0.1 > 0.1)
```
nigoroll: feeling bad about being so picky:

```
interval -= 0.1;
if (interval < 0.1)
    interval = 0.1;
```

doc/sphinx/reference/varnishstat.rst
```diff
@@ -109,6 +109,13 @@ The following keys control the interactive display:
 <CTRL+T>
 	Sample now.
 
+<+>
+   Increase refresh interval.
```
nigoroll: tab intendentation

lkarsten:
Hi.

Thanks for the feedback. I'll look into it.

When I started adding a test case for this, I noticed that the text was hidden in 80x25. My terminal is usually larger.

I need to redo that bit as well, to get proper test case cover.

lkarsten:
@nigoroll Will you please take another look at this?

lkarsten:
Just saw the notification on skipping varnishtest. that part is trivial. i'm happy either way.

This work is in part sponsored by the kind man in seat 97D that allowed me legroom to work on this on my way home from vacation. :-)

nigoroll:
can you please rebase, squash and force-push?

PR #2744, opened 2018-08-17
Don't mess with C-L when responding to HEAD

gquintard:
all HEAD responses can have a C-L, not just pass+HEAD


bin/varnishtest/tests/b00065.vtc
```diff
@@ -0,0 +1,20 @@
+varnishtest "Test that C-L on HEAD request aren't nuked"
+
+server s1 {
+	rxreq
+	txresp -nolen -hdr "content-length: 17"
+	expect req.method == "HEAD"
```
fgsch: You have this backwards.

Dridi: I personally prefer to have the `rxreq`'s `expect` happening after a `txresp` when relevant: that removes opportunities for tests timing out after a long time when someone else is waiting for that response and prevents the overall test case from failing fast.

bsdphk:
Looks good.

Should be mentioned clearly in relnotes

Issue #2754, opened 2018-08-30
high number of objects with hit-for-miss


nigoroll:
early ticket, details still missing:

For a popular object with high request rates, we are seeing a very high number of objects (~500K - ~800K) behind a single objhead.

slimhazard:
Possibly related to (or a dup of?) #2653 and #2654.

martin-uplex:
In a cluster of 8 servers with varnish-varnish recursion we observed n_objecthead * 4 = n_object on two servers, while all other roughly showed n_objecthead = n_object. Of course Vary was our best guess, but we could not figure our any issue using varnishlog. So we decided to narrow the problem down using bans. This very specific ban dropped the number of objects down to the number of n_objectheads - by about 1,2m (from 1,6m to about 400k):

varnishadm ban 'obj.http.Vary ~ ^X-DevClass,Accept-Encoding,User-Agent$ && obj.http.X-Varnish-Bereq-Host == www.example.com && obj.http.X-Varnish-Bereq-URL == /onespecificbusyurl'

We double checked non of the Vary-Headers could have caused the large amount of objects (X-DevClass has a variety of only 3, Accept-Encoding is handled by varnish and User-Agent is deleted in recv). Also hash_data is not extended in this case.

It turned out these requests are all hit_for_miss, so it looks like there is a problem cleaning up hit for miss (grace) objects in certain conditions?


Details:

varnishd -V (Version of slink, 4 Aug. 2018):
varnishd (varnish-trunk revision 029751ba81d5dd2105b86cb86da0626fa71a5407)
Copyright (c) 2006 Verdens Gang AS
Copyright (c) 2006-2015 Varnish Software AS

request rate to /onespecificbusyurl is about 50/s
ttl = 120s
grace = 16h
uncacheable + deliver -> hit_for_miss


Here's a modified/anonymized except of varnishlog:

```
*   << Request  >> 718341316 
-   Begin          req 706582603 rxreq
-   Timestamp      Start: 1535635332.634153 0.000000 0.000000
-   Timestamp      Req: 1535635332.634153 0.000000 0.000000
-   ReqStart       XXX.XXX.XXX.XXX 27395 a3
-   ReqMethod      GET
-   ReqURL         /onespecificbusyurl
-   ReqProtocol    HTTP/1.1
-   ReqHeader      Host: www.example.com
-   ReqHeader      User-Agent: Mozilla/5.0 XXX
-   ReqHeader      Accept: */*
[...]

PR #2755, opened 2018-08-31
Hsh cleanup - no semantic changes

nigoroll:
in preparation of #2754 but I also think generally sensible and overdue

None of these patches should change anything semantically, the actual fixes and improvements are to follow.


nigoroll:
fixed, the last commit is to be squashed before merge

nigoroll:
@bsdphk is currently working on a minimal for for #2754, so I will reopen the cleanup PR once that's done.

Issue #2763, opened 2018-09-05
panic in vbf_stp_mkbereq(): Condition((bo->stale_oc->flags & OC_F_PASS) == 0) not true.


nigoroll:
```
Panic at: Wed, 05 Sep 2018 12:39:08 GMT
Assert error in vbf_stp_mkbereq(), cache/cache_fetch.c line 198:
  Condition((bo->stale_oc->flags & OC_F_PASS) == 0) not true.
version = varnish-trunk revision 99cd38f0b9f12f1d78c1f8789dad5d5a4d8b4f5a, vrt api = 7.0
ident = Linux,3.10.0-327.13.1.el7.x86_64,x86_64,-jnone,-smalloc,-sdefault,-hcritbit,epoll
now = 74308522.401266 (mono), 1536151106.734342 (real)
Backtrace:
  0x44298a: /opt/varnish/sbin/varnishd() [0x44298a]
  0x442e7b: /opt/varnish/sbin/varnishd() [0x442e7b]
  0x4c0a35: /opt/varnish/sbin/varnishd(VAS_Fail+0x4d) [0x4c0a35]
  0x42bc17: /opt/varnish/sbin/varnishd() [0x42bc17]
  0x42e987: /opt/varnish/sbin/varnishd() [0x42e987]
  0x467d92: /opt/varnish/sbin/varnishd() [0x467d92]
  0x466e3e: /opt/varnish/sbin/varnishd() [0x466e3e]
  0x467ec8: /opt/varnish/sbin/varnishd() [0x467ec8]
  0x7fcac773de25: /lib64/libpthread.so.0(+0x7e25) [0x7fcac773de25]
  0x7fcac7467bad: /lib64/libc.so.6(clone+0x6d) [0x7fcac7467bad]
thread = (cache-worker)
thr.req = (nil) {
},
thr.busyobj = 0x7fc9b7908020 {
  ws = 0x7fc9b7908060 {
    id = \"bo\",
    {s, f, r, e} = {0x7fc9b7909f60, +1920, (nil), +123032},
  },
  retries = 0, flags = {do_stream},
  vfc = 0x7fc9b7909f18 {
    failed = 0,
    req = (nil),
    resp = (nil),
    wrk = (nil),
    oc = (nil),
    obj_flags = 0x0,
  },
  director_req = 0x7fcac6812680 {
    cli_name = boot.a0unset,
    health = sick,
    admin_health = probe, changed = 1536150990.251284,
    type = backend {
[...]

PR #2766, opened 2018-09-06
HFM+IMS Variant A: Do not copy the IMS attribute

nigoroll:
see #2763 for discussion


nigoroll:
merged in even simpler form as addea755c1d41e0431730da666363f76ad00990a

PR #2767, opened 2018-09-06
HFM+IMS Variant B: keep the body and use for 304

nigoroll:
See #2763 for disussion


nigoroll:
to be picked up again in october

nigoroll:
ditched after minimal discussion on IRC. phk verdict: 

> I really dont think we should IMS on non-cacheable

PR #2771, opened 2018-09-07
Clarify and test object slimming for hfp+hfm

### LINKED ISSUE #2768: recheck object slimming (not a bug)
quick note before I forget:
https://github.com/varnishcache/varnish-cache/blob/5e838844ab3772ad4c961c94292efff68b974dd3/bin/varnishd/cache/cache_req_fsm.c#L420
Depending on #2763 we might want to add `| OC_F_HFP` ?

nigoroll:
issuing PR with respect to the pre-release freez-ish period

The previous code was correct already, but we can make it clearer that HFP implies OC_F_PASS

Also test explicitly that both HFM and HFP have their objects slimmed.

Closes #2768


nigoroll:
merged because it is just a clarification

PR #2776, opened 2018-09-14
Simple addition on req.grace

hermunn:
As far as I can see, the documentation is correct, but I suggest a slight update to the text. Take it or leave it.


Dridi:
LGTM

Issue #2775, opened 2018-09-13
net::ERR_CONNECTION_CLOSED on Chrome 69


CarloCannas:
Some resources requested on h2 may fail to load with error net::ERR_CONNECTION_CLOSED using Chrome 69.

I received yesterday the latest Chrome upgrade (Chrome 69) and I started noticing some missing images on a website that uses Varnish with HTTP/2 enabled. The problem is not always reproducible: it may load everything fine without errors, but on some pages the error shows up after a few reloads.
I was able to reproduce it only on Chrome 69, previous Chrome versions or Firefox seems unable to trigger the bug.
I dumped the HTTP/2 traffic between the browser and Varnish: it showed a PRIORITY frame from Chrome, followed immediately by a GOAWAY from Varnish and the closing of the TCP stream.
The PRIORITY frame was sent to a stream just closed.
That's exactly the problem: Varnish doesn't accept PRIORITY frames on a closed stream and treats them as a protocol violation.
This is however not a protocol violation: [rfc7540,l,1130,1131](https://tools.ietf.org/html/rfc7540#page-21) states that

> PRIORITY can be sent and received in any stream state

[bin/varnishd/http2/cache_http2_proto.c:L928](https://github.com/varnishcache/varnish-cache/blob/6fdc89c1fdd12bb83fadd8412faa1d7703761e36/bin/varnishd/http2/cache_http2_proto.c#L928) is the culprit: that return ultimately triggers the GOAWAY, this codepath is reached whenever a PRIORITY frame is received and the corresponding stream doesn't appear in h2->streams.

A minimal testcase which shows the issue:

```
varnishtest "PRIORITY on closed stream"

server s1 {
	rxreq
	txresp
}

varnish v1 -vcl+backend {} -start

varnish v1 -cliok "param.set feature +http2"
varnish v1 -cliok "param.set debug +syncvsl"

client c1 {
	stream 1 {
		txreq
		rxresp

		delay .1

		txprio
	} -run
	stream 3 {
		txreq
		rxresp
	} -run
} -run
```

dugwood:
Seems that 6.0.1 and 6.1.0 are both affected. I've tried to find out if other «big» websites are impacted, for example https://www.fastly.com/customers and can't reproduce.

Perhaps they're still using 5.2 release? Hum, no, just tested 5.2.1 and it's buggy too.

Any plans to release this shortly?

Thanks for the patch!

CarloCannas:
[...]

PR #2781, opened 2018-09-24
Allow PRIORITY frames on closed streams

CarloCannas:
As described in #2775, Varnish doesn't allow PRIORITY frames on closed streams, but it should.

Since there's no priority handling at all and PRIORITY frames are essentially discarded I thought that preventing the whole [stream creation section](https://github.com/varnishcache/varnish-cache/blob/6fdc89c1fdd12bb83fadd8412faa1d7703761e36/bin/varnishd/http2/cache_http2_proto.c#L926-L941) to be run upon receiving PRIORITY frame could be a good way to solve the issue.

Reading the specifications carefully I also noticed that this solution fixes some PRIORITY frame related issues:
Currently when Varnish receives a PRIORITY frame it creates a new h2_req object, which increases the concurrent streams counter. So we can get a REFUSED_STREAM if enough PRIORITY frames are sent.
This is however a wrong behaviour: [the spec](https://tools.ietf.org/html/rfc7540#section-5.1.2) states that only streams in "open" or "half-closed" states count toward the concurrent limit.
Since the proposed fix doesn't create new h2_req objects on PRIORITY frames, this is resolved.
Another thing is highest_stream: it is now updated even on PRIORITY frames, but the [RFC only lists](https://tools.ietf.org/html/rfc7540#section-5.1.1) HEADERS and PUSH_PROMISE.
This is also solved.

These changes caused some test to fail, so two preparatory commits were needed.

I'm only unsure on what to do for https://github.com/varnishcache/varnish-cache/blob/6fdc89c1fdd12bb83fadd8412faa1d7703761e36/bin/varnishd/http2/cache_http2_proto.c#L1028-L1033: the current changes make that untrue and unneeded.
Maybe something like this?
```patch
@@ -990,7 +990,6 @@ static int
 h2_sweep(struct worker *wrk, struct h2_sess *h2)
 {
        int tmo = 0;
-       int nprio = 0;
        struct h2_req *r2, *r22;
 
        ASSERT_RXTHR(h2);
@@ -1022,20 +1021,18 @@ h2_sweep(struct worker *wrk, struct h2_sess *h2)
                        }
                        break;
                case H2_S_IDLE:
-                       /* This stream was created from receiving a
-                        * PRIORITY frame, and should not be counted
-                        * as an active stream keeping the connection
-                        * open. */
-                       AZ(r2->scheduled);
-                       nprio++;
-                       break;
+                       /* Current code make this unreachable: h2_new_req is
+                        * only called inside h2_rx_headers, which immediately
+                        * sets the new stream state to H2_S_OPEN */
+                       /* FALLTHROUGH */
                default:
+                       WRONG("Wrong h2 stream state");
                        break;
                }
        }
        if (tmo)
                return (0);
-       return ((h2->refcnt - nprio) > 1);
+       return (h2->refcnt > 1);
 }

```

Please let me know what you think.

### DIFF
--- a/bin/varnishd/http2/cache_http2_proto.c
+++ b/bin/varnishd/http2/cache_http2_proto.c
@@ -382,7 +382,7 @@ h2_rx_priority(struct worker *wrk, struct h2_sess *h2, struct h2_req *r2)
 
 	(void)wrk;
 	ASSERT_RXTHR(h2);
-	xxxassert(r2->stream & 1);
+	(void)r2;
 	return (0);
 }
 
@@ -616,7 +616,21 @@ h2_rx_headers(struct worker *wrk, struct h2_sess *h2, struct h2_req *r2)
 	size_t l;
 
 	ASSERT_RXTHR(h2);
+
+	if (r2 == NULL) {
+		if (h2->rxf_stream <= h2->highest_stream)
+			return (H2CE_PROTOCOL_ERROR);	// rfc7540,l,1153,1158
+		if (h2->refcnt >= h2->local_settings.max_concurrent_streams) {
+			VSLb(h2->vsl, SLT_Debug,
+			     "H2: stream %u: Hit maximum number of "
+			     "concurrent streams", h2->rxf_stream);
+			return (H2SE_REFUSED_STREAM);	// rfc7540,l,1200,1205
+		}
+		h2->highest_stream = h2->rxf_stream;
+		r2 = h2_new_req(wrk, h2, h2->rxf_stream, NULL);
+	}
 	AN(r2);
+
 	if (r2->state != H2_S_IDLE)
 		return (H2CE_PROTOCOL_ERROR);	// XXX spec ?
 	r2->state = H2_S_OPEN;
@@ -927,23 +941,6 @@ h2_procframe(struct worker *wrk, struct h2_sess *h2, h2_frame h2f)
 	    !(r2 && h2->new_req == r2->req && h2f == H2_F_CONTINUATION))
 		return (H2CE_PROTOCOL_ERROR);	// rfc7540,l,1859,1863
 
-	if (r2 == NULL && h2f->act_sidle == 0) {
-		if (h2->rxf_stream <= h2->highest_stream)
-			return (H2CE_PROTOCOL_ERROR);	// rfc7540,l,1153,1158
-		if (h2->refcnt >= h2->local_settings.max_concurrent_streams) {
-			VSLb(h2->vsl, SLT_Debug,
-			     "H2: stream %u: Hit maximum number of "
-			     "concurrent streams", h2->rxf_stream);
-			// rfc7540,l,1200,1205
-			h2_tx_rst(wrk, h2, h2->req0, h2->rxf_stream,
-				H2SE_REFUSED_STREAM);
-			return (0);
-		}
-		h2->highest_stream = h2->rxf_stream;
-		r2 = h2_new_req(wrk, h2, h2->rxf_stream, NULL);
-		AN(r2);
-	}
-
 	h2e = h2f->rxfunc(wrk, h2, r2);
 	if (h2e == 0)
 		return (0);
@@ -993,7 +990,6 @@ static int
 h2_sweep(struct worker *wrk, struct h2_sess *h2)
 {
 	int tmo = 0;
-	int nprio = 0;
 	struct h2_req *r2, *r22;
 
 	ASSERT_RXTHR(h2);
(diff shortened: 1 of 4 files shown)


bin/varnishd/http2/cache_http2_proto.c
```diff
@@ -382,7 +382,7 @@ h2_rx_priority(struct worker *wrk, struct h2_sess *h2, struct h2_req *r2)
 
 	(void)wrk;
 	ASSERT_RXTHR(h2);
-	xxxassert(r2->stream & 1);
+	(void)r2;
```
Dridi: Maybe something like this instead of not using `wrk` and `r2`?

```c
ASSERT_RXTHR(h2);
CHECK_OBJ_ORNULL(r2, H2_REQ_MAGIC);
if (r2 != NULL)
        AN(r2->stream & 1);
```

CarloCannas: I only touched that line because the changes may make r2 null there and since a [odd stream number check](https://github.com/varnishcache/varnish-cache/blob/6fdc89c1fdd12bb83fadd8412faa1d7703761e36/bin/varnishd/http2/cache_http2_proto.c#L911) is already in place in h2_procframe I removed it. It looks useless to me, and a bit wrong: ok, Varnish doesn't support push promise streams, but PRIORITY frames aren't incompatible with these. Also, why perform a check like this only in h2_rx_priority and not in any other h2_rx_*?
I digged into commit history: that assert [got added in the very first commit](https://github.com/varnishcache/varnish-cache/blob/66d2682647b8fc00c32f69bb4b1856d7d4dcc30c/bin/varnishd/http2/cache_http2_proto.c#L333) of HTTP2, at the time there wasn't the check in h2_procframe (in fact, there was no h2_procframe, only h2_rxframe) but a similar assert was [inside h2_rx_headers](https://github.com/varnishcache/varnish-cache/blob/66d2682647b8fc00c32f69bb4b1856d7d4dcc30c/bin/varnishd/http2/cache_http2_proto.c#L395). Then, [a few months later](https://github.com/varnishcache/varnish-cache/commit/77cab3676155ac3f45db4176795f5621e12be8e6) the assert in h2_rx_headers got removed and [the check we have today appeared in h2_rxframe](https://github.com/varnishcache/varnish-cache/blob/77cab3676155ac3f45db4176795f5621e12be8e6/bin/varnishd/http2/cache_http2_proto.c#L524-L531), it got essentially untouched since then.
I think the assert in h2_rx_priority should have been removed in that same commit, along with the one in h2_rx_headers.

If we do not want to leave h2_rx_priority completely empty we could add something else: for example we could enforce [rfc7540,l,1952,1953](https://tools.ietf.org/html/rfc7540#section-6.3)
```patch
 static h2_error v_matchproto_(h2_frame_f)
 h2_rx_priority(struct worker *wrk, struct h2_sess *h2, struct h2_req *r2)
 {

        (void)wrk;
        ASSERT_RXTHR(h2);
        (void)r2;
+       if (h2->rxf_len != 5) // rfc7540,l,1952,1953
+               return (H2SE_FRAME_SIZE_ERROR);
        return (0);
 }
```

The CHECK_OBJ_* asserts look nice, however if we are going to add them I think they should go in a commit of their own. Also, for consistency they should be in every h2_rx_* function, not only here. In fact, i resorted to `(void)r2;` because h2_rx_push_promise and h2_rx_goaway (which do not use r2) do so.

Dridi: I wasn't asking this to you in particular, it just caught my eye during the review (so rather a comment for fellow reviewers=). And since we already check the stream number beforehand, then the `xxxassert` statement could really be turned into a definitive assertion.

Anyway, thanks for the thorough research, and like I said: LGTM!

daghf:
I agree with the suggested change in `h2_sweep` - could you include a commit for that as well?

With that, I'm very much in favor of this PR.

CarloCannas:
@daghf: the change in `h2_sweep` as I proposed it in my first message? Is it okay if I `git commit --amend` it instead of creating a separate commit?

daghf:
@CarloCannas I think what you proposed in your first message is OK.

--amend should be fine :+1:

CarloCannas:
Done.

Dridi:
Thanks! Could you please solve the conflicts too? Otherwise I can take care of it. Basically, rebase against current master, solve conflicts, push -f again.

Discussed with @daghf, I will add the bells and whistles to the `h2_frame_f` callbacks myself after that.

CarloCannas:
Rebased against master.

CarloCannas:
Should I rebase again on the current master or should I wait until https://github.com/varnishcache/varnish-cache/commit/4e7e349949d7305dd8d21539b128b0225f7357b7#comments get sorted?

Dridi:
One more time please and apologies for that, we were too late.

daghf:
Merged.

 Thanks a lot, @CarloCannas!

PR #2791, opened 2018-10-05
typedefs for real/mono time and durations

nigoroll:
this is far from complete, but before going through the whole tree and determining the proper type for all the uses of our doubles, @bsdphk please let me know if the simple basics are ok

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -252,7 +252,7 @@ struct worker {
 
 	struct pool_task	task;
 
-	double			lastused;
+	vtim_real		lastused;
 	int			strangelove;
 
 	struct v1l		*v1l;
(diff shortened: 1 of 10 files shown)


bsdphk:
I think we should stick to the vtim_ prefix, I don't really see a big advantage of introducing vtm_ ?

nigoroll:
@bsdphk the motivation was to keep the VTIM_* functions with the existing signature and introduce new functions with struct-wrapping of alternative representations.
If we prefer a breaking change, keeping the VTIM prefix would be better, yes.

bsdphk:
I think we should stick with VTIM

Dridi:
Let's handle this cleanly this time: major soname bump if the new types diverge away from `double`.

nigoroll:
@Dridi yes

bsdphk:
Bugwash decision:

We'll stick with doubles, but typedef them.

Formatting with "%ju.%06ju" seems faster than %.6f, so we head that way next.

PR #2792, opened 2018-10-08
faster vtim printf macros

nigoroll:
how @bsdphk 's idea could look like with macros.
I've only changed one place (the most relevant one) to illustrate before the PR is accepted, if at all.

### DIFF
--- a/bin/varnishd/cache/cache_shmlog.c
+++ b/bin/varnishd/cache/cache_shmlog.c
@@ -375,6 +375,7 @@ void
 VSLb_ts(struct vsl_log *vsl, const char *event, vtim_real first,
     vtim_real *pprev, vtim_real now)
 {
+	double d_first, d_prev;
 
 	/* XXX: Make an option to turn off some unnecessary timestamp
 	   logging. This must be done carefully because some functions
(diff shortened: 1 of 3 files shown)


nigoroll:
why is it only after pushing/PRing, that I notice room for improvement? Will update in a minute

nigoroll:
the minute is over now

fgsch:
For the format part, I'd prefer to have it in there rather than via a macro; for the arguments it's ok. 
Do we need different precisions?

fgsch:
I still think the format should not be a macro. It makes it very hard to read, especially when is intermixed with other strings.
I'd rather have a define that better reflects the intent if you really want to avoid using the exact format.

nigoroll:
rebased, force-pushed

nigoroll:
bugwash conclusion:
* We only want precision 3 and 6, 9 to be replaced with 6
* phk wants to ponder if there are less ugly macro names

nigoroll:
FTR, my last suggestion for macro names:
* `TF6 TA6(t)`
* `TF3 TA3(t)`

Dridi:
What about `VTIMf` and `VTIMa(t)` ? It's got the `VTIM` prefix so it's hard to confuse with anything else.

Also, do we really need ms resolution? If yes, what about `VTIMf6` and `VTIMf3`? I believe the logic wouldn't change for the "argument" macro so `VTIMa(t)` would work for both.

In the VCC code we do have an argument-only macro for tokens and we still have a macro-less format everywhere. But I think that in this case we probably want the format macro to be different depending on where we are: code vs documentation. I don't think the `vsl(7)` manual should show `%jd.%06jd` for timestamps.

nigoroll:
updated to Tf/Ta, rebased, force pushed

nigoroll:
bugwash:
* we want to unify all timestamps to us precision
* @bsdphk to make the final decision

bsdphk:
I have applied this only to VSLb_ts(), pending other possible VSL work, I don't want it to spread elsewhere.

Issue #2418, opened 2017-09-07
H/2 thread starvation deadlock


daghf:
Exhausting the available worker threads can make Varnish go completely unresponsive to client traffic indefinitely. This happens when the h/2 session threads are taking up all the available threads and request threads have been queued.

The issue is trivially reproducible by running an h2 load testing tool with a number of connections exceeding the available number of threads.

For a simple test I ran with a low-ish thread count (`-p thread_pool_min=40 -p thread_pool_max=40`)

and then run an h/2 load test with a number of connections exceeding the available threads,

```
$ ./h2load https://localhost:9443/ -c 100 -n 2000
```

(h2load is part of nghttp2, https://github.com/nghttp2/nghttp2)

The above h2load command will not finish. After killing the load testing tool, Varnish will stay unresponsive indefinitely - also for h/1 traffic.

Attaching gdb at this point, we find that the threads are stuck like follows:

```
#0  pthread_cond_timedwait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_timedwait.S:225
#1  0x000000000044dba3 in Lck_CondWait (cond=0x7fcfeb7b7e40, lck=0x7fcfd9a85240, when=1504773044.7735426) at cache/cache_lck.c:207
#2  0x00000000004a0acb in h2_new_session (wrk=0x7fcfeb7b7dd0, arg=0x7fcfd9a8e020) at http2/cache_http2_session.c:344
#3  0x000000000048487c in Pool_Work_Thread (pp=0x7fcfe280e140, wrk=0x7fcfeb7b7dd0) at cache/cache_wrk.c:379
#4  0x00000000004840f1 in WRK_Thread (qp=0x7fcfe280e140, stacksize=49152, thread_workspace=2048) at cache/cache_wrk.c:132
#5  0x0000000000483d7a in pool_thread (priv=0x7fcfe180e100) at cache/cache_wrk.c:412
#6  0x00007fcfea63a494 in start_thread (arg=0x7fcfeb7b8700) at pthread_create.c:333
#7  0x00007fcfea37caff in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:97
```

This is in the request cleanup part of `h2_new_session`. The deadlock occurs here because the sessions are signaling and waiting for requests that haven't been dispatched on a worker yet. 

In summary,
- h/2 session threads can't finish because they are waiting for request threads that haven't been dispatched yet
[...]

PR #2796, opened 2018-10-09
reserve a fraction of idle threads for each priority class

nigoroll:
Previously, we used a minimum number of idle threads (the reserve) to ensure that we do not assign all threads with client requests and no threads left over for backend requests.
    
This was actually only a special case of the more general issue exposed by h2: Lower priority tasks depend on higher priority tasks (for h2, sessions need streams, which need requests, which may need backend requests).
    
To solve this situation, we divide the reserve by the number of priority classes and schedule lower priority tasks only if there are enough idle threads to run higher priority tasks eventually.
    
This change does not guarantee any upper limit on the amount of time it can take for a task to be scheduled (e.g. backend requests could be blocking on arbitrarily long timeouts), so the thread pool watchdog is still warranted. But this change should guarantee that we do make progress eventually.


nigoroll:
bugwash conclusion: phk to ponder a bit longer

nigoroll:
rebased, force pushed

bsdphk:
Bugwash generally in favour.

@Dridi to give it a whirl and feedback

Dridi:
TL;DR this patch improves the situation.

I wrote a test scenario where Varnish is serving uncacheable responses from a synthetic backend and in my naive setup I took the numbers from @daghf's #2418 for both varnish thread pool sizing and `h2load` concurrent traffic. In addition, I set the watchdog interval to its minimum value.

`h2load` will output among other things a summary like this one:

```
requests: 2000 total, 2000 started, 2000 done, 638 succeeded, 1362 failed, 1362 errored, 0 timeout
```

I failed with #2418 numbers to run into the deadlock so something must have improved somewhere else in terms of scheduling.

In this very unscientific benchmark the master branch got results where the number of succeeding request varied between 600 and 1200.

The same master branch with the patch from #2796 applied to it varied between 1000 and 1200 successful requests. However, I can easily run into a case where `h2load` stalls and I have yet to walk through the logs, and yet the watchdog never kicks in.

This need to be checked under more realistic settings now.

This test case:

```
varnishtest "#2418 vs #2796"

shell {
cat >backend.vcl <<EOF
vcl 4.0;

backend dummy {
	.host = "${bad_backend}";
}

sub vcl_recv {
	return (synth(200));
}

sub vcl_synth {
	set resp.http.Cache-Control = "no-cache";
}
EOF
}

process p_backend {
	varnishd -F -n ${tmpdir}/backend -f ${tmpdir}/backend.vcl -a 127.0.0.1:0
} -start

shell {
varnishadm -n ${tmpdir}/backend debug.listen_address >backend.sock

cat >proxy.vcl <<EOF
vcl 4.0;

backend be {
	.host = "$(cat backend.sock)";
}
EOF
}

process p_proxy {
	varnishd -F -n ${tmpdir}/proxy -f ${tmpdir}/proxy.vcl -a 127.0.0.1:0 \
		-p feature=+http2 -p thread_pool_min=40 -p thread_pool_max=40 \
		-p thread_pool_watchdog=0.1
} -start

shell {
varnishadm -n ${tmpdir}/proxy debug.listen_address | tr ' ' : >proxy.sock

h2load -c 100 -n 2000 http://$(cat proxy.sock)/ || :

varnishadm -n ${tmpdir}/proxy panic.show && exit 1

# process -kill sends a signal to all processes...
kill ${p_backend_pid}
kill ${p_proxy_pid}
}

process p_backend -wait
process p_proxy -wait
```

Dridi:
I prematurely posted my first observations with an incorrect TL;DR and deleted that comment. This was intentional and I was planning to come back here with better results than a synthetic setup based on arbitrary settings from #2418. It started with a note that this PR is improving the situation while concluding otherwise, and this is because I conducted other tests, slightly more realistic, but still not sufficient.

Apologies for the noise, please disregard until I come back with better conclusions, this was meant to be logged somewhere else but force of habit kicked in...

nigoroll:
@Dridi wanted to look at this is more detail, but did not get around to it and will now be away until the bugwash.
My take on this is that the PR fixes a real problem of priority inversion starvation and should thus go in, even if there were additional issues which we still need to look at.

Dridi:
So, I have done a fair amount of housekeeping before I leave and didn't want to engage in anything too time consuming. This is why 6.0.3 was released despite not catching up with current master (but well, there will be a 6.0.4 and more eventually) and when I was doing housekeeping involving long-running tasks I let them run in the background to look at this patch in the mean time.

My conclusion so far is that this patch improves the reserve, but there seems to be a problem somewhere in the thread pools that becomes more apparent once you apply it. It's harder to reproduce without, but possible, on both 6.0 and master.

@mbgrydeland and @hermunn also seem to think there is a problem lurking in this vicinity, but they were working on something else, and it's unlikely I will get to the bottom of it by tomorrow.

I have witnessed improvements in sort-of real-life scenarios (on a very modest setup though) and shrinking down the thread pools significantly (going further away from real life) to trigger thread starvation more easily exhibits less stability.

My main worry was that I ran into a lockup situation where the watchdog didn't kick in, despite an aggressive timeout, but can't explain yet why. Not being able to dedicate a 100% focus didn't help.

nigoroll:
Dridi, so do you you agree that this patch should go in even if there may be
more issues ahead?

Dridi:
I will wave from the fence and say I don't disagree :)

Dridi:
Works fine (and doesn't merge-conflicts) with #2942 during my tests. Either way I was no longer able to lock varnish up after repeated attempts.

It doesn't offer a clear cut improvement compared to current master but tends to lead to at least slightly better results, but most importantly stable results. Trying to put master under heavy load I get similar error rates when we run out of threads, but a higher variance.

Overall, it makes semantically more sense to me that the reserve would be per priority.

bsdphk:
Discussed at bugwash.

It does not seem like enough of an improvement to be worth it.

We wil leave this PR around with "help wanted" until 2019-11-01 for visibility.

nigoroll:
I do not understand why it is not "enough of an improvement" to solve the deadlocks caused by priority inversion.
If anyone could explain to be which issue is seen unresolved, I would be interested in continuing to work on this suggestion.
My suspicion is that reviews may confuse general overload with the issues caused by priority inversion.

nigoroll:
actually, this is not really 'priority inversion', just similar. I should have stuck to the initial description:

> Lower priority tasks depend on higher priority tasks (for h2, sessions need streams, which need requests, which may need backend requests).

bin/varnishd/mgt/mgt_pool.c
```diff
@@ -57,6 +57,8 @@ static int
 tweak_thread_pool_min(struct vsb *vsb, const struct parspec *par,
     const char *arg)
 {
+	MCF_ParamConf(MCF_MINIMUM, "thread_pool_min",
+	    "%u", 5 /* TASK_QUEUE_END */);
```
Dridi: If we hard-code the minimum, I don't think it belongs here, but in the param spec instead.

nigoroll: the field is a `char *`, so a literal "5" triggers a compiler error

nigoroll: sorry, that's the dyn_min.
/me bad

Dridi: I don't understand, I'm talking about the `min` field here:

https://github.com/varnishcache/varnish-cache/blob/c59b80661bf4f331e908ce4e9f1e85038c929829/bin/varnishd/mgt/mgt_param.h#L38-L44

Dridi: I posted before seeing your reply, I'll let you fix this.

nigoroll: fixed, thank you

Dridi:
I managed to salvage r2418.vtc from the grave, following my intuition that the new reserve logic would not remove the deadlock.

<details><summary>Show test case</summary>

```vcl
varnishtest "h2 queuing deadlock"

barrier b1 cond 2

# dridi> I didn't work out exactly how I tickled the watchdog, because unlike
# the current reserve system on the master branch I run into a deadlock before
# before even involving a 3rd client. Analysis for later, I was sidetracked by
# a bug (049ce0e73) in varnishtest.

varnish v1 -cliok "param.set thread_pools 1"
varnish v1 -cliok "param.set thread_pool_min 5"
varnish v1 -cliok "param.set thread_pool_max 5"
varnish v1 -cliok "param.set thread_pool_reserve 0"
varnish v1 -cliok "param.set thread_pool_watchdog 2"
varnish v1 -cliok "param.set feature +http2"
varnish v1 -cliok "param.set feature +no_coredump"

varnish v1 -vcl {
	backend be none;
	sub vcl_recv {
		return (synth(200));
	}
} -start

logexpect l1 -v v1 -g raw -q "vxid == 0" {
	expect * 0	Error	"Pool Herder: Queue does not move"
} -start

client c1 {
	txpri
	stream 0 rxsettings -run

	barrier b1 sync

	stream 1 {
		txreq
		# dridi> not sure how relevant this still is
	} -run
} -start

client c2 {
	barrier b1 sync
	txpri
	delay 3
} -start

client c1 -wait
client c2 -wait

logexpect l1 -wait

varnish v1 -cliok panic.show
varnish v1 -cliok panic.clear

varnish v1 -expectexit 0x20
```

</details>

daghf:
I'm no longer able to deadlock Varnish with this patch set in place. Good job :-)

@Dridi the test case does not deadlock, the reason we trigger the watchdog is due to the ``delay 3`` being longer than the watchdog timeout.

I wonder if we might still have a theoretical deadlock possibility for h/2 via ``Upgrade: h2c``, where the upgraded request gets dispatched as a stream with prio ``TASK_QUEUE_REQ`` in ``h2_ou_session()``. We should probably turn this into ``TASK_QUEUE_STR``.

Dridi:
Thank you @daghf for your analysis, when I produced the test case I wasn't sure what was happening yet and when I tried to analyze what was going on I couldn't focus properly.

Besides your suggestion, do you have an idea on how to fix the spurious watchdog trigger? This is caused by a client, so this is something we should either fix as part of this pull request or soon after if we decide to merge it in its current form.

But this is not something we should allow to fall on the back burner.

nigoroll:
@daghf a big thank you also from my side. As @Dridi has passed the maintainer hat on to you for this one, would you mind leaving an approval review?

bin/varnishd/cache/cache.h
```diff
@@ -214,17 +214,15 @@ struct pool_task {
...
- * prios up to TASK_QUEUE_RESERVE are run from the reserve
- *
  * TASK_QUEUE_{REQ|STR} are new req's (H1/H2), and subject to queue limit.
  *
- * TASK_QUEUE_RUSH is req's returning from waiting list, they are
- * not subject to TASK_QUEUE_CLIENT because we cannot safely clean
- * them up if scheduling them fails.
+ * TASK_QUEUE_RUSH is req's returning from waiting list
+ *
+ * NOTE: When changing the numner of classes, update places marked with
```
daghf: "number"

nigoroll:
review session with @bsdphk and @Dridi :
* rename `TASK_QUEUE_END` -> `TASK_QUEUE__END`
* add an assert `thread_pool_min.min is >= TASK_QUEUE__END`

OK otherwise

nigoroll:
FTR:

> add an assert `thread_pool_min.min is >= TASK_QUEUE__END`

from the commit message:

the respective places are marked and an assert is in place to ensure we do not start a worker with too small a number of workers. I decided against checks in the manager to avoid include pollution from the worker (cache.h) into the manager.

Issue #2789, opened 2018-10-03
No counter on v1d_error() ?


bsdphk:
There doesn't seem to be a VSC for late delivery errors (WS overflow) which triggers v1d_error()'s 500 response ?

daghf:
Bikeshed time - 

Should this be the same counter for h/1.1 and h/2? (I think yes)

What's a good name? Some ideas: `resp_ws_overflow`, `sess_ws_overflow`, `deliver_ws_overflow` ?

bsdphk:
We want a "500-counter" for delivery like client_req_400

We want per-ws-type (client/be/wrk) overflow counters

daghf:
Fixed via #2798

PR #2798, opened 2018-10-11
Workspace overflow counters

daghf:
This adds overflow counters for each of `workspace_client`, `workspace_backend`, `workspace_session` and `workspace_thread`. 

In addition, it also adds a `client_resp_500` counter, which is the number of times we failed an h/1.1 or h/2 response at delivery due to insufficient workspace.


daghf:
Ref #2789

bsdphk:
Should the overflow counters have a higher level than "diag" ?

daghf:
Merged after discussing with phk on irc.

PR #2799, opened 2018-10-11
Backports 6.1.1

hermunn:
These are my suggested *code* back ports for the 6.1.1 release. I have included most of the changes and tried not to break VRT, but I would like someone to look at it before we merge this into the 6.1 branch.

When this is merged, I will do the changelog etc. and roll the release.

Feedback welcome. Please don't merge until everyone has had a chance to have a look.


hermunn:
If I am not mistaken, contributors were asked to have a look at this. There are only a few of these with conflicts, and they should be relatively simple. I will merge this some time during the week-end unless I hear otherwise, so that I can present the changelog + "prepare for" patches for review on Monday.

PR #2803, opened 2018-10-18
Save writev() calls for high numbers of http headers

nigoroll:
Since 7bf932f3a674fa934c1bfd586686686003eec550 we might exceed the
number of iovs for many http headers (>31 with default parameters), so
we would issue more than one writev() for the http headers.

While sending http headers, we can use all of the thread workspace for
IOVs, so if we know that we would need a seond writev call anyway, we
re-open V1L to ensure we send as many headers at once as possible.

### DIFF
--- a/bin/varnishd/http1/cache_http1_deliver.c
+++ b/bin/varnishd/http1/cache_http1_deliver.c
@@ -85,8 +85,8 @@ void v_matchproto_(vtr_deliver_f)
 V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
 {
 	int err = 0, chunked = 0;
-	unsigned u;
-	uint64_t hdrbytes, bytes;
+	unsigned u, niov;
+	uint64_t hdrbytes, bytes = 0;
 
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 	CHECK_OBJ_ORNULL(boc, BOC_MAGIC);
@@ -129,9 +129,15 @@ V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
 		return;
 	}
 
+	if (HTTP1_Estimate(req->resp) > cache_param->http1_iovs)
+		niov = 0;
+	else
+		niov = cache_param->http1_iovs;
+
+	u = 0;
 	AZ(req->wrk->v1l);
 	V1L_Open(req->wrk, req->wrk->aws,
-		 &req->sp->fd, req->vsl, req->t_prev, cache_param->http1_iovs);
+		 &req->sp->fd, req->vsl, req->t_prev, niov);
 
 	if (WS_Overflowed(req->wrk->aws)) {
 		v1d_error(req, "workspace_thread overflow");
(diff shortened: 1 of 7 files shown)


bin/varnishd/http1/cache_http1_deliver.c
```diff
@@ -129,9 +129,13 @@ V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
 		return;
 	}
 
+	niov = HTTP1_Estimate(req->resp) > cache_param->http1_iovs ? 0
+		: cache_param->http1_iovs;
```
fgsch: My eyes!

nigoroll:
bugwash conclusion:

> (13:39:40) slink: can we wrap this up? I don't think the V1L_Reopen is complex, we only use the code when we need to and we can save a writev(). If that doesn't convince you, we should just close this, it's not worth any more time. I just thought it would be a sensible optimization.
> (13:40:02) phk: yeah, I'll ponder it a bit.

nigoroll:
@bsdphk 

as this is somewhat related, here are some preliminary figures from the NIOV debug on a test system with live-ish traffic (no ESI):

```
$ varnishlog -cdg raw -I Debug:NIOV | awk '{ print $5 }'| sort -n | uniq -c
    179 1
  11451 24
      1 28
     34 30
   9417 32
    127 34
     27 36
     26 42
   3950 44
      1 45
    125 46
   4129 48
   4609 52
      8 54
      3 56
      1 60
      8 61
```
```
$ varnishlog -bdg raw -I Debug:NIOV | awk '{ print $5 }'| sort -n | uniq -c
     95 29
     25 31
   4660 33
  14442 35
   1466 37
    144 38
      2 39
   1545 40
      4 41
      9 43
```

http1_niovs is at 64 (default).

This is _not_ the system under control of my friend the header-maniac

nigoroll:
Production system data with this patch (which explains niovs > 64)

```
$ varnishlog -cg raw -I Debug:NIOV | head -100000 | awk '{ print $5 }'| sort -n | uniq -c
   5712 1
    937 2
  56434 3
    774 4
     84 5
     21 6
     14 7
      8 8
      6 9
      7 10
      3 11
      4 12
      5 14
      4 15
      3 16
    986 17
      1 18
      2 19
      3 20
     10 21
      3 22
    888 23
    187 24
      8 25
    640 26
    780 27
    168 28
   1044 29
    171 30
    379 31
     39 32
   1375 33
     82 34
    517 35
    437 36
    914 37
    338 38
   2874 39
    273 40
   1648 41
    597 42
    563 43
   3179 44
    571 45
   2452 46
    167 47
    549 48
    665 49
   3083 50
    189 51
    960 52
     46 53
   1984 54
    708 55
   1132 56
    297 57
    434 58
    202 59
     88 60
    142 61
     17 62
   1887 63
     62 64
    847 65
    731 67
    649 69
     16 71
```

nigoroll:
bugwash conclusion: I failed to convince @bsdphk of the usefulness of this patch, so we'll close and carry on.

PR #2806, opened 2018-10-25
Blob nocopy: should we introduce struct vmod_blob ?

nigoroll:
TBD - opening a PR because of the unavoidable `TRUST_ME()` required with the current `struct vmod_priv`:

Should we rather introduce a separate struct for blobs as?

```
struct vrt_blob {
        unsigned                magic;
        int                     len; // or size_t ?
        const void              *blob;
        vmod_priv_free_f        *free;
};

```


nigoroll:
FTR edited the initial note for clarity: renamed the struct to `struct vrt_blob`

nigoroll:
bugwash: write the patch

nigoroll:
continued in #2827

PR #2813, opened 2018-10-30
manage dynamic privs as a red/black tree

nigoroll:
O(n) does not scale for a high number of dynamic privs, so trade some
workspace for O(lg n)

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -49,6 +49,7 @@
 #include "miniobj.h"
 #include "vas.h"
 #include "vqueue.h"
+#include "vtree.h"
 
 #include "vapi/vsl_int.h"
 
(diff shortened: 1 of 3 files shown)


gquintard:
Out of curiosity, would you have some numbers about the performance impact?

nigoroll:
@gquintard it's just what the theory predicts, but here are some numbers from the trivial benchmark I have just checked in:

* RB tree (this PR)

\# of priv_tasks |  avg time per VRT_priv_task()
-- | --
1 | 57.785399986 ns
10 | 73.406120000 ns
100 | 86.081872000 ns
1000 | 95.048375300 ns
10000 | 108.457345140 ns

* List (original implementation)

\# of priv_tasks |  avg time per VRT_priv_task()
-- | --
1 | 35.247899996 ns
10 | 58.420890000 ns
100 | 230.740021000 ns
1000 | 2,006.253359200 ns

So the list is the better option for ~10-ish priv_tasks due to the lower constant overhead, but for higher numbers the tree scales much better. Because the max potential saving between a list and rbtree access in favor of the list is ~20ns, I think the higher initial cost of the rbtree is justified for irrelevance.

in `vtree.h` there is also a splay tree implementation which actually could work better in practice because it increases the locality of recently used objects. But I'd have a bit of a preference for more predictable behavior of the rbtree.

nigoroll:
rebased, force-pushed

PR #2816, opened 2018-11-01
Vstrerror

nigoroll:
trivial thing in principle, but submitting as a PR because of the involvement

### DIFF
--- a/bin/varnishadm/varnishadm.c
+++ b/bin/varnishadm/varnishadm.c
@@ -49,7 +49,6 @@
 #  error missing readline.h - this should have got caught in configure
 #endif
 
-#include <errno.h>
 #include <math.h>
 #include <fcntl.h>
 #include <poll.h>
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -33,7 +33,6 @@
 
 #include "config.h"
 
-#include <errno.h>
 #include <stdlib.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
@@ -526,7 +525,7 @@ vca_accept_task(struct worker *wrk, void *arg)
 
 			VSL(SLT_SessError, 0, "%s %s %s %d %d %s",
 			    wa.acceptlsock->name, laddr, lport,
-			    ls->sock, i, strerror(i));
+			    ls->sock, i, vstrerror(i));
 			(void)Pool_TrySumstat(wrk);
 			continue;
 		}
@@ -650,12 +649,12 @@ ccf_start(struct cli *cli, const char * const *av, void *priv)
 			if (i)
 				VSL(SLT_Error, 0,
 				    "Kernel TCP Fast Open: sock=%d, ret=%d %s",
-				    ls->sock, i, strerror(errno));
+				    ls->sock, i, vstrerror(errno));
 		}
 		if (listen(ls->sock, cache_param->listen_depth)) {
 			VCLI_SetResult(cli, CLIS_CANT);
 			VCLI_Out(cli, "Listen failed on socket '%s': %s",
-			    ls->endpoint, strerror(errno));
+			    ls->endpoint, vstrerror(errno));
 			return;
 		}
 		vca_tcp_opt_set(ls->sock, ls->uds, 1);
@@ -665,7 +664,7 @@ ccf_start(struct cli *cli, const char * const *av, void *priv)
 			if (i)
 				VSL(SLT_Error, 0,
 				    "Kernel filtering: sock=%d, ret=%d %s",
-				    ls->sock, i, strerror(errno));
+				    ls->sock, i, vstrerror(errno));
 		}
 	}
 
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -33,7 +33,6 @@
 #include "config.h"
 
 #include <stdlib.h>
-#include <errno.h>
 
 #include "cache_varnishd.h"
 
@@ -154,7 +153,7 @@ vbe_dir_getfd(struct worker *wrk, struct backend *bp, struct busyobj *bo,
 		VBE_Connect_Error(bp->vsc, err);
 		VSLb(bo->vsl, SLT_FetchError,
 		     "backend %s: fail errno %d (%s)",
-		     VRT_BACKEND_string(bp->director), err, strerror(err));
+		     VRT_BACKEND_string(bp->director), err, vstrerror(err));
 		VSC_C_main->backend_fail++;
 		bo->htc = NULL;
 		return (NULL);
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -39,7 +39,6 @@
 
 #include <poll.h>
 #include <stdio.h>
-#include <errno.h>
 #include <string.h>
 #include <stdlib.h>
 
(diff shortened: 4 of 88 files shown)


fgsch:
besides the placement of verror.h it looks fine.

fgsch:
Actually not sure why we want an inline here?

fgsch:
I still don't see the need to inline this. 
And AFAICT you have replaced a bunch of errno.h that have no strerror() calls with verror.h.

nigoroll:
as a basis for comparison, here's the errno handling code of the `pan_ic` function from `objdump -S <varnishd>` for the two options compiled with `gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1)` on 64bit. `CFLAGS = -g -O2 -pthread -Wall -Werror -Wno-error=unused-result` for both cases

The alternative code is here: https://github.com/nigoroll/varnish-cache/tree/vstrerror_libvarnish

### `static inline const char * vstrerror(int e)`

```
        if (err)
   4bd29:       83 7c 24 14 00          cmpl   $0x0,0x14(%rsp)
   4bd2e:       0f 85 e2 01 00 00       jne    4bf16 <pan_ic+0x4b6>

...

   4bf16:       48 8b 44 24 18          mov    0x18(%rsp),%rax
        p = strerror(e);
   4bf1b:       8b 7c 24 14             mov    0x14(%rsp),%edi
        int oerrno = errno;
   4bf1f:       8b 18                   mov    (%rax),%ebx
        p = strerror(e);
   4bf21:       e8 ba 60 fd ff          callq  21fe0 <strerror@plt>
        if (p != NULL)
   4bf26:       48 85 c0                test   %rax,%rax
   4bf29:       0f 84 f9 07 00 00       je     4c728 <pan_ic+0xcc8>
                VSB_printf(pan_vsb, "errno = %d (%s)\n", err, vstrerror(err));
   4bf2f:       8b 54 24 14             mov    0x14(%rsp),%edx
   4bf33:       48 8b 3d 8e 64 2e 00    mov    0x2e648e(%rip),%rdi        # 3323c8 <pan_vsb>
   4bf3a:       48 8d 35 7d ec 0b 00    lea    0xbec7d(%rip),%rsi        # 10abbe <__func__.3220+0x86>
   4bf41:       48 89 c1                mov    %rax,%rcx
   4bf44:       31 c0                   xor    %eax,%eax
   4bf46:       e8 95 fa 06 00          callq  bb9e0 <VSB_printf>
   4bf4b:       e9 e4 fd ff ff          jmpq   4bd34 <pan_ic+0x2d4>
```

```
(gdb) print 0x4bf4b-0x4bf16
$3 = 53
```

...


```
   4c728:       48 8b 44 24 18          mov    0x18(%rsp),%rax
   4c72d:       89 18                   mov    %ebx,(%rax)
        return ("strerror(3) returned NULL");
   4c72f:       48 8d 05 e6 ad 07 00    lea    0x7ade6(%xrip),%rax        # c751c <_IO_stdin_used+0x51c>
   4c736:       e9 f4 f7 ff ff          jmpq   4bf2f <pan_ic+0x4cf>
```


```
(gdb) print 0x4c736-0x4c728
$2 = 14
```

### `const char * vstrerror(int e)`


```
       if (err)
   4bb89:       83 7c 24 0c 00          cmpl   $0x0,0xc(%rsp)
   4bb8e:       0f 85 e2 01 00 00       jne    4bd76 <pan_ic+0x4a6>
        q = THR_GetName();

...

   4bd76:       8b 5c 24 0c             mov    0xc(%rsp),%ebx
   4bd7a:       89 df                   mov    %ebx,%edi
   4bd7c:       e8 af 9b 06 00          callq  b5930 <vstrerror>
   4bd81:       48 8b 3d 40 66 2e 00    mov    0x2e6640(%rip),%rdi        # 3323c8 <pan_vsb>
   4bd88:       48 8d 35 0f e9 0b 00    lea    0xbe90f(%rip),%rsi        # 10a69e <__func__.3220+0x86>
   4bd8f:       48 89 c1                mov    %rax,%rcx
   4bd92:       89 da                   mov    %ebx,%edx
   4bd94:       31 c0                   xor    %eax,%eax
   4bd96:       e8 55 f7 06 00          callq  bb4f0 <VSB_printf>
   4bd9b:       e9 f4 fd ff ff          jmpq   4bb94 <pan_ic+0x2c4>
```

```
(gdb) print 0x4bd9b-0x4bd76
$3 = 37
```

```
00000000000b5930 <vstrerror>:

#include "verrno.h"

const char *
vstrerror(int e)
{
   b5930:       41 54                   push   %r12
   b5932:       55                      push   %rbp
   b5933:       89 fd                   mov    %edi,%ebp
   b5935:       53                      push   %rbx
        const char *p;
        int oerrno = errno;
   b5936:       e8 f5 bb f6 ff          callq  21530 <__errno_location@plt>

        p = strerror(e);
   b593b:       89 ef                   mov    %ebp,%edi
        int oerrno = errno;
   b593d:       48 89 c3                mov    %rax,%rbx
   b5940:       44 8b 20                mov    (%rax),%r12d
        p = strerror(e);
   b5943:       e8 b8 c6 f6 ff          callq  22000 <strerror@plt>
        if (p != NULL)
   b5948:       48 85 c0                test   %rax,%rax
   b594b:       74 0b                   je     b5958 <vstrerror+0x28>
                return (p);

        errno = oerrno;
        return ("strerror(3) returned NULL");
}
   b594d:       5b                      pop    %rbx
   b594e:       5d                      pop    %rbp
   b594f:       41 5c                   pop    %r12
   b5951:       c3                      retq   
   b5952:       66 0f 1f 44 00 00       nopw   0x0(%rax,%rax,1)
        errno = oerrno;
   b5958:       44 89 23                mov    %r12d,(%rbx)
        return ("strerror(3) returned NULL");
   b595b:       48 8d 05 9e 55 05 00    lea    0x5559e(%rip),%rax        # 10af00 <vct_typtab+0x200>
}
   b5962:       5b                      pop    %rbx
   b5963:       5d                      pop    %rbp
   b5964:       41 5c                   pop    %r12
   b5966:       c3                      retq   
   b5967:       66 0f 1f 84 00 00 00    nopw   0x0(%rax,%rax,1)
   b596e:       00 00 
```

```
# normal path:
(gdb) print 0xb5951-0xb5930
$2 = 33
```
### summary

So the data path for the normal case is:

* inlined code is 53 bytes, 12 instructions.
* extra function 37 + 33 = 70 bytes, 25 instructions

Inlining saves the additional function call and avoids unused function compiler errors for a static. It can save an additional call to the `errno` function. Needless to say that inlining is optional.

Regarding the `verrno.h` include, I think we should keep the include consistent, it does not cause any harm and removes special casing. But I'm also fine with including `verrno.h` only where `vstrerror()` is required.

As a general note, I feel that my limited lifetime is not well spent on such a very minor aspect. So basically I am fine with any solution and would have preferred to get this done quicker.

Disclaimer: My caclulations most likely include off-by-small-amount-errors

fgsch:
I agree with you that we should not spend too much time on minor things but IMO that's not the point here. My reaction to inlining is not about the size of the assembly code. This is in the error path so I don't think speed is a concern either. I just don't see a reason to increase the binary size for no apparent gain. It feels to me that you went for the quick approach, not the correct one (replacing all errno.h, adding a single .h file and therefore using inline).

Anyway, I will stop here. @bsdphk can decide.

Dridi:
VTLA suggestion: `VERR_str()` in `verr.h` in `libvarnish` and possibly exported by `libvarnishapi`. This opens the possibility of more error handling facilities both in and out of tree.

nigoroll:
* function in libvarnish: ```9741 -rwxr-xr-x 1 slink slink 4899856 Nov  2 11:51 ./bin/varnishd/varnishd```
* inlining: ```9741 -rwxr-xr-x 1 slink slink 4912304 Nov  2 11:49 ./bin/varnishd/varnishd```

Beware of bikesheds. I'll also stop at this point. Let's just get a decision and move on.

Dridi:
@nigoroll the VTLA suggestion is orthogonal to the function inlining, I was not commenting on that specific point. Sorry if it felt that way, it's rather about spreading the usefulness outside of `varnishd` because we have other things that may run out of memory (like `varnishlog` with session grouping) and circling back to `varnishd` it could come in handy for VMODs too on Solaris derivatives.

nigoroll:
@dridi I agree, this should be available to vmods: https://github.com/varnishcache/varnish-cache/pull/2816/commits/56b9f37b84c0d2d74fdbc0484a759cf895f155b7#diff-9b39bc57e0a9f5b39f437ac40f59bdd7R63

nigoroll:
bugwash conclusion: 
* cleanup https://github.com/nigoroll/varnish-cache/tree/vstrerror_libvarnish
* move the include into vas.h

PR #2818, opened 2018-11-02
Add a mechanism to interrupt VSM_Attach

### LINKED ISSUE #2657: varnishncsa -t waiting period is unbreakable
Issue: When running varnishncsa (or another tool like varnishhist) as a user that does not have access to the shared memory file, the period where it is retrying the process is not possible to stop with Ctrl+c.

Expected: the process should abort the wait and exit back to the shell.

```
$ date; time varnishncsa -t 10; date
Mon Apr 23 16:46:56 CEST 2018
^C^C.^C.........
VSM: Could not get hold of varnishd, is it running?

real	0m10.246s
user	0m0.002s
sys	0m0.001s
Mon Apr 23 16:47:06 CEST 2018
```

Workaround: bringing out the big hammer and send sigquit with Ctrl+4 does work.

Running Varnish Cache 5.2.1 from packagecloud on ubuntu bionic. (xenial packages)

fgsch:
Call it from the VUT code so tools using it can be interrupted.
Fixes #2657.

### DIFF
--- a/bin/varnishadm/varnishadm.c
+++ b/bin/varnishadm/varnishadm.c
@@ -73,6 +73,7 @@
 
 
 static double timeout = 5;
+static struct vsm *vsm;
 
 static void
 cli_write(int sock, const char *s)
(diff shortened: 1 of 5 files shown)


fgsch:
@bsdphk This is the idea I mentioned a few weeks ago. The description for the two functions need to be added but besides that it should be complete. Thoughts?

Dridi:
Missing an update of `libvarnishapi.map`, otherwise LGTM.

fgsch:
bugwash outcome: @bsdphk to ponder over this.

fgsch:
OBE.

PR #2827, opened 2018-11-12
Change type of VCL_BLOB to newly introduced struct vrt_blob

nigoroll:
Re-using struct vmod_priv helped simplicity, but had the following
shortcomings:

- in ddfda3d7c996e67a1894fc3955eb546ebd0b721b we documented what was
  long implied: return values of vmod functions are assumed immutable
  and so are blobs.

  So the blob pointer should be const, yet the struct vmod_priv pointer
  can't be.

- the struct vmod_priv free pointer implies that there would be
  automatic memory management provided by varnish core, yet this does not
  exist for a good reason: We would otherwise need to track all blobs
  ever returned by vmod functions/methods.

So we turn the data pointer into a const and remove the free function
callback.

We also add a type field, which is to be viewed similar to the miniobj
magic, except that it should not be asserted upon. The type field is
intended for additional (yet unreliable) checks of vmods using BLOBs
to carry around vmod-specific private data.


nigoroll:
I forgot to point out that the patch also changes the blob-operations to not copy.

nigoroll:
force-pushed with minor fix of an assertion

nigoroll:
bugwash: make blob length size_t and merge

bsdphk:
Length should be size_t, otherwise OK

PR #2838, opened 2018-11-19
After the divorce from vtest, marry it again as a submodule

nigoroll:
please note:

* for now, the submodule is https://github.com/nigoroll/VTest
  because of additional changes to vtest required
  (PR is https://github.com/vtest/VTest/issues/3 )

* This will change the procedure to check out the sources:
  (not yet docmented)

	git clone --recursive

	or, if already cloned

	git submodule update --init

* To pull

	git pull --recurse-submodules


nigoroll:
travis: ```make[3]: *** No rule to make target `vtest/src/vtc_h2_dectbl.h', needed by `distdir'.  Stop.```
seems like we need to teach it how to work with submodules

bsdphk:
Thanks for doing this Nils.

I think my conclusion is that the time is not ripe for this yet. 

How we integrate vtest ultimately depends on what the vtest project develops into.

Right now it is sort of a "source-code library" and the manual integration is not a big deal.

If vtest development picks up, that calculus will change and if so, we will make new decisions.

nigoroll:
VDD: revive this

nigoroll:
re-opening as a reminder to myself that I had promised a long time ago to do this.
The http workshop just brought vtest back closer to the top of my agenda

nigoroll:
Oh sorry I overlooked #3983

Issue #2840, opened 2018-11-22
VRT_fail from VMOD call used in an if test expression not working as advertised


daghf:
Ran into this with a VCL like the following, where ``myvmod.fail()`` did a `VRT_fail()`:

```
sub vcl_recv {
    if (myvmod.fail()) {
        return (hash);
    }
}
```

I expected this to end up as a ``fail`` return. However, it actually ends up as a ``return (hash)``.

We may need to teach VCC to sprinkle even more ``ctx->handling`` checks.

Test case needs additional ``vmod_debug`` support, so it's included here as a git diff,

```
diff --git a/bin/varnishtest/tests/v00051.vtc b/bin/varnishtest/tests/v00051.vtc
index 664af71f7..ec5479a7c 100644
--- a/bin/varnishtest/tests/v00051.vtc
+++ b/bin/varnishtest/tests/v00051.vtc
@@ -425,3 +425,37 @@ varnish v1 -expect sc_vcl_failure == 10
 
 logexpect l1 -wait
 
+#######################################################################
+# Fail in vmod call used in an if test
+
+server s1 {
+       rxreq
+       txresp
+} -start
+
+varnish v1 -vcl+backend {
+       import debug;
+       sub vcl_recv {
+               if (debug.fail2()) {
+                       return (hash);
+               }
+       }
+}
+
+logexpect l1 -v v1 -g raw {
+       expect * 1001   VCL_call        "RECV"
+       expect 0 1001   VCL_Error       "Forced failure"
+       expect 0 1001   VCL_return      "fail"
+} -start
+
+client c1 {
+       txreq
+       rxresp
+       expect resp.status == 503
+       expect resp.reason == "VCL failed"
+} -run
+
+varnish v1 -expect vcl_fail == 14
+varnish v1 -expect sc_vcl_failure == 11
+
+logexpect l1 -wait
diff --git a/lib/libvmod_debug/vmod.vcc b/lib/libvmod_debug/vmod.vcc
index 3e0cc4faa..590971aa1 100644
--- a/lib/libvmod_debug/vmod.vcc
+++ b/lib/libvmod_debug/vmod.vcc
@@ -130,6 +130,10 @@ $Function VOID fail()
 
 Function to fail vcl code.  (See also: RFC748)
 
+$Function BOOL fail2()
+
+Function to fail vcl code. Always returns true.
+
 $Object dyn(STRING addr, STRING port, PROBE probe=0)
 
[...]

PR #2842, opened 2018-11-22
Fix VRT_fail for 'if'/'elseif' conditional expressions

daghf:
This adds a ``ctx->handling`` test at the beginning of any compound statement, to catch the cases where VRT_fail was invoked as part of an if test condition.

A side-effect of this approach is that we will end up with one extra ``ctx->handling`` test at the start of any ``sub``. Ideally we could restrict adding this to the cases where the compound statement was an if/elseif block. From the look of things, that would require us to either redo how/where we insert the curly brackets, or carry state that tells us "this is an if block". Having the extra ``ctx->handling`` test seems like the lesser evil. 

Fixes: #2840

### DIFF
--- a/bin/varnishtest/tests/v00051.vtc
+++ b/bin/varnishtest/tests/v00051.vtc
@@ -425,3 +425,32 @@ varnish v1 -expect sc_vcl_failure == 10
 
 logexpect l1 -wait
 
+#######################################################################
+# Fail in vmod call used in an if test
+
+varnish v1 -vcl+backend {
+	import debug;
+	sub vcl_recv {
+		if (debug.fail2()) {
+			return (hash);
+		}
+	}
+}
+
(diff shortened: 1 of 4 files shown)


nigoroll:
_superficial_ in my review should have read _superfluous_

nigoroll:
not sure if this is relevant, but FTR: This removes the option to custom-handle the failure, e.g. as with
```
if (vmod.fail()) {
  return (synth(599, "my failure"));
}
```
Ref: https://code.uplex.de/uplex-varnish/libvmod-blobdigest/commit/bd7c6346b2522e6b949237be45cc17932e9784d2

I consider the fact that this worked more an unintended side effect, but opinions may vary

daghf:
Merged via 05f3ee5ef

Issue #1841, opened 2016-03-04
r01801.vtc panic on OSX: vwk_thread(), waiter/cache_waiter_kqueue.c line 109:


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1841

lkarsten:
backport review: this is in 4.1.2-beta1 and newer.

PR #2843, opened 2018-11-22
use our knowledge about master file descriptors also for VSUBs

### LINKED ISSUE #2730: Closing all FD's even those which aren't open
Possibly related commit: https://github.com/varnishcache/varnish-cache/commit/c60b4f81391c6d162543ff4471d86e423e187e66
Possibly related issue: #669

## Expected Behavior
Faster loading of VCL's

## Current Behavior
Trying to close all FD's even if they do not exist
(Code https://github.com/varnishcache/varnish-cache/blob/master/lib/libvarnish/vsub.c#L58)

## Possible Solution
After master starts up keep track of opened FD's and close them when needed. Maybe we could prevent VSUB_closefrom from running for each VCL load. Or maybe this is a solution:

    void
    VSUB_closefrom(int fd)
    {
    
        assert(fd >= 0);

    #ifdef HAVE_CLOSEFROM
            closefrom(fd);
    #else

    DIR* dirFile = opendir("/proc/self/fdinfo");
    int fdClose;
    if (dirFile)
    {
       struct dirent* fdFile;
       while ((fdFile= readdir(dirFile)) != NULL)
       {
           sscanf(fdFile->d_name, "%d", &fdClose);
           if (fdClose > fd) {
               (void)close(fdClose);
           }
       }
       closedir(dirFile);
    }

    //      int i = sysconf(_SC_OPEN_MAX);
    //      assert(i > 0);
    //      for (; i > fd; i--)
    //              (void)close(i);
    #endif
    }

_(I'm not a C developer)_

Strace on varnish and load random VCL

## Context
When we've had a server crash we had to load ~1400 VCL's, our NOFILE limit was set to 900.000, it took 0.9 seconds per VCL to load (without compile time) because it was trying to close all 900.000 FD's

## Your Environment
CentOS 7 (also tested on 6)
Varnish 5.2.1 (but code seems to be the same in 6.0)

nigoroll:
In the master process, we record file descriptors to be inherited to the worker child and a maximum file descriptor ever used. We long used this knowledge when starting a worker child to selectively not close file descriptors required by the child and to limit the maximum file descriptor number to close.

This patch makes the same knowledge available to `VSUB_run()` which is used by the master process to start the VCC- and C-Compilers and the .so test-load while executing `vcl.load`. The benefit is to avoid long `vcl.load` times on systems which do not have an efficient `closefrom()` implementation. In addition, the issue addressed in #1841 should, in principle, also apply to processes started via `VSUB_run()` and this change should fix those as well.

Regarding the information leakage argument made in #2730, we would have the same issue for the varnish worker child, so I do not think that we are taking on additional risks. Also, since 1be6ad1d5a807038adcaa39168ce69a5a76bc8e1 we can now be more confident not to be missing to close any file descriptors which should be closed.

Fixes #2730

### DIFF
--- a/bin/varnishd/common/heritage.h
+++ b/bin/varnishd/common/heritage.h
@@ -95,6 +95,7 @@ extern struct heritage heritage;
 
 /* Really belongs in mgt.h, but storage_file chokes on both */
 void MCH_Fd_Inherit(int fd, const char *what);
+void MCH_Fd_closefrom(int);
 
 #define ARGV_ERR(...)						\
 	do {							\
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -184,6 +184,8 @@ mch_cli_panic_clear(struct cli *cli, const char * const *av, void *priv)
  *
  * For added safety, we check that we see no file descriptor open for
  * another margin above the limit for which we close by design
+ *
+ * XXX as this also applies to VSUB, move to a more generic source file?
  */
 
 static int		mgt_max_fd;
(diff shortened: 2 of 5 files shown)


bsdphk:
I think the cleaner way is to put the _closefrom in VSUB and make it possible to pass in a bitmap of Fd's which should not be touched, so that the child can call that function too ?

nigoroll:
ok I finally understood @bsdphk\ 's last comment, will work on it

nigoroll:
#2913 contains an alternative suggestion. I would like to keep this one open until we know which way to go please.

nigoroll:
overtaken by 14721f5379ea644cec138dc73958ecbfcee09888 / 6ab26126194fc81017c1847925b7ea4fc2b10969

PR #2844, opened 2018-11-22
engage tighter master file descriptor checks

nigoroll:
In the master process, we record file descriptors to be inherited to the worker child and a maximum file descriptor ever used.
Previously, there was a tolerance margin of 100 file descriptors which we close on the suspicion that they could be used. 6bead571da6a2027a518af81e14d87a247f38156 hopefully has fixed an issue where we failed to track the max fd, so I presume that we now got accurate knowledge.
This simple change removes the margin and thus more tightly engages the assertions from 1be6ad1d5a807038adcaa39168ce69a5a76bc8e1

### DIFF
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -188,9 +188,7 @@ mch_cli_panic_clear(struct cli *cli, const char * const *av, void *priv)
 
 static int		mgt_max_fd;
 
-#define CLOSE_FD_UP_TO	(mgt_max_fd + 100)
-// XXX should work now - engage?
-//#define CLOSE_FD_UP_TO	mgt_max_fd
+#define CLOSE_FD_UP_TO	mgt_max_fd
 #define CHECK_FD_UP_TO	(CLOSE_FD_UP_TO + 100)
 
 void


nigoroll:
notice that @fgsch confirmed this also looks ok on MacOS, I personally have run make check on linux.

IRC:
(13:45:48) slink: fgs only if you have time, could you check 6bead571da6a2027a518af81e14d87a247f38156 on macos with #define CLOSE_FD_UP_TO       mgt_max_fd in mgt_child.c ?
...
(15:08:03) fgs: slink: no issues

nigoroll:
bugwash: keep margin, lower to 10

nigoroll:
sorry, phk did not say 10. That came from my imagination, he said "cut it down from 100"

bsdphk:
Background:  The 100 is a finagle constant to cope with any FD's which a library might have opened which we do not know about, so absent a "give me highest open fd" system-call, we cannot avoid that.

100 is probably excessively defensive.

nigoroll:
Yes. Some background on the reasoning:
* since https://github.com/varnishcache/varnish-cache/commit/1be6ad1d5a807038adcaa39168ce69a5a76bc8e1 we will notice any (untracked) fd above `CLOSE_FD_UP_TO`
* As far as I could track this back, the motivation for the margin was #1841 which happened on OSX and @fgsch checked the patch.

In summary, I am not against the margin, but with the assertions based on `CHECK_FD_UP_TO` I think we can pull it tight, unless we see the assertion failing.

PR #2845, opened 2018-11-23
make proxy protocol header formatting available in VRT

nigoroll:
the use case is to send proxy headers from backend vmod code

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -989,7 +989,8 @@ VRT_VSA_GetPtr(VRT_CTX, const struct suckaddr *sua, const unsigned char ** dst)
 }
 
 void
-VRT_Format_Proxy(struct vsb *vsb, VCL_INT version, VCL_IP sac, VCL_IP sas)
+VRT_Format_Proxy(struct vsb *vsb, VCL_INT version, VCL_IP sac, VCL_IP sas,
+    VCL_STRING auth)
 {
-	VPX_Format_Proxy(vsb, (int)version, sac, sas);
+	VPX_Format_Proxy(vsb, (int)version, sac, sas, auth);
 }
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -47,6 +47,7 @@
 
 #include "common/heritage.h"
 #include "common/vsmw.h"
+#include "proxy/cache_proxy.h"
 
 const void * const vrt_magic_string_end = &vrt_magic_string_end;
 const void * const vrt_magic_string_unset = &vrt_magic_string_unset;
@@ -985,4 +986,10 @@ VRT_VSA_GetPtr(VRT_CTX, const struct suckaddr *sua, const unsigned char ** dst)
 		return (-1);
 	}
 	return (VSA_GetPtr(sua, dst));
+}
+
+void
+VRT_Format_Proxy(struct vsb *vsb, VCL_INT version, VCL_IP sac, VCL_IP sas)
+{
+	VPX_Format_Proxy(vsb, (int)version, sac, sas);
 }
--- a/bin/varnishd/proxy/cache_proxy_proto.c
+++ b/bin/varnishd/proxy/cache_proxy_proto.c
@@ -632,6 +632,34 @@ vpx_format_proxy_v1(struct vsb *vsb, int proto,
 	AZ(VSB_finish(vsb));
 }
 
+static void
+vpx_format_proxy_v2(struct vsb *vsb, int proto,
+    const struct suckaddr *sac, const struct suckaddr *sas)
+{
+	AN(vsb);
+	AN(sac);
+	AN(sas);
+
+	VSB_bcat(vsb, vpx2_sig, sizeof(vpx2_sig));
+	VSB_putc(vsb, 0x21);
+	if (proto == PF_INET6) {
+		VSB_putc(vsb, 0x21);
+		VSB_putc(vsb, 0x00);
+		VSB_putc(vsb, 0x24);
+	} else if (proto == PF_INET) {
+		VSB_putc(vsb, 0x11);
+		VSB_putc(vsb, 0x00);
+		VSB_putc(vsb, 0x0c);
+	} else {
+		WRONG("Wrong proxy v2 proto");
+	}
+	vpx_enc_addr(vsb, proto, sac);
+	vpx_enc_addr(vsb, proto, sas);
+	vpx_enc_port(vsb, sac);
+	vpx_enc_port(vsb, sas);
+	AZ(VSB_finish(vsb));
+}
+
 void
 VPX_Send_Proxy(int fd, int version, const struct sess *sp)
 {
@@ -649,7 +677,6 @@ VPX_Send_Proxy(int fd, int version, const struct sess *sp)
 	AZ(SES_Get_server_addr(sp, &sas));
 	AN(sas);
 	proto = VSA_Get_Proto(sas);
-	assert(proto == PF_INET6 || proto == PF_INET);
 
 	if (version == 1) {
 		VTCP_name(sas, ha, sizeof ha, pa, sizeof pa);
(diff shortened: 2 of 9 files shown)


Dridi:
Why not exposing the feature directly in vmod-proxy instead of vmod-debug? That would give end users a way to pass a `BLOB` around in VCL.

The very test case you modified fails on Travis CI.

Dridi:
Pass the blob to some other vmod, maybe? In its current form, LGTM.

Dridi:
When I said "current" I hadn't seen yet that you moved the vmod bit from debug to proxy. Having the VRT function only accessed by vmod-debug in-tree is aslo ok. And you're probably right, the use case of vmods passing around PROXY header BLOBs is a bit stretched.

One question though, PROXY v2 can pass much more information, I don't think your code currently forwards anything but IPs (can be optional).

nigoroll:
done, moved to vmod_proxy

nigoroll:
@Dridi so what's your final word? I now moved it to vmod_proxy...

Yes, no TLV support for now, but that's optional anyway so we can add it later. The first step is to get agreement on adding this to VRT and get the merge.

nigoroll:
bugwash: debate about the use case for this

nigoroll:
as this is a prerequisite for #2850 which appears to take on momentum, can we please try to get this one ticked off?

nigoroll:
travis: I do not see why a02022 should fail because of this change and I do not see it failing locally either

nigoroll:
squashed two commits, changed order and force-pushed

bin/varnishd/proxy/cache_proxy_proto.c
```diff
@@ -591,11 +591,36 @@ vpx_enc_port(struct vsb *vsb, const struct suckaddr *s)
 	VSB_bcat(vsb, b, sizeof(b));
 }
 
+/* short path for stringified addresses from session attributes */
```
Dridi: Typo: stringified

bin/varnishd/proxy/cache_proxy_proto.c
```diff
@@ -642,11 +661,12 @@ VPX_Send_Proxy(int fd, int version, const struct sess *sp)
 		vpx_enc_addr(vsb, proto, sas);
 		vpx_enc_port(vsb, sac);
 		vpx_enc_port(vsb, sas);
+		AZ(VSB_finish(vsb));
 	} else
 		WRONG("Wrong proxy version");
 
-	AZ(VSB_finish(vsb));
-	(void)VSB_tofile(fd, vsb);	// XXX: Error handling ?
+	(void)write(fd, VSB_data(vsb), VSB_len(vsb));
```
Dridi: `VSB_tofile(fd, vsb)`

Not sure about the error handling.

bin/varnishd/proxy/cache_proxy_proto.c
```diff
@@ -705,11 +705,11 @@ VPX_Send_Proxy(int fd, int version, const struct sess *sp)
 	} else
 		WRONG("Wrong proxy version");
 
-	(void)write(fd, VSB_data(vsb), VSB_len(vsb));
+	r = write(fd, VSB_data(vsb), VSB_len(vsb));
```
Dridi: Forget my previous comment regarding error handling, you took care of it but simply removed the comment too soon.

lib/libvmod_proxy/vmod.vcc
```diff
@@ -1,8 +1,10 @@
 #-
 # Copyright (c) 2018 GANDI SAS
+# Copyright 2018 UPLEX - Nils Goroll Systemoptimierung
```
Dridi: "Copyright (c)" for consistency?

lib/libvmod_proxy/vmod_proxy.c
```diff
@@ -1,8 +1,10 @@
 /*-
  * Copyright (c) 2018 GANDI SAS
+ * Copyright 2018 UPLEX - Nils Goroll Systemoptimierung
```
Dridi: "Copyright (c)" for consistency?

bin/varnishd/proxy/cache_proxy_proto.c
```diff
@@ -692,7 +692,6 @@ VPX_Send_Proxy(int fd, int version, const struct sess *sp)
 	CHECK_OBJ_NOTNULL(sp, SESS_MAGIC);
 	assert(version == 1 || version == 2);
 	AN(VSB_new(vsb, buf, sizeof buf, VSB_FIXEDLEN));
-	AN(vsb);
```
Dridi: Please squash in 26045e48b83ba276b00ee4056581b33294b6eb49.

bin/varnishd/cache/cache_vrt.c
```diff
@@ -989,7 +989,8 @@ VRT_VSA_GetPtr(VRT_CTX, const struct suckaddr *sua, const unsigned char ** dst)
 }
 
 void
-VRT_Format_Proxy(struct vsb *vsb, VCL_INT version, VCL_IP sac, VCL_IP sas)
+VRT_Format_Proxy(struct vsb *vsb, VCL_INT version, VCL_IP sac, VCL_IP sas,
+    VCL_STRING auth)
```
Dridi: Should we make the last argument generic?

```c
VRT_Format_Proxy(vsb, version, sac, sas, "auth=varnish.test[,other=value,...]");
```

This way we future-proof the addition of future TLVs without planning to break VRT.

lib/libvmod_proxy/vmod.vcc
nigoroll: I once learned that *Copyright (c)* actually is wrong. No idea, but I would like to stick to it, each their own :)

bin/varnishd/cache/cache_vrt.c
nigoroll: Is a parsed sting the best option? Because the TLVs are well defined, I would have thought passing a struct would be appropriate?

bin/varnishd/proxy/cache_proxy_proto.c
nigoroll: will do

bin/varnishd/cache/cache_vrt.c
Dridi: Well, on the VCL side (via `vmod_proxy`) I'd rather avoid pushing the knowledge of what TLVs we support. But something like this could work too:

```
$Function BLOB header(INT version, IP client, IP server, [STRING authority])
```

And in the future we could add other `[optional]` arguments if we want to support more TLVs.

nigoroll: see https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt#L535
The TLVs do not need to be strings, though I agree that the ones which I would consider useful at this point in time are: `PP2_TYPE_ALPN`, `PP2_TYPE_AUTHORITY` and `PP2_TYPE_NETNS`
But already if we wanted to generate `PP2_TYPE_SSL`, a string serialization would not seem to make much sense, IMHO.
I think, for a VRT function to be useful, it should actually save some work for the caller, otherwise it would make more sense for a caller to fill out the tlv structures and just pass them.

Dridi: TLVs don't need to be strings, but they can be represented as such. Maybe this is a good case for VMOD varargs?

```
$Function BLOB header(INT version, IP client, IP server, STRING... tlv)
```

Where the TLV vararg expects a TLV name followed by the value:

```vcl
proxy.header(1, client.ip, server.ip);
proxy.header(2, client.ip, server.ip, "authority", "example.com", "alpn", "http/1");
proxy.header(2, client.ip, server.ip, "ssl", "client,cert", "ssl.version", "TLS1.2"); # involves some parsing
```

Before you say anything, I know, evil.

By the way, do we want `ENUM { 1, 2 } version`? `vmodtool.py` will happilly let you but the libvcc parser will expect an ID token. `ENUM { v1, v2 }`?

nigoroll: @Dridi please let us not create a dependency on yet another feature which might take even more time to get in, after this ticket will celebrate its first anniversary tomorrow.
My suggestion at this point is to mark `VRT_Format_Proxy()` as having a transitional interface and come up with more improvements later. Right now, there is exactly one use case and I do not understand why we would need to keep back that one use case just because we do not know how the ideal interface would look like for hypothetical use cases.
Regarding representing stuff as strings: Sure, we all know what serialization is, but really do we want to `blob.encode(BASE64)` arguments to `proxy.header`?
The enum idea is a good one, thank you!

nigoroll: @Dridi I feel an urge to pick on one aspect: The suggestion to add this to vmod_proxy comes from here: https://github.com/varnishcache/varnish-cache/pull/2845#issuecomment-441276492 If the interface question turned out to be holding this up, I would rather move the vmod function to vmod_debug again in order to get ahead. For now, it really only exists for coverage testing.

nigoroll:
So after some back and forth with @Dridi I have  now move `proxy.header()` back into `vmod_debug`, where I had originally put it. We can still debate useful interfaces later.

Dridi:
I'm sorry for bothering you one more time with this, but you initially put it in `vmod_vtc` probably to make it available out of tree. In which case I would advise against falling back to `vmod_debug`.

nigoroll:
moved to vtc

nigoroll:
@Dridi I think your request for changes is OBE

nigoroll:
OKd during review session with @bsdphk and @Dridi

PR #2848, opened 2018-11-28
Tolerate null IP addresses for ACL matches

Dridi:
A vmod may return a null IP. This relaxes the check in `VRT_acl_match` to
fail the transaction instead of crashing.

Refs #2842


bsdphk:
Is it workable to allow NULL VCP_IP's in general ?

Dridi:
I think we don't want null IPs in general and we have the bogo IP for that. We should however gracefully reject them at the boundaries. There may be more opportunities to sneak in a null IP, I only patched the one I ran into.

slimhazard:
With the UDS feature we had quite a bit of discussion about NULL IPs (a consideration was for VCL ``*.ip`` be NULL is the address is in fact a UDS), and we ended up putting quite a bit of effort into disallowing it.

Although I think the issue here is to not crash on ACL matches if a VMOD returns a NULL IP. In that case I see it more of a robustness issue.

bsdphk:
Bugwash:  No NULL VCL_IP's should flow around.

OK on patch.

More similar patches might be a good idea.

PR #2850, opened 2018-11-30
"via" backends: dynamic backend connections via haproxy

nigoroll:

So the obvious way to connect to TLS backends is to use a TLS"onloader" (a term coined by @slimhazard as in the opposite of "offloader"), which turns a clear connection into a TLS connection to some other fixed config address.
    
haproxy, however, offers a convenient and elegant way to avoid this configuration overhead: The PROXY protocol can also be used to transport the destination address which haproxy is to connect to if a server's address is unspecified (IN_ADDR_ANY / 0.0.0.0). The configuration template for this use case looks like this (huge thank you to @wtarreau for pointing out this great option in haproxy):

```
    listen clear-to-ssl
            bind /my/path/to/ssl_onloader accept-proxy
            balance roundrobin
            stick-table type ip size 100
            stick on dst
            server s0 0.0.0.0:443 ssl ca-file /etc/ssl/certs/ca-bundle.crt
            server s1 0.0.0.0:443 ssl ca-file /etc/ssl/certs/ca-bundle.crt
            server s2 0.0.0.0:443 ssl ca-file /etc/ssl/certs/ca-bundle.crt
            # .. approximately as many servers as expected peers
            # for improved tls session caching
```

With this setup, by connecting to `/my/path/to/ssl_onloader` and sending the address to make a TLS connection to in a PROXY header (as the server address / port), we can reduce the configuration overhead outside varnish substantially. In particular, we do not require a path / port per destination and get back the option to use dynamic backends.
    
This patch, which has yet to undergo final polishing if reviewed positively, implements the basis for simple means of configuring such an ssl onloader:
    
backends can be created with an additional "via" director, which has to resolve to a simple backend. The connection is then made to that address and the actual endpoint address is sent in an additional PROXY header.  Notice that sending yet another proxy header to the actual backend is unaffected. Despite using the same format, the two proxy headers are semantically different: The first, here coined the "preamble", is the address to make the connection to while the (optional) second proxy header continues to contain the addresses of the connection to varnish.
    
This patch is based upon two other PRs:
    
* #2845 for generic proxy header formatting
* #2680 for resolving a director
    
Again, I do not consider this PR final. There are lots of other options to implement the same functionality, but this one is the most direct one I could come up with so far.
    
Reviews and comments are highly appreciated.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -169,8 +169,27 @@ vbe_dir_getfd(struct worker *wrk, struct backend *bp, struct busyobj *bo,
 	bp->vsc->req++;
 	Lck_Unlock(&bp->mtx);
 
+	err = 0;
 	if (bp->proxy_header != 0)
-		VPX_Send_Proxy(*fdp, bp->proxy_header, bo->sp);
+		err += VPX_Send_Proxy(*fdp, bp->proxy_header, bo->sp);
+	if (err < 0) {
+		VSLb(bo->vsl, SLT_FetchError,
+		     "backend %s: proxy write errno %d (%s)",
+		     VRT_BACKEND_string(bp->director),
+		     errno, vstrerror(errno));
+		// account as if connect failed - good idea?
+		VSC_C_main->backend_fail++;
+		bo->htc = NULL;
+		VTP_Close(&pfd);
+		AZ(pfd);
+		Lck_Lock(&bp->mtx);
+		bp->n_conn--;
+		bp->vsc->conn--;
+		bp->vsc->req--;
+		Lck_Unlock(&bp->mtx);
+		return (NULL);
+	}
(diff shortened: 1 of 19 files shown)


nigoroll:
updated, force-pushed

nigoroll:
input from bugwash:
* which timeouts to apply when (via backend vs. actual backend)
* propagate health: if via probe is down, consider all backends down

Still waiting for a go/nogo on the generic idea and feedback on the interface

bsdphk:
Bugwash feeling is that this is generally a good idea, many questions about exact implementation.

nigoroll:
my todo: continue with implementation

nigoroll:
as discussed with @bsdphk :

I see no use in continuing to work on this before #2845 and #2680 are in

nigoroll:
continued in fresh branch -> #2957

PR #2854, opened 2018-12-04
Reduce locking in HSH_Lookup

mbgrydeland:
This patch set reduces the number of times the objhead mutex needs to be acquired for many types of passes through HSH_Lookup(), including the simple cache hit. It does this by changing the functions to deref objhead to take an already locked objhead as input, and then unlocking it as part of the operation. This saves the need to take the lock again during deref.

This change will likely give a measurable performance increase for caches seeing a large number of hits to the same object.

### DIFF
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -77,6 +77,7 @@ static void hsh_rush1(const struct worker *, struct objhead *,
     struct rush *, int);
 static void hsh_rush2(struct worker *, struct rush *);
 static int hsh_deref_objhead(struct worker *wrk, struct objhead **poh);
+static int hsh_deref_objhead_unlock(struct worker *wrk, struct objhead **poh);
 
 /*---------------------------------------------------------------------*/
 
@@ -979,7 +980,7 @@ HSH_DerefObjCore(struct worker *wrk, struct objcore **ocp, int rushmax)
 }
 
 static int
-hsh_deref_objhead(struct worker *wrk, struct objhead **poh)
+hsh_deref_objhead_unlock(struct worker *wrk, struct objhead **poh)
 {
 	struct objhead *oh;
 	struct rush rush;
@@ -988,9 +989,10 @@ hsh_deref_objhead(struct worker *wrk, struct objhead **poh)
 	TAKE_OBJ_NOTNULL(oh, poh, OBJHEAD_MAGIC);
 	INIT_OBJ(&rush, RUSH_MAGIC);
 
+	Lck_AssertHeld(&oh->mtx);
+
 	if (oh == private_oh) {
 		assert(VTAILQ_EMPTY(&oh->waitinglist));
-		Lck_Lock(&oh->mtx);
 		assert(oh->refcnt > 1);
 		oh->refcnt--;
 		Lck_Unlock(&oh->mtx);
@@ -1005,7 +1007,6 @@ hsh_deref_objhead(struct worker *wrk, struct objhead **poh)
 	 * just make the hold the same ref's as objcore, that would
 	 * confuse hashers.
 	 */
-	Lck_Lock(&oh->mtx);
 	while (oh->refcnt == 1 && !VTAILQ_EMPTY(&oh->waitinglist)) {
 		hsh_rush1(wrk, oh, &rush, HSH_RUSH_ALL);
 		Lck_Unlock(&oh->mtx);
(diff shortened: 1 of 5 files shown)


bin/varnishd/cache/cache_hash.c
```diff
@@ -458,10 +459,11 @@ HSH_Lookup(struct req *req, struct objcore **ocp, struct objcore **bocp,
 					oc->hits++;
 				retval = HSH_HIT;
 			}
-			Lck_Unlock(&oh->mtx);
 			*ocp = oc;
```
nigoroll: (very minor): can move below the unlock

bin/varnishd/cache/cache_hash.c
```diff
@@ -538,12 +540,10 @@ HSH_Lookup(struct req *req, struct objcore **ocp, struct objcore **bocp,
 		assert(oh->refcnt > 1);
 		assert(exp_oc->objhead == oh);
 		exp_oc->refcnt++;
-		Lck_Unlock(&oh->mtx);
 		*ocp = exp_oc;
```
nigoroll: (same)

bin/varnishd/hash/hash_classic.c
```diff
@@ -171,13 +171,16 @@ hcl_lookup(struct worker *wrk, const void *digest, struct objhead **noh)
...
-hcl_deref(struct objhead *oh)
+static int v_matchproto_(hash_deref_unlock_f)
+hcl_deref_unlock(struct worker *wrk, struct objhead *oh)
 {
 	struct hcl_hd *hp;
 	int ret;
 
 	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
+	Lck_AssertHeld(&oh->mtx);
+	Lck_Unlock(&oh->mtx);
```
nigoroll: I think this unlock is too early, we would at least need to do the oh->refcnt under the oh mtx

bin/varnishd/hash/hash_critbit.c
```diff
@@ -346,14 +346,16 @@ hcb_start(void)
...
 {
+	int r;
 
+	(void)wrk;
 	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
-	Lck_Lock(&oh->mtx);
+	Lck_AssertHeld(&oh->mtx);
 	assert(oh->refcnt > 0);
-	oh->refcnt--;
+	r = --oh->refcnt;
```
nigoroll: here it's correct, I think

bin/varnishd/hash/hash_simple_list.c
```diff
@@ -107,18 +107,24 @@ hsl_lookup(struct worker *wrk, const void *digest, struct objhead **noh)
...
+hsl_deref_unlock(struct worker *wrk, struct objhead *oh)
 {
 	int ret;
 
+	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
+	Lck_AssertHeld(&oh->mtx);
+	Lck_Unlock(&oh->mtx);
+
 	Lck_Lock(&hsl_mtx);
 	if (--oh->refcnt == 0) {
```
nigoroll: and same as in hash_classic

bin/varnishd/hash/hash_critbit.c
```diff
@@ -364,7 +366,7 @@ hcb_deref(struct objhead *oh)
 #ifdef PHK
 	fprintf(stderr, "hcb_defef %d %d <%s>\n", __LINE__, r, oh->hash);
 #endif
-	return (1);
+	return (r);
```
nigoroll: HSH_DeleteObjHead(wrk, oh); missing for r==0 ?

bin/varnishd/hash/hash_classic.c
mbgrydeland: The classic and simple hashes differ from the critbit in that they have a single global mutex that protects the refcounts of all the objheads, while critbit uses the objhead's own mutex to protect it's refcount. So I believe the code is correct as is.

bin/varnishd/hash/hash_simple_list.c
mbgrydeland: Same reply as in comment on hash_classic.

bin/varnishd/hash/hash_critbit.c
mbgrydeland: Don't think so. Critbit puts it on the cooling list when r==0, and it will be picked up and free'd later. The cooling list is what enables the critbit code to use the objhead's own mutex for protecting it's refcount.

nigoroll: I see it now, thank you. The reason for my confusion is that in the old HSH_DerefObjHead there was a HSH_DeleteObjHead for r==0, but that never happened, and which you now removed. So I thought it had to go here, but I missed the fixed return(1) here. Thank you, I agree now

bin/varnishd/hash/hash_classic.c
nigoroll: you are right.

bin/varnishd/hash/hash_simple_list.c
nigoroll: yes, I was wrong, thank you

bsdphk:
There is no need to change the hash_slinger->deref name.

_{locked|unlocked} suffixes only make sense when both exist and here there is only one.

Because the typedef changes, we're sure to break compilation of any private hash implementations, so the name change doesn't buy us anything there either.

PR #2855, opened 2018-12-04
finalize return from HSH_Lookup() in a separate function

nigoroll:
This splits the classification ("which return value is it") from preparation of the response and thus removes some code duplication and minor differences between the code paths.

There probably is more potential for simplification along this path, but I have tried to closely stick to the original semantics while shuffling code around.

should come after #2854 but I'd appreciate some early review


nigoroll:
will report as separate functions

PR #2856, opened 2018-12-04
Hsh lookup returns with separate functions

nigoroll:
for review only, in a polished version, 
calls to `hsh_lookup_return()` would be replaced with individual function calls.


bsdphk:
I agree that HSH_Lookup() had become too unreadable, but I didn't think any of the two attempts you made today really improved it enough, so I've taken a slightly different take on it in four separate commits (to make bisection easier if I messed up).

PR #2857, opened 2018-12-05
Clean up dead code in hsh_deref_objhead_unlock

daghf:
While we're doing a proper christmas cleaning of cache_hash.c, I figured we should deal with this one too.

Contrary to what the comment in ``hsh_deref_objhead_unlock`` says, req's on waiting list do indeed hold an objhead ref (``req->hash_objhead``), so the condition for the while loop in there will always evaluate to false.

This gets rid of the needless rushing code and replaces it with an assert.


bsdphk:
Good catch

daghf:
Thanks. Merged via b81bb71db0ce1655a0a23bdb47e5b8b3f8315749.

PR #2859, opened 2018-12-05
Retire return(miss) from vcl_hit {}

nigoroll:
Ref #1799


hermunn:
I am very much in favor of this. However, strictly forbidding `return (miss)` in `sub vcl_hit` is obviously existing VCL code, so we probably want a new version of VCL. I think that (actually I am afraid that) the best solution is to emit  a compilation error for code that does `return (miss)` in `sub vcl_hit` when we are running VCL 4.2, and only a warning for VCL 4.1 and earlier.

At the same time we can still change the default VCL for all versions, since the part does not do anything at the moment (due to #2519 and #2547).

nigoroll:
FTR: the travis check is failing because the FSM SVGs would need to be rebuild yet travis does not have graphviz. I Intentionally did not include the generated SVGs in this PR to not add changes without any semantic value.

Re @hermunn I am not sure if we ever wrote down our policy regarding VCL versions, but what I remember is that we wanted to increment it where we actually changed behavior under the hood. Yet a VCC error is an obvious signal that the VCL needs to be adjusted.

The advantage of forcing users to look at their VCL would be to raise awareness that assumed semantics actually never existed.

Dridi:
I think such a change belongs in VCL 4.2, and maybe that would be also be an opportunity to change the boolean semantics of an empty string (#2846). However, we'd need more than just this to bump a new VCL version. It'd be nice to identify what changes we may need in VCL to avoid bumping it every other release.

I think this should be postponed until we have enough hindsight of what h3 will need in terms of VCL changes, like unix domain socket did for VCL 4.1 previously.

nigoroll:
@dridi, if we stuck to what I this was our agreement on the vcl version (only bump it for semantic changes which are not exposed by syntax), we do not need a vcl version bump. And #2846 won't qualify anyway because we agreed that the current semantics is consistent.
To reiterate: In this case we immediately drop an error, so the only reason to bump the vcl version would be backwards compatibility with existing vcls. And my counter-argument on this idea is that users should rather be aware of `return(fetch)` not doing what they might assume.

bsdphk:
Bugwash consensus:  Yes, and without vcl version bump.

PR #2861, opened 2018-12-07
Allow now to be used in vcl_init

rezan:
So I noticed that the following VCL:

```vcl
sub vcl_init {
	std.log("NOW: " + now);
}
```

Produces: `NOW: Thu, 01 Jan 1970 00:00:00 GMT`

This is due to `ctx->now` not being initialized. This PR initializes it.

Now you get: `NOW: Fri, 07 Dec 2018 01:57:06 GMT`

I didn't bother with a VTC, seems pretty straight forward.


bsdphk:
OK, good catch.

PR #2863, opened 2018-12-10
Add bytes() and bytes2string() to vmod std

steven-j-wojcik:
I ran into a use case where I needed to store bytes in a string (i.e `3MB`) to convert to `BYTES` and pass it to `cache_req_body()` later on in the VCL. This was needed because `std.integer()` would throw an error when passing to `cache_req_body()` about using the wrong data type. For consistency's sake, I added the inverse function `bytes2string()`.

The API is as follows:
```
BYTES bytes(STRING b, BYTES fallback)
bytes2string(BYTES b, ENUM {B, KB, MB, GB, TB, PB} unit = "B")
```
`bytes()` utilizes `VNUM_2bytes()` to parse the string and return the converted string in bytes.

While `bytes2string()` takes in a BYTES and a type. The ENUM of types is taken from the possible values that can be parsed in `VNUM_2bytes()`. This function cannot fail without a hard failure in the bytes conversion, so no fallback is needed. The goal of this function is to be the inverse of `bytes()`.

I thought these functions were useful and might be worth adding to VMOD std.

### DIFF
--- a/lib/libvmod_std/vmod_std_conversions.c
+++ b/lib/libvmod_std/vmod_std_conversions.c
@@ -44,6 +44,80 @@
 #include "vtim.h"
 #include "vcc_if.h"
 
+VCL_BYTES v_matchproto_(td_std_bytes)
+vmod_bytes(VRT_CTX, VCL_STRING b, VCL_BYTES d)
+{
+	uintmax_t ret;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+
+	if (VNUM_2bytes(b, &ret, 0)) {
+		return (d);
+	}
+
+	return (ret);
+}
+
+VCL_STRING v_matchproto_(td_std_bytes2string)
+vmod_bytes2string(VRT_CTX, VCL_BYTES b, VCL_ENUM unit)
+{
+	int i;
+	char *ret;
+	double value;
+	size_t len;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+
+	if (!strcmp(unit, "B")) {
+		value = b;
+	} else if (!strcmp(unit, "KB")) {
+		value = b / 1024.;
+	} else if (!strcmp(unit, "MB")) {
+		value = b / (1024. * 1024.);
+	} else if (!strcmp(unit, "GB")) {
+		value = b / (1024. * 1024. * 1024.);
+	} else if (!strcmp(unit, "TB")) {
+		value = b / (1024. * 1024. * 1024. * 1024.);
+	} else if (!strcmp(unit, "PB")) {
+		value = b / (1024. * 1024. * 1024. * 1024. * 1024.);
+	} else {
+		VRT_fail(ctx, "vmod std.bytes2string(): bad unit");
+		return (NULL);
+	}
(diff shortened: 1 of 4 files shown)


lib/libvmod_std/vmod_std_conversions.c
```diff
@@ -44,6 +44,80 @@
...
+vmod_bytes2string(VRT_CTX, VCL_BYTES b, VCL_ENUM unit)
+{
+	int i;
+	char *ret;
+	double value;
+	size_t len;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+
+	if (!strcmp(unit, "B")) {
```
nigoroll: the enum comparison could be more efficient by comparing against the enum pointers

lib/libvmod_std/vmod_std_conversions.c
```diff
@@ -44,6 +44,80 @@
...
+	if (ret[len - i] != '.') {
+		i--;
+	}
+
+	strncpy(&ret[len - i], unit, 2);
+	if (value == b) {
+		ret[len - i + 1] = '\0';
+	} else {
+		ret[len - i + 2] = '\0';
+	}
```
nigoroll: I think this could be written more concisely. For instance, by first checking for a conversion remainder and then using two WS_Printfs based on that resolt.

lib/libvmod_std/vmod_std_conversions.c
slimhazard: If we're really going after efficiency, it could be a switch on the first char in unit -- they happen to be all distinct.

Dridi:
I think we should wait and see the fallout of #2882 before merging this.

bsdphk:
We talked about this at bugwash today and the consensus was to do it.

However, @dridi has pointed out that this area is a bit inconsistent and messy, so we should survey the area first, to make sure this doesn't make it worse.

lib/libvmod_std/vmod_std_conversions.c
steven-j-wojcik: Clearly the best option updated in new commit, thanks.

Dridi:
@steven-j-wojcik can you revisit your patch and try again with `libvarnish`? https://github.com/varnishcache/varnish-cache/issues/2882 was just closed.

nigoroll:
FTR, an alternative suggestion for the bytes formatting is to convert to the highest possible scale which results in a zero remainder.
Here's some code which does the same for durations: https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_ban.c#L770

slimhazard:
```
#include <math.h>
static const char *pfx = {"", "K", "M", "G", "T", "P", "E"};

// Assume bytes is unsigned and no wider than 64-bit.
// Has issues with the int->double conversion when bytes > 2^53 (= 8PB).
int scale = (int)(log2((double)bytes)) / 10;
bprintf(outbuf, "%.3f %sB", ldexp((double)bytes, scale * -10),  pfx[scale]);
```

steven-j-wojcik:
This was already merged into master: e8ffdc452a80835acdae47bc7bdff65baf9c5a26

PR #2868, opened 2018-12-17
Initialize ESIs from the correct header set

daghf:
Ensure ESI children are set up from preq->http0.

This broke in 2c59afe22ba108c7f2ab5ff75b2e30baf4ef9ec7


bsdphk:
Good catch.

PR #2872, opened 2018-12-21
 Make Backend_health fields work in VSL queries

Dridi:
This is just a quick proof of concept. The VSL query code relies too much on asserts and maybe we should add a negative return value to `vslq_test_rec` to gracefully recover from malformed fields and print an error when we encounter one.

I ran into this use case:

```
varnishtest "Backend_health fields"

server s1 {
        rxreq
        txresp
} -start

varnish v1 -vcl {
        backend s1 {
                .host = "${s1_sock}";
                .probe = {
                        .interval = 1m;
                }
        }
} -start

server s1 -wait

shell -expect "HTTP/1.1 200 OK" {
        varnishlog -d -n ${v1_name} -g raw -q 'Backend_health[2] ~ healthy'
}
```

I will update this change based on review feedback.

### DIFF
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -181,7 +181,7 @@ VBP_Update_Backend(struct vbp_target *vt)
 	vt->backend->sick = i;
 
 	AN(dir->vcl_name);
-	VSL(SLT_Backend_health, 0, "%s %s %s %s %u %u %u %.6f %.6f %s",
+	VSL(SLT_Backend_health, 0, "%s %s %s %s %u %u %u %.6f %.6f \"%s\"",
 	    dir->vcl_name, chg ? "Went" : "Still",
 	    i ? "sick" : "healthy", bits,
 	    vt->good, vt->threshold, vt->window,
--- a/lib/libvarnishapi/vsl_query.c
+++ b/lib/libvarnishapi/vsl_query.c
@@ -122,7 +122,7 @@ vslq_test_rec(const struct vex *vex, const struct VSLC_ptr *rec)
 	double lhs_float = 0.;
 	const char *b, *e, *q;
 	char *p;
-	int i;
+	int i, dq;
 
 	AN(vex);
 	AN(rec);
@@ -144,15 +144,40 @@ vslq_test_rec(const struct vex *vex, const struct VSLC_ptr *rec)
 
 	/* Field */
 	if (vex->lhs->field > 0) {
-		for (e = b, i = 0; *e && i < vex->lhs->field; i++) {
+		for (e = b, i = 0, dq = 0; *e && i < vex->lhs->field; i++) {
+			/* Skip end of previous field */
+			if (dq) {
+				assert(e > b);
+				assert(*e == '"');
+				dq = 0;
+				e++;
+				if (*e == '\0')
+					break;
+				if (!isspace(*e))
+					return (0);
+			}
+
 			b = e;
 			/* Skip ws */
 			while (*b && isspace(*b))
 				b++;
+
+			dq = (*b == '"');
+			if (dq)
+				b++;
 			e = b;
-			/* Skip non-ws */
-			while (*e && !isspace(*e))
-				e++;
+
+			if (dq) {
+				/* Find end of string */
+				while (*e && *e != '"')
+					e++;
+				if (*e != '"')
+					return (0);
+			} else {
+				/* Skip non-ws */
+				while (*e && !isspace(*e))
+					e++;
+			}
 		}
 		assert(b <= e);
 		if (*b == '\0' || i < vex->lhs->field)
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -508,7 +508,7 @@ vca_accept_task(struct worker *wrk, void *arg)
 				    lport, VTCP_PORTBUFSIZE);
 			}
 
-			VSL(SLT_SessError, 0, "%s %s %s %d %d %s",
+			VSL(SLT_SessError, 0, "%s %s %s %d %d \"%s\"",
 			    wa.acceptlsock->name, laddr, lport,
 			    ls->sock, i, vstrerror(i));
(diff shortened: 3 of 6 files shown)


Dridi:
I discussed it with @mbgrydeland on IRC and there are a couple things.

The problem I'm trying to solve is not clear, so let's clarify: Currently VSL fields are separated by spaces (I think it should be blanks by the way) but at least in one place we document fields that contain spaces (`Backend_health` records). The proposed solution is to support quoted strings for such spaces.

Once the problem is solved, the test case passes. This test case should fail on any recent branch today.

In @mbgrydeland's review we came to the conclusion that the existing assertions are correct, and only the `xxxassert` I temporarily introduced may be a problem. We already fail to match malformed VSL, so I should turn this assert into a conditional return.

I will update the changes accordingly.

Dridi:
Updated accordingly.

fgsch:
What if we change the space to underscore for the second field?

Dridi:
I think supporting spaces inside a field would be better. VMOD authors could produce structured `VCL_Log` records, usually with a prefix, that could contain either empty fields or fields with spaces. And we keep the ability to have such fields for error messages and whatnots.

fgsch:
I'm not against supporting spaces within fields but it adds more complexity and I'd prefer to avoid it iff we can. 
For this particular case, and specifically about the second field, I think it's fine having e.g. Still_healthy or still-healthy.

Dridi:
I agree, and I prefer tokens like `still-healthy` or `went-sick` for `Backend_health[2]` fields. I can open a separate pull request for that but overall that's a breaking change.

This change (allowing spaces inside fields by means of double quotes) doesn't technically break. It fixes the fact that we can't use `Backend_health` fields in a VSL query today. I'd like to keep both changes separate.

bsdphk:
Please, not another unrolled/handwritten "take a string apart at the spaces" implementation.

Either use VAV or come up with a suitable library function for this kind of work.

Dridi:
Bugwash consensus:

- VAV uses malloc so we'd rather avoid it
- docfix to split Backend_health[2] in two
- we may evolve the VSL binary format in the future
- @slimhazard @mbgrydeland and @Dridi to figure what to do now
- @Dridi to update the documentation accordingly

fgsch:
@Dridi ping

Dridi:
Bugwash consensus:

- solve merge conflicts accumulated over time
- require fields with spaces to be quoted
- fix other fields than `Backend_health` (the latter was docfixed instead)

Dridi:
It turned out `Backend_health` still needed fixing for its last field, which I kept as part of the rebase against master (from more than 700 patches ago!) and ultimately I swept all VSL tags hunting for field issues and found some.

Ready again for review.

lib/libvarnishapi/vsl_query.c
```diff
@@ -144,15 +144,40 @@ vslq_test_rec(const struct vex *vex, const struct VSLC_ptr *rec)
 
 	/* Field */
 	if (vex->lhs->field > 0) {
-		for (e = b, i = 0; *e && i < vex->lhs->field; i++) {
+		for (e = b, i = 0, dq = 0; *e && i < vex->lhs->field; i++) {
+			/* Skip end of previous field */
+			if (dq) {
```
nigoroll: why this block...

lib/libvarnishapi/vsl_query.c
```diff
@@ -144,15 +144,40 @@ vslq_test_rec(const struct vex *vex, const struct VSLC_ptr *rec)
...
-			while (*e && !isspace(*e))
-				e++;
+
+			if (dq) {
+				/* Find end of string */
+				while (*e && *e != '"')
+					e++;
+				if (*e != '"')
+					return (0);
+			} else {
```
nigoroll: ... and not just remove the else here, also setting `dq=0` ?

bin/varnishd/cache/cache_backend_probe.c
```diff
@@ -181,7 +181,7 @@ VBP_Update_Backend(struct vbp_target *vt)
 	vt->backend->sick = i;
 
 	AN(dir->vcl_name);
-	VSL(SLT_Backend_health, 0, "%s %s %s %s %u %u %u %.6f %.6f %s",
+	VSL(SLT_Backend_health, 0, "%s %s %s %s %u %u %u %.6f %.6f \"%s\"",
```
nigoroll: we should probably mark here with an XXX that the `%s` must not contain any quotes

lib/libvarnishapi/vsl_query.c
Dridi: This block checks what's after the closing dquote of the previous field, skips the dquote `e` is pointing to and makes sure a proper "whitespace" field delimiter follows the closing dquote.

lib/libvarnishapi/vsl_query.c
Dridi: We need to remember `dq` before we move on to the next field, so both branches really are mutual-exclusive.

See 681da75565cc897ffcb7a1921b96e17c308483ac for the two syntax errors this `if` and the other `if` blocks each catch.

bin/varnishd/cache/cache_backend_probe.c
Dridi: This is something we get from the backend, if the reason of a backend response contains double quotes there's little I'm willing to do to preserve the last field of `Backend_health`. The reason is only meant to be human-readable so as long as we have the protocol version and status this is fine.

Please note that there is no `:reason` pseudo header field since h2.

Dridi:
Since I have both @nigoroll's approval and Chief VSL Officer @mbgrydeland (who thought this was merged already) I'm going ahead and merging this as soon as CI passes. We can revisit the details later to simplify the field dissection now that I added more test coverage.

lib/libvarnishapi/vsl_query.c
nigoroll: I find this confusing, especially the "field""nospace" case, tried to find a simpler solution, failed to get to one within ~20 minutes and give up for now because it's probably not relevant.
But I do wonder if we need to handle escaped quotes....

PR #2886, opened 2019-01-25
Relaxed UDS backend restrictions

### LINKED ISSUE #2884: Fails to start when backend's listen socket does not (yet) exist
## Expected Behavior

Varnish should start irrespective of backend's listen socket availability. 
Similar to how it does if HTTP TCP socket of the backend is not / yet available.

## Current Behavior

Varnish halts running immediately if the backend's socket does not exist:

```
varnishd[11705]: Error:
varnishd[11705]: Message from VCC-compiler:
varnishd[11705]: Backend path: Cannot stat: No such file or directory
varnishd[11705]: ('/etc/varnish/default.vcl' Line 8 Pos 13)
varnishd[11705]: .path = "/var/run/nginx.sock";
varnishd[11705]: ------------#####################-
varnishd[11705]: In backend specification starting at:
varnishd[11705]: ('/etc/varnish/default.vcl' Line 7 Pos 1)>
```

## Steps to Reproduce

Define backend's listen socket in VCL:

```
vcl 4.1;
backend default {
    .path = "/var/run/nginx.sock";
}
```

## Context

Varnish fails to start when for example NGINX has not started yet, so one has to define order of running services in a way that NGINX runs before Varnish. Instead, it would be good to have Varnish do the same thing it does with non-unix-socket based backends.

## Environment

* Varnish 6.0.2
* CentOS 7

slimhazard:
VCC issues a warning, but does not fail, if stat(UDS backend) fails with ENOENT or EACCES.

This makes it possible to start the UDS-listening peer, or set the socket file's permissions, after starting Varnish or loading VCL with a UDS backend.

If the file at the specified path does exist and is accessible, then it must be a socket, otherwise the VCL load fails.

Closes #2884

### DIFF
--- a/bin/varnishtest/tests/v00038.vtc
+++ b/bin/varnishtest/tests/v00038.vtc
@@ -123,16 +123,39 @@ varnish v1 -errvcl "Must be an absolute path:" {
 	}
 }
 
-shell { rm -f ${tmpdir}/foo }
-
-varnish v1 -errvcl "Cannot stat:" {
-	backend b1 {
-		.path = "${tmpdir}/foo";
-	}
-}
-
 varnish v1 -errvcl "Not a socket:" {
 	backend b1 {
 		.path = "${tmpdir}";
 	}
 }
+
+# VCC warns, but does not fail, if stat(UDS) fails with ENOENT.
+shell { rm -f ${tmpdir}/foo }
+
+varnish v1 -vcl { backend b {.host="${bad_ip}";} }
+
+varnish v1 -cliexpect "(?s)Cannot stat:.+That was just a warning" \
+	{vcl.inline test "vcl 4.1; backend b {.path=\"${tmpdir}/foo\";}"}
+
+# VCC also warns but doesn't fail for EACCES. Tested in c00086.vtc.
+
+# The following test verifies that Varnish continues connecting to a
+# socket file, even if the listener at that location changes.
+
+server s1 -listen "${tmpdir}/server.sock" {
+	rxreq
+	txresp -hdr "Connection: close" -hdr "Cache-Control: max-age=0"
+} -start
+
+varnish v1 -vcl {
+	backend b {.path = "${s1_sock}"; }
+} -start
+
+client c1 {
+	txreq
+	rxresp
+} -run
+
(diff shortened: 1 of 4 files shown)


Dridi:
I think we may need one more test case:

- server s1 listens to a uds
- varnish makes a successful bereq to s1
- server s2 takes over s1's uds' path with a new socket
- varnish makes a successful bereq to s2

Otherwise LGTM except for a nitpick. This results in a message like this:

```
%s: Cannot stat: %s\n

# or

%s: Cannot stat: %s\n
(That was just a warning)\n
```

I'd rather have something like:

```
%s: Cannot stat: %s\n

# or

Warning: %s: Cannot stat: %s\n
```

Dridi:
About the nitpick, I see it's the same as other places so feel free to ignore it. The test case on the other hand I'd rather have in to make sure that a backend restarting while Varnish is running is not killing backend traffic until the next VCL reload or whatnot.

slimhazard:
The ``(That was just a warning)`` output was chosen to match the warnings generated when VCC checks references, finds unused backends or vcl subs, but ``vcc_err_unref`` is false.

@Dridi not saying you're wrong, but if we want to change the warning output, the other one should probably be changed as well. And maybe that's a subject for another issue?

slimhazard:
* Rebased to current master
* Added a test as @Dridi described
* Removed a minor superfluousness in the code (checking a variable for == 0 that is always 0 in that context)
* Force-pushed

IMO ready for merge now, but I'll wait until after bugwash.

bin/varnishtest/tests/v00038.vtc
```diff
@@ -123,16 +123,39 @@ varnish v1 -errvcl "Must be an absolute path:" {
...
+	txresp -hdr "Connection: close" -hdr "Cache-Control: max-age=0"
+} -start
+
+varnish v1 -vcl {
+	backend b {.path = "${s1_sock}"; }
+} -start
+
+client c1 {
+	txreq
+	rxresp
```
Dridi: Might want to check the status too

slimhazard:
Merged to master, travis is green.

PR #2888, opened 2019-01-25
Output correct director health in CLI

### LINKED ISSUE #2887: director health wrong in cli
for directors which implement the `health` callback method (and do not set the `struct director` member `sick`), the health state as reported in the cli is wrong.

Example https://code.uplex.de/uplex-varnish/libvmod-all_healthy/blob/master/src/vtc/vmod_all_healthy.vtc

for the case 

```
varnish v1 -cliok "backend.set_health be_a sick"

varnish v1 -cliok "backend.list -j"

client c1 {
        txreq
        rxresp
        expect resp.status == 200
        expect resp.http.a == false
        expect resp.http.b == true
        expect resp.http.be == false
        expect resp.http.name == be
} -run
```

The `backend.list -j` output is:

```
**** v1    0.6 CLI RX|    "vcl1.be": {
**** v1    0.6 CLI RX|      "type": "all_healthy",
**** v1    0.6 CLI RX|      "admin_health": "probe",
**** v1    0.6 CLI RX|      "probe_message": "healthy",
**** v1    0.6 CLI RX|      "last_change": 1548431491.324
**** v1    0.6 CLI RX|    }
```

When in fact it should be sick.

nigoroll:
As `VRT_Healthy` needs a `VRT_CTX`, we move `vcl_get_ctx` and `vcl_rel_ctx` to `cache_varnishd.h` scope as `VCL_Get_CliCtx` and `VCL_Rel_CliCtx`.

It has been discussed before if all varnishd cli commands should have a `VRT_CTX`, so the natural solution would seem  to add one in or close to `VCLS_Poll()`

But: CLI commands need specific ctxes (e.g. `vcl.load` needs `VCL_MET_INIT`), so (for now) it appears sensible to limit the ctx' scope to individual backend method calls.

Fixes #2887

### DIFF
--- a/bin/varnishd/cache/cache_director.c
+++ b/bin/varnishd/cache/cache_director.c
@@ -277,6 +277,20 @@ struct list_args {
 	const char	*jsep;
 };
 
+static const char *
+cli_health(struct director *d)
+{
+	struct vrt_ctx *ctx;
+	VCL_BOOL healthy;
+
+	ctx = VCL_Get_CliCtx(0,0);
+	healthy = VRT_Healthy(ctx, d, NULL);
+	VCL_Rel_CliCtx(&ctx);
+	AZ(ctx);
+
+	return (healthy ? "healthy" : "sick");
+}
+
 static int v_matchproto_(vcl_be_func)
 do_list(struct cli *cli, struct director *d, void *priv)
 {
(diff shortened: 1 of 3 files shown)


bsdphk:
OK

As mentioned in bugwash, "sick/healthy" is very binary, and maybe we should think about giving the backends/directors a chance to write the text for CLI usage, so they can bring more nuance ?

That way a director with 10 backends could return a health of "4/10 healthy" to indicate that yes, it can still do its job, but things are far from perfect.

"Sick" and "Healthy" should probably still be reserved for "nothing works" and "everything is OK" in that case.

PR #2890, opened 2019-01-28
Fail on NULL IPs in VRT

### LINKED ISSUE #2860: Null IP addresses
Followup to #2848.

For the remaining VRT symbols involving `VCL_IP`s that could be NULL:

- `VRT_IP_string` returns a null string
- `VRT_ipcmp` fails the comparison, not the transaction
- `VRT_VSA_GetPtr` may be #2800 material

I can patch the first two to make them fail the transaction.

In addition, symbols that indirectly take a `VCL_IP`:

- `VRT_new_backend_clustered` panics
- `VRT_new_backend` calls `VRT_new_backend_clustered`

Again, I think they should fail the transaction, but they are less straightforward to deal with. There's probably some `ctx->handling` checking that needs to happen in core code too.

Dridi:
This breaks the API since some functions didn't take a VRT_CTX until now.

Closes #2860

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -499,10 +499,12 @@ VRT_new_backend_clustered(VRT_CTX, struct vsmw_cluster *vc,
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	CHECK_OBJ_NOTNULL(vrt, VRT_BACKEND_MAGIC);
-	if (vrt->path == NULL)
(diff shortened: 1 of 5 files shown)


bsdphk:
Can we make the message "Illegal IP" instead of "null IP" ?

Dridi:
Done!

bsdphk:
Sorry for overlooking this one.

OK to commit, but please make sure the include/vrt.h comment does not time-travel.

Dridi:
No worries, I will resolve the conflict and push directly.

Dridi:
1a6af67c731cd295899a5598057d63d3b9f93836

PR #2895, opened 2019-02-05
centralize definition of backend.list fields width

nigoroll:
Yes, this is fugly, yet formatting of the backend.list output
is spread all over the place (including vmods via the vdi_list_f
callback), so until we decide it is worth spending time on separating
formatting from content, I see no other option than making the
field widths known to all vrt.


nigoroll:
IRC:
> (11:36:18) phk: slink, yeah, that's good enough for now.
> (11:36:42) phk: slink, we may want to have some generic columnizing code later, but I suspect the easier way to do that is to reformat the json output

PR #2908, opened 2019-02-17
add vrt & vmod support to look up a director by name

nigoroll:
This allows for relevant simplification of vcl with varying backend / director names, as when identical VCL is deployed to several nodes.
Toy example:
```
vcl_init {
    new myself = directors.round-robin();
    myself.add_backend(directors.lookup(server.identity));
}
```

### DIFF
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -261,6 +261,35 @@ VRT_DisableDirector(VCL_BACKEND d)
 	vdir->health_changed = VTIM_real();
 }
 
+VCL_BACKEND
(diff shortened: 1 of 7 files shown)


fgsch:
I've had a similar diff to this but doesn't increase lock contention considerably?

nigoroll:
@fgsch yes. My use case is just vcl_init, so it's not that relevant, but I left this comment:
https://github.com/varnishcache/varnish-cache/blob/500dda49903c70458a625365ab31b5f0b420a9b0/bin/varnishd/cache/cache_vrt_vcl.c#L277

bsdphk:
This makes sense, but it must be limited to vcl_init{} only.

PR #2909, opened 2019-02-17
atomic write for vbe32end

nigoroll:
If accepted, the same technique should be applied to all the vend.h macros

Fixes #2904

### DIFF
--- a/include/vend.h
+++ b/include/vend.h
@@ -98,12 +98,17 @@ vbe16enc(void *pp, uint16_t u)
 static __inline void
 vbe32enc(void *pp, uint32_t u)
 {
-	uint8_t *p = (uint8_t *)pp;
+	union {
+		uint8_t p[4];
+		uint32_t i;
+	} d;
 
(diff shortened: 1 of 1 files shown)


mbgrydeland:
I do not think that will fix the issue. This is because OA_GZIPBITS is not aligned, and may (quite likely I believe) span cachelines in the CPU.

nigoroll:
@mbgrydeland Interesting, I was not aware that unaligned writes would not be atomic on i386/x64. Do you have a reference?

But you definitely have a point here: Even if unaligned writes were atomic, I believe that my proposal would fail on platforms forcing aligned writes (sparc comes to mind).

Yet is this the right way to go then? Wouldn't it make more sense to align at least the fixed object attributes?

nigoroll:
And just to clarify: with this patch, I do not see #2904 happening on my machine at least

bsdphk:
And the C code proposed does not result in guaranteed atomic writes once the compiler is done with it.

This is not how we should be solving the underlying problem.

PR #2913, opened 2019-02-18
take2: use our knowledge about master file descriptors also for VSUBs

### LINKED ISSUE #2730: Closing all FD's even those which aren't open

nigoroll:
Follow up on #2843: @bsdphk is this like what you had in mind?

Regarding the information leakage argument, we would have the same issue
for the varnish worker child, so I do not think that we are taking on
additional risks.

Fixes #2730

### DIFF
--- a/bin/varnishd/mgt/mgt.h
+++ b/bin/varnishd/mgt/mgt.h
@@ -71,11 +71,13 @@ void MAC_Arg(const char *);
 int MAC_reopen_sockets(void);
 
 /* mgt_child.c */
+struct vbitmap;
 void MCH_Init(void);
 int MCH_Running(void);
 void MCH_Stop_Child(void);
 int MCH_Start_Child(void);
 void MCH_TrackHighFd(int fd);
+void MCH_FdInfo(int *, int *, const struct vbitmap **);
 void MCH_Cli_Fail(void);
 
 /* mgt_cli.c */
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -50,6 +50,7 @@
 #include "vfil.h"
 #include "vlu.h"
 #include "vtim.h"
+#include "vsub.h"
 
 #include "common/heritage.h"
 #include "common/vsmw.h"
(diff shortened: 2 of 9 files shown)


bsdphk:
I've added a hack to read the /proc/`getpid`/fd directory if we have no closefrom() (see 14721f537)

Let me know if that helps enough.

nigoroll:
`strace -f` tells me that for `ulimit -n 4096`, the number of `close()` calls is down from 17056 to 707 for a trivial vcl.

PR #2915, opened 2019-02-19
Stop shell scripts on autotools errors.

lkarsten:
When configure is still around, it looks like your automake.am changes
were used, but in fact nothing happened on disk.

### DIFF
--- a/autogen.des
+++ b/autogen.des
@@ -1,6 +1,7 @@
 #!/bin/sh
 #
 # Use this when doing code development
+set -e	# Stop on errors.
 
 SRCDIR=${SRCDIR:-.}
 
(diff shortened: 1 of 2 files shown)


lkarsten:
There may be cleaner ways to do this, so here is a PR. Feedback appreciated.

(I think `set -e` is the portable way, but only tested on recent linux and and freebsd7.)

autogen.des
```diff
@@ -23,7 +24,11 @@ else
 fi
 
 rm -f configure
-(cd $SRCDIR && . ./autogen.sh 2>&1 | egrep -v "(subdir-objects|is in a subdirectory)")
+TREEDIR=`pwd`
```
nigoroll: You are removing the output filter here. And I think a subshell is not that bad after all, just the return code needed to be saved.

autogen.sh
```diff
@@ -47,5 +48,9 @@ set -ex
 $LIBTOOLIZE --copy --force
 aclocal -I m4 -I .
 autoheader
-automake --add-missing --copy --foreign
+AM_OUTPUT=$(automake --add-missing --copy --foreign 2>&1 | egrep -v "(subdir-objects|is in a subdirectory)")
```
nigoroll: by moving the automake into a subshell, the `-e` option has no effect.

lkarsten: The main issue seems to be that warnings from automake does not return an error, yet it does not complete.

I wonder if deleting configure on top is the easier fix, to avoid the error masking.

lkarsten:
No, this whole thing is flawed. Closing unmerged.

autogen.sh
nigoroll: Sounds simple and clean

Issue #2899, opened 2019-02-06
VIP12bis (extend std.conversions)


bsdphk:
Should we extend functionality of std.conversions:

    TIME std.time([STRING input], [TIME fallback], [REAL real_input], [INT int_input])

See discussion in: https://github.com/varnishcache/varnish-cache/wiki/VIP12:-vmod-polymorphism-(for-type-conversions)

nigoroll:
I would prefer if we kept the argument names short, but otherwise sgtm
```
TIME std.time([STRING string], [TIME fallback], [REAL real], [INT int])
```

bsdphk:
Bugwash consensus: Yes we should.

bsdphk:
VIP12 updated, now we just need somebody to do it.

nigoroll:
I've started working on it

PR #2921, opened 2019-02-27
make vmod_std conversions more flexible

nigoroll:
In order to avoid the cross product problem with conversion functions
(from x to y types would require x*y functions), we add flexibility
regarding the input parameters to conversion functions:

Each convertion function named after the destination type now takes all
sensible arguments by name.

int is named integer because of the reserved symbol name "int" in C.

All functions should be fully backwards compatible (existing vtcs
continue to work), but compile time checks are now effectively removed.
The conversion functions now trigger vcl errors if used incorrectly or
if conversion errors occur and no failback is provided. However, if a
failback is provided, vcl errors are only raised for usage errors.

For consistency, the conversion functions now only ever truncate if
necessary. std.round() is added for explicit rounding where required.

Existing functions which are now obsolete are marked deprecated.

Ref #2899
Ref https://github.com/varnishcache/varnish-cache/wiki/VIP12:-vmod-polymorphism-(for-type-conversions)

### DIFF
--- a/bin/varnishtest/tests/m00005.vtc
+++ b/bin/varnishtest/tests/m00005.vtc
@@ -9,8 +9,36 @@ varnish v1 -vcl+backend {
 	import std;
 
 	sub vcl_deliver {
-		set resp.http.duration = std.duration(req.http.duration, 1s) + 1000000s;
-		set resp.http.bytes = std.bytes(req.http.bytes, 10K) + 512B;
+		if (req.http.d-real) {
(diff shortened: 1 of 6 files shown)


nigoroll:
Actually I do not see much need to discuss this - we had already decided that this is what we want, labeling bugwash just to make sure it gets attention.

bsdphk:
Cross ref to #2863 because we are in the same general "what kind of language is VCL" space.

PR #2925, opened 2019-03-03
Ban lurker dying objcores

hermunn:
Not evaluating dying objects against bans seems like a good idea, and I propose a way of fixing that here.

The *holdoff* behavior change in the last commit is there to sleep less (for example) when a lone object between the markers dies while the ban lurker is *holdoff* sleeping. When it removes the dying object on the next run, we do not want it to sleep again before returning `NULL`.

It could be that we should also do something smarter with *busy* objects, maybe move them past the end marker instead of consider them "contested", but that is not clear to me.

This is code very hard to test in a VTC. Perhaps a debugging VDP which sleeps a lot, can allow us to have dying objects in the ban lists without too much fuzz, and make this commit testable.

### DIFF
--- a/bin/varnishd/cache/cache_ban_lurker.c
+++ b/bin/varnishd/cache/cache_ban_lurker.c
@@ -169,6 +169,21 @@ ban_lurker_getfirst(struct vsl_log *vsl, struct ban *bt)
 		if (!Lck_Trylock(&oh->mtx)) {
 			if (oc->refcnt == 0 || oc->flags & OC_F_BUSY) {
 				Lck_Unlock(&oh->mtx);
+			} else if (oc->flags & (OC_F_DYING | OC_F_FAILED)) {
+				/*
+				 * For a dying object we seize the opportunity
+				 * to remove it completely off the ban list,
+				 * now that we have both the oh and ban
+				 * mutexes.
+				 */
+				AZ(oc->flags & OC_F_BUSY);
+				noc = VTAILQ_NEXT(oc, ban_list);
+				VTAILQ_REMOVE(&bt->objcore, oc, ban_list);
+				oc->ban = NULL;
+				bt->refcount--;
+				Lck_Unlock(&oh->mtx);
+				oc = noc;
+				continue;
 			} else {
 				/*
 				 * We got the lock, and the oc is not being
--- a/bin/varnishd/cache/cache_ban_lurker.c
+++ b/bin/varnishd/cache/cache_ban_lurker.c
@@ -365,7 +365,7 @@ ban_lurker_work(struct worker *wrk, struct vsl_log *vsl)
 
 	/*
 	 * conceptually, all obans are now completed. Remove the tail. If it
-	 * containted the first oban, all obans were on the tail and we're
+	 * contained the first oban, all obans were on the tail and we're
 	 * done.
 	 */
 	if (ban_cleantail(VTAILQ_FIRST(&obans)))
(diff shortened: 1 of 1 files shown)


bin/varnishd/cache/cache_ban_lurker.c
```diff
@@ -169,6 +169,21 @@ ban_lurker_getfirst(struct vsl_log *vsl, struct ban *bt)
...
+				 * mutexes.
+				 */
+				AZ(oc->flags & OC_F_BUSY);
+				noc = VTAILQ_NEXT(oc, ban_list);
+				VTAILQ_REMOVE(&bt->objcore, oc, ban_list);
+				oc->ban = NULL;
+				bt->refcount--;
+				Lck_Unlock(&oh->mtx);
+				oc = noc;
+				continue;
```
nigoroll: This does look like a good idea, but if we add this, we should also optimize https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_ban.c#L265 to not take the `ban_mtx` for `oc->ban == NULL` (and re-check under the lock)

bin/varnishd/cache/cache_ban_lurker.c
```diff
@@ -149,13 +150,15 @@ ban_lurker_getfirst(struct vsl_log *vsl, struct ban *bt)
...
 			assert(move_oc == 0);
 
-			/* hold off to give lookup a chance and reiterate */
-			VSC_C_main->bans_lurker_contention++;
-			Lck_Unlock(&ban_mtx);
-			VSL_Flush(vsl, 0);
-			VTIM_sleep(cache_param->ban_lurker_holdoff);
-			Lck_Lock(&ban_mtx);
-
+			if (contested) {
```
nigoroll: I do not understand why we would need this code. In the existing code, the holdoff only happens if there are ocs inbetween the two markers, and cleaning out dying and failed ocs does not change any of that.

hermunn: An object might be contested on the first pass, then moved to in between the markers, but on the second time it is dying and the *only* object between the markers. Then we do not want to *holdoff* sleep, since there were actually no contested objects between the markers. This is tricky stuff, and I might be wrong.

bin/varnishd/cache/cache_ban_lurker.c
hermunn: I think you are right, but we should really look for potential races if we do this. Martin had the same idea, but was not completely sure it was safe.

nigoroll: I did have my doubts as well and did check, so far I think it is safe because once we have a ban, we only ever modify it and never transition from `NULL` to `!NULL` except in `BAN_RefBan()` and `BAN_NewObjCore()`. So I think that if we re-check under the `ban_mtx`, it's safe.

bin/varnishd/cache/cache_ban_lurker.c
nigoroll: ok, I think I see the intention now, but I don't think it would work like this:

We want the holdoff to happen if there were any contested ocs. Yet if one contested oc was returned and _getfirst gets called the second time, contested will be 0 and we won't hold off.
Also I would consider the intention a micro-optimization not worth the complication.

bsdphk:
Postponed until after 2019-03-15 release

hermunn:
Closing in favor of #2992.

Issue #2926, opened 2019-03-04
VCL_INT, VCL_BYTES and the 64-bit range


slimhazard:
Raising this issue due to 0f7f757c478351e7a30c9820f7542b868743b7b8 and the discussion there. IMO this looks like a regression.

["Upgrading" for 6.0](https://varnish-cache.org/docs/6.0/whats-new/upgrading-6.0.html#other-changes) had this to say:

> The VCL_INT and VCL_BYTES types are now defined to be strictly 64 bit (rather than leave it to whatever your platform defines as long). But you may not get that full precision, for reasons discussed in the change log.

... linking to [changes.rst,](https://github.com/varnishcache/varnish-cache/blob/master/doc/changes.rst) which had this at the time:

> Integers in VCL are now 64 bits wide across all platforms (implemented as int64_t C type), but due to implementation specifics of the VCL compiler (VCC), integer literals' precision is limited to that of a VCL real (double C type, roughly 53 bits).

> In effect, larger integers are not represented accurately (they get rounded) and may even have their sign changed or trigger a C compiler warning / error.

My understanding at the time was that this was regarded as a bug. And I'm sure I remember understanding the work leading to 6.1 as having fixed the bug, in particular b637c111d6db652f7849c0d68fa19491f0f9c598, 5d1dd90aa55918d0d37fa87fd7a0374fa5aabe19 and da727285da6d62e70dd2038dcf6c79d91bba7564.

["Upgrading" for 6.1](https://varnish-cache.org/docs/6.1/whats-new/upgrading-6.1.html#other-changes) then said:

> We have fixed a bug that had limited the precision available for the INT data type, so you now get the full 64 bits.

That statement seems clear enough.

But 0f7f757c478351e7a30c9820f7542b868743b7b8 now defines the macros ``VCL_INT_MAX`` and ``VCL_BYTES_MAX`` in ``vrt.h`` to limit them to 53 bits, and the "to-integer" conversion functions return the fallback if they are out of that range.

The previous changes make it possible to use the full range of 64-bit literals in VCL, but not the conversion. IMO users will not find it especially plausible if we say that INT and BYTES have 64-bit range, but not if you're converting a string.

[...]

PR #2929, opened 2019-03-05
64bit integer precision for VCL_INT / VCL_BYTES

nigoroll:
Ref #2926

### DIFF
--- a/bin/varnishd/mgt/mgt_param_tweak.c
+++ b/bin/varnishd/mgt/mgt_param_tweak.c
@@ -261,20 +261,20 @@ tweak_generic_bytes(struct vsb *vsb, volatile ssize_t *dest, const char *arg,
 
 	if (arg != NULL && arg != JSON_FMT) {
 		if (min != NULL) {
-			p = VNUM_2bytes(min, &rmin, 0);
+			p = VNUM_2bytes(min, &rmin);
 			if (p != NULL) {
 				VSB_printf(vsb, "Invalid min-val: %s\n", min);
 				return (-1);
 			}
 		}
 		if (max != NULL) {
-			p = VNUM_2bytes(max, &rmax, 0);
+			p = VNUM_2bytes(max, &rmax);
 			if (p != NULL) {
 				VSB_printf(vsb, "Invalid max-val: %s\n", max);
 				return (-1);
 			}
 		}
-		p = VNUM_2bytes(arg, &r, 0);
+		p = VNUM_2bytes(arg, &r);
 		if (p != NULL) {
 			VSB_cat(vsb, "Could not convert to bytes.\n");
 			VSB_printf(vsb, "%s\n", p);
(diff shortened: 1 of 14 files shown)


slimhazard:
Getting way pedantic here: do we have any rules to the effect that a VCL INT is only decimal, or could it be octal or hex, by the usual lexical rules? (Leading '0' for octal, prefix '0x' for hex.)

I seem to recall that the lexer for numbers in a VSL query string might interpret them as octal or hex (but I didn't find the code that I thought I remember, maybe if I search a little longer). Of course VCL INTs and integers in a log query are not the same thing, but users might well expect the same rules.

Back when I went a little overboard re-implementing VSL queries in another language, I did [this here](https://code.uplex.de/uplex-varnish/varnishapi/blob/master/pkg/log/bytesconv.go) to parse numbers in a query as float, decimal, octal or hex, because I was sure that Varnish VSLQ does the same thing (test vectors [here](https://code.uplex.de/uplex-varnish/varnishapi/blob/master/pkg/log/bytesconv_test.go)).

nigoroll:
@slimhazard `VNUMpfx()` does decimal https://github.com/varnishcache/varnish-cache/blob/master/lib/libvarnish/vnum.c#L72
At this point I would (only) want to replace it. Other bases should be easy to add and we could do that later.

On a related topic, I am working on adding support for scaled integers.

nigoroll:
I have pushed additional commits to complement (but not yet test & document) VCL_BYTES parsing as integers. I would appreciate reviews by all, but particularly @bsdphk : Any feedback if the general direction is accepted would be very helpful to reduce wasted efforts.
Patches are not yet polished!
Comments:

* *way out of precision hell & tests*  d5ceb7f 

as can be seen from the removed code, I had actually made an attempt to get to decent precision near UINT64_MAX, but rounding errors with the removed code are so massive that I think we should just refuse the combination of a large unit in combination with a high number of digits.

* relative bytes (*5% of x*) were supported by `vnum.c`, but not used anywhere, so I have removed this feature.

nigoroll:
waiting for @bsdphk feedback please. General feedback is sufficient, no detailed review required

nigoroll:
I've now addressed the integer precision thing by optimizing the int << shift / scale arithmatic

nigoroll:
I have polished the commits a bit more for separation of additions, syntactic change and semantic change

nigoroll:
Changed std.integer also

nigoroll:
conclusion from discussion with @bsdphk and @Dridi : wait for [structured headers](https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-14) to settle

Dridi:
With the march towards RFC8941 number ranges, is this still relevant?

nigoroll:
While I am not entirely happy with the choice to reduce our integer precision to below 64bit, it has always been @bsdphk s argument that if SF settles, varnish should be adjusted accordingly - and this is a relevant argument after all.
So I surrender this PR.

PR #2935, opened 2019-03-08
Multi-instance vtest.sh

lkarsten:
This allows to run multiple instances, for example with different compilers.


bsdphk:
Sorry about the delay in getting to this.

Why not simply run two copies of vtest out of separate directories ?  (That's how I do it)

The choice to not store the sshkey in $HOME/.ssh was deliberate, it is more of an API-key than a ssh-key really.

bsdphk:
Closing this, @lkarsten will reponder when he has a chance.

PR #2939, opened 2019-03-11
Improve VDI_Resolve() FetchErrors

nigoroll:
for after release. Minor issues, except for the fact that we do not emit a FetchError for "administratively unhealthy". Do we want this in the release?


Dridi:
Closing in favor of #2680 that contains both patches, and more.

nigoroll:
ok with closing, the intention was to keep separate issues apart and there is no way (which I am aware of) to stack PRs

PR #2940, opened 2019-03-12
assert that the onearg check works and we implement all cases

nigoroll:
This avoids:

- missing to set nargs correctly
- overlooking to implement all cases
- potential fallthrough from one argument to another

Ref: 6e062834fbf261759b5edcf7c1aa0c505beb87c1

### DIFF
--- a/lib/libvmod_std/vmod_std_conversions.c
+++ b/lib/libvmod_std/vmod_std_conversions.c
@@ -91,11 +91,11 @@ vmod_duration(VRT_CTX, struct VARGS(duration) *a)
 	if (a->valid_integer)
 		return ((VCL_DURATION)a->integer);
 
-	if (a->valid_s) {
-		r = VNUM_duration(a->s);
-		if (!isnan(r))
-			return (r);
-	}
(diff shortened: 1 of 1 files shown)


nigoroll:
(12:33:15) phk: slink, I'm not actually sure I want the onearg() check.  Intuitively it makes much more sense to say that the conversions functions attempt to convert arguments L-R until one succeeds.
(12:34:21) phk: slink, for that to be intuitive, VCC would have to make valid_foo the number of the argument as specified in .VCL
(12:34:56) phk: and I fear that the conversion functions would become very complicated to implement it properly.
(12:35:13) phk: so for now I kept the onearg() check.
(12:35:41) slink: Intuitively I would say that the best effort scanning would complicate debugging
(12:35:48) phk: but I simplified the stylistic convolution of the assert for it.
(12:36:14) phk: I'm not sure, it's something I want to think more about without a release hanging over my head.
(12:36:30) slink: So you see no value in the tighter assertions?
(12:36:50) phk: not a big enough value to make the source style "lumpy"
(12:37:11) slink: so be it...

PR #2941, opened 2019-03-13
Keep error messages from additional VRT_fail() invocations

nigoroll:
this is particularly important for testing failure conditions as in

	vmod.test_error_handling(vmod.failing())

where `vmod.failing()` fails and `vmod.test_error_handling()` is expected to produce an error message for the case that `vmod.failing()` returns a invalid value (such as `NULL`).

Real life example: https://code.uplex.de/uplex-varnish/libvmod-blobdigest/commit/5a057fb51a5a2ea99490293fae3e92c4b015c9a4

fixes regression from 5735a543c720655c484ef11c70b6cde7f201d5ca

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -508,9 +508,7 @@ VRT_fail(VRT_CTX, const char *fmt, ...)
 
 	assert(ctx->vsl != NULL || ctx->msg != NULL);
 	AN(ctx->handling);
-	if (*ctx->handling == VCL_RET_FAIL)
-		return;
-	AZ(*ctx->handling);
+	assert(*ctx->handling == 0 || *ctx->handling == VCL_RET_FAIL);
 	AN(fmt);
 	AZ(strchr(fmt, '\n'));
 	va_start(ap, fmt);


nigoroll:
(09:58:00) phk: slink, It was sort of deliberate to only report the first and triggering Fail
(09:58:22) phk: slink, the expectation was that the possible cascading errors following from that would just be confusing.
(09:58:35) slink: I did see the commit at the time and also found it reasonable, but it simply prevents proper error message checking for some cases
(09:58:56) slink: not sure about the cascade
(09:59:22) slink: For vtc, we could add extra vtc functions just for error simulations which do not fail themselves
(09:59:30) phk: so why not issue a SLT_Debug in that case ?
(09:59:42) phk: and test for that in the .vtc ?
(09:59:43) slink: But this made me think if avoiding the additional errors is helpful after all
(10:00:14) phk: slink, one thing to keep in mind here is "failure inversion"
(10:00:15) slink: SLT_Debug would not test the actual error
(10:00:44) phk: slink, in HPC you should never have an error path be much more expensive than normal path
(10:00:58) phk: slink, because that leads to the "fall over the edge of the cliff" performance collaps
(10:01:12) slink: Yes, but totally unrelated
(10:01:18) phk: No, not at all.
(10:01:20) slink: ?
(10:01:44) phk: If 1% of your transactions start failing for some reason, and suddenly become 10x more expensive to handle, you lost 10% of your performance
(10:01:59) slink: yes, but why are we talking about this?
(10:02:04) phk: VSL bandwidth
(10:02:07) slink: Are you consider recording the additional error messages relevant?
(10:02:20) phk: I said "keep in mind"
(10:02:30) slink: I think I do
(10:03:06) phk: I can see a use for having a debug flag to enable the reports for vtc-purposes, but I have a really hard time imagining them enabled by default
(10:03:44) slink: So: There is a point regarding vtc, and regarding the user, the consequences of the initial error may not be clear, so having that additional information would be helpful I think.
(10:04:12) phk: slink, so one thing we have never exploited is that we know exactly where we are in the VCL
(10:04:46) slink: True. That would be even more helpful
(10:05:38) slink: I need to gsd and would want to get the code into production. Can we resolve #2941 ?
(10:05:39) gofer: #2941: Keep error messages from additional VRT_fail() invocations [open] https://github.com/varnishcache/varnish-cache/issues/2941
(10:06:45) phk: slink, VRT_count gets called a *lot*, but I don't think stashing the integer in the wrk would cost a fortune.
(10:07:02) slink: agree
(10:07:18) phk: we would then need for VRT_Fail to translate the int to something usable
(10:07:45) lucas [lucas@89.197.101.20] entered the room.
(10:07:51) phk: lowest cost is %d, and that may be usable enough, if there is a VCL command that can translate it
(10:08:03) slink: +1
(10:08:47) phk: I guess it would have the be %s::%d, vcl->name, id
(10:09:00) phk: since we cannot rely on being in the active vcl any more
(10:09:49) slink: that or we could log the vcl once per req
(10:09:55) slink: anyway....
(10:10:07) phk: slink, we already log the VCL and it may change
(10:10:41) phk: but I think this could be the place where simply 'varnishlog -g raw -i Fail | myscript' would be preferable
(10:11:15) phk: Anyway, I'm all for improving error reporting, but I don't think unleashing the cascade by default is the way to go.
(10:11:26) phk: But for VTC I can certainly see the value
(10:12:04) slink: I disagree but I am way to much into this discussion again than is good for me
(10:13:12) slink: and yes I had never noticed VCL_use, thank you
(10:14:18) slink: as the vcl is fixed per task I we could spare emitting the vcl name for VRT_fail, but I it would be more convenient to have it

* Adding back testability for our case would require something like `vtc.nullblob()` returning NULL so I just removed the tests for now: https://code.uplex.de/uplex-varnish/libvmod-blobdigest/commit/bb1e30858c2abe268380dca027d8318945d4c38f
* SLT_Debug is not helpful. We want to test the actual error
* I think the error messages are more helpful than confusing

nigoroll:
Another point: We said we would want to replace `vsl*(SLT_Error, ...); VRT_Handling(..., VCL_RET_FAIL);` invocations with `VRT_fail().`

nigoroll:
bugwash consensus: The new behavior is the intended behavior.

PR #2942, opened 2019-03-13
 Govern thread creation by queue length instead of dry signals

mbgrydeland:
When getidleworker signals the pool herder to spawn a thread, it increases the dry counter, and the herder resets dry again when spawning a single thread. This will in many cases only create a single thread even though the herder was signaled dry multiple times, and may cause a situation where the cache acceptor is queued and no new thread is created. Together with long lasting tasks (ie endless pipelines), and all other tasks having higher priority, this will prevent the cache acceptor from being rescheduled.

c00096.vtc demonstrates how this can lock up.

To fix this, spawn threads if we have queued tasks and we are below the thread maximum level.

### DIFF
--- a/bin/varnishd/cache/cache_wrk.c
+++ b/bin/varnishd/cache/cache_wrk.c
@@ -194,13 +194,8 @@ pool_getidleworker(struct pool *pp, enum task_prio prio)
 			AZ(pp->nidle);
 	}
 
-	if (pt == NULL) {
-		if (pp->nthr < cache_param->wthread_max) {
-			pp->dry++;
-			AZ(pthread_cond_signal(&pp->herder_cond));
-		}
+	if (pt == NULL)
 		return (NULL);
-	}
 
 	AZ(pt->func);
 	CAST_OBJ_NOTNULL(wrk, pt->priv, WORKER_MAGIC);
@@ -303,6 +298,7 @@ Pool_Task(struct pool *pp, struct pool_task *task, enum task_prio prio)
 		pp->nqueued++;
 		pp->lqueue++;
 		VTAILQ_INSERT_TAIL(&pp->queues[prio], task, list);
+		AZ(pthread_cond_signal(&pp->herder_cond));
 	} else {
 		if (prio == TASK_QUEUE_REQ)
 			pp->sdropped++;
@@ -473,7 +469,6 @@ pool_breed(struct pool *qp)
 		Lck_Unlock(&pool_mtx);
 		VTIM_sleep(cache_param->wthread_fail_delay);
 	} else {
-		qp->dry = 0;
 		qp->nthr++;
 		Lck_Lock(&pool_mtx);
 		VSC_C_main->threads++;
@@ -552,7 +547,7 @@ pool_herder(void *priv)
 
 		/* Make more threads if needed and allowed */
 		if (pp->nthr < wthread_min ||
-		    (pp->dry && pp->nthr < cache_param->wthread_max)) {
+		    (pp->lqueue > 0 && pp->nthr < cache_param->wthread_max)) {
 			pool_breed(pp);
 			continue;
 		}
@@ -613,16 +608,14 @@ pool_herder(void *priv)
 			continue;
 		}
 		Lck_Lock(&pp->mtx);
-		if (!pp->dry) {
+		if (pp->lqueue == 0) {
 			if (DO_DEBUG(DBG_VTC_MODE))
 				delay = 0.5;
 			(void)Lck_CondWait(&pp->herder_cond, &pp->mtx,
 				VTIM_real() + delay);
-		} else {
+		} else
 			/* XXX: unsafe counters */
 			VSC_C_main->threads_limited++;
-			pp->dry = 0;
-		}
 		Lck_Unlock(&pp->mtx);
 	}
 	return (NULL);
--- a/bin/varnishd/cache/cache_wrk.c
+++ b/bin/varnishd/cache/cache_wrk.c
@@ -462,6 +462,7 @@ pool_breed(struct pool *qp)
 	pi->qp = qp;
 
 	if (pthread_create(&tp, &tp_attr, pool_thread, pi)) {
+		FREE_OBJ(pi);
 		VSL(SLT_Debug, 0, "Create worker thread failed %d %s",
 		    errno, vstrerror(errno));
 		Lck_Lock(&pool_mtx);
--- a/bin/varnishd/cache/cache_pool.h
+++ b/bin/varnishd/cache/cache_pool.h
@@ -48,7 +48,6 @@ struct pool {
 	struct taskhead			idle_queue;
 	struct taskhead			queues[TASK_QUEUE_END];
 	unsigned			nthr;
-	unsigned			dry;
 	unsigned			lqueue;
 	uintmax_t			sdropped;
 	uintmax_t			rdropped;
--- a/bin/varnishtest/tests/c00096.vtc
+++ b/bin/varnishtest/tests/c00096.vtc
@@ -1,0 +1,106 @@
+varnishtest "Test thread creation on acceptor thread queuing"
+
+# This tests that we are able to spawn new threads in the event that the
+# cache acceptor has been queued. It does this by starting 6 long lasting
+# fetches, which will consume 12 threads. That exceeds the initial
+# allotment of 10 threads, giving some probability that the acceptor
+# thread is queued. Then a single quick fetch is done, which should be
+# served since we are well below the maximum number of threads allowed.
+
+# Barrier b1 blocks the slow servers from finishing until the quick fetch
+# is done.
+barrier b1 cond 7
+
+# Barrier b2 blocks the start of the quick fetch until all slow fetches
+# are known to hold captive two threads each.
+barrier b2 cond 7
+
+server s0 {
+	rxreq
+	txresp -nolen -hdr "Content-Length: 10" -hdr "Connection: close"
+	send "123"
+	barrier b1 sync
+	send "4567890"
+	expect_close
+} -dispatch
+
+server stest {
+	rxreq
+	txresp -body "All good"
+} -start
+
+varnish v1 -arg "-p debug=+syncvsl -p debug=+flush_head"
+varnish v1 -arg "-p thread_pools=1 -p thread_pool_min=10"
+varnish v1 -vcl+backend {
+	sub vcl_backend_fetch {
+		if (bereq.url == "/test") {
+			set bereq.backend = stest;
+		} else {
+			set bereq.backend = s0;
+		}
+	}
+} -start
+
+# NB: we might go above 10 threads when early tasks are submitted to
+# the pool since at least one idle thread must be kept in the pool
+# reserve.
+varnish v1 -expect MAIN.threads >= 10
+
+client c1 {
+	txreq -url /1
+	rxresphdrs
+	barrier b2 sync
+	rxrespbody
+} -start
+
+client c2 {
+	txreq -url /2
+	rxresphdrs
+	barrier b2 sync
+	rxrespbody
+} -start
+
+client c3 {
+	txreq -url /3
+	rxresphdrs
+	barrier b2 sync
+	rxrespbody
+} -start
+
+client c4 {
+	txreq -url /4
+	rxresphdrs
+	barrier b2 sync
+	rxrespbody
+} -start
+
+client c5 {
+	txreq -url /5
+	rxresphdrs
+	barrier b2 sync
+	rxrespbody
+} -start
+
+client c6 {
+	txreq -url /6
+	rxresphdrs
+	barrier b2 sync
+	rxrespbody
+} -start
+
+client ctest {
+	barrier b2 sync
+	txreq -url "/test"
+	rxresp
+	expect resp.status == 200
+	expect resp.body == "All good"
+} -run
+
+barrier b1 sync
+
+client c1 -wait
+client c2 -wait
+client c3 -wait
+client c4 -wait
+client c5 -wait
+client c6 -wait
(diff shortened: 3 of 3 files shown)


Dridi:
Your patch's test case is failing on Travis CI.

nigoroll:
besides the test case, this sounds very reasonable to me, in particular because we still have the `wthread_add_delay`: if after the delay there still is a queue, we really should start more threads.

Dridi:
FYI: combining this patch set with @nigoroll's patch from #2796 I can still lock varnish up with `h2load`.

mbgrydeland:
@Dridi What does the relevant varnish counters look like during that test? Do you see threads_limited increase?

mbgrydeland:
Hmm, looking at the Travis log it seems my patch is not doing the trick good enough, it seems one of the thread eating backend fetches was not started properly.

At my local machine I had the test case failing before the patch, and completing fine after. But it is likely that the conditions are not exactly the same. Especially the outset conditions are hard to control. Varnish will start up one acceptor thread for each of the listening interfaces of the host, which makes the number of idle threads at the test case start dependent on the number of interfaces.

mbgrydeland:
Rebased and force-pushed to see if #2986 was the cause of Travis' failures

Dridi:
@mbgrydeland, it may be related to #2796 too. See https://github.com/varnishcache/varnish-cache/pull/2796/commits/6826070d61d25aba16392137b8db1c25e5ac28c2 as the test case mentioned in the commit log exhibits the same behavior without @nigoroll's change, but less frequently.

bin/varnishd/cache/cache_wrk.c
```diff
@@ -194,10 +194,8 @@ pool_getidleworker(struct pool *pp, enum task_prio prio)
 	}
 
 	if (pt == NULL) {
-		if (pp->nthr < cache_param->wthread_max) {
-			pp->dry++;
+		if (pp->nthr < cache_param->wthread_max)
 			AZ(pthread_cond_signal(&pp->herder_cond));
```
nigoroll: I am not sure if waking up the herder makes any sense here without `dry`: We have not yet incremented `lqueue`, which happens in `Pool_Task()`, which is where I would think the signal should be posted.

nigoroll:
@mbgrydeland close now and revisit later?

Dridi:
Looking at the CI problem, I tried to stress c96 locally and ran into a different failure:

```
**   top  === varnish v1 -expect MAIN.threads == 10
---- v1   Not true: MAIN.threads (14) == 10 (10)
```

Dridi:
I was also able to reproduce the CI failure locally. In both cases simply running c96 with `-n100` and `-j32` is enough to eventually run into either.

Dridi:
With this change I'm able to improve the test case readability, and stabilize the timeout problem:

```diff
diff --git a/bin/varnishtest/tests/c00096.vtc b/bin/varnishtest/tests/c00096.vtc
--- a/bin/varnishtest/tests/c00096.vtc
+++ b/bin/varnishtest/tests/c00096.vtc
@@ -29,7 +29,9 @@ server stest {
        txresp -body "All good"
 } -start
 
-varnish v1 -arg "-p debug=+syncvsl" -arg "-p thread_pools=1" -arg "-p thread_pool_min=10" -vcl+backend {
+varnish v1 -arg "-p debug=+syncvsl -p debug=+flush_head"
+varnish v1 -arg "-p thread_pools=1 -p thread_pool_min=10"
+varnish v1 -vcl+backend {
        sub vcl_backend_fetch {
                if (bereq.url == "/test") {
                        set bereq.backend = stest;
```

The addition is the `flush_head` debug bit, needed to make `rxresphdrs` more reliable in `varnishtest`. I will try to understand the expectation failure.

Dridi:
Final diff to pass the test after multiple runs with `-n1000 -j40`:

```diff
diff --git i/bin/varnishtest/tests/c00096.vtc w/bin/varnishtest/tests/c00096.vtc
--- i/bin/varnishtest/tests/c00096.vtc
+++ w/bin/varnishtest/tests/c00096.vtc
@@ -29,7 +29,9 @@ server stest {
        txresp -body "All good"
 } -start
 
-varnish v1 -arg "-p debug=+syncvsl" -arg "-p thread_pools=1" -arg "-p thread_pool_min=10" -vcl+backend {
+varnish v1 -arg "-p debug=+syncvsl -p debug=+flush_head"
+varnish v1 -arg "-p thread_pools=1 -p thread_pool_min=10"
+varnish v1 -vcl+backend {
        sub vcl_backend_fetch {
                if (bereq.url == "/test") {
                        set bereq.backend = stest;
@@ -39,7 +41,10 @@ varnish v1 -arg "-p debug=+syncvsl" -arg "-p thread_pools=1" -arg "-p thread_poo
        }
 } -start
 
-varnish v1 -expect MAIN.threads == 10
+# NB: we might go above 10 threads when early tasks are submitted to
+# the pool since at least one idle thread must be kept in the pool
+# reserve.
+varnish v1 -expect MAIN.threads >= 10
 
 client c1 {
        txreq -url /1
```

What really need here instead of `-expect` is a synchronization point:

```
varnish v1 -await 3s MAIN.threads == 10
```

If you think this would make sense, I can attempt a patch. Otherwise I couldn't find a problem with the non-VTC part of the pull request besides what @nigoroll already mentioned.

bin/varnishd/cache/cache_wrk.c
Dridi: I agree, I also think we don't want to wake up the pool herder in the `Pool_Task_Arg` case.

mbgrydeland: The lock is held all through the call to getidleworker and past the increment of lqueue, so the herder would not be able to run until after lqueue had been increment and would see it. It is a point for the case of having reached the queue limit, and would avoid a spurious herder wakeup in those cases. And it makes the code easier to read, so I agree with that change.

mbgrydeland:
Thank you @Dridi and @nigoroll  for looking after this.

Dridi:
With @mbgrydeland we agreed that I would update the pull request with the suggested diffs. This was also rebased against current master.

Dridi:
Following up my offline discussion regarding `-expect` vs `-await` with @mbgrydeland: there is currently a 5-ish seconds (50 times 100ms) grace period for the counter to show up, not to match the expected value. So we currently don't have the `-await` behavior I'm suggesting. I will open a separate pull request for that.

Dridi:
Actually no, @mbgrydeland was right, so I'm wondering what may take up to 21 threads even after 5 seconds.

Dridi:
I looked a bit closer and realized that at this point only one task is really added tot the pool but there is a window between the moment a thread is created and the moment it goes idle. If that single acceptor task is added during that window, we may create \*dozens\* of threads beyond `thread_pool_min` because until we have two idle threads (one in the reserve with this configuration, and one not subject to the reserve constraint) we might keep creating more threads.

So this fixes a bug (try the test case on the master branch if you have any doubts) but at the same time introduces a creation race.

With this diff, the test case is very stable:

```diff
diff --git i/bin/varnishtest/tests/c00096.vtc w/bin/varnishtest/tests/c00096.vtc
--- i/bin/varnishtest/tests/c00096.vtc
+++ w/bin/varnishtest/tests/c00096.vtc
@@ -31,6 +31,7 @@ server stest {
 
 varnish v1 -arg "-p debug=+syncvsl -p debug=+flush_head"
 varnish v1 -arg "-p thread_pools=1 -p thread_pool_min=10"
+varnish v1 -arg "-p thread_pool_add_delay=0.01"
 varnish v1 -vcl+backend {
        sub vcl_backend_fetch {
                if (bereq.url == "/test") {
@@ -41,10 +42,7 @@ varnish v1 -vcl+backend {
        }
 } -start
 
-# NB: we might go above 10 threads when early tasks are submitted to
-# the pool since at least one idle thread must be kept in the pool
-# reserve.
-varnish v1 -expect MAIN.threads >= 10
+varnish v1 -expect MAIN.threads == 10
 
 client c1 {
        txreq -url /1
```

I think the bug fix might be incomplete, but I'm not sure how to close the race.

Dridi:
More digging: I thought the same race would exist with dry signals considering that this patch is not ground-breaking. Looking closer at the code we have the same kind of logic where we increment dry but never decrement it. Instead we wipe it, which may lead to the bug @mbgrydeland fixed.

So the same race exists, and I was able to exhibit it:

```vcl
varnishtest "over-breeding"

varnish v1 -arg "-p debug=+syncvsl -p debug=+flush_head"
varnish v1 -arg "-p thread_pools=1 -p thread_pool_min=10"
varnish v1 -vcl {
        backend be none;
} -start

varnish v1 -expect MAIN.threads == 10
```

If you stress this VTC enough, the last line may fail because 11 threads are accounted for. Relying on the queue length instead opens the gates and increases the opportunity for over-breeding, and in some cases I observed 50 threads created for one acceptor task.

My suggestion is that in addition to governing thread creation by queue length, new threads should start as idle, and go to sleep if the queue was empty. Alternatively we can dismiss over-breeding as a problem since we technically always stay within min/max bounds.

Dridi:
I think there is more to it.

I tried to make the `(struct pool).nidle` counter eventually consistent: it increments before the thread is created and decremented if that failed. Once a thread goes through its first task lookup, it adjusts `nidle` accordingly (but only for the very first iteration). With that I lowered the probability of over-breeding by an order of magnitude, but it still happened.

Looking closer, I believe we need to always guard accesses to `nidle`, `nthr` and `lqueue`, which is not always the case, so a data race exists.

Dridi:
I still would like to give @mbgrydeland an opportunity to have a look before merging if that's fine with you.

nigoroll:
discussed with @Dridi : we should add `sched_yield()` for `wthread_add_delay == 0`

Dridi:
On my system `sched_yield()` doesn't seem to bring any improvement.

nigoroll:
I see different figures, see #3107 
This is still opportunistic and the primary goal is not to prevent overbreeding, I just noticed that, especially on small systems, we might increase the responsiveness by yielding from the herder

Dridi:
To clarify, @nigoroll and I weren't looking at the same metric. It so appears that yielding reduces the probability to over-breed, but when I see it happen, I still easily see dozens of needless threads creation.

Dridi:
I discussed this with @mbgrydeland and he will eventually revisit this pull request.

The plan is to do the following:

- check the locking correctness of the thread pools
- look at the over-breeding bug
- revisit the under-breeding fix if any of the above resulted in a change

nigoroll:
@mbgrydeland if this will be reworked, could we close the PR in the meantime?

mbgrydeland:
I had a quick look at the overbreeding effect that this patch introduces. That is the result of the herder not taking into account the threads it has already created, and unless these new threads have already dequeued tasks by the next loop iteration of the herder, it will continue to create more.

One way to fix this race would be for the herder to dequeue tasks as part of the thread creation, and handing these tasks to the newly created threads as the first task, before becoming a pool worker once that task is finished. That way the lqueue would be decremented by the herder and it would have the complete right picture when making the decision to create another thread. Though this solution will require some refactoring and I don't really like it.

A simpler approach would be for the herder to keep track of how many newly created threads there are, that have not yet reached the first task dequeue point in Pool_Work_Thread(). So the herder does newborn++ in pool_breed, and Pool_Work_Thread does newborn-- once and only once. Then the breed condition would change from 'lqueue > 0' to 'lqueue - newborn > 0'. Some locking changes would also be needed for the tests of lqueue and newborn, instead of testing them without like the current patch does.

Dridi:
This is similar to my experiment, except that I went for `lqueue > newborn` and repurposed `nidle` instead of introducing a new field. Either way, you also reached the same conclusions with regards to locking.

How do we proceed? And who does the implementation? (@mbgrydeland, please tell my you are planning to do this=)

nigoroll:
This change fixes a serious underbreeding problem (which, in fact, I ran into just yesterday).
My take it that we should fix this and not care about the occasional overbreeding, because fixing it would add more locking overhead.

So, back to first principles, why is overbreeding a problem?

mbgrydeland:
> So, back to first principles, why is overbreeding a problem?

Not a problem per se, but something that is more likely to happen when the system is busy than when not, which would make the system even more busy and even more likely.

Dridi:
> So, back to first principles, why is overbreeding a problem?

One reason is also that we should avoid creating spurious VSC spikes that may result in puzzling graphs or misleading alerts from monitoring systems.

nigoroll:
I see your points, but in this case I really lean towards avoiding the additional complications, and, in particular, the additional `pp->mtx` ops.

If there are no other objections, I will go with this. But my personal point of view is that avoiding this rather exceptional situation, mostly seen with artificial testing, should not be handled with additional overhead.
Regarding the argument that creating even more threads in an overload situation was a bad idea: This is exactly what I had in mind when adding the `sched_yield()`: If the machine _is_ actually saturated, it will simply not get back to the herder in the first place. On the contrary, if it does, this means we do have CPUs available to create more threads.

nigoroll:
@mbgrydeland @Dridi To make progress here, because, as mentioned, this PR is important: Can we maybe slit out the overbreed-avoidance and agree on the patch as without the last commit first?

mbgrydeland:
I haven't been involved in all the discussions on this and maybe I missunderstood the previous comments, but I thought sched_yield() was found to not solve the overbreeding problem? If it does then I have no problem using that instead of the proposed patch.

On the pp->mtx concern, I do not think that is a concern, with one more lock/unlock per MIN(thread_add_delay, thread_timeout).

nigoroll:
@mbgrydeland the `sched_yield()` does improve on it, but not _solve_ it as far as @Dridi s test is concerned.

my concern regarding `pp->mtx` is that it is central to all scheduling functions, so we should simply avoid any additional load on it - or, in other words, have a very good reason to increase its use. In particular, we need  `pp->mtx` to have our newly created threads begin useful work, so grabbing it one more time in the herder will increase chances of further delays.
Note that `wthread_add_delay` defaults to 0

Also please rebase on master to see the `sched_yield()` in your branch

mbgrydeland:
Force update rebase on master and removed overbreeding patch

mbgrydeland:
I think this PR is good to go at this point. Marking for bugwash for that purpose.

PR #2947, opened 2019-03-14
kill stale oc for backend synth

### LINKED ISSUE #2946: "backend synth" does not kill stale_oc
Noticed while tracking down #2945 - unclear if related

The following vcl with a bad backend shows `n_objcore` increasing with each bgfetch:

```
sub vcl_backend_error {
	set beresp.status = 200;
	set beresp.ttl = 1s;
	set beresp.grace = 1h;
}
```

I am not certain which behavior is to be expected:

* For a backend error, do we want to keep the previous object in case the "backend synth" object expires?
* Or do we want to handle the "backend synth" object like any other object and kill the stale one?

nigoroll:
Fixes #2946 in the sense that we want to treat backend synth like any
other object and kill the stale object it replaces.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -776,6 +776,8 @@ vbf_stp_error(struct worker *wrk, struct busyobj *bo)
 	VSB_destroy(&synth_body);
 	HSH_Unbusy(wrk, bo->fetch_objcore);
 	ObjSetState(wrk, bo->fetch_objcore, BOS_FINISHED);
+	if (bo->stale_oc != NULL)
+		HSH_Kill(bo->stale_oc);
 	return (F_STP_DONE);
 }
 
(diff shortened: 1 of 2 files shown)


bsdphk:
I think this is a false positive.

At the bottom of vbf_fetch_thread() the stale_oc is deref'ed, so it should die on its own whenever its keep runs out.

If I modify the testcase as follows:

        varnishtest "#2946 - objcore leak for backend_synth"

        varnish v1 -vcl {
         backend bad { .host = "${bad_backend}"; }

         sub vcl_backend_error {
          set beresp.status = 200;
          set beresp.ttl = 0.0001s;
          set beresp.grace = 5s;
         }
        } -start

        client c1 -repeat 20 -keepalive {
         txreq
         rxresp
         expect resp.status == 200

         delay 0.001
        } -run

        delay 5

        varnish v1 -vsc MAIN.n_*

        varnish v1 -expect n_objectcore == 0

It passes

PR #2948, opened 2019-03-14
Mention in the etc/default.vcl that vcl >= 4.1; is required with UDS

dmatetelki:
With UDS, VCL needs to be 4.1


Dridi:
Nitpick for the commit message, it's `etc/example.vcl`, otherwise LGTM.

rezan:
Why not just make the default 4.1?

dmatetelki:
@Dridi commit msg changed to `etc/example.vcl`.

bsdphk:
Make it 4.1 syntax.

PR #2950, opened 2019-03-17
respect max_restarts and fix c00028.vtc

nigoroll:
Fixes #2949

PR #2951, opened 2019-03-17
backend synth vs. stale oc take 2

### LINKED ISSUE #2946: "backend synth" does not kill stale_oc

nigoroll:
includes #2950 as a prerequisite

I have now made up my mind about #2946 and would fix/clarify as follows:

Fixes #2946 in the sense that we want to treat a cacheable backend synth like any other object and kill the stale object it replaces.
  
We do not replace the stale oc when
* abandoning the bereq
* leaving vcl_backend_error with return (deliver) and beresp.ttl == 0s
* there is a waitinglist on this object because in this case the default ttl would be 1s, so we might be looking at the same case as the previous

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -683,6 +683,7 @@ vbf_stp_error(struct worker *wrk, struct busyobj *bo)
 	vtim_real now;
 	uint8_t *ptr;
 	struct vsb *synth_body;
+	struct objcore *stale = NULL;
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(bo, BUSYOBJ_MAGIC);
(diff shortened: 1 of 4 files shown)


nigoroll:
edit: I changed my mind. My previous suggestion change change `builtin.vcl` `vcl_backend_error {}` was too invasive because, for existing vcl, the client side processing would change from `vcl_deliver {}` to `vcl_synth {}`.
Due to the ttl check, the patch works also for an unchanged `builtin.vcl`.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -720,6 +720,7 @@ vbf_stp_error(struct worker *wrk, struct busyobj *bo)
 		bo->fetch_objcore->ttl = 1;
 		bo->fetch_objcore->grace = 5;
 		bo->fetch_objcore->keep = 5;
+		bo->stale_oc = NULL;
```
mbgrydeland: I believe this leaks the stale_oc reference that is held at this point.

nigoroll:
Bugwash:
* check martin's comment
* add comment that we might need VCL control later

bin/varnishd/cache/cache_fetch.c
nigoroll: yes, that was a stupid oversight from my side, thank you @mbgrydeland

PR #2956, opened 2019-03-22
make VRT_fail() work with any initializier

nigoroll:
Historically, all initializers called from VGC_Load() would handle
failures individually (return (1) from VGC_Load) or not fail at
all. Now that we got VRT_fail(), it makes sense to generalize.

The upcoming use case is to properly handle failures from
VRT_new_backend_clustered().

### DIFF
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -311,6 +311,10 @@ EmitInitFini(const struct vcc *tl)
 		assert(p->n > 0);
 		if (VSB_len(p->ini))
 			Fc(tl, 0, "\t/* %u */\n%s\n", p->n, VSB_data(p->ini));
+		if (p->ignore_errors == 0) {
+			Fc(tl, 0, "\tif (*ctx->handling == VCL_RET_FAIL)\n");
+			Fc(tl, 0, "\t\treturn(1);\n");
+		}
 		Fc(tl, 0, "\tvgc_inistep = %u;\n\n", p->n);
 		VSB_destroy(&p->ini);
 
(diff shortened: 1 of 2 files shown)


Dridi:
+1 for this, but it's currently reliably breaking m22.vtc so I suspect some `logexpect` needs to be adjusted.

nigoroll:
@Dridi thank you and sorry for my oversight, I've quickly cherry-picked this from a longer series of commits and missed to test properly.

So the question behind this is: If vcl_init fails, do we want to run vcl_fini ? Right now we do, but is this what we want?

Dridi:
I have seen one VMOD in the wild that makes use of `vcl_finì` instead of tearing things down when the discard event is sent so I suppose we might.

nigoroll:
@Dridi thank you for the lightning fast answer. Someone who works with varnish code regularly would have been aware of this comment ;) https://github.com/varnishcache/varnish-cache/blob/master/lib/libvcc/vcc_compile.c#L667
I'll try to come up with a solution...

nigoroll:
updated...

bsdphk:
Do we need to cover this in a vtc ?

nigoroll:
@bsdphk for this to be vtc-testeable, we would need a fail-path for some of the ini-steps like `VRT_new_backend_clustered` or `VRT_re_init` which are not detected by VCC already.
My motivation for this PR is https://github.com/varnishcache/varnish-cache/pull/2957/files#diff-8dfab2c266cb7e7270373622c65bcd86R604 and I would rather add a VTC when we get this in.

nigoroll:
> (13:27:06) phk: go on the patch, but please change "errignore" to "ignore_errors"

PR #2957, opened 2019-03-22
via backends continued

nigoroll:
#2850 continued:
* General cleanup
* Split out features into individual commits
* Fix: Send preamble only for new connections
* Add Basic VCC Support for via

There are some obvious next steps (mostly from feedback on #2850 - see bottom of this description), but I think the patch is viable as-is now.

Prerequisite PRs have been merged explicitly (with `--no-ff`):
* #2845 for generic proxy header formatting
* #2680 for resolving a director
* #2956 make VRT_fail() work with any initializier for `VRT_fail()` handling from `VRT_new_backend_clustered()`

Overview info from 5fdb1a20089520515cb8725e51e4e7915cc52925:

----

So the obvious way to connect to TLS backends is to use a TLS "onloader" (a term coined by @slimhazard as in the opposite of "offloader"), which turns a clear connection into a TLS connection to some other fixed config address.

haproxy, however, offers a convenient and elegant way to avoid this configuration overhead: The PROXY protocol can also be used to transport the destination address which haproxy is to connect to if a server's address is unspecified (`IN_ADDR_ANY` / `0.0.0.0`). The configuration template for this use case looks like this (huge thank you to @wtarreau for pointing out this great option in haproxy):

```
    listen clear-to-ssl
            bind /my/path/to/ssl_onloader accept-proxy
            balance roundrobin
            stick-table type ip size 100
            stick on dst
            server s0 0.0.0.0:443 ssl ca-file /etc/ssl/certs/ca-bundle.crt
            server s1 0.0.0.0:443 ssl ca-file /etc/ssl/certs/ca-bundle.crt
            server s2 0.0.0.0:443 ssl ca-file /etc/ssl/certs/ca-bundle.crt
            # .. approximately as many servers as expected peers
            # for improved tls session caching
```

With this setup, by connecting to `/my/path/to/ssl_onloader` and sending the address to make a TLS connection to in a PROXY header (as the server address / port), we can reduce the configuration overhead outside varnish substantially. In particular, we do not require a path / port per destination and get back the option to use dynamic backends (see also https://github.com/nigoroll/libvmod-dynamic/issues/34 )

This patch implements the basis for simple means of configuring such an ssl onloader:
    
backends can be created with an additional "via" director, which has to resolve to a simple backend. The connection is then made to that address and the actual endpoint address is sent in an additional PROXY header.  Notice that sending yet another proxy header to the actual backend is unaffected. Despite using the same format, the two proxy headers are semantically different: The first, here coined the "preamble", is the address to make the connection to while the (optional) second proxy header continues to contain the addresses of the connection to varnish.

Obvious next steps:

* resolution of the "via" backend at the time the connection is made: This will allow for fault tolerance and load balancing of via backends

* Cascade the health check: If the "via" backend is probed / set down, any backends using it should be down immediately

* Timeouts: The "via" backend's timeouts should define maximum values for any connections made through it


bin/varnishd/cache/cache_backend.c
```diff
@@ -652,9 +652,8 @@ VRT_new_backend_clustered(VRT_CTX, struct vsmw_cluster *vc,
 		AN(sa);
 		AN(bogo);
 		be->tcp_pool = VTP_Clone(viabe->tcp_pool, sa);
```
nigoroll: this needs to be changed for autority

nigoroll:
FTR from VDD:
* @bsdphk wants to change the VRT backend interface to https://etherpad.wikimedia.org/p/vdd19q3 `VRT_new_backend()`
* this PR needs to be adjusted once that is in
* also the existing tcp pool id code is from for now, but that will be fixed with the VRT change

PR #2965, opened 2019-04-04
Include vnum.h in public headers

steven-j-wojcik:
I ran into a case with an out of tree VMOD where I wanted to convert a string to a number. I wound up including `vnum.h` by hand to convert the string but I thought including `vnum.h` in the public headers would be a useful addition.

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -67,6 +67,7 @@ nobase_pkginclude_HEADERS += \
 	vcs.h \
 	verrno.h \
 	vmod_abi.h \
+	vnum.h \
 	vqueue.h \
 	vre.h \
 	vdef.h \
@@ -97,7 +98,6 @@ nobase_noinst_HEADERS = \
 	vjsn.h \
 	vlu.h \
 	vmb.h \
-	vnum.h \
 	vpf.h \
 	vsc_priv.h \
 	vsl_priv.h \


nigoroll:
+1 in general, I would just appreciate that we could finish #2929 first as making `vnum.h` public is a stronger commitment towards stability.

Dridi:
> 13:18 < phk> and if we want to do that, are the API/functions sensible ?
13:19 < dridi> i'm still not happy with the duration parsing supporting spaces between the number and the unit

Now it is :)

bsdphk:
Consensus: Vnum.h makes sense as API.  Some spit & polish needed before then. (And tickets like #2929 fixed)

nigoroll:
suggestion for names:
* VNUMpfx -> VNUMdouble
* VNUMpfxint from #2929 -> VNUMint
* VNUM_2bytes -> VNUMbytes

bsdphk:
I'm trying slowly to converge on a more systematic naming, so I would prefer something like:

    VNUM_Double
    VNUM_Int
    VNUM_Bytes

nigoroll:
@bsdphk +1

nigoroll:
FTR, I would ask to tick off #2929 first and then get back to this one.
@steven-j-wojcik can we close until then?

steven-j-wojcik:
@nigoroll no issues with me .

Closing as #2929 should be settled first.

PR #2966, opened 2019-04-06
Add information about vcl object instances to the panic output

nigoroll:
In the absence of a core dump, we do not have any information yet in the
panic output about vcl object instances, for example to find out which
object a priv belongs to when the instance address is used for
per-instance priv state.

To make this information available at the time of a panic, we add the
following:

* A struct vrt_ii (for instance info), of which a static gets
  filled in by VCC to contain the pointers to the C global variable
  instance pointers at VCC time

* A pointer to this struct from the VCL_conf to make it available to
  the varnishd worker

* dumps of the instance info for panics

### DIFF
--- a/lib/libvcc/vcc_xref.c
+++ b/lib/libvcc/vcc_xref.c
@@ -318,6 +318,29 @@ vcc_xreftable_len(struct vcc *tl, const struct symbol *sym)
 }
 
 static void v_matchproto_(symwalk_f)
+vcc_instance_info(struct vcc *tl, const struct symbol *sym)
+{
+
+	CHECK_OBJ_NOTNULL(sym, SYMBOL_MAGIC);
+	CHECK_OBJ_NOTNULL(sym->kind, KIND_MAGIC);
+	if (sym->kind != SYM_INSTANCE)
+		return;
+	AN(sym->rname);
+	Fc(tl, 0, "\t{ .p = &%s, .name = \"", sym->rname);
+	VCC_SymName(tl->fc, sym);
+	Fc(tl, 0, "\" },\n");
+}
+
+void
+VCC_InstanceInfo(struct vcc *tl)
+{
+	Fc(tl, 0, "\nconst struct vpi_ii VGC_instance_info[] = {\n");
+	VCC_WalkSymbols(tl, vcc_instance_info, SYM_NONE);
+	Fc(tl, 0, "\t{ .p = NULL, .name = \"\" }\n");
+	Fc(tl, 0, "};\n");
+}
+
+static void v_matchproto_(symwalk_f)
 vcc_xreftable(struct vcc *tl, const struct symbol *sym)
 {
 
--- a/lib/libvcc/vcc_xref.c
+++ b/lib/libvcc/vcc_xref.c
@@ -322,9 +322,6 @@ vcc_instance_info(struct vcc *tl, const struct symbol *sym)
 {
 
 	CHECK_OBJ_NOTNULL(sym, SYMBOL_MAGIC);
-	CHECK_OBJ_NOTNULL(sym->kind, KIND_MAGIC);
-	if (sym->kind != SYM_INSTANCE)
-		return;
 	AN(sym->rname);
 	Fc(tl, 0, "\t{ .p = &%s, .name = \"", sym->rname);
 	VCC_SymName(tl->fc, sym);
@@ -335,7 +332,7 @@ void
 VCC_InstanceInfo(struct vcc *tl)
 {
 	Fc(tl, 0, "\nconst struct vpi_ii VGC_instance_info[] = {\n");
-	VCC_WalkSymbols(tl, vcc_instance_info, SYM_NONE);
+	VCC_WalkSymbols(tl, vcc_instance_info, SYM_INSTANCE);
 	Fc(tl, 0, "\t{ .p = NULL, .name = \"\" }\n");
 	Fc(tl, 0, "};\n");
 }
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -44,6 +44,7 @@
 #include "cache_vcl.h"
 #include "vcli_serve.h"
 #include "vtim.h"
+#include "vcc_interface.h"
 
 const struct vcltemp VCL_TEMP_INIT[1] = {{ .name = "init", .is_cold = 1 }};
 const struct vcltemp VCL_TEMP_COLD[1] = {{ .name = "cold", .is_cold = 1 }};
@@ -208,6 +209,7 @@ vcl_find(const char *name)
 void
 VCL_Panic(struct vsb *vsb, const char *nm, const struct vcl *vcl)
 {
+	const struct vpi_ii *ii;
 	int i;
 
 	AN(vsb);
@@ -232,6 +234,15 @@ VCL_Panic(struct vsb *vsb, const char *nm, const struct vcl *vcl)
 		VSB_indent(vsb, 2);
 		for (i = 0; i < vcl->conf->nsrc; ++i)
 			VSB_printf(vsb, "\"%s\",\n", vcl->conf->srcname[i]);
+		VSB_indent(vsb, -2);
+		VSB_cat(vsb, "instances = {\n");
+		VSB_indent(vsb, 2);
+		ii = vcl->conf->instance_info;
+		while (ii != NULL && ii->p != NULL) {
+			VSB_printf(vsb, "\"%s\" = %p,\n", ii->name,
+			    (const void *)*(const uintptr_t *)ii->p);
+			ii++;
+		}
 		VSB_indent(vsb, -2);
 		VSB_cat(vsb, "},\n");
 	}
--- a/bin/varnishtest/tests/v00010.vtc
+++ b/bin/varnishtest/tests/v00010.vtc
@@ -26,6 +26,12 @@ server s1 {
 
 varnish v1 -arg "-sdefault,1m" -vcl+backend {
 	import vtc;
+	import debug;
+
+	sub vcl_init {
+		new foo = debug.obj("foo");
+		new bar = debug.concat("bar");
+	}
 
 	sub vcl_backend_response {
 		if (beresp.http.panic == "fetch") {
--- a/include/vcc_interface.h
+++ b/include/vcc_interface.h
@@ -64,5 +64,11 @@ struct vrt_acl {
 	const char	*name;
 };
 
+/* vmod object instance info */
+struct vpi_ii {
+	const void *			p;
+	const char * const		name;
+};
+
 VCL_STRANDS VPI_BundleStrands(int, struct strands *, char const **,
     const char *f, ...);
--- a/lib/libvcc/generate.py
+++ b/lib/libvcc/generate.py
@@ -656,6 +656,7 @@ struct VCL_conf {
 	const char		**srcbody;
 
 	int			nvmod;
+	const struct vpi_ii	*instance_info;
 
 	vcl_event_f		*event_vcl;
 """)
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -488,6 +488,7 @@ EmitStruct(const struct vcc *tl)
 #define VCL_MET_MAC(l,u,t,b) \
 	Fc(tl, 0, "\t." #l "_func = VGC_function_vcl_" #l ",\n");
 #include "tbl/vcl_returns.h"
+	Fc(tl, 0, "\t.instance_info = VGC_instance_info\n");
 	Fc(tl, 0, "};\n");
 }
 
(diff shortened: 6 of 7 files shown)


bsdphk:
Yes in principle, but possibly not this implementation.

So I think we need to grow VCL in a more OO direction, and this is something I'm spending a lot of scratch paper on right now. (Many drivers for this: H3, SH etc.)

I think the overall scheme will probably be something like:

    struct vcl_instance { // the struct formerly known as vmod_priv
          unsigned magic etc.
          void *priv;
          size_t len;
          struct vcl_class *methods;
    };
    
    struct vcl_methods {
        unsigned magic etc.
        const struct vcl_priv_methods *priv_methods;
        const int n_priv_methods;
        /* Global methods */
        vcl_method_panic(struct vcl_instance, struct vsb *);
        ...
    };

The priv_methods may actually not live in .[ch] land but in .vcc land, not sure yet.

I'm not quite ready to pull the trigger on this yet, so can we hold of until we have talked more about the OO thing ?

nigoroll:
@bsdphk as the intention is solely to add a debugging aid, we sure can postpone this.
But as we are not committing to any API nor create any internal dependencies, I do not see why we should not adjust (or, as a last resort even remove) something like this when we change the OO model.

WRT your OO scheme, this looks perfectly plausible to me. I might have some items for the wishlist, but I do not want to pile up even more on your list.

Dridi:
It looks like some of the object orientation recently landed, and this is merely about adding information to the panic output. Why not move `struct vrt_ii` to `VPI` instead so we don't commit to any kind of API? This way we can move forward with this pull request and have complete liberty to change the implementation details along with the remaining OO changes?

nigoroll:
For now, I have updated the PR to work with the recent changes, but yes, as the VCL OO interface evolves, this will need more attention.
Moving to VPI sounds like a good idea irrespective of this, yes.

lib/libvcc/vcc_xref.c
```diff
@@ -326,6 +326,29 @@ vcc_CheckUses(struct vcc *tl)
...
 /*---------------------------------------------------------------------*/
 
+static void v_matchproto_(symwalk_f)
+vcc_instance_info(struct vcc *tl, const struct symbol *sym)
+{
+
+	CHECK_OBJ_NOTNULL(sym, SYMBOL_MAGIC);
+	CHECK_OBJ_NOTNULL(sym->kind, KIND_MAGIC);
+	CHECK_OBJ_NOTNULL(sym->type, TYPE_MAGIC);
+	if (sym->rname == NULL || sym->kind != SYM_INSTANCE)
```
Dridi: Can a `SYM_INSTANCE` not have an `rname`?

nigoroll: I guess you might have misread the code. We do only add to the instance info `SYM_INSTANCE` _with_ an `rname`.

nigoroll:
@Dridi my apologies, I had said before that I would move the struct to VPI but did not. That is done now.

lib/libvcc/vcc_xref.c
Dridi: Let me rephrase, in which case would a `SYM_INSTANCE` not have an `rname`?

Dridi: I meant that my understanding of the symbol table for this case is this:

```diff
--- lib/libvcc/vcc_xref.c
+++ lib/libvcc/vcc_xref.c
@@ -333,8 +333,9 @@ vcc_instance_info(struct vcc *tl, const struct symbol *sym)
        CHECK_OBJ_NOTNULL(sym, SYMBOL_MAGIC);
        CHECK_OBJ_NOTNULL(sym->kind, KIND_MAGIC);
        CHECK_OBJ_NOTNULL(sym->type, TYPE_MAGIC);
-       if (sym->rname == NULL || sym->kind != SYM_INSTANCE)
+       if (sym->kind != SYM_INSTANCE)
                return;
+       AN(sym->rname);
        Fc(tl, 1, "\t{ .p = &%s, .name = \"", sym->rname);
        VCC_SymName(tl->fc, sym);
        Fc(tl, 1, "\" },\n");
```

nigoroll: Do we want such assertions for code which only generates linkage for debugging?

nigoroll: anyway, nevermind. I will just add your suggestion, it's not worth it to discuss

Dridi: Also to be consistent we shouldn't auto-indent the generated code:

```diff
--- lib/libvcc/vcc_xref.c
+++ lib/libvcc/vcc_xref.c
@@ -332,12 +332,12 @@ vcc_instance_info(struct vcc *tl, const struct symbol *sym)
 
        CHECK_OBJ_NOTNULL(sym, SYMBOL_MAGIC);
        CHECK_OBJ_NOTNULL(sym->kind, KIND_MAGIC);
-       CHECK_OBJ_NOTNULL(sym->type, TYPE_MAGIC);
-       if (sym->rname == NULL || sym->kind != SYM_INSTANCE)
+       if (sym->kind != SYM_INSTANCE)
                return;
+       AN(sym->rname);
-       Fc(tl, 1, "\t{ .p = &%s, .name = \"", sym->rname);
+       Fc(tl, 0, "\t{ .p = &%s, .name = \"", sym->rname);
        VCC_SymName(tl->fc, sym);
-       Fc(tl, 1, "\" },\n");
+       Fc(tl, 0, "\" },\n");
 }
 
 void
```

This doesn't change the output, but it's consistent with how the end-of-ii marker is generated.

Dridi:
@nigoroll I didn't even remember that we had the VPI discussion before you brought it up, and it's only a few comments above :)

nigoroll:
> @nigoroll I didn't even remember that we had the VPI discussion before you brought it up, and it's only a few comments above :)

You had already mentioned it in June https://github.com/varnishcache/varnish-cache/pull/2966#issuecomment-499808296

Dridi:
Since this is meant for the panic output, would it make sense to add a "NULL_OK" flag in the struct too?

nigoroll:
@Dridi yes this might make sense. But let's get the basics ticked off first please

bsdphk:
Isn't this implementation a dead end if we want to instantiate objects other places than in vcl_init ?

At one point we talked about having panics call into the applicable VCL's VMODs, for, more or less, this kind of reporting, isn't that a more solid approach ?

Another related idea:  Give VRT_CTX a `const struct vpi_ref*` member, have VCC maintain it and add a suitable VRT function to report it.

That way the VMOD could say not only which objects, but also give details about them and tell where it came into existence.

nigoroll:
Re @bsdphk 

> Isn't this implementation a dead end if we want to instantiate objects other places than in vcl_init ?

It depends on how exactly we are going to do it. The way I see it from the current perspective, this implementation would still work for per-vcl objects and would need to be complemented for per-task objects.

> At one point we talked about having panics call into the applicable VCL's VMODs, for, more or less, this kind of reporting, isn't that a more solid approach ?

I am not saying this implementation is the last word. It is just indefinitely more than what we have now (as that is zero).
Also, this implementation has no API impact and we are free to remove or improve it any time.

Dridi:
> Isn't this implementation a dead end if we want to instantiate objects other places than in vcl_init ?

I'm not worried about dead-ends when it comes to VPI, because we can always adjust whenever this becomes unfit for new changes.

> That way the VMOD could say not only which objects, but also give details about them and tell where it came into existence.

Now that worries me because the panic code really needs to be failsafe, something I wouldn't entrust to third-party VMODs, and this would bind us to a public interface that we'll need to think twice about once we need to make changes.

nigoroll:
> panic code really needs to be failsafe
+1

bsdphk:
> Now that worries me because the panic code really needs to be failsafe, something I wouldn't entrust to third-party VMODs, and this would bind us to a public interface that we'll need to think twice about once we need to make changes.

This is actually less of an issue than you may think.

The panic VSB goes into memory mapped shared with the management process, so as long as we call the vmods's panic hooks last, all the normal panic will be safe.

Dridi:
I'm still worried that a third-party VMOD might deadlock when it tries to access object information and lock the child up instead of swiftly panicking.

And while this is something we could expose in a very well documented "here be dragons" form like a `$Panic` stanza in VCC files I see no reason to delay this pull request since at the very least getting a pointer for each object symbol in the panic output would make things easier to audit a core dump. And this being hidden behind VPI, we wouldn't have any form of ABI commitment.

Dridi:
I played a bit with this and out of the box getting the symbols from a panic is not trivial. The symbols are defined by the VCL's `vgc.so` so I should be able to access `vo_bar` to get the pointer to the VCL `bar` object's pointer. However, the way we handle DSOs makes it non available out of the box, probably because we load them using a relative path.

In particular, getting the pointer alone offers little value in a debugger because unless we have debug information from the DSOs (vmod and vgc) we get an opaque pointer with little value.

I'll check whether using an absolute path makes this trivial, in which case I would move back on the fence and see little benefit to this pull request in its current form.

nigoroll:
@Dridi as explained in the PR description, my use case was to find out the object for a `PRIV_TASK`

Debugging when you have a varnish working directory is easy (just cd into it with `gdb`), so I am not exactly sure what you are referring to in the first paragraph.
This PR specifically is for the case when you only have the panic string to correlate the PRIV info to objects.

Dridi:
I will need to dig further but this is registering `SYM_INSTANCE` symbols for the panic output, I missed the part where PRIVs play in. The only interesting thing I see is getting pointers for the failing VCL, in case you have more than one VCL with the same symbols (which would be rather common across VCL reloads).

nigoroll:
> The only interesting thing ...

I wonder why I fail to get the basics across: This is not a perfect debugging aid, it is just something to maybe correlate some pointers in the panic string.

Really, why do we spend so much time on such a trivial thing which may or may not provide very helpful information when trying to make sense of a panic string?

Dridi:
Because we both want this to be useful? I submitted #3161 to improve the debugging aid a bit further. I believe we are better off with both pull requests merged.

Please note that I never withdrew my approval, although I said I could end up neutral here (on the fence). I have no intention in vetoing this change, I said it had little value to what I initially perceived. So I researched this further and found a way to further improve the debugging experience but this falls short for the reasons explained (and probably a lack of debugging proficiency on my end). This PR turns out to still be very helpful in the process.

lib/libvcc/vcc_xref.c
```diff
@@ -317,6 +317,29 @@ vcc_xreftable_len(struct vcc *tl, const struct symbol *sym)
...
+	Fc(tl, 0, "\t{ .p = &%s, .name = \"", sym->rname);
+	VCC_SymName(tl->fc, sym);
+	Fc(tl, 0, "\" },\n");
+}
+
+void
+VCC_InstanceInfo(struct vcc *tl)
+{
+	Fc(tl, 0, "\nconst struct vpi_ii VGC_instance_info[] = {\n");
+	VCC_WalkSymbols(tl, vcc_instance_info, SYM_NONE);
```
Dridi: Why don't we limit the scope to `SYM_INSTANCE` directly here instead of the callback? Did you try? (I haven't, yet)

Dridi: This seems to be enough:

```diff
diff --git i/lib/libvcc/vcc_xref.c w/lib/libvcc/vcc_xref.c
index 21ede56e3..89fb35cdc 100644
--- i/lib/libvcc/vcc_xref.c
+++ w/lib/libvcc/vcc_xref.c
@@ -342,9 +342,6 @@ vcc_instance_info(struct vcc *tl, const struct symbol *sym)
 {
 
        CHECK_OBJ_NOTNULL(sym, SYMBOL_MAGIC);
-       CHECK_OBJ_NOTNULL(sym->kind, KIND_MAGIC);
-       if (sym->kind != SYM_INSTANCE)
-               return;
        AN(sym->rname);
        Fc(tl, 0, "\t{ .p = &%s, .name = \"", sym->rname);
        VCC_SymName(tl->fc, sym);
@@ -355,7 +352,7 @@ void
 VCC_InstanceInfo(struct vcc *tl)
 {
        Fc(tl, 0, "\nconst struct vpi_ii VGC_instance_info[] = {\n");
-       VCC_WalkSymbols(tl, vcc_instance_info, SYM_NONE);
+       VCC_WalkSymbols(tl, vcc_instance_info, SYM_INSTANCE);
        Fc(tl, 0, "\t{ .p = NULL, .name = \"\" }\n");
        Fc(tl, 0, "};\n");
 }
```

nigoroll:
@Dridi thank you for the sensible suggestion. As agreed before, please feel free to commit to my branches.
Regarding the meta-discussion, I apologize for my impatience, but I just think that we could make better use of our time than discussing and polishing in such depth trivial changes like this one.

Dridi:
> Regarding the meta-discussion, I apologize for my impatience, but I just think that we could make better use of our time than discussing and polishing in such depth trivial changes like this one.

I could probably have chose better wording myself, I also apologize :)

nigoroll:
OKd during discussion with @Dridi and @bsdphk

Issue #2967, opened 2019-04-07
WS_Reserve() return inconsistent


nigoroll:
Ref: https://github.com/varnishcache/varnish-cache/blob/f106ed834f31f2ab5c798c56a6ca48225615dc45/bin/varnishd/cache/cache_ws.c#L242

* case a: If `WS_Reserve()` is called with `bytes > 0` and the reservation cannot be fulfilled, we return 0 with `ws->r == NULL`.
* case b: if `WS_Reserve()` is called with `bytes == 0` and there are 0 bytes available, we also return 0, yet with `ws->r == ws->e`.

`WS_Release()` asserts `ws->r != NULL` (same for `WS_ReleaseP()`)
https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_ws.c#L270
so error handling should only call `WS_Release()` for case b

In varnish-cache core, there are only a few cases where we call `WS_Reserve()` with `bytes > 0` and they either get it right or assert (e.g. for the session where I think it is reasonable to assert that the session workspace is large enough), so there is no real issue here. Yet still I think we should make this consistent.

I think the most POLA way would be to always leave with `ws->r != NULL` such that calling `WS_Release()` is always the right error handling.

slimhazard:
To put this another way (since it was my bug in a VMOD that got caught): I thought it was an error to call ``WS_Release()`` after ``WS_Reserve()==0``, because in the overflow case, ``ws->r==NULL``, and ``WS_Release()`` would fail the assertion.

But in fact, ``WS_Reserve(ws, 0)==0`` can *only* happen if the workspace is right at the limit: ``ws->f==ws->e``. The 0 return is not because of overflow, but because ``ws->r`` gets set, and ``pdiff(ws->f, ws->r)==0``. Then you *have to* call ``WS_Release()`` to reset ``ws->r==NULL``.

So you better know that: ``WS_Release()`` MUST NOT be called after ``WS_Reserve(ws, bytes)`` if ``bytes != 0``, but MUST be called if ``bytes == 0``. I got POLA'd.

Dridi:
I don't think `ws->f==ws->e` is the only case where this may happen because of memory alignment, but close enough.

At the same time I don't consider this inconsistent since the purpose of reserving memory is to use what's available and in the case of a zero-length reservation it doesn't trigger my POLA as I consider it fail-safe (unless there was already an ongoing reservation).

[...]

PR #2969, opened 2019-04-08
Deprecate WS_Reserve() and replace it with WS_ReserveSize()

nigoroll:
`WS_ReserveSize()` does not leave the workspace reserved when the reservation fails, so afterwards `WS_Release()` must be called for the success case (`retval > 0`) only.

Re valid concerns from @Dridi, this does not touch the existing `WS_Reserve()` so we retain full binary compatibility until we remove `WS_Release()`.

As `WS_Release()` is marked deprecated, existing code can still be compiled unchanged by adding `-Wno-deprecated` to `CFLAGS` for the time being.

Besides the debug string, `WS_ReserveSize()` it is identical to `WS_Reserve()` except for
```
        assert(bytes > 0);
```

Follow-up #2967

Additional notes: I chose this approach over macro-wrapping `WS_Release()` because that would only work for a literal `bytes = 0` argument

### DIFF
--- a/bin/varnishd/cache/cache_session.c
+++ b/bin/varnishd/cache/cache_session.c
@@ -112,7 +112,7 @@ ses_reserve_attr(struct sess *sp, enum sess_attr a, void **dst, int sz)
 	assert(a < SA_LAST);
 	assert(sz >= 0);
 	AN(dst);
-	o = WS_Reserve(sp->ws, sz);
+	o = WS_ReserveSize(sp->ws, sz);
 	assert(o >= sz);
 	*dst = sp->ws->f;
 	o = sp->ws->f - sp->ws->s;
@@ -415,7 +415,7 @@ ses_handle(struct waited *wp, enum wait_event ev, vtim_real now)
 	case WAITER_ACTION:
 		pp = sp->pool;
 		CHECK_OBJ_NOTNULL(pp, POOL_MAGIC);
-		assert(sizeof *tp <= WS_Reserve(sp->ws, sizeof *tp));
+		assert(sizeof *tp <= WS_ReserveSize(sp->ws, sizeof *tp));
 		tp = (void*)sp->ws->f;
 		tp->func = xp->unwait;
 		tp->priv = sp;
@@ -453,7 +453,7 @@ SES_Wait(struct sess *sp, const struct transport *xp)
 	/*
 	 * put struct waited on the workspace
 	 */
-	if (WS_Reserve(sp->ws, sizeof(struct waited))
+	if (WS_ReserveSize(sp->ws, sizeof(struct waited))
 	    < sizeof(struct waited)) {
 		SES_Delete(sp, SC_OVERLOAD, NAN);
 		return;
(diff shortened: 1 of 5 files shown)


bin/varnishd/cache/cache_session.c
```diff
@@ -112,7 +112,7 @@ ses_reserve_attr(struct sess *sp, enum sess_attr a, void **dst, int sz)
 	assert(a < SA_LAST);
 	assert(sz >= 0);
 	AN(dst);
-	o = WS_Reserve(sp->ws, sz);
+	o = WS_ReserveSize(sp->ws, sz);
 	assert(o >= sz);
```
bsdphk: Notice the sz >= assert above, that seems unsound ?

bin/varnishd/cache/cache_session.c
```diff
@@ -415,7 +415,7 @@ ses_handle(struct waited *wp, enum wait_event ev, vtim_real now)
 	case WAITER_ACTION:
 		pp = sp->pool;
 		CHECK_OBJ_NOTNULL(pp, POOL_MAGIC);
-		assert(sizeof *tp <= WS_Reserve(sp->ws, sizeof *tp));
+		assert(sizeof *tp <= WS_ReserveSize(sp->ws, sizeof *tp));
 		tp = (void*)sp->ws->f;
```
bsdphk: So is this really the right form of assert now ?

bin/varnishd/cache/cache_ws.c
```diff
@@ -247,9 +247,34 @@ WS_ReserveAll(struct ws *ws)
...
+	ws->r = ws->f + b2;
+	DSL(DBG_WORKSPACE, 0, "WS_ReserveSize(%p, %u/%u) = %u",
+	    ws, b2, bytes, pdiff(ws->f, ws->r));
+	WS_Assert(ws);
+	return (pdiff(ws->f, ws->r));
+}
+
+/*
+ * XXX remove for 2020-03-15 release
  */
```
bsdphk: I wonder if we should define a standard tag for grep'ing this kind of comments ?  REL_20200315 maybe ?

bin/varnishd/cache/cache_session.c
nigoroll: Yes, this is not the only place where we do not yet handle allocation errors properly, but I consider this a different topic. I said in #2976

> In varnish-cache core, there are only a few cases where we call WS_Reserve() with bytes > 0 and they either get it right or assert (e.g. for the session where I think it is reasonable to assert that the session workspace is large enough)

that early in the session I think it is quite unlikely that error handling will have any net effect different to DoS

bin/varnishd/cache/cache_session.c
nigoroll: I don't understand. Do you mean we should make this `==` to assert `PAOK(sizeof(...))` ?

bin/varnishd/cache/cache_ws.c
nigoroll: Yup, will change when the other conversations are resolved.

nigoroll:
OKed during bugwash

PR #2972, opened 2019-04-11
Limit the number of concurrent pipe transactions

Dridi:
This is a first of hopefully three patch series in the pipe vicinity.

This is pretty straightforward, count and limit ongoing pipe transactions only if we `return (pipe)` from `vcl_pipe`.

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -710,6 +710,7 @@ cnt_pipe(struct worker *wrk, struct req *req)
 			V1P_Leave();
 			break;
 		}
+		wrk->stats->pipe_limited++;
 		/* fall through */
 	case VCL_RET_FAIL:
 		req->req_step = R_STP_VCLFAIL;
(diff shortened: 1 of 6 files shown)


bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -695,20 +696,25 @@ cnt_pipe(struct worker *wrk, struct req *req)
...
+			bo->req = req;
+			bo->wrk = wrk;
+			SES_Close(req->sp, VDI_Http1Pipe(req, bo));
+			nxt = REQ_FSM_DONE;
+			V1P_Leave();
+			break;
+		}
+		wrk->stats->pipe_limited++;
+		/* fall through */
+	case VCL_RET_FAIL:
```
nigoroll: why the reordering?

Dridi: To reuse the failure logic by falling through.

bsdphk:
I don't understand why the last commit is required ?

If the child restarts, the VSC's should be rebuilt ?!

Dridi:
You're right, when I first tested it I must have forgotten to restart the child when I witnessed the gauge not coming back to zero. So I will remove the last commit.

PR #2979, opened 2019-04-17
Master rushing

mbgrydeland:
This PR fixes some issues with rushing of waitinglists. The first issue is a regression, where streaming delivery rushing is not propagated fast enough, but delayed until the client delivery is finished. This regression was introduced in 5.0.

The second patch deals with the same kind of problem when inserting a hit-for-pass object. I do not know if this is a regression or not.


nigoroll:
> I do not understand why we need a master PR and two individual ones, though.

Ah sorry those were issues referenced in the PR, sorry for the noise.

bsdphk:
Looks OK to me.

PR #2980, opened 2019-04-19
Start honoring [idle_]send_timeout of h2 streams

Dridi:
I entered my garage to work on something two days ago and found my pet blocking the way. It wouldn't budge until after a good shaving.

There is a lot to review here and I tried to break it down to be as granular as possible. When I first wrote my patch I found bugs I thought I had caused but no, they were already there so instead of just fixing them I tried to reorder them in a more coherent way, and tried to also add h1 coverage.

The first problem is the use of `SO_{RCV,SND}BUF` to cover the h1 timeouts. It works on my machine (obviously) but I don't know how portable this is:

- does it work on all systems in general?
- how may each system round up buffer size?
- may the rounding up lead to test failures?

I decided to leave this to the review because I won't be available next week, and don't have time today to test the various systems we care about.

There is not really a second problem, but rather open questions that may appear in individual commit messages.

And finally, I used the modest hardware from my garage to run all the h2 test cases under load (`-j32 -n1000`) a few times without finding any problem, but this hasn't seen production traffic. I also inspected passing test results as closely as I could and this is partly how I found the bugs I fixed along the way.

Summoning @xcir and @simonvik in case they could give it a try with real traffic.

Did I mention my pet is a yak?

### DIFF
--- a/bin/varnishd/http2/cache_http2_send.c
+++ b/bin/varnishd/http2/cache_http2_send.c
@@ -41,6 +41,41 @@
 
 #define H2_SEND_HELD(h2, r2) (VTAILQ_FIRST(&(h2)->txqueue) == (r2))
 
+static int
+h2_cond_wait(pthread_cond_t *cond, struct h2_sess *h2, struct h2_req *r2)
+{
+	vtim_real now, when = 0.;
+	int r;
+
+	AN(cond);
+	CHECK_OBJ_NOTNULL(h2, H2_SESS_MAGIC);
+	CHECK_OBJ_NOTNULL(r2, H2_REQ_MAGIC);
+
+	Lck_AssertHeld(&h2->sess->mtx);
+
+	now = VTIM_real();
+	if (cache_param->idle_send_timeout > 0.)
+		when = now + cache_param->idle_send_timeout;
+
+	r = Lck_CondWait(cond, &h2->sess->mtx, when);
+
+	now = VTIM_real();
+	/* NB: when we grab idle_send_timeout before acquiring the session
+	 * lock we may time out, but once we wake up both send_timeout and
+	 * idle_send_timeout may have changed meanwhile. For this reason
+	 * h2_stream_tmo() may not log what timed out and we need to check
+	 * both conditions to decide whether we cancel the stream or not.
+	 */
+	if (h2_stream_tmo(h2, r2, now) || r == ETIMEDOUT) {
+		if (r2->error == NULL)
+			r2->error = H2SE_CANCEL;
+		return (-1);
+	}
+
+	AZ(r);
+	return (0);
+}
+
 static void
 h2_send_get_locked(struct worker *wrk, struct h2_sess *h2, struct h2_req *r2)
 {
@@ -224,12 +259,11 @@ h2_do_window(struct worker *wrk, struct h2_req *r2,
 		h2_send_rel_locked(h2, r2);
 		while (r2->t_window <= 0 && h2_errcheck(r2, h2) == 0) {
 			r2->cond = &wrk->cond;
-			AZ(Lck_CondWait(r2->cond, &h2->sess->mtx, 0));
+			(void)h2_cond_wait(r2->cond, h2, r2);
 			r2->cond = NULL;
 		}
-		while (h2->req0->t_window <= 0 && h2_errcheck(r2, h2) == 0) {
-			AZ(Lck_CondWait(h2->winupd_cond, &h2->sess->mtx, 0));
-		}
+		while (h2->req0->t_window <= 0 && h2_errcheck(r2, h2) == 0)
+			(void)h2_cond_wait(h2->winupd_cond, h2, r2);
 
 		if (h2_errcheck(r2, h2) == 0) {
 			w = h2_win_limit(r2, h2);
@@ -377,4 +411,7 @@ H2_Send(struct worker *wrk, struct h2_req *r2, h2_frame ftyp, uint8_t flags,
 		counter = &dummy_counter;
 
 	h2_send(wrk, r2, ftyp, flags, len, ptr, counter);
+
+	if (h2_errcheck(r2, r2->h2sess) == H2SE_CANCEL)
+		H2_Send_RST(wrk, r2->h2sess, r2, r2->stream, H2SE_CANCEL);
 }
--- a/bin/varnishd/http2/cache_http2.h
+++ b/bin/varnishd/http2/cache_http2.h
@@ -232,6 +232,7 @@ void H2_Send(struct worker *, struct h2_req *, h2_frame type, uint8_t flags,
 /* cache_http2_proto.c */
 struct h2_req * h2_new_req(const struct worker *, struct h2_sess *,
     unsigned stream, struct req *);
+int h2_stream_tmo(struct h2_sess *, const struct h2_req *, vtim_real);
 void h2_del_req(struct worker *, const struct h2_req *);
 void h2_kill_req(struct worker *, struct h2_sess *, struct h2_req *, h2_error);
 int h2_rxframe(struct worker *, struct h2_sess *);
--- a/bin/varnishd/http2/cache_http2_deliver.c
+++ b/bin/varnishd/http2/cache_http2_deliver.c
@@ -87,6 +87,10 @@ h2_fini(struct req *req, void **priv)
 
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 	TAKE_OBJ_NOTNULL(r2, priv, H2_REQ_MAGIC);
+
+	if (r2->error)
+		return (0);
+
 	H2_Send_Get(req->wrk, r2->h2sess, r2);
 	H2_Send(req->wrk, r2, H2_F_DATA, H2FF_DATA_END_STREAM, 0, "", NULL);
 	H2_Send_Rel(r2->h2sess, r2);
--- a/bin/varnishd/http2/cache_http2_proto.c
+++ b/bin/varnishd/http2/cache_http2_proto.c
@@ -966,21 +966,20 @@ h2_procframe(struct worker *wrk, struct h2_sess *h2, h2_frame h2f)
 	return (0);
 }
 
-static int
-h2_stream_tmo(struct h2_sess *h2, const struct h2_req *r2)
+int
+h2_stream_tmo(struct h2_sess *h2, const struct h2_req *r2, vtim_real now)
 {
 	int r = 0;
 
 	CHECK_OBJ_NOTNULL(h2, H2_SESS_MAGIC);
 	CHECK_OBJ_NOTNULL(r2, H2_REQ_MAGIC);
+	Lck_AssertHeld(&h2->sess->mtx);
 
 	if (r2->t_winupd == 0 && r2->t_send == 0)
 		return (0);
 
-	Lck_Lock(&h2->sess->mtx);
 	if (r2->t_winupd != 0 &&
-	    h2->sess->t_idle - r2->t_winupd >
-	    cache_param->idle_send_timeout) {
+	    now - r2->t_winupd > cache_param->idle_send_timeout) {
 		VSLb(h2->vsl, SLT_Debug,
 		     "H2: stream %u: Hit idle_send_timeout waiting for"
 		     " WINDOW_UPDATE", r2->stream);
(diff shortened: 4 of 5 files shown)


simonvik:
I will try to build it and install on a few servers.

simonvik:
Got an assert, I'm running this patch against 85abf2044c3f501a674b594716a6af241797acf1


I'm rolling back for now.

```
200        
Panic at: Fri, 19 Apr 2019 18:51:08 GMT
Assert error in H2_Send_RST(), http2/cache_http2_send.c line 394:
  Condition(((((&(h2)->txqueue)->vtqh_first) == (r2))) != 0) not true.
version = varnish-trunk revision 85abf2044c3f501a674b594716a6af241797acf1, vrt api = 9.0
ident = Linux,4.4.134+20180824.0648.fc9acc4c.xenial,x86_64,-junix,-smalloc,-sdefault,-hcritbit,epoll
now = 15491166.651163 (mono), 1555699865.784596 (real)
Backtrace:
  0x55d2b129ace4: /usr/sbin/varnishd(+0x4ece4) [0x55d2b129ace4]
  0x55d2b1305ce0: /usr/sbin/varnishd(VAS_Fail+0x40) [0x55d2b1305ce0]
  0x55d2b12d2caf: /usr/sbin/varnishd(H2_Send_RST+0x15f) [0x55d2b12d2caf]
  0x55d2b12d0d6a: /usr/sbin/varnishd(+0x84d6a) [0x55d2b12d0d6a]
  0x55d2b12d0ffb: /usr/sbin/varnishd(h2_rxframe+0x21b) [0x55d2b12d0ffb]
  0x55d2b12d3c6a: /usr/sbin/varnishd(+0x87c6a) [0x55d2b12d3c6a]
  0x55d2b12bd1d7: /usr/sbin/varnishd(+0x711d7) [0x55d2b12bd1d7]
  0x55d2b12bd680: /usr/sbin/varnishd(+0x71680) [0x55d2b12bd680]
  0x7f28538cb6db: /lib/x86_64-linux-gnu/libpthread.so.0(+0x76db) [0x7f28538cb6db]
  0x7f28535f488f: /lib/x86_64-linux-gnu/libc.so.6(clone+0x3f) [0x7f28535f488f]
thread = (cache-worker)
pthread.attr = {
  guard = 4096,
  stack_bottom = 0x7f27eb700000,
  stack_top = 0x7f27eb730000,
  stack_size = 196608,
}
thr.req = 0x7f214226a020 {
  vxid = 45253629, transport = H2
  step = 0x0,
  req_body = R_BODY_INIT,
  err_code = 1, err_reason = (null),
  restarts = 0, esi_level = 0,
  sp = 0x7f27f940ee20 {
    fd = 1114, vxid = 45253629,
    t_open = 1555699863.222578,
    t_idle = 1555699865.784317,
    ws = 0x7f27f940ee60 {
      id = \"ses\",
      {s, f, r, e} = {0x7f27f940ee98, +224, (nil), +352},
    },
    transport = H2 {
      streams {
        0x00000000 idle  
        0x00000065 closed
        0x00000067 closed
        0x00000073 closed
        0x00000075 closed
        0x00000077 c-rem 
        0x00000079 c-rem 
        0x0000007b c-rem 
        0x0000007d c-rem 
        0x0000007f c-rem 
        0x00000081 c-rem 
        0x00000083 c-rem 
      }
    }
    client = 170.185.71.19 33404 127.0.0.1:81,
  },
  ws = 0x7f214226a170 {
    id = \"req\",
    {s, f, r, e} = {0x7f214226c0b0, +48, (nil), +122696},
  },
  http_conn = 0x7f214226c050 {
    fd = 1114 (@0x7f27f940ee44),
    doclose = NULL,
    ws = 0x7f214226a170 {
      [Already dumped, see above]
    },
    {rxbuf_b, rxbuf_e} = {0x7f214226c0b0, 0x7f214226c0dd},
    {pipeline_b, pipeline_e} = {(nil), (nil)},
    content_length = 0,
    body_status = none,
    first_byte_timeout = 0.000000,
    between_bytes_timeout = 0.000000,
  },
  http[req] = 0x7f214226a218 {
    ws = (nil) {
    },
    hdrs {
    },
  },
  vdc = 0x7f214226c028 {
    nxt = (nil),
    retval = 0,
  },
  flags = {
  },
  privs = 0x7f214226a200 {
    MAGIC at 0x7f214226a200 is 0x00000000 (Should be: VRT_PRIVS_MAGIC/0x03ba7502)
  },
  topreq = {
    req = 0x7f214226a020 {
      [Already dumped, see above]
    },
  },
},
thr.busyobj = (nil) {
},
vmods = {
  std = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
  one = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
  maxminddb = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
  directors = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
  xkey = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
  header = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 9.0},
  cookie = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 9.0},
  vsthrottle = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 9.0},
  tcp = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
  bodyaccess = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
  blob = {Varnish trunk 85abf2044c3f501a674b594716a6af241797acf1, 0.0},
},
```

Dridi:
The problem is here: https://github.com/varnishcache/varnish-cache/pull/2980/commits/fd061bfc6c2aafa15b67b22f68faa7d7a97ce4e4#diff-60b7f08c4eb5f135ecd8ad2fb3c223a4L1044

I should probably have made this change with Coccinelle to avoid that kind of mistakes :)

Fixed in e8185df577ccc1bb5b661ed10e754cbff80c87c6, we need even more test coverage.

Thank you for the early try and sorry for a silly panic.

xcir:
I'll try :)

nigoroll:
@Dridi I did not do a thorough review (for now), but I do not understand what you are saying about `SO_{RCV,SND}BUF`. I see that these could be relevant to minimize latencies with prioritization in H2, but I do not understand the relation to timeouts. Can you please explain?
Are you sure you are not confusing it with `SO_{RCV,SND}TIMEOUT`? IIUC, very early in the project the decision was made to only support timeouts on systems providing these socket options due to the overhead required otherwise. At the time I did run Varnish (2.0.3) on some Solaris 10 boxes which do not have the socket option, so timeouts simply would not work.

bin/varnishd/http2/cache_http2_send.c
```diff
@@ -41,6 +41,41 @@
...
+{
+	vtim_real now, tmo;
+	int r;
+
+	AN(cond);
+	CHECK_OBJ_NOTNULL(h2, H2_SESS_MAGIC);
+	CHECK_OBJ_NOTNULL(r2, H2_REQ_MAGIC);
+
+	now = VTIM_real();
+	tmo = cache_param->idle_send_timeout;
```
nigoroll: this does not seem to make much sense to me. We should still enforce `send_timeout` for the entirety of the response, but I guess this is incomplete because you are just trying to make progress?
If yes, mark this maybe?
(side note: `idle_send_timeout` might not be the best choice of name - rename maybe to `unacknowledged_send_timeout`?  )

Dridi: You are right, the overall `send_timeout` is treated like `idle_send_timeout` in my patch. I will try to fix this today, good catch.

Dridi:
> @Dridi I did not do a thorough review (for now), but I do not understand what you are saying about `SO_{RCV,SND}BUF`. I see that these could be relevant to minimize latencies with prioritization in H2, but I do not understand the relation to timeouts. Can you please explain?

Well, as I was embarking on something else I ended up with this patch series. I needed test coverage before I could engage with the original change I was planning. Tweaking the socket buffers is for h1 coverage, and for h2 coverage we can control a stream's window so we don't need to fiddle with the underlying transport.

The reason why we can't test h1's `send_timeout` is because socket buffers by default are so ridiculously large that we couldn't send `varnishtest` anything with its current capabilities that wouldn't immediately be buffered for `varnishd`.

Run the test case, it shows what the previous size of the rcv/snd buffer was, what the requested size is, and what the actual size will be.

> Are you sure you are not confusing it with `SO_{RCV,SND}TIMEOUT`?

Pretty sure I'm not confusing them ;-)

xcir:
I'm running e8185df commit for 24hours. not panic.

simonvik:
No panic here either, just passed 24h.

bin/varnishd/http2/cache_http2_send.c
Dridi: Fixed in 980f0f5df35a34212e3f7a0ac8890209fafbc511.

bsdphk:
Can we cherry-pick some of the trivial commits to get them out of the way first ?

Dridi:
@bsdphk, give me a list of OK commits and I will push them to master and update this PR.

nigoroll:
@Dridi for the timeouts, I would duplicate some work, of yours, so I will now cherry-pick commits which I believe should be ok. And yes, I understand now what you are doing. :)
@bsdphk I'll take all the blame

nigoroll:
So @Dridi your suspicion about `SO_RCVBUF` has proven true. On SmartOS GETSOCKOPT(3SOCKET):

> At present, lowering SO_RCVBUF on a TCP connection after it has been established has no effect.

So I will work around this for now and leave a marker.

I guess the portable fix would be to teach `VTCP_open()` / `VTCP_connect()` to set additional socket options between the `socket()` and `connect()` calls.

nigoroll:
Looks like FreeBSD behaves similarly

Dridi:
I rebased the patch series onto current master, which results in 3 commits and 1 trivial commit.

I created #2987 to deal with socket buffers limitations.

Since I made a couple mistakes in my initial patch series, I had 2 followup commits to fix a panic found thanks to @simonvik and a logic error spotted by @nigoroll during his review. Those 2 commits were squashed into the original patches they belonged to, so now we have 4 patches to review again without the noise from earlier iterations.

Dridi:
FYI @bsdphk, trivial patches merged by @nigoroll:

- 654f0b5f4
- d8421dcd5
- a0b8a7344
- 8c5118d17

Less trivial patches merged:

- 3c8ecb4cc
- 51127b460

Commit I flagged as really not trivial but also merged:

- d49025519 (see #2987)

Followup patches that work around limitations of the last patch:

- bfbd2963f
- 09785c4ee
- cc90aab3a
- 211f72da3

Dridi:
@daghf I assigned this to you, feel free to start a review despite the merge conflict. I will work on fixing that since I wasn't aware a conflict occurred. For a quick recap of what happened to some of the patches initially part of this pull request, please have a look at my previous comment. Anything before that comment is probably noise if you are looking at this ticket for the first time.

daghf:
The first three patches (https://github.com/varnishcache/varnish-cache/pull/2980/commits/299e614ac8e023ed5a0c52bf6c0f441e9ba2f7ae, https://github.com/varnishcache/varnish-cache/pull/2980/commits/6d0111e4fe0c42ace96d7a4bd2747a32ac6d0770 and https://github.com/varnishcache/varnish-cache/pull/2980/commits/bad124806194445cfc85a0a7e3f7b11b1288dd55) look trivial and ready to go in.

For https://github.com/varnishcache/varnish-cache/pull/2980/commits/5a99516aeedb83c3b0be3cad77924175f9f04b71 I've had a good look at it and I think this looks correct, aside from one particular assert.

The `XXXAZ(H2_SEND_HELD(h2, r2));` in `h2_cond_wait` is wrong. We do `h2_send_rel_locked` just prior to the first invocation in `h2_do_window`. Since that call involves a Lck_CondWait (i.e. we let go of the lock), some other stream might have grabbed the first spot in the sender list when `h2_cond_wait` is called the second time, and we will run into this assert.

I suggest we just drop this assert, what is being checked by it isn't relevant for the functionality there anyway.

The comment relating to "NB: we check the stream timeout before the lock (..)" is a little unclear and I've discussed this with @Dridi .

Aside from the merge conflict and what I've mentioned above, I think this is ready to go in.

Dridi:
Rebased against current master, I removed the offending XXX assertion (thanks for confirming my doubts) and rewrote the comment from scratch.

Please have a final review of the last commit, I pushed the 3 patches you OK'd to master. The conflict was trivial, an `acct` variable was later renamed to `counter` in 3d170c0d8fd029eb8604854b9a758404f19008c2.

daghf:
LGTM

Issue #2764, opened 2018-09-05
CLI timeout after vcl.discard with many backend (in one VSC_cluster)


ehocdet:
## Current Behavior
The issue can happen after a vcl.discard (the VCL is in cold/cold state)
```
Sep  5 09:01:07 webacc1 varnishd[3608]: CLI telnet **** 31480 ***** X Rd vcl.discard cache_1_20180905T07h40m25s
Sep  5 09:05:08 webacc1 varnishd[3608]: Child (3326) not responding to CLI, killed it.
Sep  5 09:05:08 webacc1 varnishd[3608]: CLI telnet **** 31480 **** X Wr 200 
Sep  5 09:05:08 webacc1 varnishd[3608]: Child (3326) not responding to CLI, killed it.
Sep  5 09:05:08 webacc1 varnishd[3608]: Unexpected reply from ping: 400 CLI communication error (hdr)
Sep  5 09:05:08 webacc1 varnishd[3608]: Child (3326) died signal=3
```

## Context
This issue follows #2518.
Varnish 6 with 5k+ backends and one VSC_cluster for dynamic backends (init in vcl_ini).
With VSC_cluster, vcl.load and VSC usages are fast (thanks for that).

vcl.load/vcl.discard is done very regularly, and the issue appears after a vcl.discard when 
more than 3 vcl are loaded.
To limit the impact `cli_timout`  is set to 360s.
To isolate the issue, vcl.discard is done after vcl is in cold/cold state. 

## Your Environment
. Version used: based on Varnish 6.0.0 from git + h2 fixes from master
. Operating System and version: Linux 3.14.79 and 4.9.0

## Debug:
strace child when vcl are discarded:

```
[...]
write(2420, "_.VSC_cluster.3b361cef4062598a 3"..., 105) = 105
write(2420, "_.VSC_cluster.3b361cef4062598a 3"..., 106) = 106
write(2420, "_.VSC_cluster.3b361cef4062598a 3"..., 105) = 105
close(2420)                             = 0
renameat(2393, "_._.index.719d7a460983fb48", 2393, "_.index") = 0
futex(0x7f6bb650e428, FUTEX_WAIT_PRIVATE, 2, NULL) = -1 EAGAIN (Resource temporarily unavailable)
futex(0x7f6bb650e428, FUTEX_WAIT_PRIVATE, 2, NULL) = -1 EAGAIN (Resource temporarily unavailable)
[...]

PR #2981, opened 2019-04-20
Add timeout_idle customization per session

nigoroll:
Purposes:

* Allow to optimize keepalive for trusted client connections (e.g. CDN, authenticated clients)

* Prepare for an optimization to time out connections on inactive VCLs earlier (see #2764)

A VTC and the other client timeouts will be added when accepted.


bsdphk:
Implementation sounds OK.  Wants to hear bugwash consensus on idea.

nigoroll:
bugwash OK

PR #2982, opened 2019-04-22
Kill all still running varnishtest tasks when the parent varnishtest …

andrewwiik:
By adding a list of running varnishtest jobs and catching the kill message we can make sure that all varnishtest tasks are exited when the main varnishtest is killed.

Fixes: varnishcache#2952


bsdphk:
I have a very old TODO item about this, wondering if using process groups would fix this issue.

The advantage of process groups, if they work, is that they would also take care of grand-children.

Want to play with that idea ?

andrewwiik:
I'll give it a go why not :)

PR #2983, opened 2019-04-24
Add vcl control over timeout_linger, send_timeout, idle_send_timeout

nigoroll:
and test that they work for H1

follow up #2981 

loosely related to #2980

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -575,6 +575,9 @@ struct sess {
 	vtim_real		t_open;		/* fd accepted */
 	vtim_real		t_idle;		/* fd accepted or resp sent */
 	vtim_dur		timeout_idle;
+	vtim_dur		timeout_linger;
+	vtim_dur		send_timeout;
+	vtim_dur		idle_send_timeout;
 };
 
 #define SESS_TMO(sp, tmo)					\
--- a/bin/varnishd/cache/cache_session.c
+++ b/bin/varnishd/cache/cache_session.c
@@ -389,6 +389,9 @@ SES_New(struct pool *pp)
 	sp->t_open = NAN;
 	sp->t_idle = NAN;
 	sp->timeout_idle = NAN;
+	sp->timeout_linger = NAN;
+	sp->send_timeout = NAN;
+	sp->idle_send_timeout = NAN;
 	Lck_New(&sp->mtx, lck_sess);
 	CHECK_OBJ_NOTNULL(sp, SESS_MAGIC);
 	return (sp);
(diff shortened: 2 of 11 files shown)


bsdphk:
It worries me that we have suddenly increased struct sess by 32 bytes (¼) and thereby reduced the sess->ws by same amount.

Is this a valid concern any more ?

nigoroll:
It's hard to dismiss a concern before having shown that there is no reason to it, but c00070.vtc tests for a minimum of 200 bytes free on the session workspace in `vcl_deliver {}`
On alternative options:
* @bsdphk you suggested we could make the `struct sess` members floats (and not touch the `vtim_dur`) typedef, which would save 8 bytes. We could certainly do this, but this might become an additional complication with respect to the idea of compile time switchable vtim_* implementations which I still got on my P3 todo ist.
* We could also add `PRIV_SESSION` to hold these variables. This would cost only one pointer for the `struct vrt_priv_tree`, but individual values would require significantly more memory (72 bytes) for the individual values. The advantage would be flexibility.
So at this point adding the 32bytes sounds like a viable option to me.

nigoroll:
It seems we missed to note the TODO from last bugwash:
@nigoroll and @Dridi to write up a consistent timeouts concept at least for H1

nigoroll:
current snapshot of https://etherpad.wikimedia.org/p/varnish_timeouts

Ref https://github.com/varnishcache/varnish-cache/pull/2983

* rename idle_send_timeout -> send_chunk_timeout

The current name is confusing. The actual effect is that SO_SNDTIMEO gets set such that individual write() calls time out, and it is not obvious how this would relate to _idle_. Other suggestions are welcome, the aim should be to make clear that the timeout applies to individual pieces of data. Ideally, it would also conver H2.

* default send_chunk_timeout to send_timeout / 2, limit to send_timeout

Setting send_chunk_timeout to anything larger than send_timeout effectively prolongues send_timeout, and this should be communicated to the varnish admin.
We should also use a more sensible send_chunk_timeout.

* Dynamically adjust SO_SNDTIMEO to the remainter of send_timeout

With the current code, send_timeout is not being checked unless send_chunk_timeout is being hit, so effectively send_timeout may be prolongued by just slightly less than send_chunk_timeout in the extreme case.
We should ensure that SO_SNDTIMEO is being set such that send_timeout will be in effect even if a write is issued with less than send_chunk_timeout remaining before send_timeout expires.

* record the current SO_SNDTIMEO of the file descriptor (for H1)

To save system calls, as a consequence of the previous bullet point, the file descriptor should remember its current SNDTIMEO

* duplicate send_chunk_timeout and send_timeout for the backend side

Tuning these for the backend side should also be possible.

bsdphk:
More research needed.

nigoroll:
s10 needs to be adjusted to recent changes, otherwise OKd during discussion with @bsdphk and @Dridi

PR #2984, opened 2019-04-24
Use Process Groups

### LINKED ISSUE #2952: varnishtest leaks varnishtest tasks
See below, when the main `varnishtest` process receives a `SIGINT` or `SIGTERM` it goes away but the tasks it started are not interrupted:

```
$ cat >leak.vtc <<EOF
varnishtest "varnishtest leak"

# should be plenty of time to kill varnishtest
delay 50
EOF
$ bin/varnishtest/varnishtest -i leak.vtc &
[1] 30936
$ pgrep varnishtest
30936
30959
$ kill 30936
[1]+  Terminated              bin/varnishtest/varnishtest -i leak.vtc
$ pgrep varnishtest
30959
```

andrewwiik:
This fixes #2952 by using process groups for varnishtest children and grandchildren.

### DIFF
--- a/bin/varnishtest/vtc_main.c
+++ b/bin/varnishtest/vtc_main.c
@@ -37,6 +37,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <signal.h>
 #include <unistd.h>
 
 #include "vtc.h"
@@ -85,7 +86,6 @@ struct vtc_job {
 	double			t0;
 	int			killed;
 };
-
 
 int iflg = 0;
 unsigned vtc_maxdur = 60;
@@ -359,6 +359,7 @@ tst_cb(const struct vev *ve, int what)
 				printf(" exit=%d\n", WEXITSTATUS(stx));
 			if (!vtc_continue) {
 				/* XXX kill -9 other jobs ? */
+				AZ(kill(-getpgrp(), WEXITSTATUS(stx)));
 				exit(2);
 			}
 		} else if (vtc_verbosity) {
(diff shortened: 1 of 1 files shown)


nigoroll:
lgtm

andrewwiik:
Hold this pull request, I think I am doing some things wrong.

andrewwiik:
Now it should be good to go.

bsdphk:
I don't think this works as intended, at the very least I get an assert on the "kill(negative pid)" when I try it on FreeBSD.

Staring and fumbling a bit with it, I think we may want to analyze this a tiny bit more anyway:  Which exact processes are we trying to catch here ?

The processes started with "process" have their own sessions which implies pgroups and we already kill(-pid) those, so there is little else we can do.

Things we run with "shell" is not so protected, and I guess the main risk of lingering processes there is in the argument-error tests where varnishd might start up if the args do not fail testing.  However, if the process gets far enough to call daemon(3) we can not reach it anyhow.

The implication is that testing processes through their daemon(3) calls is _hard_

Arguably "shell" could be reimplemented using the same/shared mechanism as "process", which might improve things a little bit, but doesn't solve the daemon(3) issue.

Processes started by vtc_haproxy/vtc_varnish are supposed to be well behaved, it is not my impression they are a problem ?

The only 100% sure-fire way I know to kill a random subtree of processes, is to start them in a jail(2) and kill that jail.  That will take down any process in it, no matter if it called daemon() or not.

However, jails have to be _explicitly_ killed, they do not die when the process which created the jail dies, so if you kill -9 the top varnishtest process, all the jails will survive.

Not sure what to do...

nigoroll:
when I gave my _lgtm_ I should have mentioned that I did not see this cause any regressions, but I had not tested thoroughly.
Adding to the jail(2)-ish options, on Solaris there is the [contract(4)](https://docs.oracle.com/cd/E86824_01/html/E54775/contract-4.html#REFMAN4contract-4) facility which allows to monitor process ancestry - see [ctrun](https://docs.oracle.com/cd/E86824_01/html/E54763/ctrun-1.html#REFMAN1ctrun-1) [ctwatch](https://docs.oracle.com/cd/E86824_01/html/E54763/ctwatch-1.html#REFMAN1ctwatch-1) for examples.
On Linux there are [pid namespaces](http://man7.org/linux/man-pages/man7/pid_namespaces.7.html) but creating them with `CLONE_NEWPID` requires `CAP_SYS_ADMIN` aka root-ish.

bsdphk:
Question for bugwash:  Which processes do we see lingering after varnishtest ?

andrewwiik:
So if we start varnishtest with 8 parallel tests that simply have “delay
50” then kill the main varnishtest the children do not get killed and
continue running for the entire delay.

andrewwiik:
Also I believe the current imp could also leave grandchildren also running even though the main focus of the original issue was just the children of the main varnishtest process

bsdphk:
We agreed last week to try the stdin-pipe approach instead.

PR #2986, opened 2019-04-25
Master unbusy race

mbgrydeland:
Resolve race on waitinglist rushing between OC_F_BUSY and boc->state  …
When an object is ready for delivery, HSH_Unbusy was called before calling
ObjSetState([BOS_STREAM|BOS_FINISHED]). The HSH_Unbusy() call does the
waitinglist rushing, but HSH_Lookup() wanted to look at the boc->state and
if BOS_STREAM had been reached. This could cause requests woken to find
that the stream state still hadn't been reached (ObjSetState still hadn't
executed), and go back on the waitinglist.

To fix this, this patch reverts commit 0375791, and goes back to considering
OC_F_BUSY as the gate keeper for HSH_Lookup. This eliminates the race,
because HSH_Unbusy and HSH_Lookup then uses the same mutex.

That change opens up the possiblity that req code after HSH_Lookup() sees
an object that has not yet reached BOS_STREAM. In order to not have to add
new ObjWaitState() calls (with the additional locking cost that would
bring) to wait for BOS_STREAM, the order of events is changed throughout,
and calls ObjSetState([BOS_STREAM|BOS_FINISHED]) before HSH_Unbusy(). That
way, an object returned from HSH_Lookup() is guaranteed to be at least
BOS_STREAM.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -562,9 +562,8 @@ vbf_stp_fetch(struct worker *wrk, struct busyobj *bo)
 	assert(bo->fetch_objcore->boc->state == BOS_REQ_DONE);
 
 	if (bo->do_stream) {
-		ObjSetState(wrk, bo->fetch_objcore, BOS_PREP_STREAM);
-		HSH_Unbusy(wrk, bo->fetch_objcore);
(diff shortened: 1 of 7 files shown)


bsdphk:
Looks good to me and passes mild torture on 32&64 bit FreeBSD.

PR #2991, opened 2019-05-02
2nd attempt at resolving the waitinglist rush HSH_Unbusy and boc->state race

mbgrydeland:
The previous attempt at fixing this was less than ideal. By signalling the delivery threads BOS_STREAM state before HSH_Unbusy, several test cases became fragile, and would fail under high load conditions. What would happen was that the fetch initiating request was allowed to continue (seeing BOS_STREAM) even though HSH_Unbusy had not been run yet, allowing the handling of that request to finish completely, and the test case to continue running. If the test case next step e.g.. was to purge the same URL, it would fail to purge the object (still OC_F_BUSY).

This PR reverts commits 501246e and 4130055. The third patch adds a new approach to fix the initial race.


bsdphk:
Looks good.

We may want to return to that "dont start streaming until we have body-bytes(-if any)" idea later.

PR #2992, opened 2019-05-05
Ban lurker optimalization: Remove from ban list when possible

hermunn:
This is an improvement of #2925 (which I will close). The the difference is that this PR is even more eager to remove objects from the ban list.

The goal of this PR is to not hold off when dying objects are encountered, and it turned out that we can remove `oc`s in other cases as well.

I am not sure how much effect the last commit has, but I think there are corner cases where the speedup would be significant.

### DIFF
--- a/bin/varnishd/cache/cache_ban_lurker.c
+++ b/bin/varnishd/cache/cache_ban_lurker.c
@@ -365,7 +365,7 @@ ban_lurker_work(struct worker *wrk, struct vsl_log *vsl)
 
 	/*
 	 * conceptually, all obans are now completed. Remove the tail. If it
-	 * containted the first oban, all obans were on the tail and we're
+	 * contained the first oban, all obans were on the tail and we're
 	 * done.
 	 */
 	if (ban_cleantail(VTAILQ_FIRST(&obans)))
--- a/bin/varnishd/cache/cache_ban_lurker.c
+++ b/bin/varnishd/cache/cache_ban_lurker.c
@@ -167,8 +167,23 @@ ban_lurker_getfirst(struct vsl_log *vsl, struct ban *bt)
 		oh = oc->objhead;
 		CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
 		if (!Lck_Trylock(&oh->mtx)) {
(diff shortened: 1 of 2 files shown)


bin/varnishd/cache/cache_ban_lurker.c
```diff
@@ -149,13 +150,15 @@ ban_lurker_getfirst(struct vsl_log *vsl, struct ban *bt)
...
 
-			/* hold off to give lookup a chance and reiterate */
-			VSC_C_main->bans_lurker_contention++;
-			Lck_Unlock(&ban_mtx);
-			VSL_Flush(vsl, 0);
-			VTIM_sleep(cache_param->ban_lurker_holdoff);
-			Lck_Lock(&ban_mtx);
-
+			if (contested) {
+				/* hold off to give lookup a chance and reiterate */
```
nigoroll: My comment from #2925 still holds:


We want the holdoff to happen if there were any contested ocs. Yet if one contested oc was returned and _getfirst gets called the second time, contested will be 0 and we won't hold off.
Also I would consider the intention a micro-optimization not worth the complication.

bin/varnishd/cache/cache_ban_lurker.c
```diff
@@ -167,7 +167,23 @@ ban_lurker_getfirst(struct vsl_log *vsl, struct ban *bt)
...
-			if (oc->refcnt == 0 || oc->flags & OC_F_BUSY) {
+			if (oc->refcnt == 0 ||
+			    oc->flags & (OC_F_DYING | OC_F_FAILED)) {
+				/*
+				 * We seize the opportunity to remove
+				 * the object completely off the ban
+				 * list, now that we have both the oh
+				 * and ban mutexes.
+				 */
+				AZ(oc->flags & OC_F_BUSY);
```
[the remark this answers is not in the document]
hermunn: Not sure, and I agree that I can just as well change the order.

bin/varnishd/cache/cache_ban_lurker.c
hermunn: Ok, then. I will not merge the last commit unless others protests.

hermunn:
I force pushed a new version after taking @nigoroll 's comments into account.

PR #2993, opened 2019-05-06
Work around old and new getaddrinfo quirks

Dridi:
Starting with glibc 2.29 I lost the behavior we consistently rely upon when it comes to resolving names and services. A `VSS_ResolveOne` function was introduced as the blessed wrapper around `getaddrinfo`, but is not consistent with the expected behavior.

This patch introduces a `vss_resolve` function shared by both old and new interfaces, and aligns `VSS_ResolveOne` with the established behavior, treating its port as a default port overriden if one is present in the address.

<details>
<summary>Previous pull request description</summary>
Its first purpose is to deduplicate results as it may happen at least
with glibc depending on your nsswitch configuration. In particular if
one has more than one source for name and service lookup the exact same
entry may show up more than once, especially with special names like
localhost.

That lead us to choose that binding for -a types of parameters should be
"at least one" to be considered successful. We could now reconsider to
make "all results" the condition to succeed. That would move the error
for ports already bound to the user if multiple -a parameters create an
overlap, or something is bound by another process.

The second problem is that since glibc 2.29 our ability to pass both
name and service as the node parameter and ignore the service parameter
was geopardized and is no longer reliable. We rely on it implicitly in
several places so the wrapper decomposes service and name when it runs
into such occurrence.

I ran into this trying to build the 6.0 branch where d00007.vtc would
fail inconditionally. This was not apparent in trunk because of a patch
that had been back-ported slightly differently:

- ead72749b78d99f941649e879deadbb683891d10 (master)
- b302a2cf4d4f3863d4ebe9d8c0d6092e5b401079 (6.0)

As this is mostly used in the test suite, this could explain why we
haven't heard of the second problem yet. But as of today glibc 2.29 is
fairly young so maybe few systems are using it and running Varnish.

For this reason d00007.vtc was changed to exhibit the usage of a
"name service" node with and without a service name to show that we
definitively get the implicit behavior we were reyling on so far.
I also found that debug.dyn() failed if at least one of name or service
was omitted and changed that since getaddrinfo doesn't require both,
especially if we pass them via the node argument.

Finally, to make this change throughout the tree I wrote a Coccinelle
patch. If we decide to expose this facility via libvarnishapi it may
become useful out of tree too.
</details>

### DIFF
--- a/bin/varnishtest/tests/m00011.vtc
+++ b/bin/varnishtest/tests/m00011.vtc
@@ -7,25 +7,76 @@ server s1 {
 
 varnish v1 -vcl+backend {
 	import std;
+	import debug;
 
 	sub vcl_deliver {
 		std.timestamp("t0");
-		set resp.http.foo0 = std.ip("8.8.8.*", client.ip);
+
+		debug.store_ip(std.ip("8.8.8.*", client.ip));
+		set resp.http.ip0 = debug.get_ip();
+		set resp.http.port0 = std.port(debug.get_ip());
 		std.timestamp("8.8.8.*, client.ip");
-		set resp.http.foo1 = std.ip("9.9.9.*", server.ip);
+
+		debug.store_ip(std.ip("9.9.9.*", server.ip));
+		set resp.http.ip1 = debug.get_ip();
+		set resp.http.port1 = std.port(debug.get_ip());
 		std.timestamp("9.9.9.*, server.ip");
-		set resp.http.foo2 = std.ip("1.2.3.*", "127.0.0.2");
+
+		debug.store_ip(std.ip("1.2.3.*", "127.0.0.2"));
+		set resp.http.ip2 = debug.get_ip();
+		set resp.http.port2 = std.port(debug.get_ip());
 		std.timestamp("1.2.3.*");
-		set resp.http.foo3 = std.ip("1.2.3.5", "127.0.0.3");
+
+		debug.store_ip(std.ip("1.2.3.5", "127.0.0.3"));
+		set resp.http.ip3 = debug.get_ip();
+		set resp.http.port3 = std.port(debug.get_ip());
 		std.timestamp("1.2.3.5");
-		set resp.http.foo4 = std.ip("2001:db8::", "[::1]");
+
+		debug.store_ip(std.ip("2001:db8::", "[::1]"));
+		set resp.http.ip4 = debug.get_ip();
+		set resp.http.port4 = std.port(debug.get_ip());
 		std.timestamp("2001:db8::");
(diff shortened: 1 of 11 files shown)


Dridi:
Updated to:

- move the code to `vss.c`
- discard `ai_canonname` altogether

bsdphk:
I am not satisfied with the mechanics of this change.

In order to simplify things a bit, I have gone through the uses in the tree, and abstracted those out which want a single unique result to a new VSS_ResolveOne() function.

This creates conflicts, but I trust it is easy to resolve.

I'm not sure I see where getaddrinfo is being called with combined host/port argument, can you elaborate on that ?

Dridi:
We are discussing it offline, I will update the pull request once we agree on the way forward.

Dridi:
I pushed a couple post-`VSS_ResolveOne` commits that should be uncontroversial. I forced-push a change that should be enough to solve my original problem, and I can iterate on a `VSS_ResolveAll` function later if we decide to move forward with `suckaddr` deduplication.

I updated the pull request description accordingly.

Dridi:
I attempted a back-port of this pull request and the commits it's build upon on the 6.0 branch to test the waters. I think it's ready now, but I had to introduce a `VSS_ResolveFirst` too.

bsdphk:
OK

Dridi:
I have one more thing before I consider this OK.

With the changes already committed and this pull request we now officially open the door to resolving both an address and a port with `std.ip()`. It was already possible before since to some extent `getaddrinfo` supported this case unconditionally until my last glibc upgrade.

My proposal is to:

- add a `[STRING default_port]` optional argument to `std.ip()`
- document that address may contain a service or port number
- also pass `AI_NUMERICSERV` when `[resolve]` is `false`

Caveats for VCL: the address passed to `std.ip()` may require some sanitizing beforehand, but that is already the case with implicit address+service support in `getaddrinfo`.

Benefits for VCL: it is now possible to meaningfully use `std.port()` on an `std.ip()`-built IP, and we have a better-defined behavior.

I won't attend the bugwash today, please let me know what the consensus is and I can either update this pull request to add the missing bits or merge it and push the rest directly.

bsdphk:
Consensus: Do it.

Dridi:
Done, ready for a last review.

Dridi:
A conflict landed in `vmod_debug.c`, I can take care of resolving it by myself if approved.

bsdphk:
OK

Dridi:
ea3c109c5...5aea717e2

Dridi:
7d1ded3aa...41d303866 in 6.0

PR #2996, opened 2019-05-13
Kill strcat and strcpy usage in VIN_n_Arg

Dridi:
If an absolute path is provided as n_arg with a length of exactly
PATH_MAX-1 then the combination of strcpy and strcat for the trailing
slash '/' overflows dn by one byte, writing its new null-terminating
character '\0' right after dn's upper bound.

By using a fixed-length VSB we can simply ensure that we stay within
bounds at a reasonable cost. Guarding VSB operations should silence
Flexelint as a nice side effect.

VIN_n_Arg is not exposed outside of the source tree, and both callers
today provide a valid dir argument, so we can now make it part of the
contract with an assertion, simplifying the strdup error handling.

### DIFF
--- a/lib/libvarnish/vin.c
+++ b/lib/libvarnish/vin.c
@@ -39,15 +39,19 @@
 
 #include "vdef.h"
 
-#include "vas.h"	// XXX Flexelint "not used" - but req'ed for assert()
+#include "vas.h"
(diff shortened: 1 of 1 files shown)


bsdphk:
OK

Issue #2997, opened 2019-05-15
panic in VRT_fail() due to ctx->handling missing in director context


nigoroll:
This panic happened on code based on 1ad4bb344cc16f17e05a2d350834f5f202922b97 - the code can be found here: https://github.com/nigoroll/varnish-cache/tree/handling_issue

This basically is the same issue as #2692 but now caused by the changes in 5735a543c720655c484ef11c70b6cde7f201d5ca : Insisting on `ctx->handling` being present basically renders the `VRT_fail()` path unusable from the director context.

So I think the options are:
* always add `ctx->handling`
* not call `VRT_fail()` from director context
* make `VRT_fail()` ignore missing `ctx->handling`

```
Panic at: Wed, 15 May 2019 04:43:06 GMT
Assert error in VRT_fail(), cache/cache_vrt.c line 511:
  Condition((ctx->handling) != 0) not true.
version = varnish-trunk revision 58e4f83ca808ad1a5302f5fce3c830330835710c, vrt api = 9.0
ident = Linux,3.13.0-126-generic,x86_64,-junix,-smalloc,-sdefault,-hcritbit,epoll
now = 7403232.980590 (mono), 1557894718.482515 (real)
Backtrace:
  0x43d0bb: /usr/sbin/varnishd() [0x43d0bb]
  0x4a0852: /usr/sbin/varnishd(VAS_Fail+0x42) [0x4a0852]
  0x44f340: /usr/sbin/varnishd(VRT_fail+0x170) [0x44f340]
  0x7f1fdbba4927: ./vmod_cache/_vmod_directors.885a314a7f37617696cf8fadadd2c7c323784baddc40e9588ebc503d9c9d71b0(+0x6927) [0x7f1fdbba4927]
  0x7f1fdbba5388: ./vmod_cache/_vmod_directors.885a314a7f37617696cf8fadadd2c7c323784baddc40e9588ebc503d9c9d71b0(+0x7388) [0x7f1fdbba5388]
  0x7f1fdbba54af: ./vmod_cache/_vmod_directors.885a314a7f37617696cf8fadadd2c7c323784baddc40e9588ebc503d9c9d71b0(+0x74af) [0x7f1fdbba54af]
  0x459ee1: /usr/sbin/varnishd(VRT_DirectorResolve+0x41) [0x459ee1]
  0x422b00: /usr/sbin/varnishd() [0x422b00]
[...]

PR #2998, opened 2019-05-15
add ctx->handling and ctx->vsl for cli

nigoroll:
This implements one of the possible options to approach #2997

### DIFF
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -68,6 +68,7 @@ static struct vrt_ctx ctx_cli;
 static unsigned handling_cli;
 static struct ws ws_cli;
 static uintptr_t ws_snapshot_cli;
+static struct vsl_log vsl_cli;
 
 /*--------------------------------------------------------------------*/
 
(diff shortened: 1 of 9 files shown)


bin/varnishd/cache/cache_vcl.c
```diff
@@ -85,6 +86,8 @@ VCL_Bo2Ctx(struct vrt_ctx *ctx, struct busyobj *bo)
 	ctx->sp = bo->sp;
 	ctx->now = bo->t_prev;
 	ctx->ws = bo->ws;
+	ctx->handling = &bo->wrk->handling;
+	*ctx->handling = 0;
```
bsdphk: Isn't this already done by vcl_call_method() in cache_vrt_vcl ?

nigoroll: `vcl_call_method()` uses a local `struct vrt_ctx`, as does `VDI_GetHdr()`, so there is no effect of the former on the latter.
But you rightly pointed out that my suggestion duplicates these two lines, I will check if they could be removed from `vcl_call_method()`.

nigoroll: sorry, I was looking at master, I actually had already suggested to remove the duplication here:
https://github.com/varnishcache/varnish-cache/pull/2998/commits/218c818023dfb3446e413dd00b64c755be764876#diff-f00ee27228a92dc1471fc250affb40c9L444

nigoroll:
@bsdphk isn't this https://github.com/varnishcache/varnish-cache/pull/2998/commits/51fb3d4fef45c0911283a932123f1024ec680157 ?

nigoroll:
@Dridi I agree that we should continue in this direction.

PR #3000, opened 2019-05-16
Add Debug to the default VSL mask

Dridi:
The name implies that this is not for production usage.


bsdphk:
Hmm, I thought we already did that for releases, but I guess not... ?

Dridi:
Not that I'm aware of, either way I think users should opt in to get Debug records.

bsdphk:
Bugwash consensus: Disable SLT_Debug by default.

PR #3001, opened 2019-05-17
VSL query extensions 1 and 2

Dridi:
Better late than never, this patch series implements the easy part of my proposal from last year:

https://varnish-cache.org/lists/pipermail/varnish-dev/2018-May/009251.html

That would be step 1 and 2 from my email.

The consensus during the VDD was "send patches" and as such I took advantage of being stuck until #2993 is merged to look at my backlog and started working on that.


Dridi:
In 28a8a0ea7e2e0058ddcc7c6873d61c5773041787 the commit log says:

> This diff is better viewed with the --ignore-all-space flag.

On github that would mean adding `?w=1` to the URL: https://github.com/varnishcache/varnish-cache/commit/28a8a0ea7e2e0058ddcc7c6873d61c5773041787?w=1

bsdphk:
No opinion.

lib/libvarnishapi/vxp_lexer.c
```diff
@@ -126,6 +126,8 @@ vxp_Lexer(struct vxp *vxp)
 					q++;
 					if (q == vxp->e)
 						break;
+				} else if (*q == '\n') {
```
mbgrydeland: I am not sure I understand what is happening here. Should it not be possible to include a literal newline inside a quoted string? Agreed it makes for an awkward display, but do we need to specifically disallow it?

lib/libvarnishapi/vxp_lexer.c
```diff
@@ -54,10 +54,7 @@ vxp_add_token(struct vxp *vxp, unsigned tok, const char *b, const char *e)
 	t->tok = tok;
 	t->b = b;
 	t->e = e;
-	if (vxp->t != NULL)
```
mbgrydeland: If this change goes in, please also change name to vxp_append_token(). But I think it's better to leave as is, as it then follows the same structure of libvcc (upon which this code was based), and someone familiar with libvcc won't be confused.

lib/libvarnishapi/vxp_lexer.c
```diff
@@ -45,7 +45,7 @@
 #include "vxp.h"
 
 static void
-vxp_add_token(struct vxp *vxp, unsigned tok, const char *b, const char *e)
+vxp_append_token(struct vxp *vxp, unsigned tok, const char *b, const char *e)
```
mbgrydeland: Hehe, I guess this change belonged in the previous patch (ref my comment there)

lib/libvarnishapi/vxp_lexer.c
```diff
@@ -58,6 +58,32 @@ vxp_add_token(struct vxp *vxp, unsigned tok, const char *b, const char *e)
 	vxp->t = t;
 }
 
+static void
+vxp_add_token(struct vxp *vxp, unsigned tok, const char *b, const char *e)
+{
+
+	/* Ignore first or consecutive EOI tokens */
```
mbgrydeland: Would it be cleaner to define a EOQ (end of query) token, and leave EOI to define the end of input?

lib/libvarnishapi/vxp_lexer.c
Dridi: Here we break out of the loop if we encounter an actual end of line, not a quoted one. Since an individual query should only span a single line, encountering a new line means the same as reaching `vxp->e`: incomplete string.

Maybe this deserves some VTC coverage.

lib/libvarnishapi/vxp_lexer.c
Dridi: I suggest we discuss this and the next comment live when we get the occasion.

lib/libvarnishapi/vxp_lexer.c
Dridi: I will give this a try. I anticipated this one wouldn't pass under the radar ;)

mbgrydeland: I think the folding and stuff that happens here breaks the lexer and parser separation. We probably should have a EOL/NL token (not EOQ as I suggested), and let the lexer at the tokens it finds as normal. The separation of things into separate expressions to be OR'ed together, and the skipping of empty queries resulting from blank lines, belongs in the parser IMO.

lib/libvarnishapi/vxp_lexer.c
mbgrydeland: I'm not sure I see a problem with allowing literal newlines in quoted strings. And it would be a change of behaviour to suddenly disallow it. POLA I guess says they should be allowed. But I'm fine either way.

mbgrydeland:
One final question: Do we need the error reporting to learn about line numbers when dealing with multi-line queries and/or queries read from a file? So that it will be able to report which line caused the problem?

lib/libvarnishapi/vxp_lexer.c
Dridi: My understanding is that this comment is resolved after a brief discussion.

lib/libvarnishapi/vxp_lexer.c
Dridi: My understanding is that this comment resolves the previous one after a brief discussion.

lib/libvarnishapi/vxp_lexer.c
Dridi: I see, I initially thought you misunderstood that change. Now I guess we should gather consensus on this specific point. I will mark this conversation as resolved now that I pushed a second iteration of this patch series, we can discuss this again on the new relevant patch.

lib/libvarnishapi/vxp_lexer.c
Dridi: Marking the conversation as resolved, we can take this to the second iteration.

Dridi:
I pushed a second iteration after the first review:

- rebased against current master
- EOI folding handled by the parser (no more 223aa4291)
- removed test case noise (no more 0bb68fda3 and efc912241)

Instead of fixing the test case as part of the patch series, the test case is now properly written from the very beginning.

I didn't introduce a separate NL or EOQ token, I will leave it to a third iteration if you still think we shouldn't reuse EOI for this purpose. Not that I disagree, I wanted to make a small iteration.

This patch series still considers a quoted string incomplete when it contains a new line, it is now covered by the test case. Again, I maintained the status quo here to start the review again on a clean slate and cleaned up the first review's negative feedback (and the needless back-and-forth in the test case).

Dridi:
I pushed 0cd6008ec2c6ec8d9c1111c1a3aad849c62739fa to show that with string escaping we can already make a VSL query against a record that contains multiple lines. Besides `CLI` records and `std.log()` we have no occurrence of multi-line records and besides regex operators I don't see how a quoted string spanning multiple lines would be useful:

```
vut -q 'tag operator "quoted
string"'
```

I would personally not feel sorry about breaking this particular syntax. Please note that this pull request in general breaks the existing syntax since new lines so far were only considered as mere separators in the query:

```
vut -q 'tag operator value'

# same as

vut -q '
    tag
    operator
    value
'
```

Treating new line as separators was useful to build complex VSL queries, but I think being able to decompose and OR them is in general more useful. And multi-line quoted strings are close to useless in your average query (never ran into a case where that would solve a problem).

Dridi:
I will push one last commit with release notes before merging, this is a breaking change even though in practice it should hardly break any setup.

Dridi:
Added 4562b69d61337f747d99461be2a8eae990eee163 not initially planned (I woke up with the idea) and got it approved by @mbgrydeland.

Dridi:
I pushed the changelog update by first cherry-picking @slimhazard's commit for the 2019-03 release (very convenient FYI). I will merge this after next bugwash unless someone yells by then, feel free to review the release notes for this change.

PR #3008, opened 2019-05-31
Make backends optional and unset-able

andrewwiik:
Allow backends to be optional and unset-able. The optional backends feature has been put behind a parameter called "vcc_no_backend" that defaults to false to prevent breaking the previous behavior that users might rely on.

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -792,7 +792,6 @@ cnt_recv_prep(struct req *req, const char *ci)
 
 		/* By default we use the first backend */
 		req->director_hint = VCL_DefaultDirector(req->vcl);
-		AN(req->director_hint);
 
 		req->d_ttl = -1;
 		req->d_grace = -1;
--- a/bin/varnishd/cache/cache_vrt_var.c
+++ b/bin/varnishd/cache/cache_vrt_var.c
@@ -459,8 +459,19 @@ VRT_r_req_##nm(VRT_CTX)							\
 	return (ctx->req->elem);					\
 }
 
+#define REQ_VAR_U(nm, elem)						\
+									\
+VCL_VOID								\
+VRT_u_req_##nm(VRT_CTX)							\
+{									\
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);				\
+	CHECK_OBJ_NOTNULL(ctx->req, REQ_MAGIC);				\
+	ctx->req->elem = NULL;						\
+}
+
 REQ_VAR_L(backend_hint, director_hint, VCL_BACKEND,)
 REQ_VAR_R(backend_hint, director_hint, VCL_BACKEND)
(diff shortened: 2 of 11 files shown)


bsdphk:
A lot of discussion.

There are at least three issues here:

1) return(error) (or something like it) from vcl_backend_{} to produce synth object.  (PHK thought we had this already)

2) Ability to not declare backends (because vmods do them)

3) Ability to set variable (in this case req.backend_hint) to NULL

The first is an entirely separate PR and out of scope here.

The second can be done a number of ways (param/feature/in the vcl), for instance "backend default none;" I *think* in-vcl is preferable, because that way users do not need to remember "out-of-band" magic to load the VCL.  Homework assignment:  How to do this ?

The third is also homework:  In addition to backends, which other variables would it make sense to be able to set to a NULL value, and what restrictions should be put on doing so (and where should the restrictions apply?)

This ticket remains open, to remind us to talk about this on 20190610 bugwash.

Dridi:
For issue 2. I suggested a pragma-style approach for which I already have two and a half use cases:

```
pragma <keyword>;
```

The first use case would obviously grant the right not to declare a backend:

```vcl
# complete active vcl below
vcl 4.1;

pragma no-backend;

sub vcl_recv {
    if (req.http.host == "example.org") {
        return (vcl(org-label));
    } else if (req.http.host == "example.com") {
        return (vcl(com-label));
    }
    else {
        return (synth(404));
    }
}
```

Another use case, making the `vcc_err_unref` a per-VCL option:

```vcl
pragma allow-unref;

include "common/subroutines/i/partially/use.vcl";
```

And deprecate the global parameter as a result.

I have another idea for a `pragma` for hypothetical future developments.

bsdphk:
bugwash discussion:  Remove "there is always a default backend" for vcl 4.2, but leave "if there is a backend called default".

Dridi to write up proposal.

andrewwiik:
This commit/branch allows the ```backend default none;``` syntax: https://github.com/andrewwiik/varnish-cache/commit/96e65f4623cba7172b2351df6dd298d824710210

rezan:
So I would like to possibly propose going back to something like this:

``` vcl
backend null_backend none;

sub vcl_recv
{
    ...
    set req.backend_hint = null_backend;
}
```

One of my really nice to haves here is getting this into 6.0. So if you look at Andrews above commit (https://github.com/andrewwiik/varnish-cache/commit/96e65f4623cba7172b2351df6dd298d824710210):

* We can create a VCL with a non functional backend
* We keep the "No backends found" error intact
* We get `unset` for free by defining and using the non functional backend
* The patch is pretty simple

rezan:
Can we pick this back up again? I would say there is a large need to have optional backends in 6.0.

nigoroll:
+1 for `backend <name> none;`

daghf:
vdd discussion: To mirror VCL allowing both cases "True" and "true" for booleans, we should allow both "None" and "none"

nigoroll:
continued in #3067

PR #3010, opened 2019-06-04
Delay bereq WS_Reset() when using std.rollback()

### LINKED ISSUE #3009: std.rollback() without restart/retry unsafe
If you `std.rollback()` without a restart/retry, workspace gets reset, however, the req and bo still references these locations. This leads to a panic when that workspace gets reused.

VTC:

``` vcl
varnishtest "Rollback without restart/retry is unsafe"

server s1 {
	rxreq
	txresp
} -start

varnish v1 -vcl+backend {
	import std;

	sub vcl_recv {
		set req.http.test = "1";
	}

	sub vcl_backend_fetch {
		unset bereq.http.test;
	}

	sub vcl_backend_response {
		std.rollback(bereq);
		set beresp.http.test = bereq.http.test;
		set beresp.http.workspace = "start overwriting active workspace";
		set beresp.http.workspace = "0123456789012345678901234567890123456789";
		# panic...
	}
} -start

client c1 {
	txreq
	rxresp
	expect resp.status == 200
	expect resp.http.test == "1"
} -run
```

I looked into this, the only safe way to address this is to not reset the workspace on a rollback. Or just docfix this and state you must restart/retry?

rezan:
Resetting the workspace during the actual call exposes workspace structures to VCL writes. This delays the reset until a retry is performed.

Fixes #3009

Doing a rollback in req also has a similar problem. Can fix that once we settle on a fix for bereq.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -209,6 +209,7 @@ vbf_stp_mkbereq(struct worker *wrk, struct busyobj *bo)
 	HTTP_Setup(bo->bereq, bo->ws, bo->vsl, SLT_BereqMethod);
 	bo->ws_bo = WS_Snapshot(bo->ws);
 	HTTP_Clone(bo->bereq, bo->bereq0);
+	AZ(bo->ws_bo_rst);
 
 	if (bo->req->req_body_status == REQ_BODY_NONE) {
 		bo->req = NULL;
@@ -235,12 +236,20 @@ vbf_stp_retry(struct worker *wrk, struct busyobj *bo)
 	/* VDI_Finish must have been called before */
 	assert(bo->director_state == DIR_S_NULL);
 
+	/* Workspace hasn't been used since rollback */
+	if (bo->ws_bo_rst == WS_Snapshot(bo->ws)) {
+		CHECK_OBJ_NOTNULL(bo->bereq, HTTP_MAGIC);
+		WS_Reset(bo->bereq->ws, bo->ws_bo);
+		WS_Reset(bo->ws, bo->ws_bo);
+	}
+
 	/* reset other bo attributes - See VBO_GetBusyObj */
 	bo->storage = NULL;
 	bo->do_esi = 0;
 	bo->do_stream = 1;
 	bo->filter_list = NULL;
 	bo->was_304 = 0;
+	bo->ws_bo_rst = 0;
 
 	// XXX: BereqEnd + BereqAcct ?
 	VSL_ChgId(bo->vsl, "bereq", "retry", VXID_Get(wrk, VSL_BACKENDMARKER));
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -404,6 +404,7 @@ struct busyobj {
 
 	struct ws		ws[1];
 	uintptr_t		ws_bo;
+	uintptr_t		ws_bo_rst;
 	struct http		*bereq0;
 	struct http		*bereq;
 	struct http		*beresp;
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -497,6 +497,7 @@ struct req {
 	struct vcl		*vcl;
 
 	uintptr_t		ws_req;		/* WS above request data */
+	uintptr_t		ws_req_rst;
 
 	/* Timestamps */
 	vtim_real		t_first;	/* First timestamp logged */
--- a/bin/varnishd/cache/cache_req.c
+++ b/bin/varnishd/cache/cache_req.c
@@ -185,14 +185,17 @@ Req_Release(struct req *req)
  */
 
 void
-Req_Rollback(struct req *req)
+Req_Rollback(struct req *req, int delay)
 {
 	VCL_TaskLeave(req->vcl, req->privs);
 	VCL_TaskEnter(req->vcl, req->privs);
 	HTTP_Clone(req->http, req->http0);
 	if (WS_Overflowed(req->ws))
 		req->wrk->stats->ws_client_overflow++;
-	WS_Reset(req->ws, req->ws_req);
+	if (delay)
+		req->ws_req_rst = WS_Snapshot(req->ws);
+	else
+		WS_Reset(req->ws, req->ws_req);
 }
 
 /*----------------------------------------------------------------------
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -96,6 +96,7 @@ cnt_transport(struct worker *wrk, struct req *req)
 
 	req->ws_req = WS_Snapshot(req->ws);
 	HTTP_Clone(req->http0, req->http);	// For ESI & restart
+	AZ(req->ws_req_rst);
 	req->req_step = R_STP_RECV;
 	return (REQ_FSM_MORE);
 }
@@ -205,7 +206,7 @@ cnt_vclfail(const struct worker *wrk, struct req *req)
 	AZ(req->objcore);
 	AZ(req->stale_oc);
 
-	Req_Rollback(req);
+	Req_Rollback(req, 0);
 
 	req->err_code = 503;
 	req->err_reason = "VCL failed";
@@ -751,11 +752,15 @@ cnt_restart(struct worker *wrk, struct req *req)
 		req->req_step = R_STP_SYNTH;
 	} else {
 		// XXX: ReqEnd + ReqAcct ?
+		if (req->ws_req_rst == WS_Snapshot(req->ws))
+			WS_Reset(req->ws, req->ws_req);
+
 		VSLb_ts_req(req, "Restart", W_TIM_real(wrk));
 		VSL_ChgId(req->vsl, "req", "restart",
 		    VXID_Get(wrk, VSL_CLIENTMARKER));
 		VSLb_ts_req(req, "Start", req->t_prev);
 		req->err_code = 0;
+		req->ws_req_rst = 0;
 		req->req_step = R_STP_RECV;
 	}
 	return (REQ_FSM_MORE);
@@ -851,7 +856,7 @@ cnt_recv(struct worker *wrk, struct req *req)
 
 	VCL_recv_method(req->vcl, wrk, req, NULL, NULL);
 	if (wrk->handling == VCL_RET_VCL && req->restarts == 0) {
-		Req_Rollback(req);
+		Req_Rollback(req, 0);
 		cnt_recv_prep(req, ci);
 		VCL_recv_method(req->vcl, wrk, req, NULL, NULL);
 	}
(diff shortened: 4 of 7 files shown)


rezan:
Had to amend the PR to account for the following case:

``` vcl
sub vcl_backend_response {
	std.rollback(bereq);
	set bereq.http.header = "workspace123";
	return (retry);
}
```

Basically if you rollback(), but then decide to use workspace, then retry, we cannot reset the workspace anymore. So snapshot is taken at rollback() and if the snapshot shows no workspace was used when you reach a retry, reset.

Dridi:
I had a test case ready before lunch to show the unsafety of the initial reset, but by the time I came back you had already fixed that bug.

<details>
<summary>Show test case</summary>

```
varnishtest "unsafe std.rollback() with pure vcl"

server s1 {
        rxreq
        txresp

        rxreq
        close
        expect resp.http.retries == 1
        expect resp.http.retried == "something went wrong"
} -start

varnish v1 -vcl+backend {
        import std;

        sub vcl_backend_fetch {
                if (bereq.retries > 0) {
                        set bereq.http.cto = bereq.connect_timeout;
                        set bereq.http.fbto = bereq.first_byte_timeout;
                        set bereq.http.bbto = bereq.between_bytes_timeout;
                        set bereq.http.isbg = bereq.is_bgfetch;
                        set bereq.http.retries = bereq.retries;
                }
        }

        sub vcl_backend_response {
                if (bereq.retries == 0) {
                        std.rollback(bereq);
                        set bereq.http.retried = "something went wrong";
                        return (retry);
                }
        }
} -start

client c1 {
        txreq
        rxresp
        expect resp.status == 503
} -run
```

</details>

The thing we should closely look at now is PRIV_TASK if you start allowing arbitrary code between a rollback and a retry. Again, I would rather add optional parameters to `std.rollback()` to allow narrowing down what is rolled back (defaulting to everything to maintain compatibility) and allow for example only undoing changes to `(be)?req` whilst still making progress in the current transaction.

nigoroll:
I do not think that delaying the rollback and silently not doing it if something has changed on the workspace is a good solution.
A rollback should just work and additional changes should be possible after it whether or not followed by a retry/restart.
If this does not work at the moment (I have tried to reproduce it yet), we should understand why and make sure the relevant data structures are reset.

I dug out the most relevant email from a conversation with @Dridi back in 2017 which summarizes more or less what we were aiming for at the time (but did not get around implementing). This is a slightly edited version:

```
$Function VOID rollback(ENUM {all, req, bereq} what, [BLOB survivor])
```

what:
* all: Everything in req/bereq context (including PRIVs & workspace)
* req/bereq: Only on client/backend side: (be)req.*

Note: only "all" includes a `WS_Reset()`. To be able to `WS_Reset()` with PRIVs and other vmod data on the workspace, we would need to make all of it relocatable, which I think we shouldn't do for too much overhead.

survivor:

A blob to be restored after the rollback. This is particularly relevant for `what=all` where the workspace gets reset to improve memory efficiency.

Dridi Example to save method/url/proto across a rollback - this includes a hypothetical "vdr = Varnish Data Representation" VMOD (iow, this does _not_ exist yet)

```
	sub vcl_init {
	    # mup = Method URL Protocol / vdr = Varnish Data Representation
	    new mup = vdr.create();
	    mup.field("method", STRING);
	    mup.field("url", STRING);
	    mup.field("proto", STRING);
	    mup.finalize();
	}

	sub vcl_recv_start_all_over {
	    # will look much better with polymorphism
	    mup.set_string("method", req.method);
	    mup.set_string("url", req.url);
	    mup.set_string("proto", req.proto);

	    mup.unserialize(
		std.rollback(req, survivor=mup.serialize()))

	    # will look much better with polymorphism
	    req.method = mup.get_string("method");
	    req.url = mup.get_string("url");
	    req.proto = mup.get_string("proto");
	}
```

rezan:
> A rollback should just work and additional changes should be possible after it whether or not followed by a retry/restart.

The problem is that since we store req and bereq structures in the workspace. These structures aren't accounted for when you do the rollback and basically expose core structures to a use after free state. You would have to allocate them pre-snapshot and this all might work.

nigoroll:
@rezan IIUC, the intention of the existing code is to reset (`HTTP_Clone()`) to the initial state.

rezan:
The problem is with the `WS_Reset()`. The `HTTP_Clone()` is ok.

nigoroll:
If what I am trying to say does not make any sense, I will try to get back to this and submit a PR eventually. But probably not very soon.

rezan:
ok, we still perform the `HTTP_Clone()` and reset the headers. This happens immediately. We just delay the workspace operations and sometimes skip it entirely. Did you read the PR title :)

Dridi:
We (@bsdphk, @mbgrydeland, @daghf and myself) reviewed this yesterday and weren't overall satisfied with the proposal. I was tasked to attempt a plan moving forward, taking into account what was discussed yesterday, and figuring out the rest myself (so it took me some time to do my research).

Below, when I say restart I mean either `return(restart)` or `return(retry)` in a generic fashion when I'm not talking specifically about a client or backend transaction.

One thing we wanted to clarify, is that a rollback+restart implies that dynamic privs owned by a transaction should be finished before we reenter the state machine. With PRIV_TOP in particular, it would be nice to have test coverage involving rollbacks, which I requested in https://github.com/varnishcache/varnish-cache/pull/3011#issuecomment-499427811.

We agreed on a problem statement, that today `std.rollback()` must be followed by a restart, but as I stated above this is not a solution we are satisfied with. We are considering moving the rollback+restart synergy to the VCL itself:

- `return (restart)` vs `return (restart(rollback))`
- `return (retry)` vs `return (retry(rollback))`

The lack of a `rollback` argument to a restart transition keeps the same semantics as today. The presence of a `rollback` argument makes `std.rollback()` entirely redundant.

Currently `std.rollback()` takes an `HTTP` argument, it *MUST* be `(be)?req` or `std.rollback()` will panic (`VRT_Rollback 'hp' invalid`) if we try to pass either `resp` or `beresp`.

Actionable items:

- docfix that `std.rollback()` *MUST* be followed by a restart
- make the `HTTP` argument optional, defaults to TheRightThing(tm), to lessen ceremony
- `VRT_fail()` the transaction for `(be)?resp`

There are legitimate cases where we may want to roll back a `(be)?req`, possibly restart the transaction, but in general not roll back the transaction itself and make further progress without wiping dynamic privs and workspace allocations.

Possible solution:

- introduce a VCL 4.2 syntax (review what else may need a behavior change)
- change the `std.rollback()` behavior for VCL 4.2 only
- allow the rollback of `(be)?resp`? (we don't keep track of them as of today)

Alternative solution:

- introduce a new `$Function` in std for `HTTP`-only rollback
- possibly one for client and one for backend transactions (eg. `std.rollback_(be)?req()`)
- allow the rollback of `(be)?resp` with the new `$Function`?

The last items end with a question mark for both, on purpose. Probably a discussion of its own, but unless we teach the rollback how to deal with more than `(be)?req` it is pointless to even pass an `HTTP` argument.

If we go for a specific function like `std.rollback_bereq()` for example, we may want to revisit [VIP4](https://github.com/varnishcache/varnish-cache/wiki/VIP4%3A-Restrict-VMOD-function-call-sites).

To make the `HTTP` argument relevant, we could imagine new functions:

```
$Function VOID rollback_http([HTTP dest, HTTP src])
$Function HTTP req0()
$Function HTTP bereq0()
```

The following statements would be equivalent in a client context:

```
std.rollback_http();
std.rollback_http(req, std.req0());
```

With such an API [VIP4](https://github.com/varnishcache/varnish-cache/wiki/VIP4%3A-Restrict-VMOD-function-call-sites) would also be relevant to revisit.

A VMOD could then "snapshot" a `(be)?resp` to later be used with `std.rollback_http()`. Going this route would make the existing `std.rollback()` definitively redundant, so we could deprecate it, and remove it when we settle on the next stable branch.

rezan:
> docfix that `std.rollback()` MUST be followed by a restart

Does this mean we are going to tolerate user VCL generating a panic? The one thing that makes me nervous about this panic is that you can rewrite part of the req, so there are security implications here (you can rewrite `struct http_conn` inside of the busyobj using a header). So im ok with re-designing rollback(), but I do feel we need to close this hole in 6.0.

rezan:
Added a second commit to do the same delayed reset on the client side. There was no safety issue that I could see on the client side, so this was mostly done for future safety and parity with the backend patch.

The goal here would be to patch this bug without any functionality changes and leave the door open for a future enhancement to make straighten out workspaces and rollbacks.

Anyway, I will be around tomorrow US time on IRC if anyone wants to chat about this.

rezan:
This PR simply fixes the workspace unsafeness issue without any VCL, behavior, or functionality side effects.

If you want to propose something new, then I would recommend doing that in a separate PR. I figured it would be prudent to patch this sooner than later given this has security implications. This would then remove any time constraints on pushing out a new feature fix.

nigoroll:
@rezan here you are: #3077

nigoroll:
@rezan for completeness (and to hopefully make my reasons easier to understand) regarding my dismissed review, please let me quote again the relevant bits:

> I do not think that delaying the rollback and silently not doing it if something has changed on the workspace is a good solution.
> A rollback should just work and additional changes should be possible after it whether or not followed by a retry/restart.
> If this does not work at the moment (I have tried to reproduce it yet), we should understand why and make sure the relevant data structures are reset.

so I did not request changes for "new features", but because your suggestion:
* did not address the root cause
* silently did not execute the rollback if a change was found

rezan:
> silently did not execute the rollback if a change was found

No, the rollback was executed immediately. Its just the workspace reset which got delayed. This has no effect on the actual rollback. See the title of the PR :)

> did not address the root cause

Right. I do like your PR better because, as you state, it fixes the actual issue which is that we leave pointers on the workspace.

If everyone else is on board with this, then we should be able to finally resolve this issue!

bin/varnishd/cache/cache_fetch.c
```diff
@@ -235,12 +236,20 @@ vbf_stp_retry(struct worker *wrk, struct busyobj *bo)
 	/* VDI_Finish must have been called before */
 	assert(bo->director_state == DIR_S_NULL);
 
+	/* Workspace hasn't been used since rollback */
+	if (bo->ws_bo_rst == WS_Snapshot(bo->ws)) {
```
nigoroll: @rezan isn't this silently not executing the rollback if anything else was changed?

nigoroll:
added another review comment just to clarify.
It's good if we agree on a solution, I just wanted to get this straight because I do not want to criticize for no good reason.

bin/varnishd/cache/cache_fetch.c
rezan: Right, but you said *rollback* in your previous comment. Those are 2 very different words in the context of this PR.

nigoroll: Yes, this is confusing, but the VCL rollback actually is a WS Reset plus restoring the http struct.

So yes, you are right that I should have/could have been more precise.

Thanks for clarifying

rezan:
Closing in favor of https://github.com/varnishcache/varnish-cache/pull/3077.

Issue #3003, opened 2019-05-22
PRIV_TOP conflicts with PRIV_TASK


Dridi:
I was struggling to back-port a bunch of commits in this area to 6.0, eventually decided to test trunk, and I realized the conflict exists there too:

```
varnishtest "Test PRIV_TOP vs PRIV_TASK"

varnish v1 -vcl {
        import debug;

        backend bad { .host = "${bad_backend}"; }

        sub vcl_recv {
                return (synth(200));
        }

        sub vcl_synth {
                set resp.http.priv_task = debug.test_priv_task("task");
                set resp.http.priv_top = debug.test_priv_top("top");
        }
} -start


client c1 {
        txreq -url
        rxresp
        expect resp.http.priv_task == task
        expect resp.http.priv_top == top
} -run
```

The last `expect` statement fails:

```
**   c1    0.5 === expect resp.http.priv_task == task
**** c1    0.5 EXPECT resp.http.priv_task (task) == "task" match
**   c1    0.5 === expect resp.http.priv_top == top
---- c1    0.5 EXPECT resp.http.priv_top (task) == "top" failed
```

Swapping the two statements inside `vcl_synth` makes `vmod-debug` panic, which is a less friendly way to exhibit the priv conflict. It boils down to both using the (so to speak) top `req` address as the `priv` id in this specific scenario.

A possible solution could be to resurrect `struct reqtop`, and add a field that would hold the `priv_top` id or make the workspace-allocated struct address the id itself, similar to what is done with `req` for client `priv_task`.

I'm well aware why it was retired (d67993451bda7eeb7c0ea3c5528a8353f283a190) but it may have been a necessary evil all along.

Dridi:
By the way, I was able to confirm that back-porting e8985f22711fe9a342b27a10052ae3655f64435c to 6.0 with the merge conflict in `VRT_priv_top` resolved as below is enough to fix this bug.

```diff
diff --git a/bin/varnishd/cache/cache_vrt_priv.c b/bin/varnishd/cache/cache_vrt_priv.c
index 52fd156b7..55802769a 100644
--- a/bin/varnishd/cache/cache_vrt_priv.c
+++ b/bin/varnishd/cache/cache_vrt_priv.c
@@ -152,18 +152,25 @@ VRT_priv_top(VRT_CTX, const void *vmod_id)
 {
        uintptr_t id;
        struct vrt_privs *vps;
+       struct req *req;
 
[...]

PR #3011, opened 2019-06-05
fix PRIV_TOP for real

nigoroll:
#3003 exposed a bug which I introduced when I removed the additional id from `struct vrt_priv`, unfortunately.
This series of patches ultimately implements an own head for `PRIV_TOP` which only exists once per top request.
As also the `vcl0` pointer is once per top request, we now reintroduce `struct topreq` (which was added and removed again recently) to contain the pointer to the top request, `vcl0` and the `PRIV_TOP` head.
One particular complication regarding `struct topreq` is that we want to allocate it only on demand, yet it needs to persist a rollback. `Req_MoveTop()` moves it to the beginning of the workspace after a `WS_Reset()`

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -452,6 +452,7 @@ struct reqtop {
 #define REQTOP_MAGIC		0x57fbda52
 	struct req		*topreq;
 	struct vcl		*vcl0;
+	struct vrt_privs	privs[1];
 };
 
 struct req {
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -447,6 +447,11 @@ struct busyobj {
 
 /*--------------------------------------------------------------------*/
 
+struct reqtop {
+	unsigned		magic;
+#define REQTOP_MAGIC		0x57fbda52
+	struct req		*topreq;
+};
 
 struct req {
 	unsigned		magic;
@@ -456,8 +461,8 @@ struct req {
 	volatile enum req_body_state_e	req_body_status;
 	enum sess_close		doclose;
 	int			restarts;
-	unsigned		esi_level;
-	struct req		*topreq;	/* esi_level == 0 request */
+	int			esi_level;
+	struct reqtop		*top;	/* esi_level == 0 request */
 	struct vcl		*vcl0;
 
 #define REQ_FLAG(l, r, w, d) unsigned	l:1;
@@ -538,7 +543,7 @@ struct req {
 	struct vcf		*vcf;
 };
 
-#define IS_TOPREQ(req) ((req)->topreq == (req))
+#define IS_TOPREQ(req) ((req)->top == NULL || (req)->top->topreq == (req))
 
 /*--------------------------------------------------------------------
  * Struct sess is a high memory-load structure because sessions typically
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -451,6 +451,7 @@ struct reqtop {
 	unsigned		magic;
 #define REQTOP_MAGIC		0x57fbda52
 	struct req		*topreq;
+	struct vcl		*vcl0;
 };
 
 struct req {
@@ -463,7 +464,6 @@ struct req {
 	int			restarts;
 	int			esi_level;
 	struct reqtop		*top;	/* esi_level == 0 request */
-	struct vcl		*vcl0;
 
 #define REQ_FLAG(l, r, w, d) unsigned	l:1;
 #include "tbl/req_flags.h"
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -263,15 +263,8 @@ ved_vdp_esi_init(struct req *req, void **priv)
 	*priv = ecx;
 	RFC2616_Weaken_Etag(req->resp);
 
-	if (IS_TOPREQ(req)) {
-		Req_MakeTop(req);
-		if (req->top == NULL) {
-			VSLb(req->vsl, SLT_Error,
-			    "(top)request workspace overflow");
-			Req_Fail(req, SC_OVERLOAD);
-			return (-1);
-		}
-	}
+	if (IS_TOPREQ(req) && Req_MakeTop(req))
+		return (-1);
 
 	req->res_mode |= RES_ESI;
 	if (req->resp_len != 0)
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -170,8 +170,9 @@ ved_include(struct req *preq, const char *src, const char *host,
 	req->req_body_status = REQ_BODY_NONE;
 
 	AZ(req->vcl);
-	if (req->vcl0)
-		req->vcl = req->vcl0;
+	AN(req->top);
+	if (req->top->vcl0)
+		req->vcl = req->top->vcl0;
 	else
 		req->vcl = preq->vcl;
 	VCL_Ref(req->vcl);
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -113,7 +113,7 @@ ved_include(struct req *preq, const char *src, const char *host,
 	enum req_fsm_nxt s;
 
 	CHECK_OBJ_NOTNULL(preq, REQ_MAGIC);
-	CHECK_OBJ_NOTNULL(preq->topreq, REQ_MAGIC);
+	CHECK_OBJ_NOTNULL(preq->top, REQTOP_MAGIC);
 	sp = preq->sp;
 	CHECK_OBJ_NOTNULL(sp, SESS_MAGIC);
 	CHECK_OBJ_NOTNULL(ecx, ECX_MAGIC);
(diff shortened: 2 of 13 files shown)


Dridi:
I find it funny how this one got full circle since I initially thought we'd need to bring back `struct reqtop` to fix #3003. Regarding the reasons why `struct reqtop` was dropped, VRT was irrelevant since this is exposed in `cache.h` but the overhead might be relevant. However between its first introduction and retirement I haven't seen any numbers.

I find a391deb36ba9186f88746d2767fedb43a8bad677 misleading since reqtop doesn't seem to interact with vcl0 the way the commit message suggests. Later when it is effectively moved as described in 17ea57e71112c00e08e88b1530c0d6477ee71846, I'm starting to wonder why a sub-request may see a NULL `vcl0` (I only superficially read the diffs). Is `vcl0` set only if the active VCL switched to a label?

`Req_MoveTop()` looks like a clever solution for restarts, how does it compose with `std.rollback()`? I had a semi-hard time with e7e6f2b20d9cec68658aabb1928894f06d5939e8, but didn't have time for a thorough review. See #3010 and #3009 for the backend side.

I guess it boils done to a lack of test coverage in this area, like #3003 before I realized we didn't exercise both TOP and TASK privs in the same transaction. In general LGTM, but with the all the grains of salt disseminated above.

nigoroll:
@Dridi thank you for having a look

> the overhead might be relevant.

I don't think it matters: this is basically just a `WS_Alloc()` and it only happens for ESI and switching.

Also, on the positive side, we save a bit of workspace by not having the contents of `struct reqtop` for non-ESI non-switching requests.

> I find a391deb misleading since reqtop doesn't seem to interact with vcl0 the way the commit message suggests.

I have changed that bad commit message, thank you.

> Is vcl0 set only if the active VCL switched to a label?

Yes. This is nothing new, my patches should just reflect the existing behavior where `vcl0` is a member of `struct req` and the top `struct req` is being found via the `topreq` pointer.

> Req_MoveTop() looks like a clever solution for restarts, how does it compose with std.rollback()?

I tried not to get involved with #3010 and #3009 for now, but I agree that `std.rollback()` is broken. `Req_MoveTop()` is an effort to not make things worse.

> I guess it boils done to a lack of test coverage in this area, like #3003 before I realized we didn't exercise both TOP and TASK privs in the same transaction.

Agree. And thank you for having found my bug.

Dridi:
> Agree. And thank you for having found my bug.

Meh, I was trying to nudge you increase coverage in this area as part of your pull request :p

Or maybe we already have some PRIV_TOP coverage that deals with workspace resets? I was thinking about this in particular since there is an explicit change in this area (though it shouldn't change current behavior).

nigoroll:
> Meh, I was trying to nudge you increase coverage

Ah, c'était ta politesse française, just be blunt with Zejerman and he will understand :)

We really need to get the resets done properly and I would like to get back to it. But first I would have wanted to fix the bug I caused for real.

bsdphk:
Bugwash chatter:

What happens here:

    VCLA:
    sub vcl_recv {
        vmod_foo.dosomethingwith_PRIV_TOP();
        return (vcl(VCLlabeltoB));
    }
    VCLB:
    sub vcl_recv {
         vmod_foo.pickupstufffrom_PRIV_TOP():
    }

Bugwash consensus was that PRIV_TOP's fini when return(vcl) happens.

However, imagine:

     VCLB:
     import foo;
     export sub something {
            foo.munge_PRIV_TOP();
     }
     VCLA: (active VCL)
     import B:something;
     import foo; // happens to be other version
     sub vcl_recv {
            call B:something;
            foo.munge_PRIV_TOP();
     }

Two different TOP_PRIV's ?

nigoroll:
@bsdphk IMHO the hypothetical `B:something` would make that a sub of VCLA.

Other than that I do not see much of a point in discussing some other hypothetical scenario.

nigoroll:
And also the discussion about hypothetical new features would not stop at PRIV_TOP, but rather extend to all the other PRIVs (and open lots of cans of worms).

Dridi:
Fixes #3019.

Dridi:
As I was complaining regarding the lack of testing in this area, I pushed a commit to add some coverage.

bsdphk:
@nigoroll A imported sub runs in the VCL it is imported from, it is compiled there and for all we know that VCL might use a different version of some VMOD than the active VCL.

nigoroll:
@bsdphk to me this sounds like the privs (the problem extends to all the privs, not just `PRIV_TOP`) being a reason why not to add the vcl sub import.
If we really wanted something like this, I would see two options:
* require vmods to tolerate foreign privs
* make the privs per VCL

But, again, this question is out of scope for this PR. The problem existed before this PR, applies to all the PRIVs and is, IMHO, totally up to the person looking after the hypothetical VCL sub import feature to solve.

bsdphk:
@nigoroll I think the privs will have to be per VCL, already for the "return(vcl)" functionality we have now.

nigoroll:
@bsdphk for `return(vcl)`, we finalize them, which makes perfect sense for vcl switching.
Importing subs is different

Dridi:
Here is what I imagine, we automatically export VMODs:

```
 VCLB:
 import foo;
 export sub something {
        foo.munge_PRIV_TOP();
 }

 VCLA: (active VCL)
 import B:something;
 import foo; # happens to be other version
 sub vcl_recv {
        call B:something;
        B:foo.munge_PRIV_TOP(); # implicitly available
        foo.munge_PRIV_TOP(); # doesn't conflict, different priv
}
```

We may need a new `PRIV_EXPORT` if we really want to pin a `PRIV_VCL` to VCL B, otherwise anything imported would work with VCL A. Or if B may refer to a label in `import B:something` then we better make it impossible to cross the VCL boundaries via privs.

nigoroll:
@Dridi @bsdphk can we please keep this separate? This is clearly VIP material.

Dridi:
@nigoroll it is what I'm trying to do: provide ideas to avoid the hypothetical problem in the first place. I also want to merge this.

nigoroll:
@Dridi why are we sometimes applying our own rules and sometimes not?
This is a PR for an _existing_ feature and we _do not_ want to hijack PRs for new features.

Dridi:
This pull request is doing fine in production: https://github.com/varnishcache/varnish-cache/issues/3019#issuecomment-508448706

nigoroll:
I have added a commit to allow for optional thread-safe access to PRIV_TOP.
This would be a different PR altogether, but this one is still open, so there is no better place to put this.

xdecock:
Any reason why this did not make it in 6.3? we're currently handling ~200 qps of in the wild traffic per node with no issue, and we went with an average uptime of 25 seconds, withouth this MR, so we'll probably continue building the patched version, but let us know if any more feedback is required

nigoroll:
@xdecock I find this equally frustrating as you, but I hope we will implement some changes to improve the situation for the project.

Dridi:
FWIW when I approved this pull request 51e087631d7af2f8aadfc54c2ddcd9b3ec6342d9 was not present and at first glance it makes me uncomfortable.

Also I pushed some test coverage to this pull request that has been lost with subsequent force-pushes. That will teach me not to add commit hashes in comments simply because they appear in the ticket history at one point.

Dridi:
Found it in my reflog :D 58bcad66adf9273123ac03c88562c1ddea241d6b

nigoroll:
@Dridi if you are ok with the PR except for  51e087631d7af2f8aadfc54c2ddcd9b3ec6342d9 then at least this means some progress

But what is the matter with the optional locking? _at first glance it makes me uncomfortable_ does not sound like a strong argument. How else would you serialize concurrent `PRIV_TOP` access?

nigoroll:
@Dridi i have added your vtc and force-pushed

Dridi:
> But what is the matter with the optional locking? _at first glance it makes me uncomfortable_ does not sound like a strong argument. How else would you serialize concurrent `PRIV_TOP` access?

It means only looking at the diff (first glance), not doing a proper review.

Dridi:
From the commit message:

> Alternatives considered:
>
> * use a varnish lock. Possible, but would sound like additional overhead

I think that in general if we can help it we should always have a lock class, especially if we end up with `make witness` in CI (#3072).

nigoroll:
@Dridi yes, you got a point, but I am not sure it's worth the complications it brings with it:

I looked at using a varnish lock again and it really appears to me that doing so would add quite a bit of code:
* as core code would own the vsc, it would make sense to also have a VRT function to add the locking to reqtop.
* that in turn calls for cleanup code in VRT
* which calls for some version of `VCL_TaskLeave` specifically for reqtop

You might argue that vmods using the lock could implement all of this and you would be right, but this would add more friction and room for error.

As, for the foreseeable future, there will probably be exactly two users of this lock, I think that simply having the `reqtop->mtx` point at some vmod-owned pthread mtx at `esi_level == 0` is the simplest, easiest most robust option at this point.

What do you think?

Dridi:
I think I really need to highlight the "in general if we can help it" part of my previous statement, I'm not suggesting to mindlessly Lck-all-the-things :)

However I haven't reviewed 6654868 yet, but I'd like to get the rest merged, the part I confidently reviewed eons ago. We can then review the priv_top mutex question in a separate pull request and focus that discussion on concurrent sub-requests capabilities for VMODs. The first thing I would want to see is two distinct VDPs competing for a "concurrent" priv_top, ideally with coverage in vmod-debug.

I can't commit to reviewing much just yet, but I would be happy to make progress. That is, if @bsdphk is ok with this plan.

nigoroll:
@bsdphk are you ok merging _but_ the PRIV_TOP lock commit?

nigoroll:
I have force-pushed without locking change which, apparently, needs more discussion

nigoroll:
@bsdphk this needs your review and it seems due to #3019 we should really tick it off.

nigoroll:
continued in #3117 
but I would like to keep this one open until a decision has been made.

nigoroll:
superseded by #3117

PR #3014, opened 2019-06-06
Add "error" state to backend fetch and response

andrewwiik:
This adds a new return "error" to "backend_fetch" and "backend_response" that will then handle over control to "backend_error". Currently abandon and fail in these methods currently backs totally out of backend. This error return also allows an optional reason and status code to be provided, just like the synth return except it does not require them, they are optional. 

Look at v00062.vtc for all the possible use syntax for the new error return.


bin/varnishd/cache/cache_fetch.c
```diff
@@ -384,10 +387,13 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 
 	VCL_backend_response_method(bo->vcl, wrk, NULL, bo, NULL);
 
-	if (wrk->handling == VCL_RET_ABANDON || wrk->handling == VCL_RET_FAIL) {
+	if (wrk->handling == VCL_RET_ABANDON || wrk->handling == VCL_RET_FAIL || wrk->handling == VCL_RET_ERROR) {
```
Dridi: This line is too long and needs to be broken down.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -719,7 +725,11 @@ vbf_stp_error(struct worker *wrk, struct busyobj *bo)
 	// XXX: reset all beresp flags ?
 
 	HTTP_Setup(bo->beresp, bo->ws, bo->vsl, SLT_BerespMethod);
-	http_PutResponse(bo->beresp, "HTTP/1.1", 503, "Backend fetch failed");
+	if (bo->err_code > 0)
+		http_PutResponse(bo->beresp, "HTTP/1.1", bo->err_code, bo->err_reason);
+	else
+		http_PutResponse(bo->beresp, "HTTP/1.1", 503, "Backend fetch failed");
```
Dridi: It might be preferable to set `bo->err_{code,reason}` upon failure to avoid this branch.

doc/graphviz/cache_fetch.dot
```diff
@@ -39,16 +39,18 @@ digraph cache_fetch {
...
 		]
+		v_b_f:error:s -> v_b_e
 		v_b_f:fetch:s -> v_b_hdrs [style=bold]
 		v_b_hdrs [ label="send bereq,\nread beresp (headers)"]
 		v_b_hdrs -> v_b_r [style=bold]
 		v_b_hdrs -> v_b_e
 		v_b_r [
 			    shape=record
-			    label="{vbf_stp_startfetch:|{vcl_backend_response\{\}|{bereq.*|beresp.*}}|{fail|{retry|{<max>max?|<retry>ok?}}|abandon|{deliver or pass|{<fetch_304>304?|<non_304>other?}}}}"
+			    label="{vbf_stp_startfetch:|{vcl_backend_response\{\}|{bereq.*|beresp.*}}|{<error>error|{retry|{<max>max?|<retry>ok?}}|abandon|{deliver or pass|{<fetch_304>304?|<non_304>other?}}}}"
```
Dridi: Should `return (fail)` disappear from here?

doc/sphinx/users-guide/vcl-built-in-subs.rst
```diff
@@ -356,6 +356,10 @@ The `vcl_backend_fetch` subroutine may terminate with calling
   ``fetch``
     Fetch the object from the backend.
 
+  ``error(status code, reason)``
+    Abandon the backend request and control will be passed to
```
Dridi: I'm not sure we want to use the word "Abandon" here, it's confusing since we already have `return (abandon)`. We can use the same wording as `synth` instead.

doc/sphinx/users-guide/vcl-built-in-subs.rst
```diff
@@ -417,6 +421,10 @@ The `vcl_backend_response` subroutine may terminate with calling
     lookups hitting this object will be turned into passed transactions,
     as if ``vcl_recv`` had returned ``pass``.
 
+    ``error(status code, reason)``
+    Abandon the backend request and control will be passed to
```
Dridi: I'm not sure we want to use the word "Abandon" here, it's confusing since we already have `return (abandon)`. We can use the same wording as `synth` instead.

bin/varnishd/cache/cache_fetch.c
andrewwiik: This would cause more code changes as everywhere we call the error state that isn't explicitly  from return error we would have to add these lines before it.

doc/graphviz/cache_fetch.dot
andrewwiik: No, because breaking changes are bad, and fail was made as a standard return for everywhere so it should definitely stay.

doc/sphinx/users-guide/vcl-built-in-subs.rst
andrewwiik: Got it, check the new force push

doc/sphinx/users-guide/vcl-built-in-subs.rst
andrewwiik: Got it, check the new force push

doc/graphviz/cache_fetch.dot
andrewwiik: Whoops, fixed in latest force push

bin/varnishd/cache/cache_fetch.c
andrewwiik: @rezan  brought up a point saying if we change this in the struct then we lose the original user intent.

Dridi: Don't we have the handling for that?

On the same topic I would rather not execute `wrk->stats->fetch_failed++;` a few lines above if the aforementioned intent was to willingly go to `vcl_backend_error` and not a failed attempt to contact a backend.

andrewwiik: So you are saying if you use return error you don't want the fetch failed counter to be bumped?
11:41 AM is their a different counter that should be bumped instead?

Dridi: > So you are saying if you use return error you don't want the fetch failed counter to be bumped?

If I `return (error)` from either `vcl_backend_{fetch,response}` then there was either no fetch, or one succeeded. Short answer, yes.

> Is their a different counter that should be bumped instead?

Either none off the top of my head, or the other way around: we might want to have an equivalent to `MAIN.s_synth`.

andrewwiik: I would make the counter backend_error so people don't confused the counter being used for every error everywhere.

andrewwiik: I added a force push which makes the fetch_failed counter not increment when the call to ```vbf_stp_error``` was the result of a return(error) in VCL.

andrewwiik:
I have implemented the discussed change to keep using VRT_synth instead of taking the VRT_error namespace so I think we are good for a final review and merge.

bin/varnishd/cache/cache_vrt.c
```diff
@@ -55,9 +55,19 @@ VRT_synth(VRT_CTX, VCL_INT code, VCL_STRING reason)
...
+
+	if (ctx->req == NULL) {
+		CHECK_OBJ_NOTNULL(ctx->bo, BUSYOBJ_MAGIC);
+		ctx->bo->err_code = (uint16_t)code;
+		ctx->bo->err_reason = reason ? reason
+		    : http_Status2Reason(ctx->bo->err_code % 1000, NULL);
+		return;
+	}
+
+	CHECK_OBJ_NOTNULL(ctx->req, REQ_MAGIC);
```
Dridi: `CHECK_OBJ()` is enough, we have a null check line 62 already.

doc/sphinx/users-guide/vcl-built-in-subs.rst
```diff
@@ -356,6 +356,11 @@ The `vcl_backend_fetch` subroutine may terminate with calling
   ``fetch``
     Fetch the object from the backend.
 
+  ``error(status code, reason)``
+    Transition to :ref:`vcl_backend_error` with ``resp.status`` and
+    ``resp.reason`` being preset to the arguments of ``error()`` if
```
Dridi: We are dealing with `beresp.reason` in this context.

doc/sphinx/users-guide/vcl-built-in-subs.rst
```diff
@@ -356,6 +356,11 @@ The `vcl_backend_fetch` subroutine may terminate with calling
   ``fetch``
     Fetch the object from the backend.
 
+  ``error(status code, reason)``
+    Transition to :ref:`vcl_backend_error` with ``resp.status`` and
```
Dridi: We are dealing with `beresp.status` in this context.

doc/sphinx/users-guide/vcl-built-in-subs.rst
```diff
@@ -417,6 +422,11 @@ The `vcl_backend_response` subroutine may terminate with calling
     lookups hitting this object will be turned into passed transactions,
     as if ``vcl_recv`` had returned ``pass``.
 
+  ``error(status code, reason)``
+    Transition to :ref:`vcl_backend_error` with ``resp.status`` and
```
Dridi: We are dealing with `beresp.status` in this context.

doc/sphinx/users-guide/vcl-built-in-subs.rst
```diff
@@ -417,6 +422,11 @@ The `vcl_backend_response` subroutine may terminate with calling
...
+    ``resp.reason`` being preset to the arguments of ``error()`` if
```
Dridi: We are dealing with `beresp.reason` in this context.

include/vrt.h
```diff
@@ -431,6 +431,7 @@ const char *VRT_regsub(VRT_CTX, int all, const char *, void *, const char *);
 VCL_VOID VRT_ban_string(VRT_CTX, VCL_STRING);
 VCL_INT VRT_purge(VRT_CTX, VCL_DURATION, VCL_DURATION, VCL_DURATION);
 VCL_VOID VRT_synth(VRT_CTX, VCL_INT, VCL_STRING);
+VCL_VOID VRT_error(VRT_CTX, VCL_INT, VCL_STRING);
```
Dridi: This should go away.

bin/varnishd/cache/cache_vrt.c
andrewwiik: Done

doc/sphinx/users-guide/vcl-built-in-subs.rst
andrewwiik: Done

doc/sphinx/users-guide/vcl-built-in-subs.rst
andrewwiik: Done

include/vrt.h
andrewwiik: Done

doc/sphinx/users-guide/vcl-built-in-subs.rst
andrewwiik: Done

Dridi:
@bsdphk can you have a look? I think I got rid of the noise and the current patch should be ready for a more comprehensive review. I will try to have a closer look myself tomorrow.

Dridi:
Nothing else stands out for me, I would merge.

PR #3017, opened 2019-06-21
Directors can fail with a specific status

Dridi:
This merely adds test coverage and shows that since #3014 it became possible as a side effect.

I think that complex directors could benefit from going to `vcl_backend_error` with a 500 "Internal Server Error" status, this isn't meant for vmod-directors. Therefore, we should rather have coverage and not decide that core code should prevent this behavior.

Initially submitted as part of #3016.

### DIFF
--- a/bin/varnishtest/tests/d00040.vtc
+++ b/bin/varnishtest/tests/d00040.vtc
@@ -1,0 +1,23 @@
+varnishtest "Director internal error"
+
+varnish v1 -vcl {
+	import debug;
+
+	backend be { .host = "${bad_backend}"; }
+
+	sub vcl_backend_fetch {
+		set bereq.backend = debug.fail_backend();
+	}
+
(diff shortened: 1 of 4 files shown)


bsdphk:
Ok, here is my decision, relevant for #3014 as well:

We will not change the current 503 behavior in VCL4.1

We will introduce new behavior in VCL4.2

We can add a facility like pythons "from future import MAGICWORD" to enable the new behavior in branches which do not support VCL4.2. The exact syntax we use is TBD.

nigoroll:
should this one be surrendered in place of the agreement found at the VDD?

nigoroll:
@Dridi should this be surrendered in favor of #3113 ?

Dridi:
I will make a note of revisiting this once #3017 is done.

Dridi:
Closing in favor of #3016.

PR #3018, opened 2019-06-27
Ensure that we always log h2 idle_send_timeout

Dridi:
Revisiting #2980 recently compelled me to fix this.


bsdphk:
@daghf will wave this through when he is happy with it.

Dridi:
Bugwash consensus:

- explain the magic NAN value where it is used
- fix copy-pasta mistake in the test case
- @daghf to review once updated

daghf:
LGTM

PR #3020, opened 2019-07-03
Add branch prediction hint macros

fwsGonzo:
... and cover some of the most common cases such as asserts and VCC END_.

### DIFF
--- a/bin/varnishd/cache/cache_varnishd.h
+++ b/bin/varnishd/cache/cache_varnishd.h
@@ -477,7 +477,7 @@ void SMP_Ready(void);
 #endif
 
 #define FEATURE(x)	COM_FEATURE(cache_param->feature_bits, x)
-#define DO_DEBUG(x)	COM_DO_DEBUG(cache_param->debug_bits, x)
+#define DO_DEBUG(x)	UNLIKELY(COM_DO_DEBUG(cache_param->debug_bits, x))
 
 #define DSL(debug_bit, id, ...)					\
 	do {							\
--- a/include/vas.h
+++ b/include/vas.h
@@ -40,16 +40,6 @@
 
 #include "verrno.h"
 
-#ifdef __GNUC__
-// GCC/Clang/ICC
-#define LIKELY(x)   __builtin_expect(!!(x), 1)
-#define UNLIKELY(x) __builtin_expect(!!(x), 0)
-#else
-// TCC, Sun CC etc.
-#define LIKELY(x)   (x)
-#define UNLIKELY(x) (x)
-#endif
-
 enum vas_e {
 	VAS_WRONG,
 	VAS_MISSING,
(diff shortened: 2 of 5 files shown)


Dridi:
I took the liberty of moving the macro definitions to `vdef.h` (ee2283cca781b58e0ccde4ae8657e931c51f632b).

fwsGonzo:
Any reason why its not passing checks?

Dridi:
The test suite relies on an external domain for resolution, and it currently doesn't meet or expectations to fail a VCL compilation where a domain resolves to more than 1 IPv4 + 1 IPv6.

Nothing to do with your change.

Dridi:
So review-wise, I'm not sure I'm too happy with the feature detection, I think relying solely on `__GNUC__` is probably not enough.

fwsGonzo:
It is a rather crude detection, but it should work for most compilers that are around. The alternative is to detect using a test program in the configure script.

fgsch:
Perhaps you could explain why you are doing this change, why is a good idea, etc?

slimhazard:
In the spirit of @fgsch's question, do you have measurements or other evidence that the hints significantly improve branch predictions, with noticeable effect on Varnish? I think it quite likely that the branches are predicted all the time, or close to it, without the hints.

Assertion failures, by definition, don't happen, or else Varnish crashes. If the branch goes the other way just once, the process stops, so it's either predicted or we're dead anyway. I assume that the guess is right every time, don't see how it can be any other way.

The VCC case is more interesting, because of all the ways VCL can be used. But I would reason that since VCL authors strive to avoid VCL failure, branch prediction will correctly guess the non-fail case almost every time. (If someone is deliberately allowing VCL failure to happen in hard-to-predict ways, I'd say they're doing it wrong, and Varnish doesn't need to be optimized for that.) There may be a pipeline stall when VCL failure does happen, but that would be the case with or without the hints.

Without the hints, there may be mispredictions for a time after process startup, before the patterns become obvious. But does that make much difference?

Data to the contrary could prove me wrong, of course (and there's nothing wrong with that).

slimhazard:
I was only thinking about this in terms of performance optimization, but @nigoroll pointed out to me that the hints may reduce opportunities for Meltdown-style attacks. Without the hints, there may be mispredictions when branches are encountered the first time, before it becomes clear that they're always going one way. But with the hints, branches will be predicted right from the start.

There will still be mispredictions, with the Meltdown-ish risks, on VCL failure and assertion errors. So they wouldn't go away entirely, but presumably there will be fewer.

bsdphk:
I think my first requirement for this is some kind of measurement that shows it helps in some way.

nigoroll:
@fwsGonzo can we put this aside (close) until you had a chance to address the questions?

fwsGonzo:
Sure, I have to take one thing at a time here, no time for this now :) :+1:

nigoroll:
#3836 brought this back onto the table

PR #3021, opened 2019-07-03
Add a VRT_HashStrands32 utility

Dridi:
This deduplicates logic between the two hashing directors, and in the
case of vmod-shard adds a missing check for empty components in the
strands that should be skipped.

Not happy with the name, it was harder than I anticipated. Suggestions
welcome.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -37,8 +37,10 @@
 #include "vav.h"
 #include "vcl.h"
 #include "vct.h"
+#include "vend.h"
 #include "vrt_obj.h"
 #include "vsa.h"
+#include "vsha256.h"
 #include "vtcp.h"
 #include "vtim.h"
 #include "vcc_interface.h"
@@ -248,6 +250,35 @@ VRT_Strands2Bool(VCL_STRANDS s)
 		if (s->p[i] != NULL)
 			return (1);
 	return (0);
+}
+
+/*--------------------------------------------------------------------
+ * Hash a STRANDS
+ */
+
+uint32_t
+VRT_HashStrands32(VCL_STRANDS s)
+{
+	struct VSHA256Context sha_ctx;
+	unsigned char sha256[VSHA256_LEN];
+	const char *p;
+	int i;
+
+	AN(s);
+	VSHA256_Init(&sha_ctx);
+	for (i = 0; i < s->n; i++) {
+		p = s->p[i];
+		if (p != NULL && *p != '\0')
+			VSHA256_Update(&sha_ctx, p, strlen(p));
+	}
+	VSHA256_Final(sha256, &sha_ctx);
+
+	/* NB: for some reason vmod_director's shard director specifically
+	 * relied on little-endian decoding of the last 4 octets. In order
+	 * to maintain a stable hash function to share across consumers we
+	 * need to stick to that.
+	 */
+	return (vle32dec(sha256 + VSHA256_LEN - 4));
 }
 
 /*--------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


bsdphk:
Looks good.

bsdphk:
As noted in bugwash: The shard director uses the last 32 bits of the sha256, and this may affect test-cases.

There are no cryptographic/entropic reasons to use the last 32 bits, so we would end up cleaner by changing shard to use the first 32 like everybody else, but that entails adjusting the testcases.

I do not believe changing shartd this way will have consequences for sites using shard in production.

nigoroll:
Yes, the decision on which bits to use was completely arbitrary, as marked by a comment. The significance of _like everybody else_ is a bit unclear to me.
If we changed this, we would momentarily break any persistence ensured by the shard director when deployed.

simonvik:
Depending on setup it might hurt when all requests gets reshuffled when cache doesn't get hit anymore.

Dridi:
> If we changed this, we would momentarily break any persistence ensured by the shard director when deployed.

Please define persistence. When I checked for this pull request I didn't see any kind of persistence that would survive the restart of the child process. If we change the hashing that means that it will only "affect" you after the new VMOD is installed, in which case you need to restart `varnishd` entirely (which encompasses a child restart) otherwise VMOD ABI checks will prevent the new vmod-directors to even be loaded by your next `vcl.load`.

So to my knowledge there is no reason to start with a hash algorithm and change during runtime. Please let me know if I'm mistaken, otherwise I will update the test suite accordingly.

nigoroll:
@Dridi the shard director can be used to assign requests or cache objects to backend servers or other varnish servers. Just to name two examples, one may assign session id cookies to backend servers or cache objects (hashes) to varnish servers in a cluster (see for example https://code.uplex.de/uplex-varnish/libvmod-cluster/blob/master/README.rst )
For such scenarios, redeploying a newer version of the vmod with a different choice of bits taken from the sha256 hash would chance that association. Because shard hashing is stateless, it persists restarts.
Last time we made such a change, we provided the option for keeping the existing behavior, so that users could decouple deployment of a new varnish version from changing this aspect.
I do not know if there are any show stopper cases, but in cache clusters, a rolling update will become problematic if the hashing is changed. Basically, this change would force users to update all of the cluster in as short a timeframe as possible.

Dridi:
Ok, so then we should align hash director hashing with shard, and we can both fix shard's bug in strands handling and de-duplicate the logic. Too bad it has to be the complicated option, but well, my objective is not to break anyone's setup.

Thanks for the clarification.

Dridi:
Do we all agree that "breaking" the hash director hashing is fine since it doesn't maintain consistency depending on the cluster health?

nigoroll:
@Dridi 
* where is the bug in shard? I see not reference to a bug before you mentioned it now?
* in principle, users of the hash director could see similar issues as with shard, but my basic assumption would be that anyone going for this kind of load balancing in any serious way would use shard anyway because hash will not provide the same level of stability in backend assignments as consistent hashing in shard. So, in short, I think breaking hash instead of shard is the better option, yes

Dridi:
@nigoroll maybe not a bug in hindsight, but a bit of polish is applied by this patch in the sense that the shard director will hash empty components of a strands. This has no consequence, because unlike the `hash_data` function in VCL the null-terminating character isn't taken into account.

I updated the pull request accordingly.

nigoroll:
lgtm except for the merge conflict, but that's trivial

Dridi:
I just noticed that this was approved while I was away. I will solve the conflict and merge the pull request (in other words last call for yelling NO).

PR #3022, opened 2019-07-08
Turn the vmod-blob structure upside-down

Dridi:
I found the vmod-code harder to browse, a bit unfamiliar in term of code style or organization.

Hopefully this change won't be controversial, and in summary this is mostly about hiding encoding implementation details from the vmod itself, and have each encoding responsible for itself.

I found something fishy in hex encoding, but needed to clear things up a bit before I could have a proper look (I will follow up later for 0365ade5cd66b399904b16869c79e24d260ae387).

Thankfully vmod-blob has excellent test coverage (more than 90% of its code) so I'm not too worried about regressions. Please see individual commits for details as this was an incremental refactoring.


bsdphk:
Bugwash consensus yesterday:  Due to its size and nature, this PR is a bit hard to deal with as a bugwash item.

Suggested way forward:  @dridi gets a deep review from somebody else, if possible by original author, before bugwash makes decision.

lib/libvmod_blob/base64.c
```diff
@@ -31,41 +31,176 @@
 #include "vdef.h"
 #include "vrt.h"
 #include "vas.h"
+#include "miniobj.h"
 
-#include "base64.h"
```
slimhazard: IMO it's customary to include tables this large from a header file. That breaks encapsulation, but I think it's warranted for tables of numbers that now dominate the C source.

lib/libvmod_blob/hex.c
```diff
@@ -99,22 +99,35 @@ hex_encode(const enum encoding enc, const enum case_e kase,
 		*p++ = alphabet[in[i] & 0x0f];
 	}
 
+	/* XXX: the length arithmetics here are a bit complicated for such
+	 * a simple problem. It looks like the estimated encoded length
+	 * makes room for a null character, which makes sense for string
+	 * encoding. But null characters are set by the caller as documented.
+	 *
```
slimhazard: This XXX-comment is strangely placed, since it doesn't seem to be about the code that it's close to.

Don't see what's complicated or wrong. Two hex chars to a byte. hex_encode_l is 2*nbytes + 1, with the one for the terminating null. hex_decode_l is nbytes/2, rounding up (that's the odd-numbered hex string case).

lib/libvmod_blob/hex.c
```diff
@@ -99,22 +99,35 @@ hex_encode(const enum encoding enc, const enum case_e kase,
...
+	 * The result is a confusing length computation where:
+	 *
+	 *     hex_decode_l(hex_encode_l(len)) != len
+	 *
```
slimhazard: Nothing about hex or anything else in the VMOD requires this identity -- the other schemes have differing length estimates for encoding and decoding as well.

Consider especially URL, where a byte may not be changed at all, or may be encoded with three characters (%xx). The encoding length estimation makes the pessimistic assumption that *all* bytes have to be hex-encoded (nbytes*3). The decoding length estimation pessimistically assumes that *none* of them are, so that the decoding is no shorter than the encoding.

lib/libvmod_blob/hex.c
```diff
@@ -99,22 +99,35 @@ hex_encode(const enum encoding enc, const enum case_e kase,
...
+	 * Needs a proper review of "half byte" decoding. Should it be
+	 * supported in the first place?
+	 */
```
slimhazard: What's the alternative? Fail for hex encodings with an odd length? Hex strings like "123" have a common and intuitive interpretation (0x0123 == 291 decimal).

lib/libvmod_blob/hex.c
```diff
@@ -133,13 +146,15 @@ hex_decode(const enum encoding dec, char *restrict const buf,
...
+	if (inlen != -1 && len > inlen)
+		len = inlen;
 
-	if (((len+1) >> 1) > buflen) {
+	if (hex_decode_l(len) > buflen) {
 		errno = ENOMEM;
 		return (-1);
 	}
+
+	/* XXX: prepended extra nibble, not appended! */
```
slimhazard: Of course. We expect "3ab" to be decoded as 0x03ab, not 0x3ab0. This is documented, and changing it would lead to regressions, fail current VTC tests, and IMO would violate common intuitions.

lib/libvmod_blob/hex.c
```diff
@@ -161,6 +176,17 @@ hex_decode(const enum encoding dec, char *restrict const buf,
 		}
 		extranib = *s;
 	}
+
+	/* XXX: no extra nible check for the last string? */
```
slimhazard: No. If there is an extranib, and when neither of *s and *(s+1) are '\0' the last time through the while loop, then after s+=2 in the while loop, *s!='\0' and *(s+1)=='\0'. Then since if(extranib) is true, s++ gets executed. After which *s=='\0', the string is read all the way through, and the while loop is not entered.

lib/libvmod_blob/vmod_blob.h
```diff
@@ -52,8 +55,7 @@ enum case_e {
 /*
  * Length estimate interface
  */
-typedef
-size_t  len_f(size_t);
+typedef size_t len_f(size_t);
 
 /*
  * General interface for an encoder: encode the data at in of length inlen
```
slimhazard: The doc for ``enc`` in the comment becomes obsolete (no encoding enum).

lib/libvmod_blob/vmod_blob.h
```diff
@@ -80,10 +80,8 @@ size_t  len_f(size_t);
...
  */
-typedef
-ssize_t encode_f(const enum encoding enc, const enum case_e kase,
-		 char *restrict const buf, const size_t buflen,
-		 const char *restrict const in, const size_t inlen);
+typedef ssize_t encode_f(BLOB_CODEC, enum case_e kase, char *buf, size_t buflen,
+    const char *in, size_t inlen);
 
 /*
  * General interface for a decoder: decode the concatenation of strings
```
slimhazard: ``dec`` is also obsolete here.

lib/libvmod_blob/vmod_blob.h
```diff
@@ -107,34 +104,29 @@ ssize_t encode_f(const enum encoding enc, const enum case_e kase,
  *    a static constant empty BLOB
  * otherwise, the number of bytes written
  */
-typedef ssize_t decode_f(const enum encoding dec, char *restrict const buf,
-    const size_t buflen, const ssize_t inlen, VCL_STRANDS strings);
+typedef ssize_t decode_f(BLOB_CODEC, char *buf, size_t buflen, ssize_t inlen,
+    VCL_STRANDS strings);
+
+struct vmod_blob_codec {
```
slimhazard: "codec" is indeed a better name.

Consider adding a magic number to struct vmod_blob_codec, then it could be checked with CHECK_OBJ (currently just with AN).

lib/libvmod_blob/vmod_blob.c
```diff
@@ -246,54 +222,61 @@ vmod_blob_get(VRT_CTX, struct vmod_blob_blob *b)
...
-		return "";
+		return ("");
 	if (kase == DEFAULT)
 		kase = LOWER;
 
-	if (b->encoding[enc][kase] == NULL) {
-		AZ(pthread_mutex_lock(&b->lock));
-		if (b->encoding[enc][kase] == NULL) {
-			ssize_t len = func[enc].encode_l(b->blob.len);
+	if (b->encoding[idx][kase] != NULL)
```
slimhazard: Checking != NULL outside of the lock, and avoiding the double-check under the lock, is a good move.

lib/libvmod_blob/vmod_blob.c
```diff
@@ -31,11 +31,21 @@
 
 #include "cache/cache.h"
 
-#include "vcc_if.h"
 #include "vmod_blob.h"
 
 #define VMOD_BLOB_TYPE 0xfade4faa
```
slimhazard: Note: this comment is *not* about the PR, because the same is true in master.

Looking at the code I noticed that the magic numbers used in vmod_blob.c are 0xfade4faa, 0xfade4fa9 and 0xfade4fa0. They're too similar, and it defeats the purpose. If a struct is in fact corrupted at just the right byte, the CHECK_OBJ test might pass where it shouldn't.

lib/libvmod_blob/vmod_blob.c
```diff
@@ -31,11 +31,21 @@
...
 
+static const struct vmod_blob_codec *codecs[] = {
```
slimhazard: Moving the codecs into the various sources, just calling their methods and properties from this table, is a good improvement for encapsulation.

lib/libvmod_blob/vmod_blob.c
```diff
@@ -246,54 +222,61 @@ vmod_blob_get(VRT_CTX, struct vmod_blob_blob *b)
...
+				s = b->encoding[idx][kase];
+				len = enc->encode(enc, kase, s, len,
+				    b->blob.blob, b->blob.len);
+				assert(len >= 0);
+				if (len == 0) {
+					free(s);
+					b->encoding[idx][kase] = empty;
 				}
+				else
+					s[len] = '\0'; /* XXX: safe? */
```
slimhazard: Yes. ``len`` (returned from ``encode()``) is by contract the number of bytes written, not counting a terminating null byte, which is added by the caller. So if the encoder wrote "f00f", then len=4, and the '\0' is written at index 4. malloc() above ensures that enough bytes were allocated.

Just in case, I tested with ``len-1``, and the tests failed (base64 encodings were shortened by one character).

lib/libvmod_blob/base64.c
```diff
@@ -35,8 +35,6 @@
```
slimhazard: Responding to the commit message here.

const is more of a hint to programmers than to compilers, so there's no error if it's removed. But why remove the hints? The starting addresses and lengths of the input and output buffers are never changed. The contents of the input buffer are never changed. Arguably const is unnecessary for the non-pointer parameters, but IMO const matters more for the pointer parameters.

My impression from the Flexelint commits in the project is that lint favors the use of const where it is warranted. The criteria may be more than just "qualify with const if you can" -- I don't have access to Flexelint, so I can't see for myself. Maybe someone who does could run Flexelint with and without the const qualifiers -- no need to take out the const's, if Flexelint tells us to put them back in.

restrict does not guarantee non-overlapping regions, nor is it intended to. It's a contract that the caller is required to fulfill, hinting the compiler that it may take advantage of the assumption. It's been years now, but I believe I remember checking that the caller code does indeed satisfy the constraint (however I wouldn't rule out an error).

I don't have numbers about the effect of restrict and don't intend to, it would be too much work. I have gotten in the habit of using restrict when the opportunity presents, after looking at code generated with and without it, and seeing that it can lead to shorter code, because the compiler does not have to account for the case of aliasing. (And after reading reports, with numbers, indicating that it can lead to faster code.)

A consideration with the VMOD was that if it used at all, it may be hot code, churning through memory a byte at a time. Just encoding the 32 bytes in req.hash isn't all that much, but if it's happening in many requests at once, it adds up. If someone uses the trick to implement URL encoding with the VMOD, remember that URLs can be up to kilobytes long. And the VMOD is meant to be general purpose for BLOBs, which can be arbitrarily long.

So IMO a hint to the compiler that it can save some work is worth it. Yes it makes the prototypes longer, and they look a little weird, but that aesthetic is less important.

lib/libvmod_blob/hex.c
slimhazard: Just want to add that there are tests of odd-length hex encodings in the VTCs, all passing.

lib/libvmod_blob/base64.c
nigoroll: +1 on `const`/`restrict`, IMHO it makes no sense to remove correct attributes with the added benefit of producing more efficient code and implicitly documenting the code.

bsdphk:
We're closing this ticket for now, it can be reopened when the PR is reviewed and ready

Dridi:
I was trying to help bring consistency back and my apologies for removing `restrict` qualifiers that @nigoroll proved useful. I did my best to make as small as possible incremental commits, but changes had to be incrementally applied to all codecs at once. I know that +468 -432 is a lot for a single PR, but it's spread over 10 commits (and yes I know they are quite uneven).

I don't think the code is bad, but I definitely think it can be improved in terms of consistency with the rest of the code base. At the very least I'm happy to see that @slimhazard sees where I was trying to get us, feel free to push more incremental changes than I managed without breaking the optimizations I killed in the process.

slimhazard:
My initial defensive reaction was my own problem, but I did want to relate that a closer look got to me to snap out of it, which IMO is a good sign for the PR. Still in favor of what the PR was doing overall (despite picking at details), so I hope it makes its way back.

Dridi:
Did you attend the bugwash today? I don't understand the reason why it got closed instead of ping'ed if there is interest in moving forward. I have two months of catching up to do and between this ticket and others (#3051, #3057 etc, I will reply there) I don't feel an incentive to dive into vmod-blob again.

gquintard:
the takeaway was: figure it out between yourselves, come back when you are both happy with it.

The PR was closed to avoid noise, but we still want it in :-)

Dridi:
In this case I will make it more incremental and try to give @slimhazard something satisfying to review, I have an idea how to proceed (and I will fight my urge to prune `restrict` qualifiers=).

Dridi:
I pushed 92ad90fe9 and 1e1bef8fe (and unfortunately 8caa56754 too) directly to master. Hopefully we all agree that this is not controversial and will remove a bit of noise in the preparation for a second take on this change.

Dridi:
Same for 1b1e7892ecb00ddf35b305c439de8f3a5aeb109c.

PR #3026, opened 2019-07-22
Docfix: varnishlog and varnishncsa exits from SIGHUP in the foreground

dmatetelki:
Be a bit more elaborate on the expected behaviour


dmatetelki:
`make distcheck` worked. In `travis` and locally too.

fgsch:
LGTM 👍

PR #3029, opened 2019-08-01
Document 1 second TTL on (synthetic) error objects

hermunn:
The behavior described here is old (see cache_fetch.c), but has not been cleanly documented in the docs.

I am making this pull request (instead of committing directly) so that @carlosabalde can have a look first.


carlosabalde:
LGTM. Thanks! :)

PR #3032, opened 2019-08-02
Fix vcl.show for labels

mbgrydeland:
When executing vcl.show for labels, show the VCL the label points to
instead of asserting.


bsdphk:
OK

PR #3033, opened 2019-08-05
Master vsm

mbgrydeland:
Two patches to reduce the amount of work generated by updates to the VSM structure.

### DIFF
--- a/include/vapi/vsm.h
+++ b/include/vapi/vsm.h
@@ -44,7 +44,7 @@ struct vsm;
 
 struct vsm_fantom {
 	uintptr_t		priv;		/* VSM private */
-	struct VSM_chunk	*chunk;
+	uintptr_t		priv2;		/* VSM private */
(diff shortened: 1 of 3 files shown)


bsdphk:
I'm OK'ing the VSC patch.

Not sure why exposed were per point originally and dont see why it should be now.

Holding fire on the VSM patch until we have some numbers.

n4n0r:
Hi,
Following a quick discussion on the chat with hermunn, I am posting a comment here.

We have upgraded from 6.0.0 to 6.2.0 on a with a heavy config, and have encountered 3 panics when reloading the VCL. Panics: https://gist.github.com/n4n0r/bc4e31193f3e1deb33de61126f8706e9
We have not been able to reproduce this on other boxes, or after a reboot, but are worried this would happen again in the future.

We wonder if this might be linked to the subject discussed here.

PR #3034, opened 2019-08-07
Add a configure switch to skip backtrace() calls

gquintard:
musl libc doesn't doesn't support backtrace/backtrace_symbols, but there
are stubs for them, so varnish will compile/link/run, but will crash
once they are called.

There are solutions to keep the features but they all involve extra libraries,
so I went for the big hammer first: just not calling
backtrace/backtrace_symbols.

more context: https://redmine.alpinelinux.org/issues/5079

### DIFF
--- a/bin/varnishd/cache/cache_panic.c
+++ b/bin/varnishd/cache/cache_panic.c
@@ -611,6 +611,7 @@ pan_sess(struct vsb *vsb, const struct sess *sp)
 
 #define BACKTRACE_LEVELS	10
 
+#ifdef HAVE_BACKTRACE
 static void
 pan_backtrace(struct vsb *vsb)
 {
@@ -649,6 +650,7 @@ pan_backtrace(struct vsb *vsb)
 	}
 	VSB_indent(vsb, -2);
 }
+#endif
 
 #ifdef HAVE_PTHREAD_GETATTR_NP
 static void
@@ -740,7 +742,11 @@ pan_ic(const char *func, const char *file, int line, const char *cond,
 	VSB_printf(pan_vsb, "now = %f (mono), %f (real)\n",
 	    VTIM_mono(), VTIM_real());
 
+#ifdef HAVE_BACKTRACE
 	pan_backtrace(pan_vsb);
+#else
+	VSB_printf(pan_vsb, "Backtrace support disabled at compile time");
+#endif
 
 	if (err)
 		VSB_printf(pan_vsb, "errno = %d (%s)\n", err, vstrerror(err));
--- a/configure.ac
+++ b/configure.ac
@@ -343,9 +343,18 @@ esac
 AC_SUBST(JEMALLOC_LDADD)
 
 AC_CHECK_FUNCS([setproctitle])
-AC_SEARCH_LIBS(backtrace, [execinfo], [], [
-   AC_MSG_ERROR([Could not find backtrace() support])
-])
+
+AC_ARG_WITH([backtrace],
+	    [AS_HELP_STRING([--without-backtrace], [Disable backtrace() ])])
+
+if test "$with_backtrace" != no; then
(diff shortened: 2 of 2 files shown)


fgsch:
Grepping musl I can't find any stubs. 

Is there a way to detect if this build is using musl instead of adding a configure option though?

gquintard:
that would be too easy: https://www.openwall.com/lists/musl/2013/02/08/9

the other option is to compile and run a test program during configure, it just didn't seem worth the hassle.

fgsch:
I'd prefer that option then. 

Users of musl will get a working binary without having to know about a special configure parameter.

gquintard:
I feared you might say that :-) Should I leave the parameter in to force the behavior?

fgsch:
My personal opinion is no but your call.

gquintard:
I think we are good with that second commit. I ended up keeping the command-line switch for easier testing

bsdphk:
Question:  Does the stub functions crash/abort/exit or does our code not expect their dummy return values and crash/abort/exit ?

gquintard:
They just crash

bsdphk:
I just cloned MUSL source, and cannot find the stubs, where are they ?

The word 'backtrace' only ever appears in a comment in an awk script ?

gquintard:
can't find the page about stubs back, but it appears I was wrong anyway. I opened a ticket on the libexec project used by alpine (https://github.com/mikroskeem/libexecinfo/issues/2), but was redirected to the musl team (https://www.openwall.com/lists/musl/2019/08/22/2)

I'm out of my depth here, but from what I understand, the libexecinfo available on alpine won't work anyway, so we can either use something else (https://github.com/freebsd/freebsd/tree/master/contrib/libexecinfo ?) or just unplug the feature conditionally (what is done in this PR).

mikroskeem:
FYI Alpine is not related with my repository at all, they pull [libexecinfo from FreeBSD Freshports](https://git.alpinelinux.org/aports/tree/main/libexecinfo/APKBUILD).

gquintard:
@mikroskeem, apologies, "libexecinfo for musl libc" got me confused.

rofl0r:
> the other option is to compile and run a test program during configure

no, you shouldn't run a test program ever as this breaks cross-compilation. it will be sufficient to compile a .c file containing `#include <execinfo.h>` and if it compiles, it means the functionality is there. for more details see https://sabotage-linux.github.io/blog/5/ which explains in great detail how to properly configure-check for things that are cross-compilation compatible. whether it then crashes at runtime is Not Your Problem, as in that case the implementation (in this case an external libexecinfo) is buggy and needs a fix. also having an explicit configure switch like --disable-execinfo is a good idea so it can be turned off even if execinfo.h is detected.

gquintard:
@rofl0r, fair enough, I can change that to only compile.

I dug a little more and I think I found the crux of the issue: the quick-and-dirty port of libexecinfo that is used in [alpine](https://git.alpinelinux.org/aports/tree/main/libexecinfo/APKBUILD) (pulled from freeBSD) uses `__builtin_return_address()` and `__builtin_frame_address()` which are basically just dangerous. From `man gcc`:

```
       -Wframe-address
           Warn when the __builtin_frame_address or __builtin_return_address is called with an argument greater than 0.  Such calls may return indeterminate values or crash the program.  The warning is included in -Wall.
```

the question is now would: what do we do?

- be content with a way to just disable backtraces
- consider that it's alpine's job to come up with a decent and mandatory `libexecinfo` implementation
- implement #3052 and go with `libunwind` instead, but accept losing the filename in the backtrace

The first and last options sound good enough for me. `libundwind` has the advantage of being pretty much the same on all platforms, but there is some information loss.

gquintard:
handled in #3052

PR #3035, opened 2019-08-07
Bigger stack for the curses thread in hist/top

gquintard:
musl libc has a considerably smaller stack size than glibc
(https://wiki.musl-libc.org/functional-differences-from-glibc.html#Thread-stack-size)
and curses will crash right in initscr().

I looked for recommendations but couldn't find any, so I used the 2MB
announced in the musl FAQ.

### DIFF
--- a/bin/varnishhist/varnishhist.c
+++ b/bin/varnishhist/varnishhist.c
@@ -479,6 +479,7 @@ main(int argc, char **argv)
 	char *colon;
 	const char *ptag, *profile = "responsetime";
 	pthread_t thr;
+	pthread_attr_t attr;
 	int fnum;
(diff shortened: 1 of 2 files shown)


bsdphk:
It kind of surprises me that you see stack overflows, even with a 80kB stack size, none of the functions seem to spend outrageous amounts of stack.

I guess it must be the calls into curses(3) which does it ?

If so, why does varnishstat(3) not need it ?

gquintard:
hum, that's a very good question... My issues were in `initscr()` but indeed, `varnishstat` seems unaffected...

gquintard:
There's possibly something else here, I'll close this until I have a more definitive and satisfaying answer

PR #3039, opened 2019-08-09
VSM related patches

mbgrydeland:


bsdphk:
This patch is now OBE.

PR #3041, opened 2019-08-12
Expose max connection failure in v_b_e

andrewwiik:
This adds a new read-only flag that is accessible in the ```backend_error``` subroutine called ```beresp.was_max_conn``` that describes whether the backend request had failed because of the max connections threshold for the backend having been reached. This is useful information for possibly sleeping then retrying the request or for returning a more detailed error message to the client. The reason we went with a flag rather than a string to describe the general reason why it failed was because it is much easier to mess up writing a VCL check comparing a string than it is to check a flag. This only adds 1 bit to the struct and also the lifetime of the struct is only for the backend request.


bsdphk:
Well, a flag bit is fine for this particular failure mode, but shouldn't we first ask if there are other failure modes which deserve reporting and VCL visibility ?

bsdphk:
This needs work.

We do not want a flag variable for every single failure mode.

PR #3042, opened 2019-08-12
Move backend error state reason and code struct field to end of busyobj struct

andrewwiik:
This moves the error code and reason fields to the end of the struct where they should have been when these fields were added to maintain VRT compatibility (putting them in the middle would require anyone who is using the struct to recompile their VMODs apparently).


bsdphk:
OK

PR #3044, opened 2019-08-14
Show ESI request in varnishncsa

xcir:
varnishncsa skip ESI subrequests.
https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishncsa/varnishncsa.c#L894-L895

I think useful for ESI debugging, if show ESI subrequests.


bsdphk:
Bugwash consensus:  '-e' is a bit too expensive real-estate, so -E instead.  Otherwise looks good.

Issue #1853, opened 2016-03-04
Assert error in Lck_CondWait(), cache/cache_lck.c line 203 (on Mac OS X)


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1853

nigoroll:
@mjastrzebowski are you the original reporter of the (archived) ticket?

nigoroll:
The way I read the MacOS manpage, this should not be happening, also the panic trace provided shows that `errno==0` (because there is no `errno =` line present in the panic), so I'd say this smells like a macos bug.
Would you please try to reproduce on master (or at least a more current varnish version) and reopen if the issue still exists?

lkarsten:
Backport review: nothing commited.

lkarsten:
Got access to a osx instance now, and had to satisfy my curiosity.

On OSX pthread_cond_wait() can return non-null, the man page documents that EINVAL is a valid return code. There is no mention of errno being used, I think @nigoroll misread the man page on that point.

Discussed this a bit with Martin. Since this happens in VBT_Wait() this could be an issue with our kqueue waiter and/or tcp pools implementation.

It would require some effort to follow up on this, and since the original reporter has gone quiet and OSX is a nice-to-have platform, I think Nils did the right thing by closing it for now.

yousefcisco:
I'm consistently getting this error on a number of OSX machines. Is there any way I can help debug / get to the root of the issue?

lkarsten:
Reopened to get some fresh eyes on this. I'm not running Varnish on macos these days.

winstonjz:
@yousefcisco We are on the same MacOS as well I just noticed. Maybe the newest high sierra is not playing nice with Varnish.

nigoroll:
Thanks for the new panic infos. The original panic info did not have `errno = 22 (Invalid argument)` which I guess can be a helpful clue here.

yousefcisco:
I've downgraded my version of varnish and I'm still having the error. @nigoroll do you think this is a different issue?

winstonjz:
@nigoroll @yousefcisco I upgraded my version of varnish and I'm still getting the issue. 90% of my requests result in a 502 Gateway error through nginx

winstonjz:
@yousefcisco  I am still having major issues @nigoroll @fgsch  @bsdphk 
Our front end team is at a stand still. Is there any update on this issue?

[...]

PR #3048, opened 2019-08-26
cache_lck: pthread_cond_timedwait can also returns EINVAL

devnexen:
Much more so than EINTR, at least on FreeBSD/macOS, but keeping as it is harmless.


bsdphk:
We already have a ticket on this: #1853 

We need to understand *why* we see EINVAL before ignoring it.

PR #3049, opened 2019-08-28
free array returned by backtrace_symbols(3)

gquintard:
I was trying to understand why `backtrace(3)` doesn't work on alpine and noticed that we are supposed to free whatever `backtrace_symbols(3)`. From the man page:

> This array is malloc(3)ed by backtrace_symbols(), and must be freed by the caller. (The strings pointed to by the array of pointers need not and should not be freed)

### DIFF
--- a/bin/varnishd/cache/cache_panic.c
+++ b/bin/varnishd/cache/cache_panic.c
@@ -646,6 +646,7 @@ pan_backtrace(struct vsb *vsb)
 			VSB_printf(vsb, "%s", p);
 		}
 		VSB_printf (vsb, "\n");
+		free(strings);
 	}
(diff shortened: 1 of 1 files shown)


fgsch:
Since we're dying anyway, is it important?
Personally I'm a bit reluctant to add free() calls in a signal handler.

gquintard:
It probably doesn't matter and it's possibly just my OCD talking. Just
wanted to make sure it was on purpose.

fgsch:
@gquintard I know that feeling. Let's hear from the man itself, @bsdphk.

bsdphk:
I don't care much either way, but there is something to be said for DTRT.

PR #3052, opened 2019-09-04
Use libunwind instead of libexecinfo

gquintard:
offered as an alternative to #3034

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,10 @@ jobs:
             - nghttp2
             - python3-docutils
             - python3-sphinx
+            - libunwind-dev
       before_script:
         - ./autogen.sh
-        - ./configure
+        - ./configure --with-unwind
       script: &script-common
         - |
           if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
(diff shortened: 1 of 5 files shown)


nigoroll:
I have not looked at the patch, but in general I would hope that adding libunwind would help.

From a private email as of 2019-11-14:

[...] But what I would hope to get is:

* stack frame addresses for the backtrace

* portable information about stack boundaries to avoid stack overflows (e.g. before calling pcre with jit, we know we'll need ~33k at least, so we could bail out before even calling it if we know stack won't suffice or fall back to some less efficient alternative allocation)

* _maybe_ a nicer/portable longjmp replacement to handle stack overflows from VCL and _maybe_ also stuff we can't handle otherwise like SIGFPE

gquintard:
note: force-pushed to edit the travis file

bsdphk:
Do we want a hard switch or a period with the ability to pick which one ?

bsdphk:
We will move on this after the 6.3 release.

From the bugwash discussion it was not very obvious what the portability situation was, only that one probably exists.

Initially we add a configure option to use libunwind, default to status quo, then when we have portability data, we decide what to do next.

.travis.yml
```diff
@@ -18,9 +18,10 @@ jobs:
             - nghttp2
             - python3-docutils
             - python3-sphinx
+            - libunwind-dev
       before_script:
         - ./autogen.sh
-        - ./configure
+        - ./configure --with-unwind
```
fgsch: Can you also add it below?

gquintard:
thanks @fgsch, force-pushed

fgsch:
For macos things are a bit different - the library is already there but no .pc file or headers.
So perhaps for now we want to remove it from there.

gquintard:
gotcha, removing it then

gquintard:
@bsdphk : forgot to mention I force-pushed an update. Default is to keep `execinfo`, using `--with-unwind` replaced it with the new kid

gquintard:
OK'd by Poul-Henning

PR #3055, opened 2019-09-08
Import vmod_header

gquintard:
There have been some talks about using in-tree vmods to accomplish stuff the VCL won't touch. `vmod_header` has been stable for quite a long time and had proven its usefulness, so this is a decent candidate.


bsdphk:
We talked about this at a VDD long time ago, and as I recall the conclusion vmod_header and vmod_cookie were somewhat complementary.

Talked with @gquintard about it this morning, and maybe we should merge vmod_header into vmod_std, and import vmod_cookie as such.  He will look at it.

PR #3056, opened 2019-09-09
name vmod function / constructor / method RST xref like VCL

nigoroll:
as @slimhazard pointed out when reviewing a suggestion I made to some
vmod documentation, the vmodtool generated link target names are
confusing to users: vmod_foo.func is nothing you can use as VCL, while
foo.func() is.

### DIFF
--- a/doc/sphinx/reference/directors.rst
+++ b/doc/sphinx/reference/directors.rst
@@ -162,7 +162,7 @@ Health Probes
 =============
 
 It is possible in a VCL program to query the health of a director (see
-:ref:`vmod_std.healthy`). A director can report its health if it implements the
+:ref:`std.healthy()`). A director can report its health if it implements the
 ``healthy`` function, it is otherwise always considered healthy.
 
 Unless you are making a dynamic backend, you need to take care of the
--- a/doc/sphinx/whats-new/changes-6.2.rst
+++ b/doc/sphinx/whats-new/changes-6.2.rst
@@ -117,7 +117,7 @@ to make them more flexible and easier to use. The ``std.``\ *x2y*
 conversion functions are now deprecated. See
 :ref:`whatsnew_upgrading_std_conversion_2019_03`.
 
-The function :ref:`vmod_directors.lookup` has been added to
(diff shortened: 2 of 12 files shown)


bsdphk:
Ok, with me, but lets run it past the bugwash

nigoroll:
I had missed some references, hopefully fixed now after force-pushes

slimhazard:
Repeating a question I sent to @nigoroll in person -- does the resulting reference appear in ``code style``?

I haven't tried out the PR, but it doesn't look that way to me. More like regular text style, but with a link.

IMO code strings in ``code style`` contribute to better clarity. Ideally it ought to be both ``code style`` and a reference.

How can you write such a thing in RST? *Three* backticks each for opening and closing, and then the underscore following the three closing backticks?

bsdphk:
OK'ed in principle at bugwash, awaiting final version of patch...

nigoroll:
Re @slimhazard I would also prefer the links to be in ``code style`` but to my knowledge this is not possible with RST.
If anyone knows a way to achieve the suggested improvement, please let add a comment.

Otherwise, the patch is final from my perspective.

nigoroll:
I dug some more and found this: https://stackoverflow.com/questions/4743845/format-text-in-a-link-in-restructuredtext

So in our code this would look like this:

```
diff --git a/lib/libvmod_directors/vmod.vcc b/lib/libvmod_directors/vmod.vcc
index 0f27f3c29..ae759b639 100644
--- a/lib/libvmod_directors/vmod.vcc
+++ b/lib/libvmod_directors/vmod.vcc
@@ -190,7 +190,10 @@ $Object hash()
 Create a hashing backend director.
 
 The director chooses the backend server by computing a hash/digest of
-the string given to `xhash.backend()`_.
+the string given to |xhash.backend()|_.
+
+.. |xhash.backend()| replace:: ``xhash.backend()``
+
 
 Commonly used with ``client.ip`` or a session cookie to get sticky
 sessions.
```

As the formatting is in the reference, not the target, this is orthogonal to the patch.

Issue #3051, opened 2019-09-03
explicit_bzero() causing havoc


slimhazard:
tl;dr: I suggest that the project get rid of using the non-standard ``explicit_bzero()`` for ``ZERO_OBJ()``.

Longish rant follows.

With newer compiler versions, I'm finding that VMOD code has suddenly begun failing to compile with ``implicit declaration of function 'explicit_bzero'``. It happened on my dev laptop after updating to Debian buster (with both gcc and clang). A colleague also ran into the problem while building on a relatively recent version of Ubuntu.

It appears that newer versions of the include files have changed with respect to the feature test macros that determine what is and is not defined, depending on the C standard, POSIX version, GNU source,  BSD source, XOPEN source, etc etc. ``/usr/include/string.h`` on my machine currently has this:

```
#ifdef __USE_MISC
# include <strings.h>

/* Set N bytes of S to 0.  The compiler will not delete a call to this
   function, even if S is dead after the call.  */
extern void explicit_bzero (void *__s, size_t __n) __THROW __nonnull ((1));

/* ... other stuff ... */
#endif
```

``__USE_MISC`` in turn is defined if ``_DEFAULT_SOURCE`` is defined, or ``_GNU_SOURCE``, or a few other things. This is with:

```
$ ldd --version 
ldd (Debian GLIBC 2.28-10) 2.28
```
I've found that ``#define _DEFAULT_SOURCE 1`` makes the error go away, or even:
```
#define __USE_MISC 1
#include <string.h>
#undef __USE_MISC
```
But of course that is intolerably implementation-specific. Frankly, my current workaround is:
```
/* after including miniobj.h, or anything that includes it */
#undef ZERO_OBJ
#define ZERO_OBJ(to, sz) (void)memset((to), 0, (sz))
```
The autoconf macro ``AC_CHECK_FUNCS`` apparently runs with some sort of setting as to the source features, but it's not transparent at all what those settings might be. I suppose I could go digging through the m4 source to figure it out, but having to consider that at all is one of the bad signs.

The check for ``explicit_bzero()`` happens by way of the ``VARNISH_VMODS`` macro in ``varnish.m4``. So then anyone who uses that, and then includes ``config.h`` before ``miniobj.h`` will get the result of whatever ``AC_CHECK_FUNCS`` thought it found out about defining ``ZERO_OBJ``. ``ZERO_OBJ``, in turn, is called by ``FREE_OBJ``, and that is very commonly used for VMOD object finalization.

[...]

PR #3057, opened 2019-09-11
Implementing ZERO_OBJ macro proposal.

devnexen:
- According to #3051, this does not seem too obvious to be
properly detected on Linux systems.
- Some other OSes prefer to play in their own 'backyard',
having similar feature but named differently (e.g. NetBSD)
would need addition autotools check.
- Thus proposing implementing the macro with the volatile
pointer approach rather than memory fence's, sufficient to
prevent compiler optimisations.

### DIFF
--- a/include/miniobj.h
+++ b/include/miniobj.h
@@ -5,11 +5,11 @@
  *
  */
 
-#ifdef HAVE_EXPLICIT_BZERO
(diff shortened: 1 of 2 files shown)


bsdphk:
OK'ed.

PR #3065, opened 2019-09-18
Include matrix of Varnish versions to VCL versions

xordspar0:


fgsch:
Shouldn't this belong in the homepage?

gquintard:
Most possibly, we don't want to backport that information for all the
published versions on the website

xordspar0:
That makes sense. Where is a good place on the website for that?

gquintard:
possibly https://varnish-cache.org/docs/index.html ?

gquintard:
note that it possibly make sense to update the `VERSIONING` section of `man vcl` too, to explain the top VCL version allowed for one varnish version

xordspar0:
What if we put the matrix on a new page linked from the downloads page, similar to the changelog? Or we could drop the matrix and just update the latest man page. Right now the matrix is very simple, but it could get for complicated in the future.

hermunn:
I will make a table for the homepage (varnish-cache.org) and then close this.

nigoroll:
-> https://github.com/varnishcache/homepage/pull/78

xordspar0:
The reference still needs to be corrected, since it says that the current version only supports VCL 4.0. Should we remove that statement or update it just on trunk? I've updated it here: https://github.com/varnishcache/varnish-cache/pull/3078

PR #3066, opened 2019-09-20
Single-process and libfuzzer build options

fwsGonzo:
Adds varnishd support for instrumented fuzzing, PGO, callgraphs etc. by adding a single-process mode as well as an extra LIBFUZZER option. Note that I have a CMake build system I use for both varnish and varnish-plus so I didn't make any changes to autocrap.

Kinda hacky, so feel free to commit to this branch.

### DIFF
--- a/bin/varnishd/hpack/vhp_gen_hufdec.c
+++ b/bin/varnishd/hpack/vhp_gen_hufdec.c
@@ -210,6 +210,16 @@ tbl_print(const struct tbl *tbl)
 			tbl_print(tbl->e[u].next);
 }
 
+static void
+tbl_free(struct tbl* table)
+{
+	for (unsigned i = 0; i < table->n; i++) {
+		if (table->e[i].next != NULL)
+			tbl_free(table->e[i].next);
+	}
+	free(table);
+}
+
 int
 main(int argc, const char **argv)
 {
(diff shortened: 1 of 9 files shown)


bsdphk:
I'm not terribly keen on the way this is done.

I would expect that if you replaced fork(2) in mgt_child.c with pthread_create(3) and neutered the fd-frobbing code in the pid==0 case, things would just work ?

bin/varnishd/hpack/vhp_gen_hufdec.c
```diff
@@ -250,5 +260,6 @@ main(int argc, const char **argv)
 	tbl_print(top);
 	printf("};\n");
 
+	tbl_free(top);
```
Dridi: This change should be submitted as a separate patch.

Dridi:
This sounds more like a childless mode, we still have to fork VCC processes and fork+exec CC processes, unless I missed something in the diff?

Dridi:
During the bugwash I suggested that childless mode and libfuzzer integration should be separate changes, I was told that childless mode was only meant for fuzzing, and then I couldn't continue the discussion.

When I see an `#ifdef SINGLE_PROCESS_MODE` block immediately followed by an `#ifndef LIBFUZZER_ENABLED` block then my intuition tells me that both could be introduced in separate commits. And as I commented earlier, you can either submit the change in `vhp_gen_hufdec.c` in a separate pull request or push it directly to master, it's self-contained and can be taken out of context.

PR #3067, opened 2019-09-24
Added the syntax 'backend name none;'

andrewwiik:
This brings in the feature for allowing null backend declarations via the syntax discussed [here](https://github.com/varnishcache/varnish-cache/pull/3008). The syntax is as follows:

```
backend default none;
```

### DIFF
--- a/bin/varnishtest/tests/v00060.vtc
+++ b/bin/varnishtest/tests/v00060.vtc
@@ -1,0 +1,45 @@
+varnishtest "NULL backend allowed"
+
+server s1 {
+	rxreq
+	txresp
+
+} -start
+
+varnish v1 -vcl {
+	backend default none;
+} -start
+
+client c1 {
+	txreq
+	rxresp
+	expect resp.status == 503
+} -run
+
+# Test NULL none default backend
+
+varnish v1 -vcl+backend {
+	backend null_backend none;
+        backend null_backend_uppercase None;
+        sub vcl_recv {
+                if (req.url ~ "/no_backend_lowercase") {
+                        set req.backend_hint = null_backend;
+                } else if (req.url ~ "no_backend_uppercase") {
+                        set req.backend_hint = null_backend_uppercase;
+                }
+        }
+}
+
+client c1 {
+        txreq
+	rxresp
+	expect resp.status == 200
+
+	txreq -url "/no_backend_lowercase"
+	rxresp
+	expect resp.status == 503
+
+        txreq -url "/no_backend_uppercase"
+	rxresp
+	expect resp.status == 503
+} -run
--- a/lib/libvcc/vcc_backend.c
+++ b/lib/libvcc/vcc_backend.c
@@ -321,6 +321,16 @@ vcc_ParseHostDef(struct vcc *tl, const struct token *t_be, const char *vgcname)
 	    "?proxy_header",
 	    NULL);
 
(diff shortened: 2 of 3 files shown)


lib/libvcc/vcc_backend.c
```diff
@@ -321,6 +321,16 @@ vcc_ParseHostDef(struct vcc *tl, const struct token *t_be, const char *vgcname)
...
+		tl->fb = vsb;
+		Fb(tl, 0, "\n\t%s = (NULL);\n", vgcname);
+		vcc_NextToken(tl);
+		SkipToken(tl, ';');
+		return;
+	}
+
 	SkipToken(tl, '{');
 
 	vsb = VSB_new_auto();
```
nigoroll: very minor: could save three lines by moving before the check for `none`

bin/varnishtest/tests/v00060.vtc
```diff
@@ -0,0 +1,45 @@
...
+client c1 {
+	txreq
+	rxresp
+	expect resp.status == 503
+} -run
+
+# Test NULL none default backend
+
+varnish v1 -vcl+backend {
+	backend null_backend none;
```
nigoroll: indentation

bin/varnishtest/tests/v00060.vtc
```diff
@@ -0,0 +1,45 @@
...
+                if (req.url ~ "/no_backend_lowercase") {
+                        set req.backend_hint = null_backend;
+                } else if (req.url ~ "no_backend_uppercase") {
+                        set req.backend_hint = null_backend_uppercase;
+                }
+        }
+}
+
+client c1 {
+        txreq
```
nigoroll: indentation

bin/varnishtest/tests/v00060.vtc
```diff
@@ -0,0 +1,45 @@
...
+	rxresp
+	expect resp.status == 200
+
+	txreq -url "/no_backend_lowercase"
+	rxresp
+	expect resp.status == 503
+
+        txreq -url "/no_backend_uppercase"
```
nigoroll: indentation

lib/libvcc/vcc_backend.c
```diff
@@ -321,6 +321,16 @@ vcc_ParseHostDef(struct vcc *tl, const struct token *t_be, const char *vgcname)
 	    "?proxy_header",
 	    NULL);
 
+	if (tl->t->tok == ID && (vcc_IdIs(tl->t, "none") || vcc_IdIs(tl->t, "None"))) {
```
Dridi: Shouldn't it be a case-insensitive varnish id instead? Is there another occurrence of a keyword having two specific capitalizations?

andrewwiik:
Yes, the “True” variable

nigoroll:
more reflection:

I think this should still go in as is, but we might want to consider `None` as a generic keyword for "no value", mapping onto `NULL` in C. This could bring more clarity to comparisons and other assignments.

Example 1:

`if (bereq.backend == None) { ... }
`
synonymous for
`if (! bereq.backend) { ... }
`

Example 2:

```
vcl_init { new foo = taskvar.blob(); }
vcl_recv { foo.set(None); }

```

bsdphk:
Do we need some docs too ?

PR #3069, opened 2019-09-28
Introduce circleci instructions

gquintard:
Currently weeklies and other packagecloud items are handled via Varnish Software's jenkins setup which isn't available to the public. To improve transparency this PR adds a circleCI that drives the package builds and distcheck tests.

The packaging info is still held at https://github.com/varnishcache/pkg-varnish-cache, this only put things together for easy maintenance and review.

The goal is to build weeklies with it for a couple of weeks and once we are sure with the process we can migrate from Jenkins to cirecleCI.

Notes:
- `centos:7` currently can't `distcheck` because of an apparent bug with docutils
- `distcheck` on `debian:buster` has sanitizers disabled because the tool generating `vhp_hufdec.h` has some leaks, aborting compilation
- `alpine`'s distcheck fails because of `v00010.vtc`
- I'll try to look at what is going wrong, but any and help is appreciated
- @hermunn will actually plug circleCI sometime next week, for now this is just a a dead file

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,393 @@
+version: 2.1
+commands:
+  debian_install_build_deps:
+    description: Install build dependencies
+    steps:
+      - run:
+          name: Install build dependencies
+          command: |
+            export DEBIAN_FRONTEND=noninteractive
+            export DEBCONF_NONINTERACTIVE_SEEN=true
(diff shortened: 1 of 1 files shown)


gquintard:
the travis file already solved the sanitizer issue, shamelessly copying it solved the problem

PR #3072, opened 2019-09-30
Make debug=+witness a first-class citizen

Dridi:
This change introduces a top-level make witness target that builds a dot
graph and if graphviz is available, an SVG file as well. A shell script
replaces the previous python script that no longer works. Instead of
fixing `witness.py`, which is probably trivial, the shell script does an
intermediate pass and programmatically looks for cycles using tsort(1).

Checking lock dependencies becomes actionable in a CI context.


Dridi:
Oh, and of course I realize this once I submit the pull request: with `varnishtest -p` we can enable witness mode in `varnishd` without involving `varnishtest -W`. This is how I implemented this, by mistake. Maybe we can also retire the `-W` option?

Dridi:
> Instead of fixing `witness.py`, which is probably trivial, the shell script does an intermediate pass and programmatically looks for cycles using tsort(1).

I'm no pythonista, I wasn't going to implement the cycle detection in python. This is the improvement I had in mind, regardless of whether the python script would work or not, so I was likely to go the shell route anyway (plus I didn't get my shell fix in months and addiction symptoms started manifest themselves=).

nigoroll:
@Dridi thank you for the explanation. I am fine with moving to shell, and I do be an loc-nazi either. I just did not understand what the improvement was.

Dridi:
I reorganized the commits to remove review noise, ready to be merged if OK'd.

Dridi:
Output example:

```
$ make -sj32 witness
PASS: tests/a00006.vtc
[...]
PASS: tests/v00003.vtc
============================================================================
Testsuite summary for Varnish trunk
============================================================================
# TOTAL: 772
# PASS:  758
# SKIP:  14
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
  GEN      witness.dot
  GEN      witness.svg
```
[witness.dot.txt](https://github.com/varnishcache/varnish-cache/files/3696924/witness.dot.txt)
[witness.svg.txt](https://github.com/varnishcache/varnish-cache/files/3696925/witness.svg.txt)

Makefile.am
```diff
@@ -48,4 +54,24 @@ cscope:
...
+	@echo ==================================================
+	@echo You need graphviz installed to generate svg output
+	@echo ==================================================
+	@false
+else
+	$(AM_V_GEN) $(DOT) -Tsvg $< >$@
+endif
+
+if HAVE_DOT
+witness: witness.svg
```
nigoroll: depend on `all` maybe?

I guess this is a matter of taste, but at least I prefer to be able to call the target I intend without requiring another build target as a prerequisite.

Dridi: Good point, will do.

PR #3078, opened 2019-10-03
Update docs for supported VCL versions

xordspar0:


bsdphk:
Do we have a place in the doc we can point to, which details the difference between 4.0 and 4.1 ?

PR #3080, opened 2019-10-04
RFC: Ban strcpy and strcat variants from the code base

Dridi:
This is not meant to be merged, but to illustrate how we could ban straight calls to sensitive functions such as `strcat` from the code base, or in this pull request what the replacement could look like:

- `strcpy` is replaced with `VSB_strcpy`
- it uses a VSB to get bound checks "for free"
- it behaves similarly to `bprintf`
- but it returns non-zero to signal overflow

It can't be applied mechanically throughout the source tree, but it's a trivial task when done manually.

I think this is the kind of hardening overhead we can afford, although I didn't check that every single use of those functions is not in a critical code path. I'm curious how the rest of the team feels about such a change.

I think that those functions are as safe as their bound checks, and those can become incorrect without us noticing as illustrated by #2996 where the code in question used to be correct, and turned into a minor severity bug.

To my knowledge, there is no way to enforce that the `dest` argument of `VSB_strcpy()` is a fixed-size buffer, but it also applies to `bprintf`.

### DIFF
--- a/bin/varnishd/cache/cache_ws.c
+++ b/bin/varnishd/cache/cache_ws.c
@@ -104,8 +104,7 @@ WS_Init(struct ws *ws, const char *id, void *space, unsigned len)
 	*ws->e = 0x15;
 	ws->f = ws->s;
 	assert(id[0] & 0x20);
-	assert(strlen(id) < sizeof ws->id);
-	strcpy(ws->id, id);
+	AZ(VSB_strcpy(ws->id, id));
 	WS_Assert(ws);
 }
 
--- a/include/vsb.h
+++ b/include/vsb.h
@@ -87,8 +87,11 @@ void		 VSB_quote_pfx(struct vsb *, const char*, const void *,
 void		 VSB_quote(struct vsb *, const void *, int len, int how);
 void		 VSB_indent(struct vsb *, int);
 int		 VSB_tofile(int fd, const struct vsb *);
+int              VSB_strncpy(char *dest, const char *src, size_t len);
 #ifdef __cplusplus
 };
 #endif
+
+#define VSB_strcpy(dest, src) VSB_strncpy(dest, src, sizeof(dest))
 
 #endif
--- a/lib/libvarnish/vsb.c
+++ b/lib/libvarnish/vsb.c
@@ -640,3 +640,16 @@ VSB_tofile(int fd, const struct vsb *s)
 	sz = write(fd, s->s_buf, s->s_len);
 	return (sz == s->s_len ? 0 : -1);
 }
+
+/*
+ * string.h
+ */
+
+int
+VSB_strncpy(char *dest, const char *src, size_t len)
+{
+	struct vsb s[1];
+
+	AN(VSB_new(s, dest, len, VSB_FIXEDLEN));
+	return (VSB_cat(s, src));
+}


rezan:
This is a bit confusing as there is a `VSB_*` prefix but no VSB is returned to the user.

> To my knowledge, there is no way to enforce that the dest argument of VSB_strcpy() is a fixed-size buffer.

So if its unsafe to use for most string usage, why create it? If we are going to introduce a new string library, at least make it 100% safe to use.

bsdphk:
Safe stringhandling is VSB's job, and we should use the existing facilites of VSB, rather than invent new kludges.

But it probably is a good time to do an audit of all strcat() and strcpy() calls  just in case.

Dridi:
>Safe stringhandling is VSB's job, and we should use the existing facilites of VSB, rather than invent new kludges.

We might need a new flag for `strcat`'s replacement to deal with the fact that a new VSB may not start with an empty string in the destination buffer.

> But it probably is a good time to do an audit of all strcat() and strcpy() calls just in case.

Some of the `strcpy` calls would need to be turned into `VSB_strncpy` (with an n) calls because some times the static buffer is declared in a function and populated in a different static function.

Not directly related to this, `VSB_bcat` usage may need more error handling.

Now that this has been added to today's buwash, I have one last bit of explanation that can only be applied if this change is made comprehensively: banning the offending functions.

My suggestion is to add something like this in `vdef.h`:

```c
#define strcpy strcpy_is_banned_from_varnish_code_see_vsb_h_for_a_replacement
...
```

bin/varnishd/cache/cache_ws.c
```diff
@@ -104,8 +104,7 @@ WS_Init(struct ws *ws, const char *id, void *space, unsigned len)
 	*ws->e = 0x15;
 	ws->f = ws->s;
 	assert(id[0] & 0x20);
-	assert(strlen(id) < sizeof ws->id);
-	strcpy(ws->id, id);
+	AZ(VSB_strcpy(ws->id, id));
```
nigoroll: wouldn't a `bstrcpy()` paralleling the existing `bprintf()` make more sense?

The VSBs are great, but adding the overhead for a single strcpy into a fixed length destination buffer seems not worth it to me.

Dridi: I opened this RFC exactly for this kind of feedback. I'm not worried by the VSB overhead but I was wondering whether someone else would. Also, does it matter whether we use a VSB if it's not performance-sensitive code?

In itself the call site doesn't change much. Since the arguments part of the signature is the same, the only thing you do in most cases is prefix the function with `VSB_` and guard it with `AZ`. I'm, fully aware that `bstrcpy` and friends would require even less changes, but the error handling would become implicit (I tend to prefer explicit error handling).

nigoroll: My answer really would be the usual _it depends_: 
* the `bprintf()`-style idea makes sense where the buffer is fixed and there is only one thing to cat/print.
* VSB makes sense where one needs multiple calls and/or a variable size buffer.

So, to summarize, I think the following would make sense:
* ban `strcpy`/`strcat`
* add `bstrcpy`

rezan:
I didn't get to say this in the bugwash, but a lot of projects have moved to strings which have a length attribute backed in. This would give us 2 things:

- A much higher level of string safety. No need to figure out the length, its always baked in and every operation can use it.
- Higher performance, no performance hit to calculate the length.

I think it would be great to see this in Varnish as a new string type.

nigoroll:
Trivial micro benchmark:

```
slink@haggis:~$ gcc bprintf.c && time ./a.out foo ; echo $?
1110000000

real	0m0.469s
user	0m0.468s
sys	0m0.000s
0
slink@haggis:~$ gcc -lbsd strcpy.c && time ./a.out foo ; echo $?
1110000000

real	0m0.070s
user	0m0.068s
sys	0m0.000s
0
```

strcpy.c:
```
#include <stdio.h>
#include <assert.h>
#include <bsd/string.h>

#define bstrcpy(dst, src)						\
	do {								\
		int istrlcpy;						\
		istrlcpy = strlcpy(buf, src, sizeof buf);		\
		assert(istrlcpy >= 0 && istrlcpy < (int)sizeof buf);    \
	} while (0)


int main(int argc, char * argv[]) {
	char buf[80];
	int r=0, i=0;

	if (argc != 2)
		return (-1);

	while (i++ < 10000000) {
		bstrcpy(buf, argv[1]);
		r += buf[1];
	}

	printf("%d\n", r);

	return (0);
}
```

bprintf.c:
```
#include <stdio.h>
#include <assert.h>

#define bprintf(buf, fmt, ...)						\
	do {								\
		int ibprintf;						\
		ibprintf = snprintf(buf, sizeof buf, fmt, __VA_ARGS__); \
		assert(ibprintf >= 0 && ibprintf < (int)sizeof buf);	\
	} while (0)


int main(int argc, char * argv[]) {
	char buf[80];
	int r=0, i=0;

	while (i++ < 10000000) {
		bprintf(buf, "%s", argv[1]);
		r += buf[1];
	}

	printf("%d\n", r);

	return (0);
}
```

PR #3081, opened 2019-10-07
Add a fixed-buffer strcpy analogous to bprintf

nigoroll:
inspired by @Dridi in #3080

### DIFF
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -501,8 +501,8 @@ vca_accept_task(struct worker *wrk, void *arg)
 			wrk->stats->sess_fail++;
 
 			if (wa.acceptlsock->uds) {
-				strcpy(laddr, "0.0.0.0");
-				strcpy(lport, "0");
+				bstrcpy(laddr, "0.0.0.0");
+				bstrcpy(lport, "0");
 			} else {
 				VTCP_myname(ls->sock, laddr, VTCP_ADDRBUFSIZE,
 				    lport, VTCP_PORTBUFSIZE);
--- a/include/vdef.h
+++ b/include/vdef.h
@@ -56,6 +56,16 @@
 		assert(ivbprintf >= 0 && ivbprintf < (int)sizeof buf);	\
 	} while (0)
 
+/* Safe strcpy into a fixed-size buffer */
+#define bstrcpy(dst, src)						\
+	do {								\
+		size_t lbstrcpy = strlen(src) + 1;			\
+		assert(lbstrcpy <= sizeof dst);				\
+		memcpy(dst, src, lbstrcpy);				\
+	} while (0)
+
+// TODO #define strcpy BANNED
+
 /* Close and discard filedescriptor */
 #define closefd(fdp)				\
 	do {					\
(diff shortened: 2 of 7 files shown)


bin/varnishd/cache/cache_acceptor.c
```diff
@@ -348,10 +348,10 @@ vca_mk_uds(struct wrk_accept *wa, struct sess *sp, char *laddr, char *lport,
...
 	SES_Set_String_Attr(sp, SA_CLIENT_PORT, "0");
 
-	strcpy(laddr, "0.0.0.0");
-	strcpy(raddr, "0.0.0.0");
-	strcpy(lport, "0");
-	strcpy(rport, "0");
+	bstrcpy(laddr, "0.0.0.0");
+	bstrcpy(raddr, "0.0.0.0");
+	bstrcpy(lport, "0");
+	bstrcpy(rport, "0");
```
Dridi: Like I said during bugwash you also need a replacement for `strncpy` because not all destination arguments are static buffers, so you need to be able to put the destination yourself and let the abstraction assert.

nigoroll: /me bad. I should not have changed this case into bstrcpy.

Dridi: I'm pretty sure it's not the only case, I think there were others when I had a first look on Friday.

nigoroll:
@Dridi sorry again for my sloppyness. I was aware of exactly the case which I ran into, and simply did not pay enough attention.
Also, I would have thought that the length assertion would fire, but `sizeof((const char *)"0.0.0.0") == sizeof(void *)`...

bsdphk:
Ok with me in principle, I'll leave the details to you two guys.

include/vdef.h
```diff
@@ -56,6 +56,14 @@
 		assert(ivbprintf >= 0 && ivbprintf < (int)sizeof buf);	\
 	} while (0)
 
+/* Safe strcpy into a fixed-size buffer */
+#define bstrcpy(dst, src)						\
+	do {								\
+		int lbstrcpy = strlen(src);				\
```
fgsch: Can we make this size_t?

include/vdef.h
```diff
@@ -56,6 +56,14 @@
...
+		assert(lbstrcpy >= 0 && lbstrcpy < sizeof dst);		\
+		strcpy(dst, src);					\
```
Dridi: It didn't occur to me until now that you used `strcpy` here. If we want to ban this family of functions once we have a replacement we can't also use them as-is under the hood.

nigoroll: I did actually write the first iteration with `strlcpy()`, but then realized that with the assert, it would not actually provide any benefit over `strcpy()`. Or do you see one?
In my mind, the requirement to not use `strcpy()` directly and use the `bstrcpy()` wrapper would make sure that the assertion is in place to improve safety.
Am I missing something?

Dridi: See the last point in this comment of my draft: https://github.com/varnishcache/varnish-cache/pull/3080#issuecomment-538880822

nigoroll: right. So `bstrcpy()` should be an inline function instead, followed by the ban-define.

nigoroll: this is getting ugly. Right now it feels like I was on a wrong track.

nigoroll:
@Dridi was right that the path I proposed first would conflict with a ban-define of strcpy. I looked at using an inline function in combination with a macro (required for the sizeof buf), but that lead to more complications, while actually we want to simplify.

But, as we need a strlen for the assert check anyway, we might as well use memcpy instead of strcpy.

force-pushed.

PR #3082, opened 2019-10-07
H2 Backports

andrewwiik:
These 3 commits are for back-porting H2 fixes and additions from master and any related features that were used in the H2 fixes/additions. The commits used for these back-ports are listed in each commit message. Testing is currently in-progress.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -1,6 +1,6 @@
 /*-
  * Copyright (c) 2006 Verdens Gang AS
- * Copyright (c) 2006-2015 Varnish Software AS
+ * Copyright (c) 2006-2019 Varnish Software AS
  * All rights reserved.
  *
  * Author: Poul-Henning Kamp <phk@phk.freebsd.dk>
@@ -567,7 +567,11 @@ struct sess {
 	vtim_real		t_open;		/* fd accepted */
 	vtim_real		t_idle;		/* fd accepted or resp sent */
 
+	vtim_dur		timeout_idle;
 };
+
+#define SESS_TMO(sp, tmo)					\
+	(isnan((sp)->tmo) ? cache_param->tmo : (sp)->tmo)
 
 /* Prototypes etc ----------------------------------------------------*/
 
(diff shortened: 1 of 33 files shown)


nigoroll:
I am not so much concerned with 6.0 backports, but I wonder if it would be helpful to keep the cherry-picked commits to facilitate future reference across branches

andrewwiik:
The reason we cannot cherry pick these is because these commits touch other files outside of the h2 folder we don't need all the commits for those files outside of h2. we put the commit references in the commit message.

nigoroll:
I think one could edit the original commits. But, again, as I am not a maintainer of 6.0, I have no strong feelings here.

Dridi:
Probably a sign that we should insist more on breaking patches down to atomic code changes, even if it's bothersome in master if only to help back-ports.

Otherwise you can cherry-pick individual patches, then a) solve conflicts and b) potentially need 6.0 specific changes and document them in the commit lo instead of dumping an untraceable list of hashes.

Some examples:

- 5882d268fc481a7a697d6f219ce5feb299b25ab2 (conflicts, partial cherry-pick)
- b302a2cf4d4f3863d4ebe9d8c0d6092e5b401079 (cherry-pick tweaks)
- 7519d53304e2b2b491060bfa89aab48141689cc2 (conflicts)

When the conflict is trivial or doesn't seem to justify an explanation in the commit log I only mention the conflicts without further comments:

- d4ff5c553a5fbf0dd816969c80e04a9e0bbd3289
- 0f8e21002c086a485dc978e7e8240940ab835388

(Un)fortunately git is very flexible in this regard.

rezan:
We are not going to be able to break this down into atomic or individual commits. We have the commit ids listed in the commit message which should provide the same result you are after.

Dridi:
When I'm talking about more atomic commits, I'm referring to the master branch. I'm not after any result, I was merely pointing out how you could do it differently. I only commented here because @nigoroll brought this up, otherwise I would have left the back-port entirely to your capable hands.

rezan:
Ran h2load on this for a good amount of time with a random pattern of requests. Everything looks good.

rezan:
We can add `Co-authored-by` lines to these commits so the original author can get credit in the commit log:

https://help.github.com/en/articles/creating-a-commit-with-multiple-authors

This way we have both author credit and commit context.

PR #3084, opened 2019-10-07
Fix rollback deliver

### LINKED ISSUE #3083: rollback in vcl_deliver{} broken
@rezan 's vtc https://raw.githubusercontent.com/rezan/varnish-cache/6ac41e8c5fcac88b09bba0b6afe3ddc06ef1e3a4/bin/varnishtest/tests/c00100.vtc from #3010 exposes another issue on the client side when calling `std.rollback()` from `vcl_deliver {}`

```
**** dT   0.992
***  c4   connected fd 17 from 127.0.0.1 37926 to 127.0.0.1 35295
**   c4   === txreq -url /6 -hdr "test: 1"
**** c4   txreq|GET /6 HTTP/1.1\r
**** c4   txreq|test: 1\r
**** c4   txreq|Host: 127.0.0.1\r
**** c4   txreq|\r
**   c4   === rxresp
...
**** c4   rxhdr|HTTP/1.1 200 OK\r
**** c4   rxhdr|Content-Length: 0\r
**** c4   rxhdr|Date: Mon, 07 Oct 2019 15:34:30 GMT\r
**** c4   rxhdr|\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\r
**** c4   rxhdr|\x00\x00\x00\x00\x00\x00\r
**** c4   rxhdr|Via: 1.1 varnish (Varnish/6.3)\r
**** c4   rxhdr|test: 1\r
**** c4   rxhdr|Accept-Ranges: bytes\r
**** c4   rxhdr|Connection: keep-alive\r
**** c4   rxhdr|\r
**** c4   rxhdrlen = 187
```

I am working on it

nigoroll:
Fix #3083 
* 1st patch is just some centralization, to allow to re-setup resp from rollback later.
* 2nd patch is that call
* 3rd patch is @rezan 's more complete vtc from #3010, slightly edited


bin/varnishtest/tests/c00018.vtc
```diff
@@ -123,11 +123,13 @@ logexpect l1 -v v1 -g raw {
 	expect 0 1011	VCL_call        {^HASH$}
 	expect 0 1011	VCL_return      {^lookup$}
 	expect 0 1011	Timestamp       {^Process:}
-	expect 0 1011	RespHeader      {^Date:}
-	expect 0 1011	RespHeader      {^Server: Varnish$}
-	expect 0 1011	RespHeader      {^X-Varnish: 1011$}
 	expect 0 1011	RespProtocol    {^HTTP/1.1$}
 	expect 0 1011	RespStatus      {^405$}
+	expect 0 1011	RespReason      {^Method Not Allowed$}
+	# XXX dup RespReason
```
Dridi: Why is this a duplicate? If you set the status it will update the reason if it's a status code Varnish knows. I think you simply ran into this log record because the order was changed where before the test stopped looking after `RespStatus`.

I think this comment is safe to remove.

bin/varnishtest/tests/m00017.vtc
```diff
@@ -67,3 +92,203 @@ client c1 {
...
+	rxresp
+	expect resp.status == 200
+	expect resp.http.fetch == "Fetch value 1"
+	expect resp.http.response == "123"
+	expect resp.http.response2 == "Another response"
+
+	txreq -url /9
+	rxresp
+	expect resp.status == 500
+}
```
Dridi: This client didn't run in CI.

bin/varnishtest/tests/m00017.vtc
```diff
@@ -67,3 +92,203 @@ client c1 {
...
+
+client c7 {
+	txreq -url /10
+	rxresp
+	expect resp.status == 200
+	expect resp.http.fetch == "Fetch value 1"
+	expect resp.http.response == ""
+}
```
Dridi: This client didn't run in CI.

bin/varnishtest/tests/m00017.vtc
```diff
@@ -20,6 +24,27 @@ server s1 {
...
+	rxreq
+	txresp
+	rxreq
+	txresp
+	rxreq
+	txresp
+	rxreq
+	txresp
+	rxreq
+	txresp
```
Dridi: I would rather redefine s1 or create a new server every time we reload VCL for a new use case.

We'd end up with a sequence like:

- set up sX
- load vclX
- run cX

Or close enough.

bin/varnishtest/tests/c00018.vtc
nigoroll: Yes, this existed before and after, I just meant to mark the fact that we might want to have a look later.

nigoroll:
@Dridi I hope to have addressed your feedback

bin/varnishtest/tests/m00017.vtc
```diff
@@ -219,13 +217,21 @@ varnish v1 -vcl+backend {
 	}
 }
 
+server s1 -repeat 1 {
```
Dridi: Very minor, I think `-repeat 1` has no effect.

nigoroll: it is required, server keeps the parameters across calls. I had a quick look if it made sense to change that (init with every call), but then we would get into the nitty gritty details of differentiating between the run, start or nothing and I thought it was not worth it.
So for now the repeat 1 basically resets the `repeat 5` from before.

Dridi: Please add an XXX comment before line 220 so we can forget about it for now.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -235,27 +272,12 @@ cnt_synth(struct worker *wrk, struct req *req)
 
 	wrk->stats->s_synth++;
 
-	now = W_TIM_real(wrk);
-	VSLb_ts_req(req, "Process", now);
+	VSLb_ts_req(req, "Process", W_TIM_real(wrk));
```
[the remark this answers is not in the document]
nigoroll: good catch

nigoroll:
force pushed to re-check c19 failure (which I do not see locally)

Dridi:
Oh, it is! I see it now, apologies for nagging you. I didn't see that the next `expect` command skipped one header!

```
 	expect 0 1011	RespReason      {^Method Not Allowed$}
 	# XXX dup RespReason
 	expect 1 1011	RespHeader      {^Date:}
               ^
      there ---'
```

nigoroll:
@Dridi this is fixed now

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -104,6 +104,74 @@ cnt_transport(struct worker *wrk, struct req *req)
  * Deliver an object to client
  */
 
+int
+Resp_Setup(struct req *req, unsigned method)
```
bsdphk: I think this should be two separate functions, the two callers share only a single source-line.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -281,14 +303,14 @@ cnt_synth(struct worker *wrk, struct req *req)
...
+		HTTP_Setup(req->resp, req->ws, req->vsl, SLT_RespMethod);
 		VSB_destroy(&synth_body);
 		req->req_step = R_STP_RESTART;
 		return (REQ_FSM_MORE);
 	}
 	assert(wrk->handling == VCL_RET_DELIVER);
 
-	http_Unset(h, H_Content_Length);
+	http_Unset(req->resp, H_Content_Length);
 	http_PrintfHeader(req->resp, "Content-Length: %zd",
```
bsdphk: Just an observation:  We should make http_Force() take printf style args for stuff like this.

bsdphk:
OK from here if Resp_Setup() is split into separate functions.

bin/varnishd/cache/cache_vrt.c
```diff
@@ -747,6 +747,8 @@ VRT_Rollback(VRT_CTX, VCL_HTTP hp)
 	if (hp == ctx->http_req) {
 		CHECK_OBJ_NOTNULL(ctx->req, REQ_MAGIC);
 		Req_Rollback(ctx->req);
+		if ((ctx->method & (VCL_MET_DELIVER | VCL_MET_SYNTH)) != 0)
+			Resp_Setup(ctx->req, ctx->method);
```
nigoroll: oops @bsdphk my previous comment further up was wrong:
This is the place where we need the common Resp_Setup

Issue #3089, opened 2019-10-09
Assert error in ved_gzgz_init(), (foo->start > 0 && foo->start < foo->olen * 8) not true.


mbgrydeland:
```
***  v1   debug|Error: Child (20724) Panic at: Wed, 09 Oct 2019 14:56:30 GMT
***  v1   debug|Assert error in ved_gzgz_init(), cache/cache_esi_deliver.c line 591:
***  v1   debug|  Condition(foo->start > 0 && foo->start < foo->olen * 8) not true.
***  v1   debug|version = varnish-trunk revision 5f039a4a464949cb378fe49a8101c273b84cb92e, vrt api = 10.0
***  v1   debug|ident = Linux,4.9.0-9-amd64,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,epoll
***  v1   debug|now = 1041565.146603 (mono), 1570632990.703138 (real)
***  v1   debug|Backtrace:
***  v1   debug|  0x458f80: varnishd() [0x458f80]
***  v1   debug|  0x458d42: varnishd() [0x458d42]
***  v1   debug|  0x518111: varnishd(VAS_Fail+0x51) [0x518111]
***  v1   debug|  0x42f109: varnishd() [0x42f109]
***  v1   debug|  0x42aea9: varnishd(VDP_Push+0x399) [0x42aea9]
***  v1   debug|  0x42e9f5: varnishd() [0x42e9f5]
***  v1   debug|  0x4660c7: varnishd() [0x4660c7]
***  v1   debug|  0x460e2c: varnishd(CNT_Request+0x9ac) [0x460e2c]
***  v1   debug|  0x430b6e: varnishd() [0x430b6e]
***  v1   debug|  0x42e24b: varnishd() [0x42e24b]
***  v1   debug|thread = (cache-worker)
***  v1   debug|pthread.attr = {
***  v1   debug|  guard = 4096,
***  v1   debug|  stack_bottom = 0x7f7ab55bf000,
***  v1   debug|  stack_top = 0x7f7ab55cd000,
***  v1   debug|  stack_size = 57344,
***  v1   debug|}
***  v1   debug|thr.req = 0x7f7aa7886020 {
***  v1   debug|  vxid = 1007, transport = ESI_INCLUDE
***  v1   debug|  step = R_STP_TRANSMIT,
***  v1   debug|  req_body = R_BODY_NONE,
***  v1   debug|  restarts = 0, esi_level = 1,
[...]

PR #3090, opened 2019-10-09
Conditional fetch wait for streaming stale object

mbgrydeland:
Currently, we allow conditional fetches to be streamed, that is if the stale object is not yet fully fetched, the new object copy will stream the body bytes from the stale object as they become available. That causes problems for OA_GZIPBITS, because that object attribute is set only at the very end of the fetch processing of the stale object, and the copy made for the new object will have stale data. This will cause problems for ESI+GZIP setups (demonstrated in the test case), and will also cause trouble for the stored away ungzip'ed length of the object.

Another problem that comes from this is partly bypassing the request coalescing towards the backend. If the object has a very short TTL, and the 200 response takes a long time to deliver the body, we will end up with a chain of fetches, each copying the very first one. For a TTL of 0.1s, and 10 second fetch time, that will be 100 threads each creating a copy of the same data, holding the data for the entire fetch duration.

To fix this I suggest to wait for the stale object to become fully fetched, so that we can catch
fetch errors, before we unbusy the new object. This serves two purposes. First it helps with request coalescing, and stops long chains of IMS-updated short-TTL objects all streaming from a single slow body fetch. Second it makes sure that all the object attributes are complete
when we copy them.

This patch OBE's r01646.vtc, and slightly patches r01648.vtc to expect a 503 instead of a 200 and a broken connection on the failing client.

Fixes: #3089


bin/varnishd/cache/cache_fetch.c
```diff
@@ -676,9 +676,37 @@ vbf_objiterator(void *priv, unsigned flush, const void *ptr, ssize_t len)
...
+		 * with request coalesching, and stops long chains of
+		 * IMS-updated short-TTL objects all streaming from a
+		 * single slow body fetch. Second it makes sure that all
+		 * the object attributes are complete when we copy them
+		 * (this would be an issue for ie OA_GZIPBITS). */
+		ObjWaitState(bo->stale_oc, BOS_FINISHED);
+		stale_state = stale_boc->state;
+		HSH_DerefBoc(bo->wrk, bo->stale_oc);
+		stale_boc = NULL;
+		if (stale_state != BOS_FINISHED) {
```
nigoroll: not to keep up the PR, but have you considered falling back to an ordinary fetch at this point? Is it possible? Would we want it?

hermunn: The thought experiment here is a huge file which streams slowly from the backend, and clients coming in at a steady pace (say once per second), so that many requests have entered before the streaming is done. The VCL writer knows that the backend is really good at 304, ie. knowing if the object is unchanged or not, so a short TTL (say 1ms) with a long keep has been decided is best. On current master, even when the gzip problem causing a panic in the test case is not in effect, this scenario will result in one copy of the streaming object per client. This will turn into a different, maybe worse, problem if the all generate bacend requests.

With the proposed solution, many requests (from the third client, I think) will end up on the waiting list. For each of the clients on the waiting list, they will deem an object deliverable if its timestamp is after the client's *request* timestamp. Since the timestamp of an inserted object is based on when it starts entering the cache, the proposed solution should delay that, and almost all of the clients on the waiting list will deliver from cache.

We should maybe have a test case which shows my though experiment, to demonstrate that the right amount of backend requests (with 304 answers) happen as a result of the horde of client requests.

This is a corner case, and a great solution is not going to be simple. We would need some kind of ref counting on bodies, so that each object created by these 304 resonses (one per request, and they can all have different sets of merged headers) references the same stored body.

If I understand correctly, this PR is an attempt at stopping the panics and, at the same time, significantly lowering the footprint of these 304 objects.

mbgrydeland: > not to keep up the PR, but have you considered falling back to an ordinary fetch at this point? Is it possible? Would we want it?

Do you mean going to F_STP_ERROR istead of doing F_STP_FAIL? That would open for the possibility of doing a 'return (restart)' in `vcl_backend_error` (builtin.vcl does not do restart). If we then had deref'ed and cleared `bo->stale_oc`, the next attempt would not have conditional bereq headers applied.

The problem with this approach is that when entering vbf_stp_condfetch, we have already set BOS_REQ_DONE, sending the request on its happy way if it had a grace object. We do not want to delay setting BOS_REQ_DONE until after we have waited for BOS_FINISHED on the stale object, as that would delay the fetch initiating req unnecessarily if it had a grace candidate. I guess it would be possible to go to F_STP_ERROR for the case of no req body, but it feels a bit convoluted to me.

nigoroll: Thank you for the two replies. I must say that I feel I have not spent enough time pondering this (and I will remove my review again for this reason), but one question which popped up:
Won't we run into timeout issues if we delay reading the new object?

mbgrydeland: > Won't we run into timeout issues if we delay reading the new object?

Sorry, I'm not sure I understand. The new object would be the 304 revalidated one right? That would copy internally the stale objects' body bytes? There are no timeouts for that - it would just hold the ref of that object preventing it from being LRU free'd.

nigoroll: would we want an early `vbf_cleanup()` to free the connection?

mbgrydeland: > would we want an early `vbf_cleanup()` to free the connection?

Right. The timing for when the backend request is released I believe comes fromo a desire to make test cases repeatable. But that may not be a valid concern in this case.

mbgrydeland:
There is one aspect of this that maybe was not made entirely clear from the initial PR.

The main and IMO most important bug to fix here is what happens with OA_GZIPBITS and how the copy of it can be stale and cause problems when it is later used for delivery. Now that the problem is understood, there are many ways to fix that bug, the most obvious of course is to copy OA_GZIPBITS again after the body.

The other problem with multiple simultaneous streaming revalidations is something that came to light while investigating the first issue. I think that is problematic, and saw the possibility to hit two bugs with one patch. But maybe I am wrong and that is actually a beautiful bird that should be shielded from rock throwing.

bsdphk:
General consensus on goodness.

Final sticking point: Is it really polite/sensible/productive to ask BE IF-* headers on an object we do not have safely rx'ed yet ?

Should the ObjWaitState go before VDI_GetHdr() instead ?

bsdphk:
Actually: Why does HSH_Lookup() even nominate a stale_oc if it is not usable yes ?

Maybe that's where the fix should go ?

bsdphk:
The more I think about it, the more I think HSH_Lookup is where we should fix this.

As I understand the scenario:  We have object X, we know the backend is slow (~ 100s) so we set long grace (~200s) and expect revalidation with short ttl (~1sec)

We we have X(1) in cache, it's fresh and we deliver it

TTL expires, we initiate a background cond fetch on X(1) into X(2) and deliver X(1) on grace

Next request sees:  unfresh X(1) and unfinished X(2) either of which could be stale_oc for a condfetch

I think the correct behaviour there is to deliver X(1) on grace and not initiate next cond-fetch until X(2) is finished.

That would have to happen in HSH_Lookup() I think

nigoroll:
@bsdphk I am not sure I understand in which way the optimal case you describe is relevant for this ticket.

IIUC, the case here is that an unfinished, out-of-grace object is the only one we got. For this case, during hash lookup, my understanding is that we got two options:

a) wait for it to become finished (on the waitinglist)
b) consider it in-grace, despite the fact that it's not

a) Would certainly only make sense if there are any validators present. The issue I see here is that by the time the fetch finishes, the object may have even run out of keep and not be eligible as a stale_oc at all.

b) would be the best option performance-wise, but wrong

Can you maybe describe again your scenario for out-of-grace within-keep again?

nigoroll:
had a good discussion with @bsdphk on IRC, short summary from my perspective:
* Waiting before VDI_Fetch() sounds like the better option because we can then still decide not to make a conditional request in the first place (if the busy_oc turns out to be unusable)
* phk is still pondering if the wait could get moved to the client side, but I cannot follow on this idea for the reason given before.
  * one point in particular: we could require BOS_FINISHED just for the stale_oc, but that would lead to non-conditional GETs where waiting for the bo and then do a refresh was the better option.
* If we ever got trailers, we would also need to wait for BOS_FINISHED, because the main use case for trailers is the validators
  * I think that this would be implicit if we followed the existing implementation idea, but anyway
* I think we should be concerned about an implicit policy decision: With something along the lines of this PR, we are implicitly judging that waiting for a stale_oc to complete is better than getting another full copy from the backend, but that might now be the varnish user's policy.
  * I am thus suggesting to add something like a `refresh_wait_timeout` which specifies "how much ttfb are we willing to sacrifice before getting a fresh object is the better option"
  *  by default, `refresh_wait_timeout` would be the remaining ttl+grace time for bgfetches

mbgrydeland:
> Waiting before VDI_Fetch() sounds like the better option because we can then still decide not to make a conditional request in the first place (if the busy_oc turns out to be unusable)

It seems to me that there is one aspect here that has been lost in the discussions, and that is what if the backend answers 200 instead of 304. It is only when there is a 304 response that the strong dependency between stale_oc and fetch_oc is established, and could be a problem if stale_oc was not fully fetched. Any solution that includes waiting for stale_oc before making the backend request would then penalize all the backend responses that weren't 304 as well, whether this was because the object had actually changed, or just simply the backend doesn't know how to do 304. In my proposal the waiting is done after the backend has answered 304 for this reason.

Not considering stale_oc as an IMS/INM source candidate if it not fully fetched is an alternative, and takes away any problems related to waiting. But for fetches that take longer than the TTL, this leaves a window of time where we will not send IMS/INM headers towards the backend. I know of use cases where this will be a problem, where it is optimised for being able very cheaply being able to revalidate and send 304, but giving the full 200 response costs a lot more. For this to work, one would have to set the TTL high enough to encompass the fetch time in normal operation. I am not thrilled with this alternative.

nigoroll:
@mbgrydeland you are right, this is yet another policy decision which we implied, and, again, I do not think we should do this. To recap:
* in `vcl_backend_fetch`
  * do we issue a conditional request when the `stale_oc` is not ready?
  * do we wait for it to become ready? For how long?
* for a 304:
  * For how long do we wait for the `stale_oc` to become ready?
  * if it turns out to be unusable, do we error out or retry?

Long story short, to me this smells like something we got VCL for.

hermunn:
My opinion:

>     * in `vcl_backend_fetch`
>       * do we issue a conditional request when the `stale_oc` is not ready?
>       * do we wait for it to become ready? For how long?

By default we do not issue a conditional fetch when `stale_oc` is not ready, and we log this in VSM. This allows us to fix this without changing `HSH_Lookup` too much with the potential it has to ruin some other depended upon behavior. Maybe it should be possible to override this by setting `bereq.do_304_stream = true`, or something.

The solution to almost all use cases where this (a 304 from the backend where the `stale_oc` is still streaming) becomes an issue, is to have a long enough TTL, possibly calculated from a Content-Length header. If your TTL *is* long enough and you *still* see this, then it is typically because the backend has problems with serving the object or the network in between. In these cases, it is unclear to me if asking again (with or without IMS headers) or backing off (`return (abandon)`) is the best idea, but it probably depends on the backend and/or network. Waiting for a bit sounds like a reasonable idea for some use cases, but it should happen inside VCL, so that the decision on what to do when we time out is in the hands of VCL.

>     * for a 304:
>       * For how long do we wait for the `stale_oc` to become ready?

I think that since the waiting happens in VCL, we can have different versions. Wait for x seconds since the start of the fetch, or x seconds from now? If we time out, and the the VCL author set `bereq.do_304_stream` to `true`, then it is up to the VCL to unset IMS or not.

>       * if it turns out to be unusable, do we error out or retry?

Typically you would want to error, but it might be too late if we started streaming the body. If you really want the error way out, you can `set do_stream = false` in `sub vcl_backend_response` whenever you get a 304, and then pick up the pieces when the `stale_oc` fails (and then you need to *restart*, not *retry*, if I am not mistaken).

Here we will need the *reason we got to vcl_synth with a 503* functionality we talked about during the Varnish Developer Day to make things possible to handle.

> Long story short, to me this smells like something we got VCL for.

Also, this is so complex that the ordinary VCL author will have no idea on what to do. If we go this route, we have a nontrivial documentation challenge on our hands.

It seems like we want a OA_GZIPBITS fix in first, maybe in a different pull request, and then we need to address these questions after.

nigoroll:
@hermunn thank you for illustrating the suggestion. I think you have given enough reasons why this should not be hardcoded.
Regarding the documentation challenge, I am not sure. For one, there is quite a lot of complexity in VCL already people who use it daily for years do not see any more, and I think this might be just another thing where varnish brings great flexibility and contrail. But also I would hope that we would be able to create VCL syntax which does not need that much documentation. At least far less than a set of knobs and switches.

mbgrydeland:
I do not think that VCL should be part of the solution to this problem. We struggle enough to even think of all the corner cases and describing the different scenarios here, and pushing that complexity down to the user is doing both them and us a disservice.

I argue that the current proposal is the POLA compliant way to address this. The waiting that can happen keeps with the way request coalescing normally works for do_stream==false, or for a slow backend header response (OC_F_BUSY), in that there is at most 1 active fetch for the same object. It just happens so that the fetch_oc becomes a dependent on stale_oc on the 304 from the backend, and thus it is the time frame of stale_oc that defines the progress. This can happen only when the TTL < fetch_time, which is an exceptional situation. And lastly, if there is active grace, client requests can continue to be served graced content.

nigoroll:
@mbgrydeland I am concerned about the lack of control over added latencies even if we do not make things worse.

bsdphk:
Somebody needs to go read the new "clarification" drafts from the HTTPwg, to see if they provide any answers or guidance here, in particular the "semantics" and "caching" documents.

See: https://github.com/httpwg/http-core

I agree that this sounds so hairy that pushing it to VCL is not indicated.

mbgrydeland:
I have done a lazy look through the HTTPwg documents, and AFAICT there is little guidance to be had from there. The concept of a half-received object isn't really something that is discussed much. As this stands, I think we should merge the patch as is.

nigoroll:
force-pushed as I am going to merge this into my "better than master" ;) branch

mbgrydeland:
This has been rebased on master, and a SLT_Notice VSL tag has been introduced and logged on dealying the condfetch.

Should be ready for merge.

Dridi:
This should be fine to back-port as-is in 6.0, as long as the `SLT_Notice` tag is appended. This doesn't break the ABI for forward compatibility so unless someone downgrades a `6.0.x` installation to `6.0.(x-1)` then we are fine. Same as VRT where symbols may be added but not removed or changed (except as we already did for the VPI special case).

But we can discuss that in the relevant thread when someones submits a back-port pull request.

PR #3091, opened 2019-10-09
vcl.list now returns 201 on response overflow. Fixes #3038

andrewwiik:
This makes vcl.list properly return a 201 status when its response size exceeds the cli_limit variable. This closes #3038

Issue #3038, opened 2019-08-08
vcl.list can be empty


rezan:
This happens on master when `vcl.list` is larger than the `cli_limit`. The problem is that the error code is 200, I believe it should be 201. Here is a VTC:

``` vcl
varnishtest "vcl.list and cli_limit"

server s1 { }

varnish v1 -vcl+backend { } -start
varnish v1 -vcl+backend { } 
varnish v1 -vcl+backend { }
varnish v1 -vcl+backend { }
varnish v1 -vcl+backend { }
varnish v1 -vcl+backend { }
varnish v1 -vcl+backend { }
varnish v1 -vcl+backend { }
varnish v1 -vcl+backend { }
varnish v1 -vcl+backend { }

varnish v1 -expect n_vcl_avail == 10

varnish v1 -cliok "param.set cli_limit 128b"

varnish v1 -clierr 201 "vcl.list"

# **   top   2.2 === varnish v1 -cliok "vcl.list"
# **** v1    2.2 CLI TX|vcl.list
# ***  v1    2.2 CLI RX  200
# **   v1    2.2 CLI 200 <vcl.list>
```

lkarsten:
May be related: #2713 .

PR #3092, opened 2019-10-10
Log truncated CLI responses as such

Dridi:
Following my own suggestion in #3091 I looked closer at CLI forwarding and CLI truncation. Looking at the truncation part I got rid of an `strcpy` shenanigan, for which I swear this is pure coincidence!

### DIFF
--- a/bin/varnishd/mgt/mgt_cli.c
+++ b/bin/varnishd/mgt/mgt_cli.c
@@ -347,7 +347,7 @@ mgt_cli_cb_after(const struct cli *cli)
 
 	MGT_Complain(C_CLI, "CLI %s Wr %03u %s",
 	    cli->ident, cli->result, VSB_data(cli->sb));
-	if (cli->priv == stderr &&
+	if (cli->priv == stderr && cli->result != CLIS_TRUNCATED &&
 	    cli->result != CLIS_OK && *VSB_data(cli->cmd) != '-') {
 		MGT_Complain(C_ERR, "-I file CLI command failed (%d)\n%s\n",
 		    cli->result, VSB_data(cli->sb));
--- a/bin/varnishtest/tests/u00000.vtc
+++ b/bin/varnishtest/tests/u00000.vtc
@@ -74,9 +74,17 @@ shell -err -expect {Warning: Ignoring deprecated second subargument} {
 
 # Code coverage of mgt_main, (VCL compiler and RSTdump etc) (former a00017)
 
-# Test -F mode with no VCL loaded
+# Test -F mode with no VCL loaded, and -I cli_limit truncation
 
-process p1 "exec varnishd -n ${tmpdir}/v1 -F -f '' -a :0 -l2m,1m" -log -start
+shell {
+	cat >truncation.cli <<-EOF
+	param.set cli_limit 128
+	param.show
+	param.reset cli_limit
+	EOF
+}
+
+process p1 "exec varnishd -n ${tmpdir}/v1 -I truncation.cli -F -f '' -a :0 -l2m,1m" -log -start
 
 delay 2
 
--- a/include/vsb.h
+++ b/include/vsb.h
@@ -61,6 +61,7 @@ struct vsb	*VSB_new(struct vsb *, char *, int, int);
 #define		 VSB_new_auto()				\
 	VSB_new(NULL, NULL, 0, VSB_AUTOEXTEND)
 void		 VSB_clear(struct vsb *);
+int		 VSB_truncate(struct vsb *, size_t);
 int		 VSB_bcat(struct vsb *, const void *, ssize_t);
 int		 VSB_cat(struct vsb *, const char *);
 int		 VSB_printf(struct vsb *, const char *, ...)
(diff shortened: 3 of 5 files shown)


Dridi:
I only checked manually :-/

Dridi:
I was partially wrong. I checked manually (because the syslog bit is inhibited in VTC mode) and indeed I can observe a difference before and after the patch:

```diff
-varnishd[14122]: CLI debug Wr 201 accept_filter                 -#012acceptor_sleep_decay          0.9 (default)#012acceptor_sleep_incr           0.000 [seconds] (default)
+varnishd[17852]: CLI debug Wr 201 accept_filter                 -#012acceptor_sleep_decay          0.9 (default)#012acceptor_sleep_incr      !#012[response was truncated]
```
However the 201 status was already present, and the only visible change is the visible "[response was truncated]" in the output (what the client actually gets).

As far as our test suite is concerned the diff below passes:

```diff
diff --git i/lib/libvarnish/vcli_serve.c w/lib/libvarnish/vcli_serve.c
index 5c95f4f74..0e57c448d 100644
--- i/lib/libvarnish/vcli_serve.c
+++ w/lib/libvarnish/vcli_serve.c
@@ -332,8 +332,7 @@ cls_exec(struct VCLS_fd *cfd, char * const *av)
        len = VSB_len(cli->sb);
        lim = *cs->limit;
        if (len > lim) {
-               if (cli->result == CLIS_OK)
-                       cli->result = CLIS_TRUNCATED;
+               assert(cli->result != CLIS_OK);
                strcpy(s + (lim - strlen(trunc)), trunc);
                assert(strlen(s) <= lim);
        }
```

So we never demote `CLIS_OK` to `CLIS_TRUNCATED` in this code path, but I'd rather keep the `if` safety net. Overall this area needs further scrutiny.

I still think that this pull request is OK to merge, but we might want to spend more brain cycles between this and #3091.

Dridi:
@nigoroll I added test coverage to the first commit, not sure how to cover the second one for the reasons posted earlier.

bsdphk:
I second the comment about more brain-cycles on this and #3091 

I am not even sure I can remember any longer why we have a max-length on CLI responses in the first place, does that even make sense (any more) ?

If we want to retain the limit, should it truncate or fail ?

If we retain limit, would it make more sense to make it something the CLI client asks for, per CLI connection, than a global parameter ?  (Or should we have both?)

If truncation happens, should that change the status code ?  (not just 200, although most other are small by accidents of design)

nigoroll:
* Failing where we would otherwise truncate sounds like the wrong thing to me
* A different status code sounds correct to me
* One reason for the truncation might be to avoid master process kills due to CLI ping timeouts

Dridi:
Í remember we got rid of `cli_buffer`, I don't remember why. Maybe we discussed the merits of `cli_limit` then? When I started my review of the CLI code I wasn't challenging `cli_limit` or the truncation status.

I noted that:
- results are inconsistent across commands (#3038)
- there might be duplicate handling of demoting OK to TRUNCATED
- maybe we should give less freedom to individual commands in that regard

I think that the TRUNCATED status doesn't help, because I might want to know that a message was truncated even if the command failed. So having this as some kind of attribute would be great, but we don't have such a free spot in the response "header" to squeeze that information.

bsdphk:
Bugwash notes: Surprising complex.

Limit probably is important, in particular for long BAN lists.

Important to break producing loop on limit - this already happens with BANs.

Per CLI-connection limit would probably be an improvement. (limit N bytes / No truncated responses useful)

We should probably have the event-loop in master-proc meter out the responses, so the CLI loop does not get stuck on slow CLI-clients.

More automated handling of truncated responses would probably be good, to avoid inconsistencies.

People to think and we'll return to this.

nigoroll:
I am OK with improving the CLI, but from my perspective the current implementation works well so I see no need for prioritizing this issue above others.
From my perspective, this PR is a fix for the current implementation and we should not make the problem bigger than it is.

Dridi:
I also noticed that VTE-able commands don't go through the "truncation" CLI code path. `backend.list` doesn't come empty, but also has a 200 status when truncated, without the `!\n[response was truncated]` trailer.

bsdphk:
I cannot say I like this patch.

Adding `VSB_truncate()` it gratuitous for just this one use.

Calling the "after" callback post-truncation seems strange to me, as the only conseqence seems to be to add complications to that function.  At the very least I would like to hear the argumentation for that change ?

@Dridi 's point about truncation of error responses is tempered by the fact error responses are not supposed to be long enough for that to become relevant in the first place.

All that said, I agree that valid critiques have been raised, I agree with @nigoroll 's proposed dog-wash priority, and I agree that the inconsistencies, where present, should be remedied.

But this PR does none of that.

Rejected.

Dridi:
I'm technically also not happy with the PR, as I started fiddling with the CLI I ran into more problems, like for example VTE'd commands not truncating properly.

I'm happy to close this until we have something better to work on, but we should really have at least a github issue where we keep track of things going wrong when truncation is involved.

Remember that this was a reaction to #3091 where the result was not quite satisfying, and the more I pulled on the thread the more problems I discovered. I currently have my head in libvcc so I don't remember the details but it shouldn't be too hard to compile a list of problems from this and #3091.

PR #3093, opened 2019-10-10
Fail fetch retries when request body has been released

mbgrydeland:
Currently we allow fetch retries with body even after we have released the
request that initiated the fetch, and the request body with it. The
attached test case demonstrates this, where s2 on the retry attempt gets
stuck waiting for 3 bytes of body data that is never sent.

Fix this by keeping track of what the initial request body status was, and
failing the retry attempt if the request was already released
(BOS_REQ_DONE).

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -389,6 +389,7 @@ struct busyobj {
 	 * is recycled.
 	 */
 	unsigned		retries;
+	enum req_body_state_e	initial_req_body_status;
 	struct req		*req;
 	struct sess		*sp;
 	struct worker		*wrk;
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -401,6 +401,7 @@ struct busyobj {
 	struct http		*bereq0;
 	struct http		*bereq;
 	struct http		*beresp;
+	struct objcore		*bereq_body;
 	struct objcore		*stale_oc;
 	struct objcore		*fetch_objcore;
 
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -241,6 +241,12 @@ vbf_stp_mkbereq(struct worker *wrk, struct busyobj *bo)
 	if (bo->req->req_body_status == REQ_BODY_NONE) {
 		bo->req = NULL;
 		ObjSetState(bo->wrk, bo->fetch_objcore, BOS_REQ_DONE);
+	} else if (bo->req->req_body_status == REQ_BODY_CACHED) {
+		AN(bo->req->body_oc);
+		bo->bereq_body = bo->req->body_oc;
+		HSH_Ref(bo->bereq_body);
+		bo->req = NULL;
+		ObjSetState(bo->wrk, bo->fetch_objcore, BOS_REQ_DONE);
 	}
 	return (F_STP_STARTFETCH);
 }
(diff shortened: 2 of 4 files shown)


mbgrydeland:
> Thank you so much for looking after this, it has bothered me for ages, but I did not take the time to actually address it.
> Yet I wonder if it would make more sense to ensure that `REQ_BODY_NONE` is kept if there is no body instead of adding the initial state?

I don't understand what you mean. Could you elaborate?

mbgrydeland:
> Shouldn't a retry of a cached request body just work? isn't this what we have request body caching for in the first place?

It really depends on the situation. The test case for this is specifically setting up a situation where the error happens late, and specifically after BOS_REQ_DONE has been set. That's the signal for the request thread to be allowed to run forward again. And that is here things gets buggy. Had the retry attempt been done before BOS_REQ_DONE, then all would have been fine, and we would have sent the cached req body. But after, we wrongly conclude that there was no request body, and attempt a retry where we send Content-Length header, but never send any body bytes. So that is the bug that this addresses.

If we had delayed setting BOS_REQ_DONE, then we could have done the retry in the scenario in the test case. But that is only true because it isn't a background fetch (ie it is a pass), and do_stream is false. So the client req wouldn't care if it was stuck waiting for BOS_REQ_DONE or BOS_FINISHED. But if this was done for ie caching something using request body, then that would have unnecessarily delayed the client req handling if there ie was a grace candidate it could deliver.

So for the specific case of (!bo->is_bgfetch && !bo->do_stream) we could have delayed setting BOS_REQ_DONE until after we had fetched the body, extending the time frame where a retry is possible to include the body fetch time. But IMO the complexity isn't worth it.

nigoroll:
Thank you very much for taking the time for the detailed explanation.
It bothers me though that we do not handle this case properly for a cached req body, and I wonder if we should just transfer ownership of the body object to the bereq if it exists. As it was then independent of the req, we could retry for all possible cases and I would hope that we might even be able to simplify some code.

mbgrydeland:
> It bothers me though that we do not handle this case properly for a cached req body, and I wonder if we should just transfer ownership of the body object to the bereq if it exists.

It does indeed feel a bit rubber banded. I don't think it would be possible to just transfer the body to the fetch side when it needs it though. The client side could still do a restart from vcl_deliver, which could again become a fetch operation that would need the body. So I think the fetch side would either have to make a copy of it, or a reference counting scheme would have to be implemented.

nigoroll:
Yes, you are right. I was also wondering if we needed refcounting, but did not immediately see the case for it. good you did.

bsdphk:
I agree that caching req.body really should cover this case.

Given that we hold the req.body in an objcore which already has a refcnt field, it wouldn't be too hard to implement proper ref-counted sharing of it, all we need to decided is which mutex controls.

The background-fetch with req.body scenario may be rare enough that having a single globale mtx for this is acceptable.

mbgrydeland:
Improving req.body caching would be great, cutting some of the rough edges in this area.

I still think this patch should go in though, as it improves the situation until we have the other solution.

bsdphk:
Bugwash consensus: We'd rather have the "correct" solution with refcounting the req.body busyobj.

This PR is on hold until 2019-11-04 to give people a chance to produce that PR

bsdphk:
Would it make sense to allow postponing req.body caching to `vcl_backend_fetch{}` ?

fgsch:
@bsdphk I don't think so. If you want to hash the body, you need to cache it earlier.

mbgrydeland:
I have updated this PR to include @bsdphk's patch for reference counting request bodies.

The test case has been slightly modified to not include the caching of request bodies. In that case, the bug that this PR addresses is present.

Updated the original patch to test for cached referenced request body, and only fail for the case where the request body wasn't cached.

nigoroll:
force-pushed to resolve the merge conflict as I am going to add this patch to my "better than master" ;) branch

mbgrydeland:
Force pushed again as one of the patches seems to have been lost in the previous forceful push dance

andrewwiik:
This introduces new fields in the middle of a struct, it was my understanding that this isn't best practice?

mbgrydeland:
When backporting, 4e6e4eedcf078795d3197792d17b1294d79ff616 should be considered as a follow-up to this PR.

PR #3099, opened 2019-10-16
RFC: Improve thread pool parameters error messages

### LINKED ISSUE #3098: param.reset fails for thread_pool_max
param.reset for thread_pool_max fails, it tries to set a default value less than minimum. There is no information about that this is caused by thread_pool_min being set to higher value than default.

## Expected Behavior
Reset to default value that seems to be 15k?

## Current Behavior
Tries to reset param to a value not allowed.

## Possible Solution
param.set thread_pool_max 15000

varnish> param.reset thread_pool_max
106
Must be at least 15000

(attempting to set param 'thread_pool_max' to '5000')
varnish> param.show thread_pool_max
200
thread_pool_max
        Value is: 16000 [threads]
        Default is: 5000
        Minimum is: 15000

## Context

Tried to reset parameters.

## Your Environment

Linux,4.9.184-linuxkit,x86_64,-junix,-smalloc,-sdefault,-hcritbit
varnish-plus-6.0.4r3 revision b12564b520cd5ff7e381c0e6f98b591583ff7556
(running docker)

Dridi:
Inform users when the parameter we try to change depends on another
parameter.

This is not meant to be merged as-is, we could have a new VSB function
to trim the end of the string if it matches the one passed as the second
argument. In this case, we would trim the trailing newline only of the
CLI response didn't overflow in the first place, adding the extra
information on the same line. Alternatively we could also simply add the
explanation for the dynamic bounds on the next line.

Feedback welcome.

Fixes #3098

### DIFF
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -708,16 +708,22 @@ MCF_ParamConf(enum mcf_which_e which, const char * const param,
 	AZ(VSB_finish(vsb));
 	switch (which) {
 	case MCF_DEFAULT:
-		pp->def = strdup(VSB_data(vsb));
-		AN(pp->def);
+		free(pp->dyn_def);
+		pp->dyn_def = strdup(VSB_data(vsb));
+		AN(pp->dyn_def);
+		pp->def = pp->dyn_def;
 		break;
 	case MCF_MINIMUM:
-		pp->min = strdup(VSB_data(vsb));
-		AN(pp->min);
+		free(pp->dyn_min);
+		pp->dyn_min = strdup(VSB_data(vsb));
+		AN(pp->dyn_min);
+		pp->min = pp->dyn_min;
 		break;
 	case MCF_MAXIMUM:
-		pp->max = strdup(VSB_data(vsb));
-		AN(pp->max);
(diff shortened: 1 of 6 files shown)


bsdphk:
The VSB functions track FreeBSD and we need _really_ good reasons to munge it.

Couln't this text-processing be done at compile time anyway ?

bin/varnishd/mgt/mgt_param.c
```diff
@@ -708,16 +708,22 @@ MCF_ParamConf(enum mcf_which_e which, const char * const param,
 	AZ(VSB_finish(vsb));
 	switch (which) {
 	case MCF_DEFAULT:
-		pp->def = strdup(VSB_data(vsb));
-		AN(pp->def);
+		free(pp->dyn_def);
+		pp->dyn_def = strdup(VSB_data(vsb));
```
fgsch: use REPLACE here?

Dridi:
@bsdphk I made it a compile-time thing, it only took a couple trillion brain cycles to click.

bsdphk:
It seems to me that at least some of those calls to tweak_generic_uint() could be calls to tweak_uint() ?

Otherwise it looks ok

Issue #3079, opened 2019-10-04
+= operator


lkarsten:
When building long strings in VCL, for example lengthy GET args for vmod-curl or a body for a callout, I'm constantly missing the += operator from Python.

```
set req.http.foo = req.http.foo + "&bar="
set req.http.foo = req.http.foo + "&baz="
```
vs  (future, wanted):
```
set req.http.foo += "&bar="
set req.http.foo += "&baz="
```

In the case of a redirect in vcl_synth, setting `resp.http.Location = resp.http.Location + "&foo"` has low SNR in 80 character wide files.

Dridi:
Should this be limited to strings on the left hand side? Or should we have a new dedicated keyword?

```
cat req.http.foo "&bar=";
cat req.http.foo "&baz=";
```

Do we have prior-art syntax at Fastly's?

nigoroll:
@fgsch can you help here?

fgsch:
Once upon a time, you had to use the `+=` to append to the hash.

Fastly has `+=` and `-=` only for addition and subtraction (https://docs.fastly.com/vcl/operators/).

bsdphk:
I think the syntax is much clearer.

Originally I did not include it, because I felt it could give the wrong impression about workspace usage.

I have not investigated what it would take to optimize the WS usage of the trivial case of

    req.http.foo += this;
    req.http.foo += that;
    req.http.foo += also;

But I guess it might be possible.

nigoroll:
+1 for adding `+=`.

speaking of which, it would be great if vmod objects could become lvalues such that

`mytaskvar += 5`

became possible. I think, a practical way to get there was to designate `$Method` as getters or setters.

But my add-on proposal should not keep this back, just mentioning it.

Re @bsdphk : While some optimizations could be added to VCC, I also think that not having them would sill be ok, we would not make VCL worse.

bsdphk:
Bugwash consensus:  Implement first, optimize later. Caveat WS consequence in docs.

PR #3100, opened 2019-10-18
+= operator

dmatetelki:
DRAFT! Do not merge

Initial stab at #3079

When "s += line", happens, the generated code is rewritten as "s = line + s"

The tokenlist is not modified.

### DIFF
--- a/bin/varnishtest/tests/b00070.vtc
+++ b/bin/varnishtest/tests/b00070.vtc
@@ -1,0 +1,94 @@
+varnishtest "Check the += operator on strings"
+
+server s1 {
+	rxreq
+	txresp
+} -start
+
+varnish v1 -vcl+backend {
+
+	sub vcl_deliver {
+
+		# NOTE: These are type: HEADER
+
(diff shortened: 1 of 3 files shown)


nigoroll:
without looking at details, my expectation would be that a simpler solution should be possible

bsdphk:
I think we should aim at getting to a point where WS optimizations are possible, and that means pushing the actuall append into VRT land.

In other words, I would prefer we add a `VRT_AppendHdr(VRT_CTX, VCL_HEADER, VCL_STRANDS)` function, and let that do the actual string composition.

nigoroll:
> In other words, I would prefer we add a `VRT_AppendHdr(VRT_CTX, VCL_HEADER, VCL_STRANDS)` function, and let that do the actual string composition.

How would this avoid the additional copies? Would it check if the existing header was the last thing on the workspace?

dmatetelki:
closing in favor of https://github.com/varnishcache/varnish-cache/pull/3116

Issue #3102, opened 2019-10-21
beresp 304 header merging makes impossible to distinguish between headers set by varnish and backend


nigoroll:
Reported by @martin-uplex in https://varnish-cache.org/lists/pipermail/varnish-dev/2019-October/009469.html

Basically, the way we merge headers for 304 responses makes it impossible to distinguish headers originating from the backend from ones which we set in VCL (for the previous response).

bsdphk:
Just to make sure I understand the scenario:  The `Cache-Control: nocache` is created in VCL ?

bsdphk:
We talked at one point about 304 responses getting their own dedicated VCL method, where both beresp.* (RW) and obj.* (RO) were available.

I wonder how much of the current C-lang 304 logic we could implement in builtin_vcl if we did that ?

martin-uplex:
> Just to make sure I understand the scenario: The `Cache-Control: nocache` is created in VCL ?

Yes, correct.

nigoroll:
@bsdphk yes to the first question: The scenario is to modify the headers of the cache-object with respect to the downstream behavior, but have a differeing ttl/cacheable status.
And yes, I think that an additional VCL method would help and IIUC this was my preferred option when we initially discussed the 304 implementation.

bsdphk:
But didnt we find out back then that we would still need C-magic to make cond-fetch work ?

nigoroll:
Discussed on IRC: I think this model should work:
* before `vcl_backend_response {}` , for a 304, call `vcl_backend_refresh {}` with `beresp.*` (r/w) and `obj.*` (r/o)
* `builtin.vcl` would just contain `sub vcl_backend_refresh { return(deliver); }`
* the header merge, as exists now, would after `vcl_backend_refresh {}`:
** `beresp.` headers always replace `obj.` headers
* 304-aware code would basically do almost all of the work (except deletion, see below) in `vcl_backend_refresh {}` and skip most of the processing in `vcl_backend_response {}` based on `beresp.was_304`
* The only special case (I would) would be header deletion, which requires cooperation between `vcl_backend_refresh {}` and `vcl_backend_response {}`: `_refresh` would set some marker (e.g. in a header or variable) and `_response` would do the actual deletion.

@martin-uplex can you please review if this suggestion would work for you?

martin-uplex:
discussed with @nigoroll, he will follow up on this

[...]

PR #3103, opened 2019-10-23
RFC: handle non-symbolic HTTP header names in pure VCL

Dridi:
This came up recently in this thread:

https://varnish-cache.org/lists/pipermail/varnish-misc/2019-October/026685.html

I still find the outcome of referring to a third-party VMOD embarrassing and now that we opened a breach with type properties and methods, I think we should enter the breach and expand VCL capabilities.

In this example, I added limited parameters capabilities to type methods (positional only, not optional, etc) and implemented a method for the HTTP type. This is particularly sneaky because I took advantage of the lack of conflicts between type methods and the symbol table, so I ended up with a very similar syntax:

```vcl
req.http("7hello");
```

See individual patches for more information.

See also #2573 and https://github.com/varnishcache/varnish-cache/issues/3102#issuecomment-544980358

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -177,6 +177,44 @@ VRT_GetHdr(VRT_CTX, VCL_HEADER hs)
 	return (p);
 }
 
+/*--------------------------------------------------------------------*/
+
+VCL_HEADER
+VRT_Header(VRT_CTX, VCL_HTTP hp, VCL_STRING name)
+{
+	struct gethdr_s *hs;
+	int len;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(hp, HTTP_MAGIC);
+	AN(name);
+
+	len = strlen(name);
+	/* TODO: check header name validity */
+
+	CHECK_OBJ_NOTNULL(ctx->ws, WS_MAGIC);
+	hs = WS_Alloc(ctx->ws, sizeof *hs);
+	XXXAN(hs);
+
+	if (hp == ctx->http_req)
+		hs->where = HDR_REQ;
+	else if (hp == ctx->http_req_top)
+		hs->where = HDR_REQ_TOP;
+	else if (hp == ctx->http_bereq)
+		hs->where = HDR_BEREQ;
+	else if (hp == ctx->http_beresp)
+		hs->where = HDR_BERESP;
+	else if (hp == ctx->http_resp)
+		hs->where = HDR_RESP;
+	else
+		hs->where = HDR_OBJ;
+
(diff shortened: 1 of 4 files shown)


nigoroll:
+1 to allow for arbitrary (and dynamic) header names, but -1 for adding it to core code as explained below:

I like the `req.http(param)` approach, but the `add_header()` / `set_header()` not so much.
As is the case so often, I think we should make up our mind about how the VCL would look like. 

As this keeps coming up, if we touch it, I really wish we take a stake at adding all the cases which we are currently missing syntax-wise. Those which come to my mind are:

* Accessing header lines: set, remove, concat
* Accessing fields: set, remove, concat
* Adding header lines

We need to consider that formats differ for different headers (think `Accept-Encodning` vs. `Cookie` vs. `Set-Cookie`)

I also think the LHS and RHS should not be different.

In order not to complicate core code more, I do think, however, that putting this into a bundled vmod makes more sense than trying to catch all the cases in core code. Also, we already made that decision in #2573 and I do not see what would have changed since then.

nigoroll:
I would like to re-phrase my suggestion a bit:
Rather than adding the functionality to _pure VCL_, I would suggest to add the missing bits to the vmod facility to make it become as syntactically clean as _pure VCL_.

Dridi:
> Also, we already made that decision in #2573 and I do not see what would have changed since then.

The difference since that decision is that types may now have properties or methods, which wasn't an option then.

> We need to consider that formats differ for different headers (think `Accept-Encodning` vs. `Cookie` vs. `Set-Cookie`)

Which is exactly why having `add_header()` as an alternative to `set` would help here. Because currently our ability to manage headers is limited to all-ish (depending on `std.collapse`) or none.

bsdphk:
Has anybody checked what fastly's VCL does here ?

As a general rule we should adopt prior art rather than diverge further

nigoroll:
ping @fgsch can you help here?

fgsch:
Both `set req.http.7hello = "...";` and `set req.http.foo.bar = "...";` work in Fastly without special incantations.

Dridi:
In other words, Fastly is still working like Varnish used to before header names became "proper" symbols, assuming "foo.bar" is a valid header name for Fastly. Which probably means you also can't deal with even more exotic header names that nobody in their right mind would use.

nigoroll:
@Dridi are you waiting for any more comments or can this one be closed?

Dridi:
We have an ongoing discussion in #3112 but I believe this specific discussion is outside of the scope, so I would rather not close this discussion before we get to the point where we may address "non symbolic" header names.

Dridi:
Two things:

- type methods (one possible outcome of VIP28)
- non-symbolic headers

As the former is a somewhat ambitious change that could enable the latter (in a better way than this RFC does) I will close this ticket and wait until type methods become actionable.

nigoroll:
see also https://github.com/varnish/varnish-modules/pull/148

PR #3105, opened 2019-10-23
param.reset to initial (-p) values

gquintard:
Add a new argument to `param.reset` to instruct it to reset not to the hard-coded default value but to the initial value specified in the command line.

I'm not a fan of the pointer dance to replace `pp->init` in `MCF_ParamSet()` but I felt it was better than removing the `const` qualifier.


gquintard:
forgot to use `autogen.des`, got bitten, I'll open a new PR

PR #3106, opened 2019-10-23
param.reset to initial (-p) values

gquintard:
Add a new argument to param.reset to instruct it to reset not to the hard-coded default value but to the initial value specified in the command line.

I'm not a fan of the pointer dance to replace `pp->init` in `MCF_ParamSet()` but I felt it was better than removing the const qualifier.

### DIFF
--- a/bin/varnishd/mgt/mgt.h
+++ b/bin/varnishd/mgt/mgt.h
@@ -173,7 +173,7 @@ enum mcf_which_e {
 void MCF_ParamConf(enum mcf_which_e, const char *param, const char *, ...)
     v_printflike_(3, 4);
 
-void MCF_ParamSet(struct cli *, const char *param, const char *val);
+void MCF_ParamSet(struct cli *, const char *param, const char *val, int init);
 void MCF_ParamProtect(struct cli *, const char *arg);
 void MCF_DumpRstParam(void);
 void MCF_AddParams(struct parspec *ps);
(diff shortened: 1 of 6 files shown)


Dridi:
What if I set my "initial value" via `-I` instead?

Dridi:
And what if I set the same parameter multiple times using either `-p` or `-I` during startup?

gquintard:
arguments in a `-I` file don't count as initial values. To be perfectly honest, I totally forgot about that case, but also, I have no idea if they should.

For multiple `-p` arguments, the last one is used as initial value.

bsdphk:
I'm not a fan of this.

Yes, I can see specific circumstances where this would be nice to have, but I cant imagine those circumstances being frequent enough to warrant anything like this.

If we really wanted to do something in this space, and I really dont think we need to, we should either add a commit/rollback facility or named sets of parameters so that one can switch from one "tuning" to another "tuning" with a single command.

So thumbs down, until/unless credible evidence that this is really a need.

nigoroll:
TL;DR: No vote from me

I do not feel like giving this a -1, because there could be a valid case which I do not see, but in my world, there is always some init/start script or parameter file which contains the initial values and I do not see much use for adding those to varnish also.
One other aspect: In my mind, what would make sort-of sense, would be a "should be" value and a "currently overridden" value. But with that model in mind, we would then also need a way to change the "should be" value, but then the question is, why have both... Sort of turning in circles.

gquintard:
I'll close :-)

PR #3111, opened 2019-11-03
add if-range logic

gquintard:


bin/varnishd/cache/cache_range.c
```diff
@@ -180,13 +181,66 @@ vrg_dorange(struct req *req, const char *r, void **priv)
...
+	if (!http_GetHdr(req->resp, H_Date, &p) || !(d = VTIM_parse(p)))
+		return (0);
+
+	/* grab the Last Modified value */
+	if (http_GetHdr(req->resp, H_Last_Modified, &p)) {
+		lm = VTIM_parse(p);
+		if (!lm)
+			return (0);
+	} else
+		AZ(ObjGetDouble(req->wrk, req->objcore, OA_LASTMODIFIED, &lm));
```
nigoroll: do we really want this? What if the LM header was removed in VCL on purpose to force the If-Range check to only ever work on Etags?

bin/varnishd/cache/cache_range.c
```diff
@@ -180,13 +181,66 @@ vrg_dorange(struct req *req, const char *r, void **priv)
 	return (NULL);
 }
 
+/*
+ * return 1 if range should be observed, based on if-range value
+ * if-range can either be a date or an ETag [RFC7233 3.1 p8]
```
nigoroll: it actually is section 3.2 https://tools.ietf.org/html/rfc7233#section-3.2

bin/varnishd/cache/cache_range.c
```diff
@@ -180,13 +181,66 @@ vrg_dorange(struct req *req, const char *r, void **priv)
...
+
+	/* grab the Last Modified value */
+	if (http_GetHdr(req->resp, H_Last_Modified, &p)) {
+		lm = VTIM_parse(p);
+		if (!lm)
+			return (0);
+	} else
+		AZ(ObjGetDouble(req->wrk, req->objcore, OA_LASTMODIFIED, &lm));
+	
+	/* Last Modified must be 60 seconds older than Date */
```
nigoroll: where does this come from?

nigoroll: ok found it https://tools.ietf.org/html/rfc7232#section-2.2.2

bin/varnishd/cache/cache_range.c
gquintard: I was (and still am) on the fence about this, I mimicked what `RFC2616_Do_Cond()` does (https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_rfc2616.c#L296).

bin/varnishd/cache/cache_range.c
gquintard: indeed, fixed, thank you

bin/varnishd/cache/cache_range.c
nigoroll: Interesting, continued in #3118

nigoroll:
@gquintard I think @Dridi wants you to put the rfc comment in the format used here:

https://github.com/varnishcache/varnish-cache/blob/bea1da6e0d2557143689acd2842b82b5b4268dec/bin/varnishd/http2/cache_http2_proto.c#L275

bin/varnishd/cache/cache_range.c
```diff
@@ -180,13 +181,66 @@ vrg_dorange(struct req *req, const char *r, void **priv)
...
+	/* strong validation needed */
+	if (p[0] == 'W' && p[1] == '/')
+		return (0);
+
+	/* ETag */
+	if (p[0] == '"') {
+		if (!http_GetHdr(req->resp, H_ETag, &e))
+			return (0);
+		if ((e[0] == 'W' && e[1] == '/'))
+			return (0);
```
nigoroll: if the etag is weak, we should fall through to LM comparison, rather than returning.

nigoroll: Ah no, I was wrong, the If-Range explicitly demands an etag, so there is no fallback. Sorry.

gquintard:
@Dridi, RFC coverage added, let me know if you want more
@nigoroll, I'll want for you for #3118 

Note that adding RFC coverage made me realize the code was wrong:

> Note that this comparison by exact match,
   including when the validator is an HTTP-date, differs from the
   "earlier than or equal to" comparison used when evaluating an
   If-Unmodified-Since conditional.

So I amended the code accordingly

nigoroll:
> @nigoroll, I'll want for you for #3118

I do not understand _want_. Do you mean _wait_?

bin/varnishd/cache/cache_range.c
```diff
@@ -180,13 +181,67 @@ vrg_dorange(struct req *req, const char *r, void **priv)
...
+	if (!http_GetHdr(req->resp, H_Date, &p) || !(d = VTIM_parse(p)))
+		return (0);
+
+	/* grab the Last Modified value */
+	if (http_GetHdr(req->resp, H_Last_Modified, &p)) {
+		lm = VTIM_parse(p);
+		if (!lm)
+			return (0);
+	} else
+		AZ(ObjGetDouble(req->wrk, req->objcore, OA_LASTMODIFIED, &lm));
```
nigoroll: Actually, coming back from a de-tour over #3118, I now think that we really should _not_ use `OA_LASTMODIFIED` here:
* While, for the `If-Modified-Since` case, the client could have used some other source to produce the time value for the greater/less than comparison, it can not do this for the `If-Range`: If there was no `Last-Modified` response header, how would it ever have guessed the exact date?

* also I wonder if _Note that this comparison by exact match, including when the validator is an HTTP-date_ from the RFC should be read as "need to be string equal"

So I think now that the modified-date check really should only succeed if `resp.http.Last-Modified` is present.

gquintard: sounds fair, I removed the `OA_LASTMODIFIED` fallback

gquintard:
> > @nigoroll, I'll want for you for #3118
> 
> I do not understand _want_. Do you mean _wait_?

That's what I meant, keyboards are hard...

Issue #2976, opened 2019-04-16
Probe interval inversion


rezan:
Myself and @steven-j-wojcik have been researching an odd probe issue and it looks like there is some interval inversion happening. Basically, the largest interval found is being used, blocking all the other probes. So if there are 3 probes with 1s, 5s, and 25s intervals respectively, the 25s interval is sometimes used, preventing the other probes from happening at 1s and 5s.

I scanned the code, its not clear how this can happen, but I am able to reproduce it with [this VTC](https://gist.github.com/rezan/0283ecd4c0fccb78ed70f5ee728805c5).

Here is a 1s probe working normally for 30s:

```
**** v1    0.4 vsl|          0 Backend_health  - (null) Went sick -------H 2 3 8 0.000000 0.000000
**** v1    0.4 vsl|          0 Backend_health  - vcl1.b1 Still sick -------H 2 3 8 0.000000 0.000000 
**** v1    0.4 vsl|          0 Backend_health  - vcl1.b1 Back healthy 4---X-RH 3 3 8 0.001059 0.000353 HTTP/1.1 200 OK
**** v1    1.4 vsl|          0 Backend_health  - vcl1.b1 Still healthy 4---X-RH 4 3 8 0.000889 0.000487 HTTP/1.1 200 OK
**** v1    2.5 vsl|          0 Backend_health  - vcl1.b1 Still healthy 4---X-RH 5 3 8 0.000454 0.000479 HTTP/1.1 200 OK
**** v1    3.5 vsl|          0 Backend_health  - vcl1.b1 Still healthy 4---X-RH 6 3 8 0.000766 0.000551 HTTP/1.1 200 OK
**** v1    4.5 vsl|          0 Backend_health  - vcl1.b1 Still healthy 4---X-RH 7 3 8 0.000829 0.000620 HTTP/1.1 200 OK
**** v1    5.5 vsl|          0 Backend_health  - vcl1.b1 Still healthy 4---X-RH 8 3 8 0.001034 0.000724 HTTP/1.1 200 OK
**** v1    6.5 vsl|          0 Backend_health  - vcl1.b1 Still healthy 4---X-RH 8 3 8 0.001044 0.000804 HTTP/1.1 200 OK
**** v1    7.5 vsl|          0 Backend_health  - vcl1.b1 Still healthy 4---X-RH 8 3 8 0.000867 0.000820 HTTP/1.1 200 OK
[...]

PR #3115, opened 2019-11-05
Fix bug in probe heap comparison function

mbgrydeland:
A case has been observed where the probe scheduling stopped. The system had some probes with very long interval (24h), and some with short intervals. It is suspected that the cause was the bug that this PR fixes, where the heap bubbled the wrong probe to the top giving a very long (24h) wait time before the next probe scheduling.

An attempt was made at creating a test case for this, but it proved very unreliable and timing sensitive so it was abandoned.

### DIFF
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -721,8 +721,8 @@ vbp_cmp(void *priv, const void *a, const void *b)
 	CAST_OBJ_NOTNULL(aa, a, VBP_TARGET_MAGIC);
 	CAST_OBJ_NOTNULL(bb, b, VBP_TARGET_MAGIC);
 
-	if (aa->running && !bb->running)
-		return (0);
+	if ((aa->running == 0) != (bb->running == 0))
+		return (aa->running == 0);
 
 	return (aa->due < bb->due);
 }


rezan:
https://github.com/varnishcache/varnish-cache/issues/2976#issuecomment-484602671

nigoroll:
I am pretty sure that at one point in time I had more or less the same code (IIUC more like `if (aa->running != bb->running) return (!aa->running)`), but somehow convinced myself that the existing code was correct.

Can you explain why it is not?

hermunn:
> Can you explain why it is not?

In my not so humble opinion, I think it should be the other way around, why would the current code work? Let me explain.

If a function has the name *compare* and is used as a *compare function* in some kind of algorithm, then it should *be* a compare function.

 In the existing code, if one probe is running and one is not, then it is possible that both `vbp_cmp(a,b)` and `vbp_cmp(b,a)` return zero, which normally would indicate that they are equal or *equivalent*. Here, the problem might be that all of `vbp_cmp(a,b)`, `vbp_cmp(b,c)` and `vbp_cmp(c,a)` can return all 0, while `vbp_cmp(b,a)` evaluates to 1, and this can happen both when exactly one or two of the three are running.

The code in the proposed fix is clear, and it clearly a true *compare* function. If one is running and the other is not, use that. If they are the same, use the due time.

It is possible that two different probes become equal in priority, but they will form equivalence classes (they all share `running == 0` and the due time).

i think it should be uncontroversial to just merge this (but I am sharing office with @mbgrydeland, so not my call).

nigoroll:
Just to illustrate, the resulting truth table change is (with aa and bb denoting the running state and due denoting the comparison of the due values)

```
aa      bb      return
 0      0       due
 0      1       due
 1      0       0
 1      1       due
```

```
aa      bb      return
 0      0       due
 0      1       1
 1      0       0
 1      1       due
```

Now I cannot understand any more why I ever thought the existing code was correct.

Issue #3114, opened 2019-11-05
Inserting binary objects in vcl_backend_error


lkarsten:
Issue: when synthing error responses in vcl_backend_error, there is no way to synth a binary body.

Expected: being able to insert a static "sorry, we have issues" jpeg into cache when we don't have a backend.

Some context: 
I have a rig which serves (many) live images from a single backend. It is very important that this returns something sensible, but never something stale, when the backend is down.

First attempt with `set beresp.body = std.fileread("error-500.jpeg")` returns the first 4 bytes of the JPEG, since the later bytes in the file includes a \0. No warnings of a partial file being returned anywhere I can find. VFIL_readfile() will read the whole file as I understand it, but std.fileread() only use up until the first \0.

Having discussed this a bit with Geoff and learning about uplex' vmod-file, I think a way forward would be to allow:
1) make a BLOB from the contents of a file on disk (with PRIV_VCL lifetime), and
2) assign this blob to beresp.body in vcl_backend_error somehow.

In VCL it could look like:
```
sub vcl_init {
   new errorblob = blob.read_file("/etc/varnish/error.jpg");
}

sub vcl_backend_error {
    if (bereq.backend == imageserver && bereq.url ~ "\*.jpg") {
        set beresp.content-type = "image/jpeg";
        set beresp.body = errorblob;
    }
}
```

Workaround: I'm experimenting with returning a base64-encoded `<img src="data:image/jpeg;, ">` page instead, but I'm not sure that will work everywhere.

nigoroll:
I totally agree that is this missing (side note: I one wrote the trivial [vmod_blobsynth](https://code.uplex.de/uplex-varnish/libvmod-blobsynth) to do the same thing on the client side).
I _tend_ towards just making `.body` a `BLOB`. Yes, it will break VCL, but `synthetic()` could stay and converting text into blob is a no-brainer.

bsdphk:
As long as we dont break existing VCL I'm fine with this.

Dridi:
I was requested as a reviewer for #3100 and I had a first stab at this when it was first brought up on IRC. With both in mind I submitted #3116, on top of which I'm convinced I can make this work without much disruption.

If nobody minds, I'm assigning this to myself.

bsdphk:
Right now we have an "ambient" rule that everything can be converted to STRING in VCL.

Right of the bat, I dont see why we could not augment that with a rule that STRING can be converted to BLOB.

[...]

PR #3116, opened 2019-11-05
Teach libvcc more += operators

Dridi:
This is an alternative take on #3100 as per my review suggestions. This is not comprehensive in terms of concatenation compared to #3100 and is only meant as a minimal proof of concept that I may complete if requested.

### DIFF
--- a/bin/varnishd/cache/cache_vrt_var.c
+++ b/bin/varnishd/cache/cache_vrt_var.c
@@ -880,13 +880,17 @@ VRT_r_resp_do_esi(VRT_CTX)
 
 #define VRT_BODY_L(which)					\
 VCL_VOID							\
-VRT_l_##which##_body(VRT_CTX, const char *str, ...)		\
+VRT_l_##which##_body(VRT_CTX, enum lbody_e type,		\
+    const char *str, ...)					\
 {								\
 	va_list ap;						\
 	const char *p;						\
 	struct vsb *vsb;					\
 								\
 	CAST_OBJ_NOTNULL(vsb, ctx->specific, VSB_MAGIC);	\
+	assert(type == LBODY_SET || type == LBODY_ADD);		\
+	if (type == LBODY_SET)					\
+		VSB_clear(vsb);					\
 	va_start(ap, str);					\
 	p = str;						\
 	while (p != vrt_magic_string_end) {			\
--- a/bin/varnishtest/tests/r03079.vtc
+++ b/bin/varnishtest/tests/r03079.vtc
@@ -1,4 +1,6 @@
-varnishtest "set VCL_??? += VCL_STRING;"
+varnishtest "set VCL_??? [+]= VCL_???;"
+
+# set STRING|HEADER += STRINGS;
 
 server s1 {
 	rxreq
@@ -22,4 +24,54 @@ client c1 {
 	rxresp
 	expect resp.status == 200
 	expect resp.http.x-powered-by == varnishtest1002
+} -run
+
+# set BODY [+]= STRINGS;
+
+varnish v1 -vcl {
+	import blob;
+	backend be none;
+	sub vcl_recv {
+		return (synth(200));
+	}
+	sub vcl_synth {
+		if (req.url ~ "synth") {
+			synthetic("hello");
+			if (req.url ~ "add") {
+				synthetic("world");
+			}
+		}
+		if (req.url ~ "string") {
+			set resp.body = "hello";
+			if (req.url ~ "reset") {
+				set resp.body = "world";
+			}
+			if (req.url ~ "add") {
+				set resp.body += "world";
+			}
+		}
+		return (deliver);
+	}
+}
+
+client c1 {
+	txreq -url "/synth"
+	rxresp
+	expect resp.body == hello
+
+	txreq -url "/synth/add"
+	rxresp
+	expect resp.body == helloworld
+
+	txreq -url "/string"
+	rxresp
+	expect resp.body == hello
+
+	txreq -url "/string/reset"
+	rxresp
+	expect resp.body == world
+
+	txreq -url "/string/add"
+	rxresp
+	expect resp.body == helloworld
 } -run
--- a/bin/varnishtest/tests/r03079.vtc
+++ b/bin/varnishtest/tests/r03079.vtc
@@ -1,0 +1,25 @@
+varnishtest "set VCL_??? += VCL_STRING;"
+
+server s1 {
+	rxreq
+	expect req.url == "/hello/world"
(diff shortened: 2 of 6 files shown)


Dridi:
Before signing off today I gave a try to the HEADER type (a1e75b1cc229fd12bea814f1c352ec62b060341d) and it turned out to be trivial with the changes so far. I'm not going to touch the BODY type because currently we can't do this:

```vcl
set [be]resp.body = [be]resp.body + "string";
```

So that wouldn't make sense:

```vcl
set [be]resp.body += "string";
```

Since it's already failing just fine, I'm doing anything special about this type. I believe at this point this PR is ready to solve #3079. It does however not deal with https://github.com/varnishcache/varnish-cache/pull/3100#issuecomment-549729001 in its current form.

nigoroll:
I definitely prefer this route over the one in #3079 
Some questions though:
* can you, at least prototypically, show how this will would with a workspace-saving append function as proposed here https://github.com/varnishcache/varnish-cache/pull/3100#issuecomment-549729001
* Are we sure that all relevant use cases of the operators are covered by VTCs?
* in https://github.com/varnishcache/varnish-cache/issues/3114#issuecomment-550220818 you seem to suggest this to be related to a `BLOB` interface for `[be]resp.body`. Can you please explain this connection?

Dridi:
> * can you, at least prototypically, show how this will would with a workspace-saving append function as proposed here https://github.com/varnishcache/varnish-cache/pull/3100#issuecomment-549729001

I don't know yet. My understanding is that `STRING_LIST` is going away eventually, so once we do everything with `STRANDS` then concatenation can become very efficient by simply creating a new `STRANDS` without touching the components of the operands (they are supposed to be constant strings) so until you collect the resulting `STRANDS` concatenation remains cheap.

My theory, I didn't focus on that aspect.

> * Are we sure that all relevant use cases of the operators are covered by VTCs?

Definitely not. My test case is only doing positive checks and does not cover expected errors.

> * in https://github.com/varnishcache/varnish-cache/issues/3114#issuecomment-550220818 you seem to suggest this to be related to a `BLOB` interface for `[be]resp.body`. Can you please explain this connection?

In current master we have this special case for the `set` action:

https://github.com/varnishcache/varnish-cache/blob/bea1da6e0d2557143689acd2842b82b5b4268dec/lib/libvcc/vcc_action.c#L123-L133

In my patch the `BODY` type is handled in a generic way:

https://github.com/varnishcache/varnish-cache/blob/a1e75b1cc229fd12bea814f1c352ec62b060341d/lib/libvcc/vcc_action.c#L87

The idea is to reintroduce a special case where we peek at the RHS expression and change the `ap` on the fly to a `BODY` assignment if we detect a `BLOB`. And of course, break the VRT signature and tweak "BODY wants STRING_LIST" and "BODY wants BLOB" assignments to account for it.

I didn't go all the way through, but I have some of the required bits in place.

nigoroll:
> > * can you, at least prototypically, show how this will would with a workspace-saving append function as proposed here [#3100 (comment)](https://github.com/varnishcache/varnish-cache/pull/3100#issuecomment-549729001)
> 
> I don't know yet. My understanding is that `STRING_LIST` is going away eventually, so once we do everything with `STRANDS` then concatenation can become very efficient by simply creating a new `STRANDS` without touching the components of the operands (they are supposed to be constant strings) so until you collect the resulting `STRANDS` concatenation remains cheap.

Interesting. I did not see this option yet, but it seems very attractive.

> The idea is to reintroduce a special case where we peek at the RHS expression and change the `ap` on the fly to a "BODY wants BLOB" assignment if we detect a `BLOB`. And of course, break the VRT signature and tweak "BODY wants STRING_LIST" and "BODY wants BLOB" assignments to account for it.

I like this idea, but I did fail with polymorphism suggestions before, so I wonder if this approach will be viable. @bsdphk what do you think?

Dridi:
> I like this idea, but I did fail with polymorphism suggestions before, so I wonder if this approach will be viable. @bsdphk what do you think?

We already have a discrepancy where you need a `.tostring`-able expression if you want to assign it to a writable `BODY`. An alternative could be to grow a `.tobody` method in `struct type` and teach it to string types and `BLOB`. When I considered this, I wasn't sure whether this would complicate expression handling on the RHS, and whether we would out of the box get something like `HEADER.tostring().tobody()` work in the generated code for this kind of assignment:

```vcl
sub vcl_synth {
    set resp.body = req.http.some-header;
```

Restoring a special case felt less intrusive after my limited research of both alternatives.

Dridi:
Upon further inspection, it turned out my initial plan regarding #3114 is relevant to this topic regardless of the `BLOB` question. I have a working PoC that I will push to this pull request and turn it into an RFC because it needs a broader discussion.

Incidentally I also solved #3114 in my PoC, so we might be able to kill two issues with one pull request.

bsdphk:
I have looked briefly at this, and I like here it is going.

Yes, STRING_LIST is being deprecated, it should only ever be used in still-not-STRANDS-upgraded vmods.

I think I can see the WS-saving trick being feasible still, but that will have to be tested.

Dridi:
I found quite the interesting VCL bug and fixed it in 6862d818a7cc9069f369363f86e2f2a65e491a0b.

I also saw my plan through, the one that turned @nigoroll in inquisitive mode. It is implemented in 475c38b664d8cad1e397386ac77e225ba1f0dcb9 and isn't completely a special case like I originally intended. It turned out to be generic because I needed both `=` and `+=` support after fixing the bug mentioned above.

However, while the two changes to `vcc_act_set()` are in a merge-able state, the latter is build on top of e9051fd54d697a94c9275f8a62faadb3050a17e5 that needs further work.

Dridi:
Summary so far: I think everything except e9051fd54d697a94c9275f8a62faadb3050a17e5 and 475c38b664d8cad1e397386ac77e225ba1f0dcb9 can be merged already.

Dridi:
Done, I will open a separate pull request for the `BODY` part.

Dridi:
Rebased against master, because I'm working on the `BODY` part in another branch that needs to be rebased against master to solve recently introduced conflicts.

Dridi:
Regarding today's bugwash, in relation to the VIPs 9 and 28 discussion, and when I picked it up the ongoing bugwash: this pull request is one step in the ambitious "no special case type system" direction that @bsdphk described.

I'd like to move forward with this because:

- it adds `+=` support to STRING
- introduces `+=` to BODY to fix the behavior of `=`
- centralizes what used to be a mixed bag of special and very special cases

After this and #3123 I can work on gradually erasing special cases throughout the parser, but this is not making the situation worse in my opinion.

PR #3117, opened 2019-11-05
fix PRIV_TOP for real with preallocation

### LINKED ISSUE #3019: VCL Return + Restart causes assertions in CNT_Embark on 	AZ(req->vcl0);
## Expected Behavior
Avoid panicking and working correctly

## Current Behavior
```
Panic at: Thu, 27 Jun 2019 16:15:21 GMT
Assert error in CNT_Embark(), cache/cache_req_fsm.c line 1023:
  Condition((req->vcl0) == 0) not true.
version = varnish-6.2.0 revision b14a3d38dbe918ad50d3838b11aa596f42179b54, vrt api = 9.0
ident = Linux,4.9.0-9-amd64,x86_64,-junix,-smalloc,-sdefault,-hcritbit,epoll
now = 615633.120472 (mono), 1561652121.833952 (real)
Backtrace:
  0x560e65033ec9: /usr/sbin/varnishd(+0x4eec9) [0x560e65033ec9]
  0x560e6509e9c0: /usr/sbin/varnishd(VAS_Fail+0x40) [0x560e6509e9c0]
  0x560e65037e16: /usr/sbin/varnishd(CNT_Embark+0xc6) [0x560e65037e16]
  0x560e65060dd3: /usr/sbin/varnishd(+0x7bdd3) [0x560e65060dd3]
  0x560e650563d5: /usr/sbin/varnishd(+0x713d5) [0x560e650563d5]
  0x560e65056910: /usr/sbin/varnishd(+0x71910) [0x560e65056910]
  0x7f3ce6755fa3: /lib/x86_64-linux-gnu/libpthread.so.0(+0x7fa3) [0x7f3ce6755fa3]
  0x7f3ce66864cf: /lib/x86_64-linux-gnu/libc.so.6(clone+0x3f) [0x7f3ce66864cf]
thread = (cache-worker)
pthread.attr = {
  guard = 4096,
  stack_bottom = 0x7f3c98abc000,
  stack_top = 0x7f3c98afc000,
  stack_size = 262144,
}
thr.req = 0x7f3c8c186020 {
  vxid = 688132, transport = HTTP/1 {
    state = HTTP1::Proc
  }
  step = R_STP_LOOKUP,
  req_body = R_BODY_NONE,
  restarts = 0, esi_level = 0,
  sp = 0x7f3ccad85020 {
    fd = 78, vxid = 688129,
    t_open = 1561652118.305774,
    t_idle = 1561652121.454280,
    ws = 0x7f3ccad85060 {
      id = \"ses\",
      {s, f, r, e} = {0x7f3ccad85098, +96, (nil), +261984},
    },
    transport = HTTP/1 {
      state = HTTP1::Proc
    }
    client = 192.168.35.49 36214 :80,
  },
  worker = 0x7f3c98afad80 {
    ws = 0x7f3c98afae28 {
      id = \"wrk\",
      {s, f, r, e} = {0x7f3c98afa380, +0, (nil), +2040},
    },
    VCL::method = HASH,
    VCL::return = lookup,
    VCL::methods = {},
  },
  ws = 0x7f3c8c186170 {
    id = \"req\",
    {s, f, r, e} = {0x7f3c8c1880b0, +664, +515912, +515912},
  },
  http_conn = 0x7f3c8c188050 {
    fd = 78 (@0x7f3ccad85044),
    doclose = NULL,
    ws = 0x7f3c8c186170 {
      [Already dumped, see above]
    },
    {rxbuf_b, rxbuf_e} = {0x7f3c8c1880b0, 0x7f3c8c188296},
    {pipeline_b, pipeline_e} = {(nil), (nil)},
    content_length = -1,
    body_status = none,
    first_byte_timeout = 0.000000,
    between_bytes_timeout = 0.000000,
  },
  http[req] = 0x7f3c8c186218 {
    ws = 0x7f3c8c186170 {
      [Already dumped, see above]
    },
    hdrs {
      \"GET\",
      \"/library/de_persgroep/fonts/2533/museo-sans-300.ttf\",
      \"HTTP/1.1\",
      \"Accept: */*\",
      \"Referer: ***** Edited *****\",
      \"Accept-Language: nl-BE\",
      \"User-Agent: Mozilla/5.0 (***** Edited *****\",
      \"Origin: https://actie.demorgen.be\",
      \"Host: files.qualifio.com\",
      \"Accept-Encoding: gzip\",
      \"X-Varnish: 14439596\",
      \"X-Forwarded-For: ***** Edited *****\",
      \"X-Original-Url: /library/de_persgroep/fonts/2533/museo-sans-300.ttf\",
    },
  },
  vdc = 0x7f3c8c188028 {
    nxt = (nil),
    retval = 0,
  },
  vcl = {
    name = \"vcl_files_20190627_161453\",
    busy = 81,
    discard = 0,
    state = warm,
    temp = warm,
    conf = {
      syntax = \"41\",
      srcname = {
        \"/etc/varnish/conf-enabled/default_files.vcl\",
        \"Builtin\",
        \"/etc/varnish/conf/common_purge.vcl\",
        \"/etc/varnish/conf/common_security_main.vcl\",
        \"/etc/varnish/conf/common_security_cleanup.vcl\",
        \"/etc/varnish/conf/common_security_clean_debug.vcl\",
        \"/etc/varnish/conf/common_security_filter_cfide.vcl\",
        \"/etc/varnish/conf/common_security_server_name.vcl\",
        \"/etc/varnish/conf/common_security_xss.vcl\",
      },
    },
  },
  vcl0 = {
    name = \"router_20190627_161453\",
    busy = 43,
    discard = 0,
    state = warm,
    temp = warm,
    conf = {
      syntax = \"41\",
      srcname = {
        \"/etc/varnish/build/router.vcl\",
        \"Builtin\",
        \"/etc/varnish/conf/common_security_sni_match.vcl\",
        \"/etc/varnish/conf/common_guru_meditation.vcl\",
      },
    },
  },
  flags = {
  },
  privs = 0x7f3c8c186200 {
  },
  topreq = {
    req = 0x7f3c8c186020 {
      [Already dumped, see above]
    },
  },
},
thr.busyobj = (nil) {
},
vmods = {
  directors = {Varnish 6.2.0 b14a3d38dbe918ad50d3838b11aa596f42179b54, 0.0},
  dynamic = {Varnish 6.2.0 b14a3d38dbe918ad50d3838b11aa596f42179b54, 0.0},
  std = {Varnish 6.2.0 b14a3d38dbe918ad50d3838b11aa596f42179b54, 0.0},
  proxy = {Varnish 6.2.0 b14a3d38dbe918ad50d3838b11aa596f42179b54, 0.0},
},
```
## Possible Solution
I suspect the cnt_restart should implement the same logic as esi to move back to the original vcl0 and remove it from here.

This is initial bug, i'll provide more detail as i read through the code, but my intution is that a pair of vcl like this one should cause the crash : 
## first_vcl
```
sub vcl_recv {
  return ("second_vcl");
}
```
## second_vcl
```
# RoutingKey: req.url == "/varnish-crash"
# VclLabel: vcl_varnish_crash
# Priority: 1
vcl 4.1;

# Dummy, MUST NOT BE USED
backend default { .host = "127.0.0.1"; }

include "conf/builtin/synth.vcl";

sub vcl_recv {
  if (req.restarts==0) {
    return (restart);
  }
  return (synth(200, "OK - restarted " + req.restarts));
}

sub vcl_synth {
    call builtin_vcl_synth;
    return (deliver);
}
```

## Your Environment

* Version used: 6.2.0 (cloudpackage)
* Operating System and version: Debian buster
* Source of binary packages used (if any)

This ticket will be updated as i can confirm my initial investigation.
thanks a lot for your support

nigoroll:
This is a follow-up to #3011 based on private feedback from @bsdphk.

The main difference is that #3011 added the `struct topreq` only when needed (which is, for vcl switching, when `PRIV_TOP` was used and for `esi_level >= 1`). The main drawback of this dynamic is that, as the allocation happens after the rollback checkpoint of the workspace gets set, `struct topreq` needs to be relocated during rollback. (see https://github.com/varnishcache/varnish-cache/blob/5bd46f4b0504e9a551225b756f2d538fa51b4182/bin/varnishd/cache/cache_req.c#L296)

To avoid this relocation, this PR goes another route: the  `struct topreq` and, consequently, the `PRIV_TOP` are now always allocated eagerly, so they are always available.

The amount of workspace used is roughly identical to is if  `struct topreq` was embedded in  `struct req`, but this way making `esi_level >= 1` requests point to the right `struct topreq` is simpler.

regarding the overall motivation:

In https://github.com/varnishcache/varnish-cache/commit/86af5ce095a6e39e313cf7c6c23a49edba4aace6 I made the mistake to overlook the fact that `PRIV_TASK` for a req and `PRIV_TOP` share the same priv tree. This was noticed in #3003 and fixed in https://github.com/varnishcache/varnish-cache/commit/d6d3416090b8b7db62d2cd0958e756e23ea92352.

There already exists a regression test for r03003

With this PR, I propose to remove the additional `(struct vrt_priv).id`  again, which came back with that regression fix, because it is not needed: All other privs already got different heads, so by giving `PRIV_TOP` a separate head also, we reduce memory overhead, simplify and clean up:

* save 8 bytes per priv: no need for duplication of the head id in each priv
* simpler lookup: no need to compare that head id
* simpler rollback: for rolling back req/topreq, we can simply fini all of the tree and be done
  * side note: This is broken in current code
* Also, as the scope differs, I think separate heads are just the right thing.

Because the (re)added `struct reqtop` is the right place for it, we also move the `vcl0` there. Being at it, I also fixed #3019 in the original PR #3011 (which came in later). This fix is now properly labeled and comes with @Dridi s PR in a single commit.


Dridi:
I'm not ready to review this yet, but it now has merge conflicts.

nigoroll:
@Dridi the merge conflict from 7532c3a7ed6c12635af1a2278e024fe61a351568 is resolved.
This is basically the same thing as #3011 except that `struct reqtop` is now always allocated such that we do not need the relocation during rollback.

Dridi:
@nigoroll you have a new merge conflict, linting.

bin/varnishd/cache/cache_vpi.c
```diff
@@ -94,6 +94,11 @@ VPI_vcl_select(VRT_CTX, VCL_VCL vcl)
...
 		return;		// Illegal, req-FSM will fail this later.
 
+	/* XXX VCL_Task* are somewhat duplicated to those in Req_Rollback called
+	 * from FSM for VCL_RET_VCL. Keeping them here to ensure there are no
+	 * tasks during calls to VCL_Rel / vcl_get
+	 */
+	VCL_TaskLeave(req->vcl, req->top->privs);
 	VCL_TaskLeave(req->vcl, req->privs);
 	if (IS_TOPREQ(req)) {
 		AN(req->top);
```
Dridi:
```c
CHECK_OBJ_NOTNULL(req->top, REQTOP_MAGIC);
```

bin/varnishd/cache/cache_vpi.c
```diff
@@ -94,6 +94,11 @@ VPI_vcl_select(VRT_CTX, VCL_VCL vcl)
 	if (IS_TOPREQ(req) && req->top->vcl0 != NULL)
 		return;		// Illegal, req-FSM will fail this later.
 
+	/* XXX VCL_Task* are somewhat duplicated to those in Req_Rollback called
+	 * from FSM for VCL_RET_VCL. Keeping them here to ensure there are no
+	 * tasks during calls to VCL_Rel / vcl_get
+	 */
```
Dridi: I'm not sure I'm comfortable with the need to use `VCL_Task*` twice for the top request. I haven't looked in details yet but maybe we need a `Req_(Enter|Leave)` functions that call `VCL_Task*` twice for each set of PRIVs?

This feels brittle and something we might forgot during a refactoring. Maybe we really need to make the `VCL_Task*` functions static and force ourselves to go through the right `Bo_` or `Req_` variants that would act as wrappers?

bin/varnishtest/tests/r02219.vtc
```diff
@@ -58,14 +58,13 @@ client c2 {
 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42
 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42
 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42
-42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42 42
-42 42 42 42 42 42 42 42 42 42 42 42 42 42
+42 42 42 42 42 42 42
```
Dridi: Did you test this on a 32bit build? I haven't reached the build-on-my-machine step of the review yet.

bin/varnishd/cache/cache_vrt_priv.c
```diff
@@ -178,7 +167,6 @@ VRT_priv_top(VRT_CTX, const void *vmod_id)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	if (ctx->req == NULL) {
-		/* XXX: should we VRT_fail here instead? */
```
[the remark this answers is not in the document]
nigoroll: I do not at all see why it would ever be ok to call VRT_priv_top from anywhere but the client side

bin/varnishd/cache/cache_vpi.c
nigoroll: I am going to add it, but I guess this is the kind of assertion we do not need that badly. If the `req` (and the `reqtop` is part of it) was broken, we would know much sooner.

bin/varnishtest/tests/r02219.vtc
nigoroll: I still see vtest as part of the process. I will know soon enough if I got this wrong.

bin/varnishd/cache/cache_vpi.c
nigoroll: Yes, we should clear this up. I will issue another PR.

PR #3122, opened 2019-11-12
Add Ctrl-f and Ctrl-b for pagedown/up in varnishstat.

lkarsten:
This adds another vim-style key binding to varnishstat.

Couldn't find any good way to extend u00008.vtc to test it. The
cursor is always at column 22 and page-downing to the bottom
will break the test case when we add another counter.

### DIFF
--- a/bin/varnishstat/varnishstat_curses.c
+++ b/bin/varnishstat/varnishstat_curses.c
@@ -62,6 +62,8 @@
 
 #define VALUE_MAX		999999999999
 
+#define CTRL(c) ((c) & 037)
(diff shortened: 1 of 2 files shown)


Dridi:
This brings the number of key bindings to 3 for both actions, so I'm not sure what the benefit is. Regardless, this lacks an update of the manual to reflect this change.

I'm wondering whether we could reuse the `CTRL()` macro for `^C`, `^T` and `^Z`.

lkarsten:
Added documentation update.

Issue #3112, opened 2019-11-04
Type specific methods, where do we go ?


bsdphk:
This is a discussion ticket for todays bugwash.

Now that we have type specific methods, where do we take them ?

    some.expression.yielding.backend.resolve() ?

    something.http.cache-control.collect() ?

Do we think that would be an improvement on VCL syntax ?

Please note: This discussion is 100% and only about VCL syntax, not the precise implementation or semantics of the examples proffered.

Dridi:
Related to #3103.

Dridi:
A bit of context, true as of today:

- some types now have `.properties` or `.methods()`
- expressions can be simple like `beresp.storage.free_space`
- they can be complex like `(req.http.host + req.url).lower()`
- there's no technical limit to dot-chaining method calls
  - example: (STRING expr).lower().upper().lower().upper()...

Bugwash summary:

I'm only reporting what I believe was the final consensus on various topics, I took the liberty of adding a couple more suggestions that didn't come up during the bugwash.

- nobody against adding more methods to types
  - general consensus is yes, that would be an improvement
- methods with arguments would definitely be more useful
- we didn't discuss type methods returning VOID
  - in other words, statements: `(expr).method(...);`
  - example: `something.http.cache-control.collect();`
- no clear consensus on vmods having the ability to add type methods
  - @bsdphk suggested a syntax: `(DURATION expr).std.format("%.3f")`
  - I suggested that any function taking a TYPE as its first parameter could translate into a type method
  - for phk's example in vmod_std `$Function STRING format(DURATION, STRING)` would be enough to recycle a plain function into a type method
  - otherwise we might prefer a new dedicated stanza (for example `$TypeMethod`)
  - this syntax opens a can of namespace worms regardless of the VCC implementation
    - post-bugwash alternative proposal: `(DURATION expr).std::format("%.3f")`

To be revisited next week.

nigoroll:
+1 for @Dridi s suggestion _that any function taking a TYPE as its first parameter could translate into a type method_
I am not quite up to speed with the type methods yet, but I still wonder if the dot is the best choice to make the syntax readable and unambiguous.
Why exactly did we decide for it and against, for example, the arrow `->`?
I would find
```resp.http.cache-control->std.collect()->myvmod.fixup_ttl()```
more readable than
```resp.http.cache-control.std.collect().myvmod.fixup_ttl()```

nigoroll:
[...]

PR #3123, opened 2019-11-12
Turn *.body variables into BLOBs

### LINKED ISSUE #3114: Inserting binary objects in vcl_backend_error

Dridi:
For this we need the BLOB type to be able to implicitly cast from a
STRING before so that any type capable of tosring may implicitly be
turned into a BLOB.

A blob generated on the fly is declared on the stack similar to a
STRANDS and only generated C code may turn a struct vrt_blob * into
a (const) VCL_BLOB usable by VRT and VMODs.

Unlike a STRANDS, the BLOB contents need to be collected so this time
it's libvcc's job to ensure that instead of VPI_Blob gets a complete
string altogether.

Some test cases needed to be updated to deal with this new capability
but now that ${TYPE} conversions to BLOB are implicit, we also use less
workspace than previously. So some vmod_blob tests also needed to be
further adjusted and some cases were marked with XXX comments for
further inspection. In particular, some test cases predating the type
change from struct vmod_priv to struct vrt_blob were still referring
to the old type.

The BODY type is still present, so *.body variables still have a type
more specific than BLOB. This was needed to keep supporting the +=
operator for the set action and doesn't have any practical effect on
bereq.body that can only be unset.

This also introduces an `std.blobread()` function instead of a function
in vmod_blob to allow sharing the logic and file cache.

Some of the patches of this pull request come from #3116.

Fixes #3114


Dridi:
I think this deserves a break down of the commit log.

Part of #3100, approved by @nigoroll and pending @bsdphk's review:

* a243fc40b Delay the code generation of the set action
* 0bedd1e67 Teach expressions to the set action's arithmetic table
* d5bea9418 Teach libvcc how to set STRING += STRING;
* c829bee71 Generalize more special cases in the set action
* 3b9e32cf7 Repurpose the arithmetic table to an assignment table
* 3b183bfc2 Teach libvcc how to set HEADER += STRING_LIST;
* 380f6793b Polish the generated C code for the set action
* c0dbe02a1 Fix the behavior of `set BODY = STRING;`

libvcc polish that I deem useful, not changing behavior:

* d39920e08 Avoid direct access to (struct vcc).sb
* 38e4732cb GC unused argument
* 91526098a That was just a warning

The bugwash consensus (from https://github.com/varnishcache/varnish-cache/issues/3114#issuecomment-552462385), the important part:

* 509b0ad15 Turn *.body variables into BLOBs
* 28c59e8ec Kill BLOB-specific error message in libvcc

My take on solving #3114, also important but trivial:

* 7413f3e70 Keep track of the file size in std.fileread()
* c5a11d540 Move the std.fileread() logic to its own function
* 6ecfc8259 New std.blobread() function

Problems caught by Travis CI:

* 953a491c2 SQUASHME: Fix portability issue
* 687408740 SQUASHME: #include <stdarg.h>

Dridi:
@slimhazard do you think you could have a closer look at the XXX comments from 509b0ad15e008c66b11c26817b8a92143aeff792?

slimhazard:
@Dridi done, let me know if you have any questions.

Dridi:
> (Although with some imagination, we could come up with implicit conversions from INT to BLOB, or just about any data type to BLOB. But if at all, that would be a subject for another day.)

This is already the case. Any type that may implicitly turn into a STRING can implicitly turn into a BLOB. I will use the HTTP type for that.

> ("BLOBs can only be used as arguments to VMOD functions") may no longer be accurate

This error message should be gone in 28c59e8ec725264eabb70a150d0a7eb5b4a04593 and I should probably move it one commit earlier.

For the GCOV part I suppose I will add a comment and check it locally. Thanks for your input!

slimhazard:
> > (Although with some imagination, we could come up with implicit conversions from INT to BLOB, or just about any data type to BLOB. But if at all, that would be a subject for another day.)
> 
> This is already the case. Any type that may implicitly turn into a STRING can implicitly turn into a BLOB. I will use the HTTP type for that.

Getting a bit off-topic here ... if it's already decided that an INT becomes a BLOB consisting of a row of ASCII digits, well then that's the solution. I was thinking of the value of the INT converted to binary, with some standardized endianness. Or say an IPv4 address as 4 bytes in network byte order, an IPv6 address as 16 bytes in network order, that sort of thing. Even an HTTP could have a binary format, if the format is specified.

Dridi:
> I was thinking of the value of the INT converted to binary, with some standardized endianness.

I'm personally not convinced by this case. But I also wasn't part of the bugwash when the consensus was reached that STRING could turn into a BLOB transitively.

> Or say an IPv4 address as 4 bytes in network byte order, an IPv6 address as 16 bytes in network order

Indeed, there could be a bridge between IP and BLOB. I think we'd need an explicit vmod_blob function to deal with it because IP falls through the STRING cast.

> Even an HTTP could have a binary format, if the format is specified.

This is basically how `obj.*` works today. Except that the HTTP type is limited to predefined variables (`req` and friends) even though it wouldn't be too hard to allow VMODs to manipulate HTTP expressions.

lib/libvcc/vcc_expr.c
```diff
@@ -189,7 +190,17 @@ vcc_expr_edit(struct vcc *tl, vcc_type_t fmt, const char *p, struct expr *e1,
...
 			    e3->nstr, tl->unique, tl->unique,
-			VSB_data(e3->vsb));
+			    VSB_data(e3->vsb));
+			tl->unique++;
+			break;
+		case 'B':
+			AZ(e2);
+			assert(e1->fmt == STRING);
+			VSB_printf(tl->curproc->prologue,
+			    "  struct vrt_blob blobs_%u;\n", tl->unique);
```
nigoroll: we cannot put blobs on the stack, this would break out contract:
https://github.com/varnishcache/varnish-cache/blob/fe687fbe9b9514e7c3a7d0d332404e8538d5a804/include/vrt.h#L191

lib/libvmod_std/vmod.vcc
```diff
@@ -202,6 +203,15 @@ case, you may need to modify the string, for example with
 
   set beresp.http.served-by = regsub(std.fileread("/etc/hostname"), "\R$", "");
 
+$Function BLOB blobread(PRIV_CALL, STRING)
```
nigoroll: I am not so happy with adding another variant of `std.fileread()` which has some relevant drawbacks, see motivation in https://code.uplex.de/uplex-varnish/libvmod-file/blob/master/README.rst
I would rather consider adding something like the above to varnish-cache

nigoroll:
> Indeed, there could be a bridge between IP and BLOB. I think we'd need an explicit vmod_blob function to deal with it because IP falls through the STRING cast.

Yes. We should have blob.serialize() the std.type functions also take blobs.

lib/libvcc/vcc_expr.c
Dridi: Good catch, that should be easy to fix: pass a `VRT_CTX` to `VPI_BLOB` instead of the on-stack blob placeholder. Will do.

lib/libvmod_std/vmod.vcc
Dridi: I agree, but this is something we can get built-in for free by leveraging the `std.fileread()` implementation. This is why I went for this approach, and anything more "serious" can rely on third-party VMODs.

Dridi:
> Besides my specific comment, I think this PR should really be broken up into smaller chunks.
Let us please get some of the prerequisite patches merged first.

Well, half of this PR is #3116, so once/if we have/had this in tree the commit list will/would shrink dramatically. I'm half confident with my last statement because I have half the approvals required to merge ;)

lib/libvcc/vcc_expr.c
nigoroll: why not just use `VRT_blob()` ?

Dridi: Because `VPI_Blob` handles the null case, and of course it can then call `VRT_blob`. I can't simply call it with `strlen()` for the length argument.

Dridi: 764207b325094a89d53af135d887ebfda41a17d8

bin/varnishd/cache/cache_vrt.c
```diff
@@ -966,17 +966,20 @@ VRT_blob(VRT_CTX, const char *err, const void *src, size_t len, unsigned type)
...
+};
+
 VCL_BLOB
-VPI_Blob(struct vrt_blob *b, const char *s)
+VPI_blob(VRT_CTX, VCL_STRING s)
 {
 
 	if (s == NULL)
-		s = "(null)";
+		return (&null_string_blob);
```
nigoroll: we should move this to VRT_blob()

Dridi: VRT_blob() doesn't take a STRING argument.

nigoroll: oh sorry, yes. /me bad

Dridi:
I reviewed the m49 VTC and initially fixed it in this branch. Then I realized it was also portable to the master branch, so I pushed my changes to `vtc.typesize()` and m49 directly:

https://github.com/varnishcache/varnish-cache/compare/4495a0e0b...175af0b91

This pull request was rebased against master, and below is the new distribution of patches.

Part of #3116, approved by @nigoroll and pending @bsdphk's review:

* a8acd8c47 Delay the code generation of the set action
* 2ec0817c3 Teach expressions to the set action's arithmetic table
* f72f21970 Teach libvcc how to set STRING += STRING;
* cdd9c4438 Generalize more special cases in the set action
* b636e56d7 Repurpose the arithmetic table to an assignment table
* cd4c42906 Teach libvcc how to set HEADER += STRING_LIST;
* 71ef1f88d Polish the generated C code for the set action
* ef5ac72ae Fix the behavior of `set BODY = STRING;`

The bugwash consensus (from #3114 (comment)), the important part:

* be57429c5 Back the BODY type with a BLOB
* 9accc5668 Kill BLOB-specific error message in libvcc

My take on solving #3114, also important but trivial:

* 5163548a0 Keep track of the file size in std.fileread()
* cb3db6ec2 Move the std.fileread() logic to its own function
* d5965dbfe New std.blobread() function

I left the libvcc polish patches aside as suggested by @nigoroll.

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -880,22 +880,17 @@ VRT_r_resp_do_esi(VRT_CTX)
...
-		if (p == NULL)					\
-			p = "(null)";				\
-		VSB_cat(vsb, p);				\
-		p = va_arg(ap, const char *);			\
-	}							\
-	va_end(ap);						\
+	if (type == LBODY_SET)					\
+		VSB_clear(vsb);					\
+	assert(type == LBODY_SET || type == LBODY_ADD);		\
+	VSB_bcat(vsb, blob->blob, blob->len);			\
```
nigoroll: **later!** not to hold up the PR!

pick up the "array of blobs" (I remember @bsdphk had a good name, but cannot remember which) idea and avoid the copying.

nigoroll:
rebased

Dridi:
#3116 was approved, so I will rebase this pull request once it's merged, but this one should wait.

For now we have a situation where we might not be able to fail properly when we run out of workspace. This was not a problem with `(be)?resp\.body` as a strands because everything would live in the stack and the VSB. So for this special case we might prefer that the `synthetic()` function still operates on a strands, possibly via a new VPI function.

I will keep you updated.

Dridi:
It turned out to be trivial to fix and even beneficial: the `synthetic(STRANDS)` action can stay as-is, it doesn't have to break since it doesn't \_directly\_ involve the BODY type.

So only actual variables with the BODY type now turn into BLOBs under the hood. The BODY type is still needed to account for the changes from #3116. Please see the details in the commit log.

One word on VIP28: this adds an explicit cast from STRINGS to BLOB in libvcc which goes against the idea of removing special cases and relying on self-contained types to know what operation to pick on a given expression.

This is particularly bad for the generalization goal because type X may convert to BLOB via STRING of X can turn into a STRING, so the transitive cast goes against what we wish to achieve in the long term. And to quote @bsdphk we don't want to turn libvcc into a SAT solver.

I don't know how we want to handle this, but the patch series is now functional and ready to be discussed.

nigoroll:
@Dridi testing your changes I noticed that std.fileread panics with https://github.com/varnishcache/varnish-cache/commit/7ae1bf244187520bc1514dd82d0bd7b1bbb9a5c1

nigoroll:
I would like to add that, in general, auto-casting strings to blobs will simplify VCL substantially when working with the `blob` and [blobdigest](https://code.uplex.de/uplex-varnish/libvmod-blobdigest) vmods where the string-to-blob conversion with `blob.decode(encoded=<string>)` will become redundant.

typical example:

```
diff --git a/src/tests/sha256.vtc b/src/tests/sha256.vtc
index d70ee6b..15e5883 100644
--- a/src/tests/sha256.vtc
+++ b/src/tests/sha256.vtc
@@ -74,7 +74,7 @@ varnish v1 -vcl {
                        set resp.status = 500;
                        return(deliver);
                }
-               if (!d4.update(blob.decode(encoded="digest"))) {
+               if (!d4.update("digest")) {
                        set resp.status = 500;
                        return(deliver);
                }
```

Dridi:
> auto-casting strings to blobs will simplify VCL substantially

Yes, this is what this PR does based on prior consensus from https://github.com/varnishcache/varnish-cache/issues/3114#issuecomment-552462385. My previous comment was more about conflicting goals we set \*afterwards\* in the aftermath of the VIP28 discussion. We already have a special case for automatic `tostring` cast, and using that to transitively cast any `tostring`-able type to BLOB goes in the opposite direction of removing special cases from libvcc.

For example somewhere in our test suite, with this pull request, we have an implicit cast from INT to BLOB via a `tostring` detour.

I'm certainly not saying we shouldn't, I'm objectively stating that there is a conflict of interest and we need to decide where we put the cursor between VCL and libvcc. If we want to carry on with implicit `tostring` to BLOB then I believe this PR is in a merge-able state.

nigoroll:
Apologies for adding more to the TODO list, but also:
Do we _really_ want the transitive `_tostring` or should other types rather cast to `BLOB` directly? If yes, with which endianness? And for the real-types by which representation? I remember @slimhazard once digging out that the binary representation of floating point numbers was by no means portable.

My take:
* one argument for directly casting to `BLOB` would be digests. Depending on the use case, one might want to hash the binary representation or a stringification. To allow for both, we basically have the choice between requiring either a `.as_blob()` or `.as_string()` call. In my mind, the POLA option is `hashfunc(intvar.as_string())` hashing the stringification and `hashfunc(intvar)` hasing the integer, so at this point I think not doing the transient cast actually makes more sense.
* regarding the binary representation, we could add explicit representation conversions for the cases where they are needed. I think the more common case would be that the same blob be re-used on the same architecture.

slimhazard:
Since my name came up ...

The binary representation of doubles is one of those issues of "what does the standard strictly say" vs "how is it actually done pretty much everywhere". The C standard does not strictly require IEEE754 binary64 as the binary format for doubles. But the fact is that on every platform that Varnish will ever be expected to support, that's how it's done. I believe that's more reliable than any of "ints are 32 bit, longs are 64 bit, and cachelines are 64 bytes", none of which are formally guaranteed, but happen to be true just about everywhere.

About the rest of the conversation (since I'm already here): using strings as a representation of information that is naturally another data type (like numbers, IP addresses etc) is a notoriously poor practice. Obviously it's going to be harder to specify a binary format for each of the VCL data types (say, an endianness for an INT-to-BLOB conversion), but I would be concerned if all X-to-BLOB conversions now become implicitly tostring(), so that it will be hard to back out of that at a later time.

Think of all the times that the Varnish project has refrained from jumping ahead to a new feature, because of the possible long-term implications. This seems like a good place to exercise that kind of caution.

Note that IP addresses have well-known and entirely specific binary representations. Maybe it's just the numeric types (REAL and INT) that present any kind of issue here?

nigoroll:
@slimhazard please note that the `VCL_IP` type is backed by `struct suckaddr`.

Dridi:
There are several things to unpack here.

On the question of transitive casting from type X to BLOB via STRING, maybe that wouldn't apply to this topic because technically there is still a BODY type in this pull request. The fact that it uses `struct vrt_blob` under the hood is almost an implementation detail.

The devel being in the details, it does matter at this point.

So instead that would be an automatic transitive cast from type X to BODY via STRING. This needs to happen if we want to prevent existing code from breaking.

On the question of casting types to BLOBs and the poor practice of stringing all the the things, I agree on principle but not with the proposal. Again, in the aftermath of the VIP28 discussion I would rather add more type methods and have something like:

```
IP.[to]blob()
```

The "to" part of the method name would not be here to be consistent with `STRING.lower()` vs `std.tolower()`. So suddenly I could write things like `client.ip.blob()`.

If we increase the capabilities of type methods (see #3103 and #3112) we could go as far as:

```
FLOAT.blob($Enum {IEEE754, ...} fmt)
INT.blob($Enum {be, le} endian=be)
```

From this morning on IRC in response to https://github.com/varnishcache/varnish-cache/pull/3147#issuecomment-561064504;

> 10:37  \* phk wonders if we should simply rename vmod_blob

Why not simply obsolete (part of) vmod_blob? We can replace things, maybe not everything. This is essentially what happened for `std.toupper()` and `std.tolower()`.

```
BLOB.sub(BYTES length=-1, BYTES offset=0)
```

With a type method we can do things directly in VCL:

```vcl
# I need the client IP as a BLOB for reasons
# And I only want a prefix for GDPR reasons
client.ip.blob().sub(2b)
```

The list goes on once this is possible...

slimhazard:
@Dridi +1 that VIP28 might be the right way to do this -- in the spirit of other features added for VIP28, and it might make it easier to define the type-specific to-binary conversions. Would probably also be easier to add the conversions incrementally to the code.

And as sentimental as I obviously am about VMOD blob, ``.encode()`` and ``.decode()`` methods for BLOB in VIP28-style also seems like a natural choice.

@nigoroll ``struct suckaddr`` may be the *Varnish* implementation of VCL_IP, but I see that as Varnish's implementation detail, which doesn't have to be exposed to VCL users. IMO it would only be POLA if ``IP.blob()`` returns network-ordered address bytes, which would be easy enough to derive from a suckaddr. And the reverse direction, ``BLOB.ip()`` to convert address bytes to a suckaddr, should also be pretty straightforward.

nigoroll:
@Dridi I do not have to add much. My intention was to ensure that we are perusing the best possible option. If that is to implement the automatic transitive cast for backwards compatibility, I go along. In fact I thought this was the best option all the way, until I realized today that there even exists a case for the alternative.
Thank you

Dridi:
> IMO it would only be POLA if IP.blob() returns network-ordered address bytes

This is also what I had in mind when I wrote my example.

> I do not have to add much

Then I probably don't need to further explain why we need to reach consensus regarding the way to go :)

nigoroll:
@Dridi can you please rebase?

Dridi:
Apologies for the delay, rebased against master.

bsdphk:
I have cherry-picked the `std.blobread()` out of this PR, then I am going to give the rest of it a thumbs down for not going far enough.

We should make the bodies a linked list of {STRAND|BLOB|STRING} elements instead.

(It may make sense to add OBJCORE, so that req.body is no longer a special case.)

The list elements themselves will be tiny (three pointers + a flag) and can be allocated from WS.

I'm not comfortable with the automatic STRING->BLOB folding, and since making *.body a BLOB was how that came up in the first place, we can avoid it with the more general approach.

Dridi:
Superseded by #3247.

PR #3125, opened 2019-11-13
Add VDP_END action for delivery processors

nigoroll:
`VDP_END` marks the end of successful processing, it is issued by `VDP_DeliverObj()` and may also be sent downstream by processors ending the stream `(for return value != 0)`

`VDP_END` must at most be generated once per processor, so any VDP sending it downstream must itself not forward it a second time.

* `VDP_bytes()`

  The only relevant change is the assertion that `VDP_END` must not be received by any VDP more than once. This comes with minor restructuring of the handling of the three actions

* `VDP_DeliverObj()`

  Here, `VDP_END` is pushed down after succesful iteration over the objects. We could also send the last storage segment with `VDP_END`, but this would require a change in all stevedores. Unless there is a good case for it, I do not see much value in this minor optimization.

* ESI:

  In `ved_vdp_esi_bytes()` we now push down `VDP_END` whenever we are done with an ESI object at `esi_level == 0`.

  `ved_bytes()` handles the pushes from higher esi levels downwards, so it now changes `VDP_END` into `VDP_FLUSH`. We need to remove the additional `VDP_FLUSH` from `ved_deliver()` because the `VDP_END` from `VDP_DeliverObj()` needs to be the last bit sent and `ved_deliver()` runs at `esi_level > 0` only (it is a transport function).

  The `gzgz` VDP actually does something which would be impossible for an `esi_level == 0` VDP: push bytes from _fini. This could be reworked, but there is also no need to.

* range VDP

  Here we now send the `VDP_END` with the last segment before the end of the range is it.

  We also take the opportunity and eleminate null `VDP_bytes()` calls before the range is reached.

* rot13 debug VDP

  Here we need to ensure that we pass on `VDP_END`

### DIFF
--- a/bin/varnishd/cache/cache_deliver_proc.c
+++ b/bin/varnishd/cache/cache_deliver_proc.c
@@ -45,6 +45,13 @@
  * VDP_bytes will return that value directly without calling the next
  * processor.
  *
+ * VDP_END marks the end of successful processing, it is issued by
+ * VDP_DeliverObj() and may also be sent downstream by processors ending the
+ * stream (for return value != 0)
+ *
+ * VDP_END must at most be received once per processor, so any VDP sending it
+ * downstream must itself not forward it a second time.
+ *
  * Valid return values (of VDP_bytes and any VDP function):
  * r < 0:  Error, breaks out early on an error condition
  * r == 0: Continue
@@ -59,15 +66,23 @@ VDP_bytes(struct req *req, enum vdp_action act, const void *ptr, ssize_t len)
 
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 	vdc = req->vdc;
-	assert(act == VDP_NULL || act == VDP_FLUSH);
 	if (vdc->retval)
 		return (vdc->retval);
 	vdpe = vdc->nxt;
 	CHECK_OBJ_NOTNULL(vdpe, VDP_ENTRY_MAGIC);
-	vdc->nxt = VTAILQ_NEXT(vdpe, list);
 
-	assert(act > VDP_NULL || len > 0);
+	/* at most one VDP_END call */
+	assert(vdpe->end == VDP_NULL);
+
+	if (act == VDP_NULL)
+		assert(len > 0);
+	else if (act == VDP_END)
+		vdpe->end = VDP_END;
+	else
+		assert(act == VDP_FLUSH);
+
 	/* Call the present layer, while pointing to the next layer down */
+	vdc->nxt = VTAILQ_NEXT(vdpe, list);
 	retval = vdpe->vdp->bytes(req, act, &vdpe->priv, ptr, len);
 	if (retval && (vdc->retval == 0 || retval < vdc->retval))
 		vdc->retval = retval; /* Latch error value */
@@ -158,6 +173,8 @@ VDP_DeliverObj(struct req *req)
 	final = req->objcore->flags & (OC_F_PRIVATE | OC_F_HFM | OC_F_HFP)
 	    ? 1 : 0;
 	r = ObjIterate(req->wrk, req->objcore, req, vdp_objiterator, final);
+	if (r == 0)
+		r = VDP_bytes(req, VDP_END, NULL, 0);
 	if (r < 0)
 		return (r);
 	return (0);
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -413,17 +413,20 @@ ved_vdp_esi_bytes(struct req *req, enum vdp_action act, void **priv,
 				/* MOD(2^32) length */
 				vle32enc(tailbuf + 9, ecx->l_crc);
 
-				(void)VDP_bytes(req, VDP_NULL, tailbuf, 13);
+				retval = VDP_bytes(req, VDP_END, tailbuf, 13);
 			} else if (ecx->pecx != NULL) {
 				ecx->pecx->crc = crc32_combine(ecx->pecx->crc,
 				    ecx->crc, ecx->l_crc);
 				ecx->pecx->l_crc += ecx->l_crc;
+				retval = VDP_bytes(req, VDP_FLUSH, NULL, 0);
+			} else {
+				retval = VDP_bytes(req, VDP_END, NULL, 0);
 			}
-			retval = VDP_bytes(req, VDP_FLUSH, NULL, 0);
 			ecx->state = 99;
 			return (retval);
 		case 3:
 		case 4:
+			assert(act != VDP_END);
 			/*
 			 * There is no guarantee that the 'l' bytes are all
 			 * in the same storage segment, so loop over storage
@@ -473,6 +476,8 @@ ved_bytes(struct req *req, struct ecx *ecx, enum vdp_action act,
     const void *ptr, ssize_t len)
 {
 	req->acct.resp_bodybytes += len;
+	if (act == VDP_END)
+		act = VDP_FLUSH;
 	return (VDP_bytes(ecx->preq, act, ptr, len));
 }
 
@@ -757,6 +762,12 @@ ved_gzgz_fini(struct req *req, void **priv)
 	CAST_OBJ_NOTNULL(foo, *priv, VED_FOO_MAGIC);
 	*priv = NULL;
 
+	/* XXX
+	 * this works due to the esi layering, a VDP pushing bytes from _fini
+	 * will otherwise have it's own _bytes method called.
+	 *
+	 * Could rewrite use VDP_END
+	 */
 	(void)ved_bytes(req, foo->ecx, VDP_FLUSH, NULL, 0);
 
 	icrc = vle32dec(foo->tailbuf);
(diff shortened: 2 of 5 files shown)


bin/varnishd/cache/cache_deliver_proc.c
```diff
@@ -43,6 +43,13 @@
  * VDP_bytes will return that value directly without calling the next
  * processor.
  *
+ * VDP_END marks the end of successful processing, it is issued by
+ * VDP_DeliverObj() and may also be sent downstream by processors ending the
+ * stream (for return value != 0)
+ *
+ * VDP_END must at most be generated once per processor, so any VDP sending it
```
slimhazard: This comment paragraph would be good in cache_filter.h -- make it part of the VDP API's public contract (and commit to staying with it, or else a major version change is necessary).

But does it mean that a VDP must not call VDP_bytes() with act=VDP_END, unless it is initiating the end of delivery? Or should (must?) a VDP push down VDP_END if it receives VDP_END? In which case, the VDP must be prepared to flush.

Unless I'm misunderstanding something, it has to be the latter, doesn't it? Otherwise it's not clear to me how lower-level VDPs know that we're at VDP_END.

bin/varnishd/cache/cache_deliver_proc.c
```diff
@@ -57,15 +64,23 @@ VDP_bytes(struct req *req, enum vdp_action act, const void *ptr, ssize_t len)
...
 		return (vdc->retval);
 	vdpe = vdc->nxt;
 	CHECK_OBJ_NOTNULL(vdpe, VDP_ENTRY_MAGIC);
-	vdc->nxt = VTAILQ_NEXT(vdpe, list);
 
-	assert(act > VDP_NULL || len > 0);
+	/* at most one VDP_END call */
+	assert(vdpe->end == VDP_NULL);
+
+	if (act == VDP_NULL)
```
slimhazard: This also would be good to document in cache_filter.h (always push non-zero bytes when it's VDP_NULL).

mbgrydeland:
There is an old PR that touches on the same functionality (though we called this VDP_LAST instead of VDP_END), just linking in for reference.
https://github.com/varnishcache/varnish-cache/pull/2673

nigoroll:
@mbgrydeland right, I had already forgot about it though I did spend some time with it back then. Thank you for refreshing my memories.
Do you think this slightly different implementation would also suit your needs? If not, what would be missing?

nigoroll:
@slimhazard 
> I'd like to see cache_filter.h document more details about the contract of both the VDP and VFP APIs.

maybe you can just push your idea of how the documentation could look like?

bin/varnishd/cache/cache_deliver_proc.c
nigoroll: what I meant to say here is
* `VDP_END` must be the last thing pushed down
* every VDP must see at most one `VDP_END`. For error cases or premature ending further downstream, there can be no `VDP_END` before VDP close.

consequently,
* A VDP must only call `VDP_bytes()` with `act=VDP_END` with or after the last bytes
* after a VDP has pushed down `act=VDP_END`, it must ignore all other `.bytes` calls to itself

Thank you for pointing out that my documentation was lacking clarity. Does the above work for you?

And yes to moving it to cache_filter.h

bin/varnishd/cache/cache_deliver_proc.c
nigoroll: +1

nigoroll:
> @slimhazard
> 
> > I'd like to see cache_filter.h document more details about the contract of both the VDP and VFP APIs.
> 
> maybe you can just push your idea of how the documentation could look like?

or rather: Let's handle this PR and then work on the docs

slimhazard:
> or rather: Let's handle this PR and then work on the docs

+1

I could submit a PR with a proposal for API docs, just have to put it on my priority queue. %^)

bin/varnishd/cache/cache_deliver_proc.c
slimhazard: I would add, if this is the correct understanding: a VDP that receives VDP_bytes(act=VDP_END) with  MUST push down VDP_bytes(act=VDP_END) calls within that invocation.

Otherwise, it's mysterious to me how the lower-level VDPs know that we're at VDP_END.

nigoroll: @slimhazard yes that or fail.

bsdphk:
So the description says what this PR implements, but not why it does so.

What purpose does VDP_END serve, that is not already served by the VDP_FLUSH+fini combo ?

bsdphk:
With respect to @slimhazard's comments about thin documentation: Yes, absolutely.

Part of the reason is that we wanted to get a bit more experience to see if the API was actually right, before we documented it.

The in-tree users are a bit on the incestuos side, so they are not very good witnesses to the soundness of the API.

I'm obviously very interested to hear more about your experience using these APIs

nigoroll:
> So the description says what this PR implements, but not why it does so.

Right, sorry and thank you for pointing this out.

> What purpose does VDP_END serve, that is not already served by the VDP_FLUSH+fini combo ?

The use case is to send data after the end of input. This is required for things like
* compession (e.g. gzip tail)
* some encrytion
* anything involving buffering (e.g. pushing vdp data through a pipeline, which needs to be closed before the final data can be real)

With the current code, we cannot react on the "end of stream" and still generate data downstream because delivery is already closed when the fini callbacks get closed. This can be seen with this simple change and then running `m00048.vtc`:

```
diff --git a/lib/libvmod_debug/vmod_debug.c b/lib/libvmod_debug/vmod_debug.c
index 1a547e172..98b4978ea 100644
--- a/lib/libvmod_debug/vmod_debug.c
+++ b/lib/libvmod_debug/vmod_debug.c
@@ -142,7 +142,7 @@ xyzzy_rot13_bytes(struct req *req, enum vdp_action act, void **priv,
 static int v_matchproto_(vdp_fini_f)
 xyzzy_rot13_fini(struct req *req, void **priv)
 {
-       (void)req;
+       VDP_bytes(req, VDP_FLUSH, "*end*", 5);
        AN(priv);
        free(*priv);
        *priv = NULL;
```

Result:

```
***  v1   debug|Error: Child (7674) Panic at: Tue, 03 Dec 2019 10:30:46 GMT
***  v1   debug|Assert error in V1L_Write(), http1/cache_http1_line.c line 254:
***  v1   debug|  Condition((v1l) != NULL) not true.
```

Notice that `ved_gzgz_fini()` does something like this, but in this case `ved_bytes()` calls the parent ESI layer which is still open at this point.

So there are two obvious ways to handle this:
* keep delivery open until _fini gets called
* add something like `VDP_END` proposed herein.

IIUC I actually asked about this on IRC, but anyway I opted for `VDP_END` to parallel the fetch side, which has `VFP_END`

slimhazard:
Another issue is that, currently for VDPs, errors may not be reported in ``fini()`` -- there's an ``AZ()`` on the return value, so a VDP ``fini()`` method may not return anything but 0, and there's no other way to fail delivery in ``fini()``. For a use case for which errors may not become apparent until the last bytes have been pushed, you have to know in the ``bytes()`` method that we're at the end.

``VDP_FLUSH`` is not sufficient for that, because (as the docs in cache_filter.h do in fact say! \o/), it has to be used if the buffer is not valid past the ``bytes()`` call, usually because it's on the stack. That's not the same as what ``VDP_END`` would say.

nigoroll:
@mbgrydeland & @daghf , Dridi has passed the review-hat on to you. I suspect that you got similar code and I would be interested in an attempt to find a joint solution

mbgrydeland:
I've looked through this, and it looks to that this is mostly future compatible with our stuff, so I have no objections to this PR.

The PR in #2673 pretty closely shows what we ended up doing in this space. I think that patch set takes it a bit further, in that the VDP_END/VDP_LAST passing down the chain isn't an after-the-iterating-body extra step, but integrated into the stevedore object iterators. This point was touched upon in the description to this PR:

> VDP_DeliverObj()
>
>Here, VDP_END is pushed down after succesful iteration over the objects. We could also send the last storage segment with VDP_END, but this would require a change in all stevedores. Unless there is a good case for it, I do not see much value in this minor optimization.

There is a good use case for this, and something that I am reminded now that was the plan to address after/if #2673 was merged. In H/2 there is the need to send a frame with a last bit set to mark the end of the stream. Today, a separate empty frame is sent after the rest of the data stream on delivery in a separate system call, just to communicate that the stream has ended. A similar optimization would be possible for H/1 and chunked encoding, though it would be less of an impact there as the fast-path (cache-hit non-streaming delivery) is always by Content-Length.

So in conclusion I think maybe we want to resurrect at least parts of #2673 in addition to this PR.

nigoroll:
@mbgrydeland do I understand correctly that this PR could be merged as-is and then the improved `VDP_END` handling added on top?

mbgrydeland:
@nigoroll Yeah, that is my conclusion

nigoroll:
@mbgrydeland could I ask you to add an approving review please?

bin/varnishd/cache/cache_deliver_proc.c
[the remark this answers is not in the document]
nigoroll: right.

bin/varnishd/cache/cache_esi_deliver.c
```diff
@@ -750,6 +755,10 @@ ved_gzgz_fini(struct req *req, void **priv)
 	CAST_OBJ_NOTNULL(foo, *priv, VED_FOO_MAGIC);
 	*priv = NULL;
 
+	/* XXX
+	 * this is only possible due to the esi layering, a VDP must not
+	 * push bytes from fini. Could rewrite to work with VDP_END
+	 */
```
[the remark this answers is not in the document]
nigoroll: I will clarify the comment.

PR #3126, opened 2019-11-13
Vpi vcl select cleanup

nigoroll:
clean up and simplification of the vcl switching facility, motivated by https://github.com/varnishcache/varnish-cache/pull/3117#discussion_r345302497

Please see the individual commit messages for explanations


bin/varnishd/cache/cache_vrt_vmod.c
```diff
@@ -167,12 +167,14 @@ VPI_Vmod_Unload(VRT_CTX, struct vmod **hdl)
 {
 	struct vmod *v;
 
+	(void) ctx;
```
Dridi: If we no longer need `ctx` we might as well remove it from the function signature. Breaking VPI is safe by definition.

PR #3130, opened 2019-11-13
Fix director list membership and temperature events

### LINKED ISSUE #3094: panic: VRT_DelDirector from VCL Event would deadlock / director locking wrong?
seen during a vcl reload with https://github.com/nigoroll/libvmod-dynamic

calling `VRT_DelDirector()` from a `VCL_COLD` event handler leads to a panic because the `temp_rwl` lock is already held when the director event handler is to be called here:

https://github.com/varnishcache/varnish-cache/blob/1f226e6d8c1c5443d9df127797c62b1e1228bfd1/bin/varnishd/cache/cache_vrt_vcl.c#L222

While this could be avoided by additional delayed destruction in the director vmod, I wonder if holding the `temp_rwl` for `VDI_Event()` is actually the right thing to do, and then if we actually need the `temp_rwl` in the first place.:

* for the actual director insertion/removal, we use the `vcl_mtx`
* we write-lock `temp_rwl` when we change the temperature in `vcl_set_state()`
* we read-lock `temp_rwl` when adding and removing directors (plus for assertions on the temperature)

We post `VDI_Event()` for adding/removing directors and from `vcl_BackendEvent()`, which is called from `vcl_set_state()`

So the purpose is to ensure that the warm/cold events from construction/destruction do not interfere with VCL temperature events.

But `vcl_BackendEvent()` only posts events for directors on the `director_list`, so it seems to me that if we get the list members right, we should not need the additional lock.

`vcl_BackendEvent()`, on the other hand, currently walks the `director_list` not holding `vcl_mtx`, so aren't we already failing to properly handle director addition/deletion concurrent to VCL events?

At this point, the following would appear sensible to me:
* hold `vcl_mtx` whenever we handle the `director_list`, IIUC the missing cases are `vcl_BackendEvent() and vcl_KillBackends()`
* by ensuring that only backends on the `director_list` receive events, we can then kill `temp_rwl`: As events happen in the cli thread (`ASSERT_CLI()`), we do not need another layer of serialization

Feedback welcome

```
Panic at: Thu, 10 Oct 2019 11:02:07 GMT
Assert error in VRT_DelDirector(), cache/cache_vrt_vcl.c line 222:
  Condition(((*__errno_location ())=pthread_rwlock_rdlock(&vcl->temp_rwl)) == 0) not true.
version = varnish-trunk revision 3fcfd4b8feb289836762902b14750d345e75ab78, vrt api = 9.0
ident = Linux,3.10.0-862.14.4.el7.x86_64,x86_64,-jnone,-smalloc,-sdefault,-hcritbit,epoll
now = 27968678.416989 (mono), 1570705183.817889 (real)
Backtrace:
  0x445790: /opt/varnish/sbin/varnishd() [0x445790]
  0x445db1: /opt/varnish/sbin/varnishd() [0x445db1]
  0x4cd44b: /opt/varnish/sbin/varnishd(VAS_Fail+0x4d) [0x4cd44b]
  0x46ac40: /opt/varnish/sbin/varnishd(VRT_DelDirector+0x207) [0x46ac40]
  0x7fc0029c4747: ./vmod_cache/_vmod_dynamic.ac90823754667b39adf868c2db7f41ad3d09769cddb736a57629336ce699ce98(+0x4747) [0x7fc0029c4747]
  0x7fc0029c6e8f: ./vmod_cache/_vmod_dynamic.ac90823754667b39adf868c2db7f41ad3d09769cddb736a57629336ce699ce98(vmod_event+0x37f) [0x7fc0029c6e8f]
  0x7fbf0f7fa565: vcl_reload-varnish_20191009135156_248082.1570629116.539489/vgc.so(+0x41565) [0x7fbf0f7fa565]
  0x4573c4: /opt/varnish/sbin/varnishd() [0x4573c4]
  0x458af1: /opt/varnish/sbin/varnishd() [0x458af1]
  0x459cf4: /opt/varnish/sbin/varnishd() [0x459cf4]
errno = 35 (Resource deadlock avoided)
thread = (cache-main)
pthread.attr = {
  guard = 0,
  stack_bottom = 0x7ffe0aef7000,
  stack_top = 0x7ffe0b6f6000,
  stack_size = 8384512,
}
thr.req = (nil) {
},
thr.busyobj = (nil) {
},
vmods = {
  std = {0x7fc0204c60f0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  header = {0x7fc0204c61d0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 9.0},
  cookie = {0x7fc0204c6240, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 9.0},
  re = {0x7fc0204c62b0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 9.0},
  re2 = {0x7fc0204c6320, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 9.0},
  dispatch = {0x7fc0204c6390, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  blob = {0x7fc0204c6400, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  blobdigest = {0x7fc0204c6470, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  blobsynth = {0x7fc0204c64e0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  *one private vmod removed*,
  dns = {0x7fc0204c65c0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 9.0},
  directors = {0x7fc0204c6630, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  all_healthy = {0x7fc0204c66a0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  cluster = {0x7fc0204c6710, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  constant = {0x7fc0204c6780, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  taskvar = {0x7fc0204c67f0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  querystring = {0x7fc0204c6860, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  vsthrottle = {0x7fc0204c68d0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 9.0},
  geoip2 = {0x7fc0204c6a20, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  dynamic = {0x7fc0204c7900, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  hoailona = {0x7fc0206044e0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  frozen = {0x7fbfbc5ca320, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
  crypto = {0x7fbfbc7f70b0, Varnish trunk 3fcfd4b8feb289836762902b14750d345e75ab78, 0.0},
},
```

nigoroll:
this PR fixes #3094 

We add missing locking for operations on the director list to ensure consistency with concurrent dynamic director additions and deletions.
We remove the temperature lock which prevented director deletions during COLD events and ensure that only backends on the director list receive events.

Please note that there is no explicit vtc, the test case has been added to vmod_debug and runs with every invocation of that vmod.

Please consider the individual commit messages for additional detail on the reasoning.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -604,7 +604,7 @@ VRT_new_backend_clustered(VRT_CTX, struct vsmw_cluster *vc,
 
 	if (be->director != NULL) {
 		/* for cold VCL, update initial director state */
-		if (be->probe != NULL && ! VCL_WARM(vcl))
+		if (be->probe != NULL && ! VCL_WARM(vcl->temp))
 			VBP_Update_Backend(be->probe);
 
 		Lck_Lock(&backends_mtx);
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -278,6 +278,7 @@ vcl_iterdir(struct cli *cli, const char *pat, const struct vcl *vcl,
 	int i, found = 0;
 	struct vcldir *vdir;
 
+	Lck_AssertHeld(&vcl_mtx);
 	VTAILQ_FOREACH(vdir, &vcl->director_list, list) {
 		if (fnmatch(pat, vdir->cli_name, 0))
 			continue;
@@ -343,8 +344,10 @@ vcl_BackendEvent(const struct vcl *vcl, enum vcl_event_e e)
 	CHECK_OBJ_NOTNULL(vcl, VCL_MAGIC);
 	AZ(vcl->busy);
 
+	Lck_Lock(&vcl_mtx);
 	VTAILQ_FOREACH(vdir, &vcl->director_list, list)
 		VDI_Event(vdir->dir, e);
+	Lck_Unlock(&vcl_mtx);
 }
 
 static void
@@ -355,6 +358,7 @@ vcl_KillBackends(struct vcl *vcl)
 	CHECK_OBJ_NOTNULL(vcl, VCL_MAGIC);
 	AZ(vcl->busy);
 	assert(VTAILQ_EMPTY(&vcl->ref_list));
+	Lck_Lock(&vcl_mtx);
 	while (1) {
 		vdir = VTAILQ_FIRST(&vcl->director_list);
 		if (vdir == NULL)
@@ -366,6 +370,7 @@ vcl_KillBackends(struct vcl *vcl)
 		FREE_OBJ(vdir->dir);
 		FREE_OBJ(vdir);
 	}
+	Lck_Unlock(&vcl_mtx);
 }
 
 /*--------------------------------------------------------------------*/
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -265,7 +265,7 @@ vcl_get(struct vcl **vcc, struct vcl *vcl)
 	(*vcc)->busy++;
 	Lck_Unlock(&vcl_mtx);
 	AZ(errno=pthread_rwlock_rdlock(&(*vcc)->temp_rwl));
-	assert(VCL_WARM(*vcc));
+	assert(VCL_WARM((*vcc)->temp));
 	AZ(errno=pthread_rwlock_unlock(&(*vcc)->temp_rwl));
 }
 
@@ -503,7 +503,7 @@ vcl_set_state(VRT_CTX, const char *state)
 	case '0':
 		if (vcl->temp == VCL_TEMP_COLD)
 			break;
-		if (vcl->busy == 0 && VCL_WARM(vcl)) {
+		if (vcl->busy == 0 && VCL_WARM(vcl->temp)) {
 			vcl->temp = VTAILQ_EMPTY(&vcl->ref_list) ?
 			    VCL_TEMP_COLD : VCL_TEMP_COOLING;
 			AZ(vcl_send_event(ctx, VCL_EVENT_COLD));
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -264,9 +264,7 @@ vcl_get(struct vcl **vcc, struct vcl *vcl)
 	AZ((*vcc)->discard);
 	(*vcc)->busy++;
 	Lck_Unlock(&vcl_mtx);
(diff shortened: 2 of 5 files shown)


Dridi:
I'm not sure why we need 22e44fe looking at everything else, wherever you use the temperature on its own it comes from a VCL.

Otherwise, I'm happy to see the temperature rwlock go away and overall this looks good to me. I will look closer.

Dridi:
I think this is generally a good opportunity to get rid of the `VCL_(COLD|WARM)` macros too:

* turn `VCL_TEMP_*` into structs
* give them `.cold` and `.warm` fields
* `s/VCL_COLD(x)/(x)->cold/`
* `s/VCL_WARN(x)/(x)->warm/`

nigoroll:
@Dridi with all due respect, I think your suggestion is overkill. Accessing a field in a struct is by no means more efficient than two pointer comparisons (to compile time values), the suggestion does not contribute to the fix and I do not see how else it would be useful.
If you really think this would be needed, can we discuss it in a separate PR?

nigoroll:
> I'm not sure why we need [22e44fe](https://github.com/varnishcache/varnish-cache/commit/22e44fedc1f648f7e7358bce098e21f8a9f79bdc) looking at everything else, wherever you use the temperature on its own it comes from a VCL.

this is used here: the temperature is saved inside the lock for consistency, then checked outside

https://github.com/varnishcache/varnish-cache/blob/be71fd82fd0b8d4fac2c8f1c1f4d41ae44a4508c/bin/varnishd/cache/cache_vrt_vcl.c#L204

nigoroll:
@Dridi thank you for the explanation. The type safety argument makes sense.

PR #3133, opened 2019-11-15
Set SA_CLIENT_{IP,PORT} only once

ehocdet:
SA_CLIENT_IP and SA_CLIENT_PORT are set twice since d7a81fe8

### DIFF
--- a/bin/varnishd/proxy/cache_proxy_proto.c
+++ b/bin/varnishd/proxy/cache_proxy_proto.c
@@ -126,8 +126,6 @@ vpx_proto1(const struct worker *wrk, const struct req *req)
 		    "PROXY1: Cannot resolve destination address");
 		return (-1);
 	}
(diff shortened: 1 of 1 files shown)


nigoroll:
Isn't the bug here that `SA_SERVER_ADDR` / `SA_SERVER_PORT` should be set from `fld[2], fld[4]` ?

nigoroll:
OK, I see it now: The point is that `SA_CLIENT_IP` / `SA_CLIENT_PORT` are special cased residues where elsewhere we use the suchaddr structs `SA_REMOTE_ADDR`/`SA_LOCAL_ADDR`/`SA_CLIENT_ADDR`/`SA_SERVER_ADDR`

nigoroll:
btw this was a copy-pasta indroduced with d7a81fe82ca09c9f3b3a9fd67265de00da7a4315

PR #3134, opened 2019-11-15
RFC: Open a constabulatory

Dridi:
*I'm going after libvcc low-hanging fruits while I have my hands deep in there.*

This patch introduces a new syntax for VCL:

    const <type> <name> = <literal>;

As demonstrated by the test case it works fine for VCL expressions, but
wherever we expect a constant in well-known VCL definitions we need to
update the parser to expect a constant. This was done only for the .host
field of backend definitions as an example.

Constants are their own kind of symbol and don't conflict with pseudo
symbols so that existing VCL won't break when new VCL transitions or
other predefined pseudo symbols are introduced. Constants can't however
be named after a reserved keyword.

The \<type\> and \<name\> tokens are case-insensitive so both definitions
are equivalent:

    const string hostname = "example.com";
    const STRING HOSTNAME = "example.com";

Not all types expect literals, so with the current constraint we can't
have this kind of constant:

    const backend my_default = my_director.backend();

It is also not yet possible to deal with compound-token types like BYTES
or DURATION but the parser can be tweaked to have pseudo-tokens on call
sites to accomodate them.


nigoroll:
did you look at https://code.uplex.de/uplex-varnish/varnish-objvar ?
VDD consensus was that we wanted to import it and only look after local variables natively in vcc

rezan:
I would prefer not re-using C syntax. Const is an ugly term for this.

Dridi:
> I would prefer not re-using C syntax. Const is an ugly term for this.

That's why this pull request is a bare-minimum PoC, and why I'm calling it an RFC. We can reach consensus on those details. If `const` is not OK, we can find something else like `constant`.

> did you look at https://code.uplex.de/uplex-varnish/varnish-objvar ?
VDD consensus was that we wanted to import it and only look after local variables natively in vcc

Variables are a different topic, but they could use a similar construct:

    var <type> <name> = <expr>;

Regarding vmod_objvar, I wasn't part of the discussion and didn't get to build an opinion.

nigoroll:
@Dridi yes, we missed you in Oslo.
But as we are at it now, please look at the constant code from vmod_objvar and your opinion would be highly valued

Dridi:
> It is also not yet possible to deal with compound-token types like BYTES
or DURATION but the parser can be tweaked to have pseudo-tokens on call
sites to accomodate them.

As I suspected this wasn't too hard to do, and this afternoon I wrote 48013248f5d5a5cea051aa874279524131283dad to prove it.

nigoroll:
@Dridi I am not against this, but no means. The reason for the vmod was that, for a long time, we made no progress with anything like native variable support in VCL, so I just wanted to get the missing feature implemented with the methods available. So I am absolutely happy to surrender the vmod in favor of a native VCC implementation.

BUT: We really should follow a concept here and not just implement some bits which happen to be easy. Again, it is very unfortunate that you could not participate in the last VDD, but there at least we came up with a concept how variable support would look like in VCL. Merging vmod_objvar is something which could be done right now, I have just not got around to it.

As this is an RFC, to summarize my opinion again:
Native variable support would be the best option, if we want to change the plan we already agreed to, we should do so following a design and with someone willing to actually do the work.

Dridi:
I read the VDD minutes again (I forgot most of it by the time I started the RFC) and came to the following conclusions:

I don't like the suggested syntax for declarations:

```
var {
      req.var.foo INT(task);
      <name> {task,const,global,top} {VCL_*};
}
```

I think we should stick to the following:

```
const <type> <name> = <literal>;
var <scope> <type> <name>[ = <literal>];
```

This syntax would only be allowed outside subroutines, as is my current constabulary implementation. For variable initialization, the optional initial value has to be literal.

Declarations being global, they must occur before the symbols get to be used, which is basically the same situation today with VMOD objects:

```vcl
import directors;
sub vcl_recv {
	set req.backend_hint = rr.backend(); # technically correct, but wrong order
}
sub vcl_init {
	new rr = directors.round_robin();
}
```

And so I think constants and variables should have their own symbol for simplicity, like VMOD objects do, and not something like `req.var.*`.

Which leads me to the scope, I think we should have the following list for starters:

- client
- backend

I have reserves regarding any scope that would imply concurrency, and I don't want the generic "task" scope because then people will invariably be confused when their variables from the client side are available in backend transactions but lose their state and vice versa.

Finally, there would be one case for the local scope: VIP2.

```
def STRING
foobar(INTEGER a1, BACKEND a2, STRING_LIST a3) {
    var <type> <name>[ = <expr>]; # not a literal here
    ...
}
```

Finally regarding @rezan's reserve about a const keyword, I looked at Python for prior art and there's none. But I remember that scala has `val` for constatns and `var` for variables.

bsdphk:
Bugwash consensus:  Please provide draft user documentation, and we'll bugwash it again.

nigoroll:
Suggestion for a generalized approach: #3141 / https://github.com/varnishcache/varnish-cache/wiki/VIP28:-designated-vmod-object-methods

There may still exist a case for special casing `const`, to be continued

Dridi:
Since this is a candidate for today's bugwash and since I was recently reminded of another use case for this with #3301 I will mention the use case of magic response status constants that could use symbols:

```vcl
const int https_redirect = 1301;

sub vcl_recv {
        if (std.port(client.ip) == 80) {
                return (synth(https_redirect));
        }
}

sub vcl_synth {
        if (resp.status == https_redirect) {
                set resp.http.location = "https://" + req.http.host + req.url;
                return (deliver);
        }
}
```

Arguably easier to follow.

gquintard:
side note: in that specific case, I would actually prefer the VCL to fail if the status code is invalid. But your point still stands in case you'd want to use `vmod_var` to the same effect.

Dridi:
> side note: in that specific case, I would actually prefer the VCL to fail if the status code is invalid.

That's perfectly valid, see #3301 and 8f3f5b57c811f750704e6e1e085706642be08e6c.

> But your point still stands in case you'd want to use vmod_var to the same effect.

Or we could have zero-overhead constants.

PR #3140, opened 2019-11-19
[Discuss] Expand VCL object support

rezan:
This is a thread for discussion about expanding VCL object support. The reason its a PR is because this is a fully working prototype. A good writeup on this is here:

https://github.com/varnishcache/varnish-cache/wiki/VIP9:-Expand-VCL-object-support

A thing to note is that this is just __expanding VCL object support__. This is not meant to solve all VCL based variable problems. This solution can live side by side with constants and real typed variables and even hybrid global/request objects. Basically, objects have been really useful in VCL and this PR expands their scope.

This VIP does address a lot of the usage questions and objections which have popped up over the years. So rather than list all of the gritty details, I will ask that you try this PR out and kick the tires a bit before coming up with the objections.

The goal here would be to get some kind of consensus that this is viable before investing more time in this. I do expect to rewrite a lot of this in the next iteration. Also, if other want to take this and run with it, go for it, I have a lot of tasks on my plate which need attention at the moment.

### DIFF
--- a/bin/varnishd/cache/cache_vrt_priv.c
+++ b/bin/varnishd/cache/cache_vrt_priv.c
@@ -160,6 +160,16 @@ VRT_priv_task(VRT_CTX, const void *vmod_id)
 	));
 }
 
+void **
+VRT_priv_task_object(VRT_CTX, void *vmod_id, vmod_priv_free_f *free)
+{
+	struct vmod_priv *priv;
+
+	priv = VRT_priv_task(ctx, vmod_id);
+	priv->free = free;
+	return (&priv->priv);
+}
+
 struct vmod_priv *
 VRT_priv_top(VRT_CTX, const void *vmod_id)
 {
--- a/bin/varnishd/cache/cache_vrt_priv.c
+++ b/bin/varnishd/cache/cache_vrt_priv.c
@@ -166,6 +166,7 @@ VRT_priv_task_object(VRT_CTX, void *vmod_id, vmod_priv_free_f *free)
 	struct vmod_priv *priv;
 
 	priv = VRT_priv_task(ctx, vmod_id);
+	// TODO this is ws, move NULL check to splice_cb
 	priv->free = free;
 	return (&priv->priv);
 }
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -538,6 +538,7 @@ struct vmod_priv {
 
 void VRT_priv_fini(const struct vmod_priv *p);
 struct vmod_priv *VRT_priv_task(VRT_CTX, const void *vmod_id);
+void **VRT_priv_task_object(VRT_CTX, void *vmod_id, vmod_priv_free_f *free);
 struct vmod_priv *VRT_priv_top(VRT_CTX, const void *vmod_id);
 
 /* Stevedore related functions */
--- a/lib/libvcc/vcc_action.c
+++ b/lib/libvcc/vcc_action.c
@@ -405,8 +405,7 @@ vcc_Action_Init(struct vcc *tl)
 	ACT(hash_data,	vcc_act_hash_data,
 		VCL_MET_HASH);
 	ACT(if,		vcc_Act_If,	0);
-	ACT(new,	vcc_Act_New,
-		VCL_MET_INIT);
+	ACT(new,	vcc_Act_New,	0);
 	ACT(return,	vcc_act_return,	0);
 	ACT(set,	vcc_act_set,	0);
 	ACT(synthetic,	vcc_act_synthetic,
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -209,6 +209,80 @@ Fb(const struct vcc *tl, int indent, const char *fmt, ...)
 }
 
 void
+Fb_splice_push(struct vcc *tl)
+{
+	struct splice *s;
+
+	AN(tl->fb);
+	ALLOC_OBJ(s, SPLICE_MAGIC);
+	AN(s);
+	s->fb = tl->fb;
+	VTAILQ_INIT(&s->callbacks);
+	VTAILQ_INSERT_HEAD(&tl->fb_splices, s, list);
+
+	tl->fb = VSB_new_auto();
+	AN(tl->fb);
+}
+
+void
+Fb_splice_print(const struct vcc *tl, int indent, const char *fmt, ...)
+{
+	struct splice *s;
+	va_list ap;
+
+	s = VTAILQ_FIRST(&tl->fb_splices);
+	AN(s);
+	if (indent)
+		VSB_printf(s->fb, "%*.*s", tl->indent, tl->indent, "");
+	va_start(ap, fmt);
+	VSB_vprintf(s->fb, fmt, ap);
+	va_end(ap);
+}
+
+void
+Fb_splice_cb(const struct vcc *tl, splice_cb_f *f, const void *priv)
+{
+	struct splice *s;
+	struct splice_cb *cb;
+
+	s = VTAILQ_FIRST(&tl->fb_splices);
+	AN(s);
+
+	ALLOC_OBJ(cb, SPLICE_CB_MAGIC);
+	AN(cb);
+	cb->callback = f;
+	cb->priv = priv;
+	VTAILQ_INSERT_TAIL(&s->callbacks, cb, list);
+}
+
+void
+Fb_splice_pop(struct vcc *tl)
+{
+	struct splice *s;
+	struct splice_cb *cb, *cb_temp;
+
+	s = VTAILQ_FIRST(&tl->fb_splices);
+	AN(s);
+
+	VTAILQ_FOREACH_SAFE(cb, &s->callbacks, list, cb_temp) {
+		AN(cb->callback);
+		cb->callback(tl, cb->priv);
+		VTAILQ_REMOVE(&s->callbacks, cb, list);
+		FREE_OBJ(cb);
+	}
+
+	VTAILQ_REMOVE(&tl->fb_splices, s, list);
+
+	AN(tl->fb);
+	AZ(VSB_finish(tl->fb));
+	VSB_bcat(s->fb, VSB_data(tl->fb), VSB_len(tl->fb));
+	VSB_delete(tl->fb);
+	tl->fb = s->fb;
+
+	FREE_OBJ(s);
+}
+
+void
 Fc(const struct vcc *tl, int indent, const char *fmt, ...)
 {
 	va_list ap;
--- a/lib/libvcc/vcc_compile.h
+++ b/lib/libvcc/vcc_compile.h
@@ -163,6 +163,7 @@ struct symbol {
 	unsigned			noref, nref, ndef;
 
 	const char			*extra;
+	const char			*extra2;
 
 	/* SYM_VAR */
 	const char			*rname;
@@ -171,6 +172,9 @@ struct symbol {
 	unsigned			w_methods;
 	const char			*uname;
 	unsigned			u_methods;
+
+	int				request_scope;
+	int				null_ok;
 };
 
 VTAILQ_HEAD(tokenhead, token);
@@ -193,6 +197,7 @@ struct proc {
 	struct vsb		*cname;
 	struct vsb		*prologue;
 	struct vsb		*body;
+	int			need_privs;
 };
 
 struct inifin {
@@ -205,6 +210,24 @@ struct inifin {
 	struct vsb		*final;
 	struct vsb		*event;
 	VTAILQ_ENTRY(inifin)	list;
+};
+
+typedef void splice_cb_f(struct vcc *, const void *);
+
+struct splice_cb {
+	unsigned		magic;
+#define	SPLICE_CB_MAGIC		0x6C34AEB9
+	VTAILQ_ENTRY(splice_cb)	list;
+	splice_cb_f		*callback;
+	const void			*priv;
+};
+
+struct splice {
+	unsigned		magic;
+#define	SPLICE_MAGIC		0x8E28B6C8
+	VTAILQ_ENTRY(splice)	list;
+	struct vsb		*fb;
+	VTAILQ_HEAD(,splice_cb)	callbacks;
 };
 
 VTAILQ_HEAD(inifinhead, inifin);
@@ -242,6 +265,7 @@ struct vcc {
 	struct vsb		*fb;		/* Body of current sub
 						 * NULL otherwise
 						 */
+	VTAILQ_HEAD(, splice)	fb_splices;
 	struct vsb		*sb;
 	int			err;
 	struct proc		*curproc;
@@ -306,6 +330,11 @@ void Fc(const struct vcc *tl, int indent, const char *fmt, ...)
     v_printflike_(3, 4);
 void Fb(const struct vcc *tl, int indent, const char *fmt, ...)
     v_printflike_(3, 4);
+void Fb_splice_push(struct vcc *tl);
+void Fb_splice_print(const struct vcc *tl, int indent, const char *fmt, ...)
+    v_printflike_(3, 4);
+void Fb_splice_cb(const struct vcc *tl, splice_cb_f *f, const void *priv);
+void Fb_splice_pop(struct vcc *tl);
 void EncToken(struct vsb *sb, const struct token *t);
 void *TlAlloc(struct vcc *tl, unsigned len);
 char *TlDup(struct vcc *tl, const char *s);
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -454,6 +454,18 @@ vcc_do_arg(struct vcc *tl, struct func_arg *fa)
 	fa->avail = 1;
 }
 
+static void v_matchproto_(splice_cb_f)
+vcc_func_sym(struct vcc *tl, const void *priv)
+{
+	const struct symbol *sym;
+
+	CAST_OBJ_NOTNULL(sym, priv, SYMBOL_MAGIC);
+	Fb_splice_print(tl, 1, "// XXX NULL? %s %s\n", sym->rname, sym->name);
+	VSB_printf(tl->curproc->prologue,
+	    "  //struct vmod_priv *lvo_priv_%u;\n", tl->curproc->need_privs);
+	tl->curproc->need_privs++;
+}
+
 static void
 vcc_func(struct vcc *tl, struct expr **e, const void *priv,
     const char *extra, const struct symbol *sym)
@@ -568,6 +580,9 @@ vcc_func(struct vcc *tl, struct expr **e, const void *priv,
 			break;
 		SkipToken(tl, ',');
 	}
+
+	if (sym->request_scope && !sym->null_ok)
+		Fb_splice_cb(tl, vcc_func_sym, sym);
 
 	if (sa != NULL)
 		e1 = vcc_mk_expr(rfmt, "%s(ctx%s,\v+\n&(%s)\v+ {\n",
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -645,6 +645,8 @@ vcc_Eval_SymFunc(struct vcc *tl, struct expr **e, struct token *t,
 	(void)fmt;
 	assert(sym->kind == SYM_FUNC);
 	AN(sym->eval_priv);
+	if(sym->r_methods)
+		vcc_AddUses(tl, t, tl->t, sym->r_methods, "Not available");
 
 	vcc_func(tl, e, sym->eval_priv, sym->extra, sym);
 	ERRCHK(tl);
@@ -1409,6 +1411,8 @@ vcc_Act_Call(struct vcc *tl, struct token *t, struct symbol *sym)
 	struct expr *e;
 
 	e = NULL;
+	if(sym->r_methods)
+		vcc_AddUses(tl, t, tl->t, sym->r_methods, "Not available");
 	vcc_func(tl, &e, sym->eval_priv, sym->extra, sym);
 	if (!tl->err) {
 		vcc_expr_fmt(tl->fb, tl->indent, e);
--- a/lib/libvcc/vcc_parse.c
+++ b/lib/libvcc/vcc_parse.c
@@ -197,7 +197,9 @@ vcc_Compound(struct vcc *tl)
 				vcc_AddUses(tl, t, NULL,
 				    sym->action_mask,
 				    "Not a valid action");
+			Fb_splice_push(tl);
 			sym->action(tl, t, sym);
+			Fb_splice_pop(tl);
 			break;
 		default:
 			/* We deliberately do not mention inline C */
@@ -266,6 +268,7 @@ vcc_ParseFunction(struct vcc *tl)
 	}
 	CHECK_OBJ_NOTNULL(p, PROC_MAGIC);
 	tl->fb = p->body;
+	VTAILQ_INIT(&tl->fb_splices);
 	Fb(tl, 1, "  /* ... from ");
 	vcc_Coord(tl, p->body, NULL);
 	Fb(tl, 0, " */\n");
@@ -277,6 +280,7 @@ vcc_ParseFunction(struct vcc *tl)
 	tl->indent -= INDENT;
 	tl->fb = NULL;
 	tl->curproc = NULL;
+	AN(VTAILQ_EMPTY(&tl->fb_splices));
 }
 
 /*--------------------------------------------------------------------
--- a/lib/libvcc/vcc_vmod.c
+++ b/lib/libvcc/vcc_vmod.c
@@ -490,6 +490,7 @@ vcc_Act_New(struct vcc *tl, struct token *t, struct symbol *sym)
 	buf = VSB_new_auto();
 	AN(buf);
 	if (request_scope)
+		// TODO this needs to move into the splice_cb
 		VSB_printf(buf, ", (%s**)\n  "
 		    "VRT_priv_task_object(ctx, &%s, &vo_free_%s),\n  \"%s\"",
 		    s_struct, sy1->rname, sy1->name, sy1->name);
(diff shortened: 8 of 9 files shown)


nigoroll:
@rezan I find it unfortunate that you open another PR with something which looks pretty similar to previous work and not addressing the feedback which I had already provided (if memory serves me right, multiple times over multiple channels).
You write that _the goal here would be to get some kind of consensus that this is viable_, but at the same time https://github.com/varnishcache/varnish-cache/wiki/VIP9:-Expand-VCL-object-support already contains a list of unsolved problems. I think it would be helpful to answer these questions with concrete proposals rather than asking if unanswered questions are a viable approach.
Also, I think it would be helpful if you tasked about which _usage questions and objections_ you have addressed and how. You seem to expect that reviewers not only try out the PR, but also find out which answers you are giving to which questions.

That said, here is my concrete feedback
* You cannot re-use the existing object constructors/destructors. The non-global object _init/_fini methods need to be aware that they are being executed concurrently and should not use space on the the heap, but rather on the workspace. IMHO, the VIP should include a proposal how to define and implement the different scopes.
* I do not understand what exactly you are proposing for `priv_task`.
* For handling uninitialized objects, you are saying _There are ways to detect this and error out_. Yes, but what exactly are you proposing? The whole point of a VIP/PR is to give answers and solutions, not questions.
* I do not understand your comment why the top scope would not be needed.

From a recent IRC concersations:

* (16:11:39) slink: - how do would you declare, use and implement the lifetime + type combinations which vmod_objvar has today?
* (16:11:39) slink: - how does the native non-global object implementation look like for task and top lifetime?

the last point is a duplication of my comment on the constructors, left here for clarity

nigoroll:
as reza left for lunch right after starting a discussion and I will leave now, here's a log FTR:

(18:09:17) reza: slink: can you explain "You cannot re-use the existing object constructors/destructors."
(18:09:56) reza: because its pretty clear to me that existing constructors/destructors are being used
(18:10:11) slink: right now, they are guaranteed to run single threaded in the cli thread. This is not the case when they run in individual workers.
(18:10:24) reza: ?
(18:10:52) slink: Also malloc/free is not efficient for task scope, so the vmod code should be aware of that
(18:11:00) reza: so objects as they are written today will continue to be global
(18:11:23) reza: if an object chooses to be request scope
(18:11:29) reza: then its request scope
(18:12:00) reza: just like when you write a vmod method, its upto the author to use the best practices for memory allocation
(18:12:14) reza: are you talking about vmod_types?
(18:12:18) reza: thats a toy example
(18:12:21) reza: brb, lunch
(18:12:43) slink: can you please point me to the place in your code where the constructor gets told that it is now being used in request context?
(18:15:58) slink: I do not see how/where "an object chooses to be request scope".

rezan:
So the comments about how does this implement `vmod_objvar`, I was pretty clear that this is just __expanding the VCL object scope__. Things that you have accomplished with objvar will remain and the problems you have solved with it are great and obviously suits you needs. This is why we have VCL, its a platform for solving problems. All this does is add another tool. This is not meant to overtake objvar.

As for all the other objections, I believe they are addressed in my writeup. So I ask that you re-read VIP9, digest it, and maybe sit on it for a bit.

rezan:
I added a `REQUEST_SCOPE` keyword to the .vcc file. It works similar to `NULL_OK`:

https://github.com/rezan/libvmod-types/blob/46d605921bcf912a240228eb38b21422af116e33/src/vmod_types.vcc#L10

rezan:
Request scope VMOD code is in its final form here:

https://github.com/rezan/libvmod-types/blob/master/src/vmod_types_integer.c
https://github.com/rezan/libvmod-types/blob/master/src/vmod_types_string.c

nigoroll:
@Dridi and myself collaborated to come up with proposal taking up the idea and filling in the missing bits: https://github.com/varnishcache/varnish-cache/wiki/VIP9B:-Extend-vmod-object-scopes
This is also my own answer to the concerns from https://github.com/varnishcache/varnish-cache/pull/3140#issuecomment-555604958

rezan:
So I have a framework for null checking all the request scoped object accesses. This covers the case where the constructor is not called or does not construct properly.

This will also cover priv_task null checking since the priv is on the workspace now and can also fail and be null. So VMOD objects will not be exposed to either failure case. Currently VMOD priv_tasks are null checked in the vcc prologue. The above framework allows the null check the happen right before the actual statement. This is needed because we may have a constructor call before this statement, so null checking in the prologue complicates things. This also allows for objects to live conditionally without error.

Just an FYI, I am traveling Monday and will not be at bugwash.

nigoroll:
@rezan did you look our (@Dridi & myself) contribution to the discussion you asked for?

rezan:
I did, I will leave my comments on that issue.

rezan:
oh sorry, that issue is for VIP28, which is related to actual variables. So VIP9B, I believe you are introducing a new sub to initialize objects? Given this PR allows objects to be initialized anywhere in *existing* subs, I dont see much value there.

As for the scopes, and in particular the top scope, that can be addressed with some kind of VCL constructor qualifier, like I pointed out in VIP9. Or you can even just use a `PRIV_TOP` in your constructor and do what you did in `objvar`.

As for `vcl_backend_init`, thats a bit dangerous because in a distributed VCL environment, client subs and backend subs might not even be executing on the same server. The only safe and recommended way to transfer data between client and backend subs is via headers.

nigoroll:
> Given this PR allows objects to be initialized anywhere in _existing_ subs, I dont see much value there.

Adding the new `*_init` subs serves two/three purposes:
* they act as a demarcation point for classes which are not `NULL_OK`: As with `vcl_init {}`, non-NULL class object can be checked when `*_init` returns. I think if we had no such demarcation point, we would need to demand any vmod to support NULL.
* they define the scope of objects without an additional keyword, reflecting the existing syntactic structure
* `vcl_backend_init` also solves the problem of getting structured data across from the client to the backend side

> As for the scopes, and in particular the top scope, that can be addressed with some kind of VCL constructor qualifier, like I pointed out in VIP9. Or you can even just use a `PRIV_TOP` in your constructor and do what you did in `objvar`.
On the latter suggestion: I agree it would be possible, but that would not make it a good concept: For local, client, backend scope the object itself would be scoped, just for top, you would use a global object with top-scoped data. And this is what we can do today, so if you argument was that this was the way to go, we would not even need this discussion to begin with.
On the first suggestion: Yes, sure, we could use constructor qualifiers. But then we should have a clear idea how they look like consistently for all the cases. [VIP9B](https://github.com/varnishcache/varnish-cache/wiki/VIP9B:-Extend-vmod-object-scopes) goes the route of implicit qualifiers mainly because this is consistent with the existing syntax. Yet I am not against explicit qualifiers, @Dridi and me also discussed a variant with them. But I do think that whatever we do, we should do it consistently. So if we use qualifiers, we should use them everywhere.

> As for `vcl_backend_init`, thats a bit dangerous because in a distributed VCL environment, client subs and backend subs might not even be executing on the same server. The only safe and recommended way to transfer data between client and backend subs is via headers.

I do not know what you might be doing in your closed source fork. But in the clustered setups I run on Varnish-Cache, also the 2nd varnish instance will process the request on the client side first.
And at any rate, we got existing cases (for example getting director parameters from the client to the backend side) where this bit of functionality is missing.
But I do agree that for distributed environments, you need to serialize control information into headers.

nigoroll:
> I added a `REQUEST_SCOPE` keyword to the .vcc file. It works similar to `NULL_OK`:
> 
> https://github.com/rezan/libvmod-types/blob/87c22a3d440158146e07d78809d1081c7de5cb7c/src/vmod_types.vcc#L12

To comment on this specific suggestion: If we made the scope a fixed compile time property, we would need to re-define classes multiple times just to cover the different scopes. So IMHO classes should define the supported scopes, but some VCL context or keyword should define which scope an object will have.

rezan:
Good, this is feedback I can work with.

> `*_init` ... act as a demarcation point for classes which are not `NULL_OK`

This has been solved. We can do null checks anywhere an object is used now. No need for a dedicated sub.

> top scope

So I am going to take the scoping to the next level here. There will be 3 things:

First, in the .vcc file, I will expose all the possible scopes. Global scope, client scope, backend scope, or top scope. The VMOD can now define any or all of these scopes. These scopes will be enforced by the vcc at compile time.

Next, VCL will have a qualifier to place an object into a specific scope. As long as the .vcc allows for it, the object will now be placed in that scope. Most of the time, the the vcc will be smart enough to figure out the scope based on where you define the object. Basically, there will be implicit default scopes for each sub. But in the cases where there are multiple scopes possible is where you would use this qualifier.

Finally, the VMOD will have runtime access to the scope being used so it can do whatever it want based on the scope used.

So if you write a VMOD and opt into all scopes and the user defines it into top scope, you can detect that and do what you need to do. Or if you write a VMOD and allow client and backend scopes, you will be able to detect which one you are in, if that matters to you.

What do you think?

nigoroll:
> > `*_init` ... act as a demarcation point for classes which are not `NULL_OK`
> 
> This has been solved. We can do null checks anywhere an object is used now. No need for a dedicated sub.

I do not see how you would do this without a dedicated sub. What would happen otherwise for a conditional construction like so?

```
if (condA) {
  new taskobj = vmod.constructor(argA);
}
# ... more code ...
if (condB) {
  new taskobj = vmod.constructor(argB);
}
```

Where in the code would be the point where the object would have to be init'ed?

Again, the suggestion in [VIP9B](https://github.com/varnishcache/varnish-cache/wiki/VIP9B:-Extend-vmod-object-scopes) is that when the `_init` sub returns, all the object which may not be NULL have to be initialized.

> Also, im going to add an optimization where we only do the null check once per sub per object.

This is not the point.

> First, in the .vcc file, I will expose all the possible scopes. Global scope, client scope, backend scope, or top scope. The VMOD can now define any or all of these scopes. These scopes will be enforced by the vcc at compile time.

I agree, this is `$Scope` in [VIP9B](https://github.com/varnishcache/varnish-cache/wiki/VIP9B:-Extend-vmod-object-scopes) 

> Next, VCL will have a qualifier to place an object into a specific scope. As long as the .vcc allows for it, the object will now be placed in that scope. Most of the time, the the vcc will be smart enough to figure out the scope based on where you define the object. Basically, there will be implicit default scopes for each sub. But in the cases where there are multiple scopes possible is where you would use this qualifier.

I can go with that. Again, in [VIP9B](https://github.com/varnishcache/varnish-cache/wiki/VIP9B:-Extend-vmod-object-scopes)  we opted for no qualifiers, then I realized that we need at least `local`, but I am also OK with required or optional qualifiers for all the scopes.

> Finally, the VMOD will have runtime access to the scope being used so it can do whatever it want based on the scope used.

Yes, there probably needs to be an additional argument added to the constructor.

rezan:
> What would happen otherwise for a conditional construction like so?

You defined `taskobj` twice, so your snippet is an immediate vcc compile error.

> Where in the code would be the point where the object would be chcked for the required initialization?

After the constructor, before the method invocation, once per sub. And conditionals are completely supported, so if you call an object method inside of an if statement, the null check is only done inside of the if statement. Basically, a null check is guaranteed to happen before calling any object method. If you define `NULL_OK`, then its simply skipped.

Dridi:
I think one aspect of VIP 9B that has been overlooked in this discussion is the fact that the new `vcl_*_init` subroutines aren't here only to force the definition of objects in a specific location.

In particular `vcl_backend_init` is very interesting because it happens while there is no concurrency \_yet\_ between the client transaction and the [bg]fetch transaction. @nigoroll mentioned it, but @rezan didn't seem to pick that up.

One thing I would like to add on `vcl_backend_init` in particular, I think `req` should be read-only, similar to what `vcl_pipe` does today.

nigoroll:
> > What would happen otherwise for a conditional construction like so?
> You defined `taskobj` twice, so your snippet is an immediate vcc compile error.

That would limit the usefulness to less than what we got today, basically this would eliminate the possibility to call a constructor from anywhere but the sub's top level.

> > Where in the code would be the point where the object would be chcked for the required initialization?
> After the constructor, before the method invocation, once per sub.

I do not think this point in the execution order is compile-time defined.

nigoroll:
Sorry I was wrong here https://github.com/varnishcache/varnish-cache/pull/3140#issuecomment-558039834
Yes, my example is not possible today. Yet objects can be defined conditionally.

Dridi:
This is why with constants and variables I was going for a declaration outside subroutines.

We could technically have something like this though:

```vcl
declare [scope?] foo as vmod.constructor;

sub vcl_init {
    if (cond) {
        new foo = vmod.constructor(args);
    } else {
        new foo = vmod.constructor(other args);
    }
}
```

That could solve the declaration+assignment constraint we have today without breaking the existing syntax.

nigoroll:
@Dridi interesting thought. So the declaration would be optional, defaulting to vcl scope?

Dridi:
My thoughts:

- VIP 9B allows us to have a natural point where we check `new`'ed objects
  - that would be the end of all the `vcl_*init` subroutines
  - a good spot for null checks (`NULL_OK`?)
- A `declare` mechanism as described above would solve conditional `new`s
  - we'd still need to guard against multiple `new` statements running for a given object
    - keeping track of that shouldn't be too hard
    - I see no sensible alternative to `VRT_fail`
    - in `vcl_backend_init` which one do we fail? both?
  - we could make the scope optional
    - nothing for vcl scope in `vcl_init`
    - `{client,top,backend}` for respective scope in `vcl_*_init`

The proposals still need some refining but I think we have a credible path ahead of us.

rezan:
Guys, we keep on coming back to dedicated `*_init` subs for object declaration. I've said this many times, there is no technical reason we need this. On top of this, no other successful language has this kind of syntax constraint. Objects can be declared anywhere, including if statements, and will be safely and efficiently null checked before method invocation.

My next push to this PR will have test objects and vtcs proving all of this. So I would like to move forward from this point. Thank you :)

Dridi:
@rezan, once again you dodge the key differentiator of `vcl_backend_init` by focusing on the constraint of where the `new` keyword would be allowed.

Can you show me one successful language that forces a keyword (like `set` or `new`) onto you for assignments? VCL does that, C doesn't, Python (apparently or prior-art consideration for new language constructs) doesn't and most (if not all) of the languages I happen to use (or used) don't. Incidentally, I'm interested in your definition of successful...

Also, what's your opinion on the prior "declare" option to allow `new` inside conditionals? This would work for either VIP 9 or 9B.

nigoroll:
@rezan first regarding the meta-discussion:

We are having a technical discussion, which does not go away by you saying that there was no problem. I am trying to explain my point, and it is unfortunate that I do not get it across. I can reassure you that I do question my own argument, but also I do not see why it would be wrong. So please try to understand my argument or explain to me why it is wrong. Trying to end a discussion with "I tell you it is not a problem" does not work.

Regarding the actual discucssion, the question is: Where in the code would the NULL check happen?

If the NULL check happened right after the initialization, we would not allow for improving the conditional initialization (Yes, I made a mistake during the discussion, but I think the current limitation should be lifted and @Dridi seems to agree).

If the NULL check happened right before the access, the error would be delayed for arbitrary amounts of time, possibly until the last bit in vcl_deliver, or not at all, which would make debugging very hard. The missing initialization is a hard error which should be caught as early as possible. Contrary to vcl_init, we cannot catch it at vcl load time, so we should at least catch it as early in the FSM and consistently, not only when the object is actually used.

That said, we have one more use case for `vcl_backend_init` and also I think that the `_init` subs can actually help vcl clarity in general. At least my own `vcl_recv` subs almost always contain some `if (req.restarts == 0)` and `if (req.esi_level == 0)` init code, which would just get moved to `vcl_client_init` and `vcl_top_init`, respectively.

Again, I am open to alternatives, but I do think that we are making a sensible, constructive suggestion and would appreciate consideration.

nigoroll:
@Dridi on `vcl_backend_init`: 

Actually we should probably make `req.*` and *client* objects read-only to avoid confusion of the workspace used. `(struct vrt_ctx *)->ws` always points to the right workspace for _task_ scope, and if `vcl_backend_init` was allowed to modify both the backend and the client side, we would complicate the code.

Yet I think that we do need to support `unset` operations in order to get _move_ semantics with respect to `client` and `backend` scoped objects.

So maybe the rule could be that no `req.*` / *client* operations are supported which require workspace allocations on the client side?

Failure in `vcl_backend_init` should only concern the backend request, or do you see a reason why it would also affect the client side?

nigoroll:
bugwash: no results

bsdphk:
I'm not at all keen on the "splice" stuff proposed in this PR, it doesnt make anything better IMO.

In this context, be it this PR or VIP9B, we may have a fundamental problem with objects created before/after `return(vcl,...)`

I'm increasingly of the opinion that allowing `return(vcl, ...` in `vcl_recv{}` was a mistake.

Maybe instead we should have a `vcl_dispatch{}`, where everything is read-only, and which can return {`fail`, `synth`, `vcl`, `recv`} the latter meaning "go to vcl_recv{}".

The default would be:

    sub vcl_dispatch {
        return (recv);
    }

VMOD use would be allowed in `vcl_dispatch{}`, which means that VMOD writers really insistent of making a mess of things can do that.

That way, at least `return(vcl,...)` does trip up this PR/VIP9B efforts.

nigoroll:
@bsdphk I do not see the problem you are seeing:
The fundamental solution is the rollback: Even today, we roll back all changes and fini all tasks/objects when we switch VCLs.
So no matter which objects are being created, we make sure that they all cease to exist when we switch vcls.
I agree that a read-only `vcl_dispatch` would also solve this problem, but it would be rather limited. Consider, for example [the re2 vmod](https://code.uplex.de/uplex-varnish/libvmod-re2) which was written for exactly that purpose: To implement a highly efficient, yet flexible "switch/case" statement for virtual hosting or url classification. If `vcl_dispatch` was read-only, such a vmod would not be possible to use or would need to work around the limitation, neither of which would make much sense.
So if `vcl_dispatch` was not read-only, which I think it can't be, then we are back to the rollback model, and then it does not make much of a difference if the sub is call `vcl_recv` or `vcl_dispatch`.

rezan:
I don't see `return(vcl...)` as an issue either. As far as i understand, objects cannot and will not span VCLs, so nothing breaks with this.

As for the splice stuff, we have no way to inspect code that we are compiling and insert init code, error checks, or even optimizations. Without this ability, then objects cannot safely exist.

Basically, we need some kind of instruction buffer and a way to inspect it before writing the final instruction logic in C. We would have to redo a lot of vcc to make that happen, no?

nigoroll:
BTW, I do not think we ware ready to discuss implementations, but they are a helpful tool for exploring options.

bsdphk:
Ok, if you guys are willing to live with rollbacks for this, the forget `vcl_dispatch{}`

Re splice thing and rewriting a lot of VCC:  Yes, it will take an effort to get this right, and shortcuts in this area have historically ended up as detours, so they are not very welcome.

rezan:
VIP9 has been updated:

https://github.com/varnishcache/varnish-cache/wiki/VIP9:-Expand-VCL-object-support

nigoroll:
@rezan that update is a step in the right direction, thank you.

Comments:

* VCL scopes

> Most of the time, the the vcc compiler will be smart enough to figure out the scope based on where you define the object.

Yes, but whenever a class has `$Scope local`, we are going to need the qualifier, so we should at least consider making it a requirement. This would also make vcl more forward-compatible for when a vmod author decides to add `$Scope local`, as that would render all declarations without a qualifier ambiguous.

* Runtime scopes

... and SCOPE would be required if `$Scope` exists, right? I like this idea.

* `NULL` checking

We need to make a decision here: Do we want to retain some demarcation line like `vcl_init` after which all objects need to be non-NULL (the proposal in [VIP9B](https://github.com/varnishcache/varnish-cache/wiki/VIP9B:-Extend-vmod-object-scopes)) or do we want to go the route of "we can `VRT_fail()` any time" which you are proposing. There are pros and cons for both. I agree with you, @rezan, that your approach is simpler, but I also think that a demarcation line would help VCL authors write better (more correct) code.

* client->backend request

Your VIP update still leaves this question open, unless I missed something. We already have an issue with this missing bit and object scopes would widen it even more. So what is your suggestion to replace `vcl_backend_init {}`?

Dridi:
> There are pros and cons for both. I agree with you, @rezan, that your approach is simpler, but I also think that a demarcation line would help VCL authors write better (more correct) code.

I beg to disagree on this one, there is a lot of "conditional" wording in the current VIP9 that hints at complicated special cases, where we actually want to erase as many special cases as possible in libvcc.

See #3147, #3116 and https://github.com/varnishcache/varnish-cache/pull/3123#issuecomment-560411831 for example.

> Your VIP update still leaves this question open, unless I missed something. We already have an issue with this missing bit and object scopes would widen it even more. So what is your suggestion to replace `vcl_backend_init {}`?

I wasn't able to join yesterday when I could have a chance to discuss this with @rezan, so I think we might still keep this part of the discussion offline for now.

rezan:
> we are going to need the qualifier, so we should at least consider making it a requirement

Sure, there are lots of scope combinations here, so lets see how this implementation ends up. We can easily add required qualifiers where needed.

> and `SCOPE` would be required if $Scope exists, right?

Not really... its upto the VMOD to decide if they care about runtime scopes. My VMODs here dont care and can live in all local, client, and backend scopes: [integer](https://github.com/rezan/libvmod-types/blob/master/src/vmod_types_integer.c) [string](https://github.com/rezan/libvmod-types/blob/master/src/vmod_types_string.c). This is why its a parameter.

> Do we want to retain some demarcation line like vcl_init after which all objects need to be non-NULL

So technically the null boundary doesn't even work today:

``` vcl
sub vcl_init {
    if (0) {
        new rr = directors.round_robin();
    }
    rr.add_backend(default);
}
```

The null check I propose would solve this. Then we have the syntax boilerplate overhead of the null boundary. The null check has no syntax boilerplate to deal with. Null checking is the elegant solution here.

> do we want to go the route of "we can `VRT_fail()` any time" which you are proposing.

The null boundary can `VRT_fail()` at anytime too. So there is nothing different here.

> There are pros and cons for both.

I am only seeing cons for null boundary. It half works today and has significant syntax boilerplate overhead. Null check fixes these things.

Also, with null checking, you can construct objects with actual VCL data. A null boundary has to init objects before running VCL code, so there is limited access to data. That's another limitation of null boundaries.

> So what is your suggestion to replace `vcl_backend_init {}`?

Nothing, its on you to show we need this and how to accomplish this. How this will work with both client and backend workers and readable and writable objects is kind of a can of worms. Then we have the fact that this breaks the entire client/backend partition. The fact that VCL can execute distributed across many servers is a huge win for the Varnish platform, so killing off this feature should not be taken lightly. So I am ok with the current VCL architecture here.

> I beg to disagree on this one, there is a lot of "conditional" wording in the current VIP9 that hints at complicated special cases

Can you be a bit more specific?

bsdphk:
I think this discussion would be helped if people provided more mock-up VCL examples to illustrate their points ?

nigoroll:
> So technically the null boundary doesn't even work today:

Yes, but I think this ticket should be concerned more with how it should work. We agree that it is broken now.

> The above code instantly segfaults. The null check I propose would solve this.

... and any other solution should solve it too.

> Then we have the syntax boilerplate overhead of the null boundary. The null check has no syntax boilerplate to deal with. Null checking is the elegant solution here.

The problem with that kind of "elegance" is, as I said before, the risk to the VCL user of unintended missed initializations. I have explained this [here](https://github.com/varnishcache/varnish-cache/pull/3140#issuecomment-558095905) and would appreciate feedback why you think my argumentation does not make sense to you.

In short: We should catch missed initializations as early as possible.

> Also, with null checking, you can construct objects with actual VCL data. A null boundary has to init objects before running VCL code, so there is limited access to data. That's another limitation of null boundaries.

Initializing objects does not need to coincide with assigning values to them.

> > So what is your suggestion to replace `vcl_backend_init {}`?
> 
> Nothing, its on you to show we need this and how to accomplish this.

I did give you the use case example in https://github.com/varnishcache/varnish-cache/pull/3140#issuecomment-557810354 . For the `shard` director, we cannot get parameters set at the client side over to the backend side.

Regarding an actual implementation, that will be done but I got enough unmerged PRs and do not want to potentially waste more time before we have an agreement on the design.

> The fact that VCL can execute distributed across many servers is a huge win for the Varnish platform, so killing off this feature should not be taken lightly.

This is the varnish-cache project and the varnish-cache project supports this setup without any such limitation.

rezan:
I have example VCLs in [VIP9](https://github.com/varnishcache/varnish-cache/wiki/VIP9:-Expand-VCL-object-support) and as stated before, this PR is 100% working. One of the strengths of this PR is that the VCL is simple, elegant, and inline with modern languages (no boilerplate).

I will continue work on fleshing this out. I will add some VTC test objects and a handful of test cases showing all the different use-cases which are covered with this PR.

rezan:
I had a look at the [VIP9B examples](https://github.com/varnishcache/varnish-cache/wiki/VIP9B:-Extend-vmod-object-scopes#examples). There is a lot going on in there...

Your touching constructed objects in your special `*_init` subs, so I assume you have accepted null checking before invoking methods? If so, then what value does `*_init` provide anymore?

Second, `tmp` morphs from an object to a string type? Basically, you are blending objects and types. This is confusing.

Finally, you are overloading `+=` with what should be plain method verbs. `+=` is for math.

In total, given you started with plain `objvar` and have now ended up with this, I have a feeling ideas are just being put together on the spot, which might explain why VIP9B is hard to wrap your head around. Next, can this even be implemented?

This is a good contrast to VIP9, which is very simple and does not introduce any fundamentally new and unseen language paradigms. It makes writing objects super simple and straight forward ([objvar](https://code.uplex.de/uplex-varnish/varnish-objvar/tree/master/src) vs [types](https://github.com/rezan/libvmod-types/tree/master/src)). However, it leaves the door open for true typed VCL variable, so it stays in its lane. The reason I say this is that typed VCL variable are dead simple C vars like `int`, `char*`, `void*`, and `double`. So I am not a fan of the level of complexity you are bringing to the table in any of these object/var VIPs.

nigoroll:
@rezan the VCL mock up comment reads:

> also contains syntax ideas from VIP28

OK, I had enough, this discussion does not lead us anyway. If my suggestions are not being taken as intended (constructive criticism and ideas for improvements), do I do not see much value in discussing this back and forth.
Please continue with the other maintainers, I will revert to a neutral position and not pursue this further unless explicitly requested.

rezan:
Closing this. Going to be actively working on this for the upcoming release deadline, so a new PR will be made when this is ready. Basically, if we plan on doing 7.0, then I think we should strongly consider including this.

VIP9 is upto date with all the takeaways from this thread and will be the spec for what I am building:

https://github.com/varnishcache/varnish-cache/wiki/VIP9:-Expand-VCL-object-support

PR #3143, opened 2019-11-23
Return the vmod_blob architecture upside-down

Dridi:
Yesterday I went through a copy-pasta of #3022, but despite an even larger change count (+606-592 vs +468-432) this time it should be easier to review and more incremental. I also had one more change and made the last include table go away.

Prior to this pull request I pushed some changes directly to master to reduce the noise.


nigoroll:
I would want to pass my review hat to @slimhazard

nigoroll:
I recevied out-of-band info from @slimhazard that he has the review on his backlog.

nigoroll:
bugwash: to be settled between @slimhazard and @Dridi 
I will also review

nigoroll:
@slimhazard's boss says that @slimhazard has important high prio work to tick off atm, but he promises that @slimhazard will get a chance to spend time on the review once this PRs gets near the bottom of the list

nigoroll:
for the failing decode, we should have a way to detect it from vcl.

nigoroll:
@Dridi I have had a serious go at your diffs, and I do not want to whine about insignificant stuff, but I can understand better now why @slimhazard had reservations to just have a go and approve. You are mixing good ideas with useless reformatting and renaming in the same commit, which gives the whole thing a bad smell also in my view. You might think other variable names are batter, you might think that representing integers in hex is better than decimal (interestingly, why for the hex encoding and not for base64?), you might think that stuff should be moved around in source files, but mixing the real improvements with all this clutter distracts me from what is really going on.

I would really like to see concise patches with the actual changes, and if we really need to move stuff around for clarity, we usually do that in a separate commit.

Do you think you could re-do this in a minimal, concise way?

Dridi:
> You might think other variable names are batter,

Yes, I needed better names to wrap my head around the code, and needed consistency.

> you might think that representing integers in hex is better than decimal (interestingly, why for the hex encoding and not for base64?),

Nothing against base64, only me having fun representing hexadecimal values in hexadecimal.

> you might think that stuff should be moved around in source files, but mixing the real improvements with all this clutter distracts me from what is really going on.

I can understand that.

> I would really like to see concise patches with the actual changes, and if we really need to move stuff around for clarity, we usually do that in a separate commit.
> 
> Do you think you could re-do this in a minimal, concise way?

I'm not motivated to spread this change over even more commits, it's hard enough to work around all the entanglement in the first place.

PR #3154, opened 2019-12-12
Add a convenience wrapper to VSA_Build sparing the additional sockaddr detour

nigoroll:
VSA_BuildFAP: Build a suckaddr from Family, Address and Port

Follow the spirit of the vsa.c top level rant and spare callers the hassle of creating sockaddrs specific to ip4/ip6 just to build a VSA, which is intended to avoid having to special-case the protocols in the first place.

Usage is intentionally simple: The interface just takes an address family, an address and a port together with length arguments for additional safely. NULL / 0 arguments are accepted and will create zero address elements.

The first commit of the PR contains the wrapper code with a basic rewrite of `vpx_proto2()` as the current use case in core code. The second commit show how `vpx_proto2()` can be further condensed because special casing is not required any more.

More use cases are to follow.

The name of the function only signifies "Family, Address and Port" and is by no means intended as a pun inspired by _suckaddr_. ;)

### DIFF
--- a/lib/libvarnish/vsa.c
+++ b/lib/libvarnish/vsa.c
@@ -187,16 +187,7 @@ const struct suckaddr *bogo_ip = &bogo_ip_vsa;
 void
 VSA_Init()
 {
-	struct addrinfo hints, *res = NULL;
-
-	memset(&hints, 0, sizeof(hints));
-	hints.ai_family = AF_INET;
-	hints.ai_flags = AI_NUMERICHOST | AI_NUMERICSERV;
-	hints.ai_socktype = SOCK_STREAM;
-	AZ(getaddrinfo("0.0.0.0", "0", &hints, &res));
-	AN(VSA_Build(&bogo_ip_vsa, res->ai_addr, res->ai_addrlen));
-	assert(VSA_Sane(bogo_ip));
-	freeaddrinfo(res);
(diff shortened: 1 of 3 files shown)


lib/libvarnish/vsa.c
```diff
@@ -264,6 +255,58 @@ VSA_Malloc(const void *s, unsigned  sal)
...
+			if (pl != sizeof(sin6.sin6_port))
+				break;
+			memcpy(&sin6.sin6_port, p, pl);
+		}
+		return (VSA_Build(d, &sin6, sizeof sin6));
+	default:
+		return (NULL);
+	}
+	errno = EINVAL;
+	return (NULL);
```
Dridi: Shouldn't the default case `break` out of the `switch` statement to also fail with `EINVAL`?

nigoroll:
pow-wow:
* The patch missed to account for (struct sockaddr).sa_len where exists
* The API way would be to go through the resolver
* Yet we want to avoid that for performance & simplicity

plan:
* add autocrap for `HAVE_STRUCT_SOCKADDR_SA_LEN`
* @Dridi kindly offered to look after this
* fill `.len` if required

Dridi:
https://github.com/varnishcache/varnish-cache/commit/98b38e9f9d67aac16e0c7da884a45361c572640a

nigoroll:
FTR @Dridi regarding _rework proxy code to avoid magic offsets_ I think the offsets actually are not magic at all, do you want to have another look maybe?

PR #3157, opened 2019-12-13
add acl entry checks for canonical form, normalize and add vcc_acl_pedantic parameter

nigoroll:
Prerequisites
-----------------

This PR builds upon #3154 . Reviewers are asked to please ignore the first two commits in the scope of this PR and tend to the other ticket with respect to addition of the `VSA_BuildFAP()` function.

Overview
------------

ACL entries with netmasks shorter than the maximum for the respective protocol represent network addresses and as such, by convention, should be written with all zero bits in the host part to avoid confusion.

For non-canonical ACL entries, this patch adds VCL compile warnings, normalizes the ACL entries and improves logging. It also adds the parameter `vcc_acl_pedantic` to optionally enforce ACL entries be written in normalized form.

The ACL matching code itself remains untouched, though the normalization from this PR will allow for a minor simplification. This is marked with an `XXX`

Discussion
-------------

For example, while 1.2.3.0/24 and 1.2.3.255/24, in CIDR notation, both specify all addresses with the first three octets matching 1, 2 and 3, using the latter can be a source of subtle confusion.

This becomes particularly apparent with netmasks outside byte boundaries: 1.2.6.0/22 specifies addresses 1.2.4.0 - 1.2.7.255, but not so experienced administrators might be tempted to think that it specified 1.2.6.0 - 1.2.9.255.

To summarize, denoting network addresses in non-canonical form is confusing, a possible source of error and additionally complicates analyses.

This patch makes sure that such mishaps do not remain unnoticed by

* issuing warnings during VCL compilation about non-canonical network addresses or optionally failing VCL compilation for such cases

* Logging ACL matches together with the canonical address


nigoroll:
pow-wow:
* improve existing ACL tests to an extent as close as possible to "100% sure it's correct"
* then revisit this PR

nigoroll:
pow-wow:
* noticed that things like "fqdn"/24 are a security risk because it could resolve to IPv6
  * this PR addresses this by requiring the host part to be zero
* using DNS for ACLs will still be possible if the host part is zero
* merge now
* turn the default to "on" before 7.0 release

PR #3158, opened 2019-12-13
Generalize type methods (take 2)

Dridi:
In the first iteration (#3147) I used the `::` notation to fit type methods in the VCC symbol table in a way that would avoid symbol conflicts:

- some reserved keywords like `acl` or `backend` are type names
- by default when vmod_blob is imported its symbol is a type name

During the review the `::` hack was understandably not well received, and after a fairly long discussion I pushed a few changes directly to master, took a slightly different direction (like unwildcarding VMOD symbols) and then solved the symbol conflict by introducing a `_type` namespace in the symbol table.

This is otherwise the same as #3147: it does not (intend to) change existing behavior and only prepare libvcc to accommodate future changes in the scope of VIP28. The difference is that we can keep vmod_blob as-is (or hopefully slightly different after #3143) and we maintain a hierarchical symbol table.

For everything else, please refer to the commit log. It's a mixed bag of new commits and cherry-picks from #3147. And some changes from #3147 already landed in master.

### DIFF
--- a/bin/varnishtest/tests/v00018.vtc
+++ b/bin/varnishtest/tests/v00018.vtc
@@ -133,3 +133,23 @@ varnish v1 -syntax 4.0 -errvcl {Symbol not found:} {
 		return (vcl(vcl_recv));
 	}
 }
+
+varnish v1 -errvcl {Syntax error} {
+	import directors;
+	sub vcl_recv {
+		set req.backend_hint = directors.round_robin.backend();
+	}
+}
+
+varnish v1 -errvcl {Syntax error} {
+	import directors;
+	sub vcl_recv {
+		directors.round_robin.backend();
+	}
+}
+
+varnish v1 -errvcl {Syntax error} {
+	sub vcl_recv {
+		_type.strings.upper();
+	}
+}
--- a/bin/varnishtest/tests/v00018.vtc
+++ b/bin/varnishtest/tests/v00018.vtc
@@ -153,3 +153,11 @@ varnish v1 -errvcl {Syntax error} {
 		_type.strings.upper();
 	}
 }
+
+varnish v1 -errvcl {Expected '.' got ';'} {
+	import directors;
+	sub vcl_init {
+		new rr = directors.round_robin();
+		rr;
+	}
+}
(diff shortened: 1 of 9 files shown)


Dridi:
On github commits are sorted by date instead of commit order, and the result looks pretty bad for this pull request. I recommend to check directly via the git command line instead...

rezan:
Just to understand, what direction are we going with this? My understanding is that VIP28 has not been accepted as anything. My person feedback is that it combines way too many ideas together. So that is why I have a hard time understanding what is being built here.

I am totally behind plain and simple VCL types. Given we already have c based VMOD types, its steaight forward to bring those types forward into VCL. I just don't see an actual proposal for this anywhere.

Dridi:
@rezan this is understandably a bit confusing to follow. When VIP28 was submitted it proposed a direction and the consensus was a different one: enrich the type system to generalize type methods, and back operations with specific methods: upon encountering the `+` operator for an INT expression, look for the `INT.__add` method. That would be `INT.__add(INT)` so the expression parser would then expect an `INT` expression on the RHS of that `+` token.

The benefit would be less hard-coded knowledge in the parser, and more information provided by the type system instead. With the possibility for VMOD objects to interact in a similar fashion with the parser, telling it how a VMOD object might be used in an expression, solely based on the methods it would provide.

This patch series carves the type system in the symbol table, enabling the generalization of expression parsing down the road.

rezan:
K, this does make sense now.

Dridi:
I have revisited this pull request in the context of #3211 and took the opportunity to rebase it against current master. In the process I realized that the `HEADER` type can't have properties or methods (see the #3211 description) but I would like to keep this question out of the scope of this pull request.

We would need a use case anyway to fix it like I did in #3211. Such a use case could be a `HEADER.count` property telling you how many occurrences there are for a given header.

Dridi:
Review summary with @bsdphk and @nigoroll: I will attempt a similar approach but with separate symbol tables instead of namespaces in the main table.

Dridi:
Closing since take 3 was approved, there might be a take 4.

PR #3161, opened 2019-12-16
Use absolute paths in dlopen(3) for VCLs and VMODs

Dridi:
This improves the out-of-the-box debugging experience for at least gdb
and lldb, allowing them to grab type information for the symbols coming
from such shared objects.

This however doesn't help further when multiple VCLs or VMODs define the
same symbols, for example in the case of similar VCLs (reload scenario)
or a VMOD upgrade (either with import+from or VMOD caching kicking in).

For the latter case, #2966 would help.

### DIFF
--- a/bin/varnishd/common/heritage.h
+++ b/bin/varnishd/common/heritage.h
@@ -73,6 +73,7 @@ struct heritage {
 	struct params			*param;
 
 	const char			*identity;
+	const char			*workdir;
 
 	char				*panic_str;
 	ssize_t				panic_str_len;
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -738,6 +738,7 @@ main(int argc, char * const *argv)
 
 	if (VIN_n_Arg(n_arg, &dirname) != 0)
 		ARGV_ERR("Invalid instance (-n) name: %s\n", vstrerror(errno));
+	heritage.workdir = dirname;
 
 	if (i_arg == NULL || *i_arg == '\0')
 		i_arg = mgt_HostName();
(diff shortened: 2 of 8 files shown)


nigoroll:
I do not see any issue with the `cd $varnish_workdir` required without this patch, but on the other hand I do not wand to be in the way if such a change is otherwise welcome, so I would appreciate more opinions.
Trying to think of possible drawbacks I wonder if this could affect folks who load binary VCLs?

Dridi:
I suppose that removes a `cd` step for you and an undefined step for me with `coredumpctl`, maybe changing directories would be enough to do the trick also for me.

nigoroll:
FTR, this discussion made me wonder if it would not be possible to include used shared libraries including their symbol information in the core dump. Results so far:
* on linux, changing `coredump_filter` (see _core(5)_) to `0xff` actually increases the size of the coredump as expected, but does not make `gdb` find symbols for `vgc.so`
* I dug a bit but could not find anything else regarding the matter.

I still think that making the core dump self contained with all the symbols should be the best option for our case, and I wonder if achieving this would be possible somehow.

Dridi:
There is probably more we can do to improve the debugging experience. To be fair, most of the time a description of the problem and the panic output is enough to come up with a VTC reproducer, so I never had an incentive to improve my `gdb` knowledge when working on Varnish.

So, this and #2966 are certainly not the only things we may need in this vicinity. Please note that this PR focused on `dlopen` in the child process, where we get the most value for core dump improvements.

nigoroll:
Thoughts about cases where the current relative path could be advantageous:
* transporting the vgc.so to servers with different paths or using it for more than one instance on the same server
* chroot
* permissions: varnishd could live in a working-directory without permissions to directories above.

So because, in my mind, this PR only saves one cd command and likely carries some drawbacks with it, I would rather improve the documentation on debugging than see this one merged.

@Dridi Other than that I do not mean that we should not aim to make debugging easier and if there is a good reason which I am overlooking, I am all for changing my mind.

Dridi:
You make a fair point, I would even argue against transporting the `vgc.so` file and instead the full work directory.

PR #3162, opened 2019-12-16
Add "error" state to backend fetch and response - backport

andrewwiik:
This adds a new return "error" to "backend_fetch" and "backend_response" that will then handle over control to "backend_error". Currently abandon and fail in these methods currently backs totally out of backend. This error return also allows an optional reason and status code to be provided, just like the synth return except it does not require them, they are optional.

Look at v00062.vtc for all the possible use syntax for the new error return.

This is backport cherry-picked from master for 6.0


andrewwiik:
I can’t merge it

PR #3163, opened 2019-12-17
a vmod type for vcl subs & VRT_call()

nigoroll:
use case
-----------

This PR implements a vmod type for vcl subs with the goal of being able to use vcl subs as parameters to vmod function/method calls, which allows them to call back into VCL subs. The primary, obvious use case is a clean interface for [vmod_dispatch](https://code.uplex.de/uplex-varnish/libvmod-dispatch). Another use case I have in mind would be the option to add iterator vmod functions, for example to tokenize a string and call a sub for each token with a task variable set.

implementation
--------------------

We add to VRT and initialize for each vcl sub a `struct vcl_sub` which contains:

* a methods bitmask defining which contexts this sub may be called from
* the name as seen from vcl
* the function pointer to the `vcl_func_f`

As the latter two are trivial, I will only go into detail about the methods bitmask:

It contains the `VCL_MET_*` bits of contexts which may potentially call this sub and is intended for later use with checks like

```
if (sub->methods & ctx->method)
    sub->func(ctx);
else
    VRT_fail(ctx, "not allowed here");
```

(see also `vmod_debug` poc code from the second commit) 

In existing code, we check if used objects and returned actions are allowed within the respective calling context.

To build the methods bitmask, we now begin with an all-one bitfield and clear the bits of methods which would not be allowed for any used object or returned action.

vcc interface
----------------

for the VCC interface, we keep the VCL sub c function name in the symbol's `lname`, and put the pointer to the `struct vcl_sub` in the `rname`. This way, we require only a few additional changes and I believe that the semantics match: The "assignable" name is the c function name, while any read access to a VCL sub has to go through the "read" name.

### DIFF
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -487,7 +487,7 @@ vcl_call_method(struct worker *wrk, struct req *req, struct busyobj *bo,
 	wrk->seen_methods |= method;
 	AN(ctx.vsl);
 	VSLb(ctx.vsl, SLT_VCL_call, "%s", VCL_Method_Name(method));
-	func(&ctx);
+	func(&ctx, VSUB_STATIC, NULL);
 	VSLb(ctx.vsl, SLT_VCL_return, "%s", VCL_Return_Name(wrk->handling));
 	wrk->cur_method |= 1;		// Magic marker
 	if (wrk->handling == VCL_RET_FAIL)
@@ -517,3 +517,51 @@ VCL_##func##_method(struct vcl *vcl, struct worker *wrk,		\
 }
 
 #include "tbl/vcl_returns.h"
+
+/*--------------------------------------------------------------------
+ */
+
+VCL_STRING
+VRT_check_call(VRT_CTX, VCL_SUB sub)
+{
+	VCL_STRING err = NULL;
+	enum vcl_func_fail_e fail;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(sub, VCL_SUB_MAGIC);
+
+	AN(sub->func);
+	sub->func(ctx, VSUB_CHECK, &fail);
+
+	switch (fail) {
+	case VSUB_E_OK:
+		break;
+	case VSUB_E_METHOD:
+		err = WS_Printf(ctx->ws, "Dynamic call to \"sub %s{}\""
+		    " not allowed from here", sub->name);
+		if (err == NULL)
+			err = "Dynamic call not allowed and workspace overflow";
+		break;
+	case VSUB_E_RECURSE:
+		err = WS_Printf(ctx->ws, "Recursive dynamic call to"
+		    " \"sub %s{}\"", sub->name);
+		if (err == NULL)
+			err = "Recursive dynamic call and workspace overflow";
+		break;
+	default:
+		INCOMPL();
+	}
+
+	return (err);
+}
+
+VCL_VOID
+VRT_call(VRT_CTX, VCL_SUB sub)
+{
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(sub, VCL_SUB_MAGIC);
+
+	AN(sub->func);
+	sub->func(ctx, VSUB_DYNAMIC, NULL);
+}
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -39,9 +39,11 @@
 
 #include "vcl.h"
 #include "vtim.h"
+#include "vbm.h"
 
 #include "cache_director.h"
 #include "cache_vcl.h"
+#include "vcc_interface.h"
 
 /*--------------------------------------------------------------------*/
 
@@ -444,10 +446,13 @@ VRT_VCL_Allow_Discard(struct vclref **refp)
 
 static void
 vcl_call_method(struct worker *wrk, struct req *req, struct busyobj *bo,
-    void *specific, unsigned method, vcl_func_f *func)
+    void *specific, unsigned method, vcl_func_f *func, unsigned track_call)
 {
-	uintptr_t aws;
+	uintptr_t rws = 0, aws;
 	struct vrt_ctx ctx;
+	struct vbitmap *vbm;
+	void *p;
+	size_t sz;
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	INIT_OBJ(&ctx, VRT_CTX_MAGIC);
@@ -468,6 +473,15 @@ vcl_call_method(struct worker *wrk, struct req *req, struct busyobj *bo,
 	assert(ctx.now != 0);
 	ctx.specific = specific;
 	ctx.method = method;
+	if (track_call > 0) {
+		rws = WS_Snapshot(wrk->aws);
+		sz = VBITMAP_SZ(track_call);
+		p = WS_Alloc(wrk->aws, sz);
+		// No use to attempt graceful failure, all VCL calls will fail
+		AN(p);
+		vbm = vbit_init(p, sz);
+		ctx.called = vbm;
+	}
 	aws = WS_Snapshot(wrk->aws);
 	wrk->cur_method = method;
 	wrk->seen_methods |= method;
@@ -484,6 +498,8 @@ vcl_call_method(struct worker *wrk, struct req *req, struct busyobj *bo,
 	 * wrk->aws, but they can reserve and return from it.
 	 */
 	assert(aws == WS_Snapshot(wrk->aws));
+	if (rws != 0)
+		WS_Reset(wrk->aws, rws);
 }
 
 #define VCL_MET_MAC(func, upper, typ, bitmap)				\
@@ -496,7 +512,7 @@ VCL_##func##_method(struct vcl *vcl, struct worker *wrk,		\
 	CHECK_OBJ_NOTNULL(vcl->conf, VCL_CONF_MAGIC);			\
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);				\
 	vcl_call_method(wrk, req, bo, specific,				\
-	    VCL_MET_ ## upper, vcl->conf->func##_func);			\
+	    VCL_MET_ ## upper, vcl->conf->func##_func, vcl->conf->nsub);\
 	AN((1U << wrk->handling) & bitmap);				\
 }
 
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -56,6 +56,19 @@ Varnish Cache Next (2021-03-15)
 * All shard ``Error`` and ``Notice`` messages now use the unified
   prefix ``vmod_directors: shard %s``.
 
+* The ``VCL_SUB`` data type is now supported for VMODs to save
+  references to subroutines to be called later using
+  ``VRT_call()``. Calls from a wrong context (e.g. calling a
+  subroutine accessing ``req`` from the backend side) and recursive
+  calls fail the VCL.
+
+* ``VRT_check_call()`` can be used to check if a ``VRT_call()`` would
+  succeed in order to avoid the potential VCL failure in case it would
+  not.
+
+  It returns ``NULL`` if ``VRT_call()`` would make the call or an
+  error string why not.
+
 ================================
 Varnish Cache 6.5.1 (2020-09-25)
 ================================
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -70,6 +70,8 @@
  *	VRT_re_fini removed
  *	VRT_re_match signature changed
  *	VRT_regsub signature changed
+ *	VRT_call() added
+ *	VRT_check_call() added
  * 12.0 (2020-09-15)
  *	Added VRT_DirectorResolve()
  *	Added VCL_STRING VRT_BLOB_string(VRT_CTX, VCL_BLOB)
@@ -339,13 +341,19 @@ struct vrt_ctx {
 #define VRT_CTX		const struct vrt_ctx *ctx
 void VRT_CTX_Assert(VRT_CTX);
 
+enum vcl_func_call_e {
+	VSUB_STATIC,	// VCL "call" action, only allowed from VCC
+	VSUB_DYNAMIC,	// VRT_call()
+	VSUB_CHECK	// VRT_check_call()
+};
+
 enum vcl_func_fail_e {
 	VSUB_E_OK,
 	VSUB_E_RECURSE, // call would recurse
 	VSUB_E_METHOD	// can not be called from this method
 };
 
-typedef void vcl_func_f(VRT_CTX);
+typedef void vcl_func_f(VRT_CTX, enum vcl_func_call_e, enum vcl_func_fail_e *);
 
 /***********************************************************************
  * This is the interface structure to a compiled VMOD
@@ -415,6 +423,10 @@ char *VRT_Strands(char *, size_t, VCL_STRANDS);
 VCL_STRING VRT_StrandsWS(struct ws *, const char *, VCL_STRANDS);
 VCL_STRING VRT_CollectStrands(VRT_CTX, VCL_STRANDS);
 VCL_STRING VRT_UpperLowerStrands(VRT_CTX, VCL_STRANDS s, int up);
+
+/* VCL_SUB */
+VCL_STRING VRT_check_call(VRT_CTX, VCL_SUB);
+VCL_VOID VRT_call(VRT_CTX, VCL_SUB);
 
 /* Functions to turn types into canonical strings */
 
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -211,6 +211,7 @@ struct vsc_seg;
 struct vsl_log;
 struct vsmw_cluster;
 struct ws;
+struct vcl_sub;
 
 /***********************************************************************
  * VCL_STRANDS:
@@ -283,6 +284,7 @@ typedef const struct vre *			VCL_REGEX;
 typedef const struct stevedore *		VCL_STEVEDORE;
 typedef const struct strands *			VCL_STRANDS;
 typedef const char *				VCL_STRING;
+typedef const struct vcl_sub *			VCL_SUB;
 typedef vtim_real				VCL_TIME;
 typedef struct vcl *				VCL_VCL;
 typedef void					VCL_VOID;
@@ -334,6 +336,8 @@ struct vrt_ctx {
 
 #define VRT_CTX		const struct vrt_ctx *ctx
 void VRT_CTX_Assert(VRT_CTX);
+
+typedef void vcl_func_f(VRT_CTX);
 
 /***********************************************************************
  * This is the interface structure to a compiled VMOD
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -332,6 +332,8 @@ struct vrt_ctx {
 	 *    synth+error:	struct vsb *
 	 */
 	void				*specific;
+	/* if present, vbitmap of called subs */
+	void				*called;
 };
 
 #define VRT_CTX		const struct vrt_ctx *ctx
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -339,6 +339,12 @@ struct vrt_ctx {
 #define VRT_CTX		const struct vrt_ctx *ctx
 void VRT_CTX_Assert(VRT_CTX);
 
+enum vcl_func_fail_e {
+	VSUB_E_OK,
+	VSUB_E_RECURSE, // call would recurse
+	VSUB_E_METHOD	// can not be called from this method
+};
+
 typedef void vcl_func_f(VRT_CTX);
 
 /***********************************************************************
--- a/lib/libvcc/generate.py
+++ b/lib/libvcc/generate.py
@@ -666,7 +666,6 @@ fo.write("""
 typedef int vcl_event_f(VRT_CTX, enum vcl_event_e);
 typedef int vcl_init_f(VRT_CTX);
 typedef void vcl_fini_f(VRT_CTX);
-typedef void vcl_func_f(VRT_CTX);
 
 struct VCL_conf {
 	unsigned		magic;
@@ -679,6 +678,7 @@ struct VCL_conf {
 	const struct vpi_ref	*ref;
 
 	int			nsrc;
+	unsigned		nsub;
 	const char		**srcname;
 	const char		**srcbody;
 
--- a/lib/libvcc/vcc_symb.c
+++ b/lib/libvcc/vcc_symb.c
@@ -471,6 +471,14 @@ VCC_GlobalSymbol(struct symbol *sym, vcc_type_t type, const char *pfx)
 	VSB_printf(vsb, "%s_", pfx);
 	VCC_PrintCName(vsb, sym->name, NULL);
 	AZ(VSB_finish(vsb));
+	sym->lname = strdup(VSB_data(vsb));
+	if (type == SUB) {
+		VSB_destroy(&vsb);
+		vsb = VSB_new_auto();
+		AN(vsb);
+		VSB_printf(vsb, "sub_%s", sym->lname);
+		AZ(VSB_finish(vsb));
+	}
 	sym->rname = strdup(VSB_data(vsb));
 	AN(sym->rname);
 	VSB_destroy(&vsb);
--- a/bin/varnishd/cache/cache_vpi.c
+++ b/bin/varnishd/cache/cache_vpi.c
@@ -35,6 +35,7 @@
 #include "cache_varnishd.h"
 
 #include "vcl.h"
+#include "vbm.h"
 
 #include "vcc_interface.h"
 
@@ -139,4 +140,49 @@ VPI_Fail(const char *func, const char *file, int line,
     const char *cond)
 {
 	VAS_Fail(func, file, line, cond, VAS_ASSERT);
+}
+
+enum vcl_func_fail_e
+VPI_Call_Check(VRT_CTX, const struct VCL_conf *conf,
+    unsigned methods, unsigned n)
+{
+	struct vbitmap *vbm;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(ctx->vcl, VCL_MAGIC);
+
+	assert(conf == ctx->vcl->conf);
+
+	vbm = ctx->called;
+	AN(vbm);
+
+	if ((methods & ctx->method) == 0)
+		return (VSUB_E_METHOD);
+
+	if (vbit_test(vbm, n))
+		return (VSUB_E_RECURSE);
+
+	return (VSUB_E_OK);
+}
+
+void
+VPI_Call_Begin(VRT_CTX, unsigned n)
+{
+	struct vbitmap *vbm;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	vbm = ctx->called;
+	AN(vbm);
+	vbit_set(vbm, n);
+}
+
+void
+VPI_Call_End(VRT_CTX, unsigned n)
+{
+	struct vbitmap *vbm;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	vbm = ctx->called;
+	AN(vbm);
+	vbit_clr(vbm, n);
 }
--- a/bin/varnishtest/tests/m00053.vtc
+++ b/bin/varnishtest/tests/m00053.vtc
@@ -1,0 +1,194 @@
+varnishtest "VCL_SUB type basics"
+
+varnish v1 -vcl {
+	import std;
+	import debug;
+
+	backend dummy None;
+
+	# call: 1 ref: 5 + 1
+	sub foo {
+		set resp.http.it = "works";
+	}
+	# call: 1 ref: 1 + 1 (from bar walk)
+	sub barbar {
+		# called from bar only, must be marked dynamic
+		call barbarbar;
+	}
+	# call: 2 (1 from bar walk, 1 from barbar walk)
+	# ref: 2 + 1
+	sub barbarbar {
+		# 2nd level call from dynamic
+	}
+	# call: 0 ref: 1
+	sub bar {
+		set beresp.http.it = "works";
+		call barbar;
+	}
+	# call: 0 ref: 1
+	sub indirect {
+		debug.call(foo);
+	}
+	# call: 0 ref: 1
+	sub direct {
+		call foo;
+	}
+	# call: 0 ref: 2
+	sub recursive {
+		std.log("recursive called");
+		debug.call(recursive);
+	}
+	# call: 1 ref: 1
+	sub recursive_indirect {
+		std.log("recursive_indirect called");
+		debug.call(recursive_indirect);
+	}
+	# call: 1 ref: 1
+	sub rollback {
+		std.rollback(req);
+	}
+	# call: 0 ref: 1
+	sub priv_top {
+		debug.test_priv_top("only works on client side");
+	}
+
+	sub vcl_recv {
+		if (req.url == "/wrong") {
+			debug.call(foo);
+		}
+		if (req.url == "/recursive") {
+			debug.call(recursive);
+		}
+		if (req.url == "/recursive_indirect") {
+			call recursive_indirect;
+		}
+		if (req.url == "/priv_top") {
+			return (pass);
+		}
+		return (synth(200));
+	}
+	sub vcl_synth {
+		if (req.url == "/foo") {
+			debug.call(foo);
+		} else if (req.url == "/direct") {
+			debug.call(direct);
+		} else if (req.url == "/indirect") {
+			debug.call(indirect);
+		} else if (req.url == "/rollback") {
+			debug.call(rollback);
+		} else if (req.url == "/callthenrollback") {
+			debug.call(foo);
+			call rollback;
+			if (! resp.http.it) {
+				set resp.http.rolledback = true;
+			}
+			debug.call(foo);
+		} else if (req.url == "/checkwrong") {
+			synthetic(debug.check_call(bar));
+			set resp.status = 500;
+		}
+		return (deliver);
+	}
+	sub vcl_backend_fetch {
+		debug.call(priv_top);
+	}
+	sub vcl_backend_error {
+		# falling through to None backend would be success
+		set beresp.status = 200;
+		return (deliver);
+	}
+} -start
+
+client c1 {
+	txreq -url "/foo"
+	rxresp
+	expect resp.status == 200
+	expect resp.http.it == "works"
+
+	txreq -url "/direct"
+	rxresp
+	expect resp.status == 200
+	expect resp.http.it == "works"
+
+	txreq -url "/indirect"
+	rxresp
+	expect resp.status == 200
+	expect resp.http.it == "works"
+
+	txreq -url "/callthenrollback"
+	rxresp
+	expect resp.status == 200
+	expect resp.http.rolledback == "true"
+	expect resp.http.it == "works"
+
+	txreq -url "/wrong"
+	rxresp
+	expect resp.status == 503
+} -start
+logexpect l2 -v v1 -g vxid -q {ReqURL ~ "^/recursive$"} {
+	expect *  *	VCL_Log	"^recursive called"
+	fail add  *	VCL_Log	"^recursive called"
+	expect 0  =	VCL_Error	{^Recursive call to "sub recursive..}
+	expect 0  =	VCL_return	"^fail"
+	expect *  =	End
+	fail clear
+} -start
+client c2 {
+	txreq -url "/recursive"
+	rxresp
+	expect resp.status == 503
+} -start
+logexpect l3 -v v1 -g vxid -q {ReqURL ~ "^/recursive_indirect$"} {
+	expect *  *	VCL_Log	"^recursive_indirect called"
+	fail add  *	VCL_Log	"^recursive_indirect called"
+	expect 0  =	VCL_Error	{^Recursive call to "sub recursive_indirect..}
+	expect 0  =	VCL_return	"^fail"
+	expect *  =	End
+	fail clear
+} -start
+client c3 {
+	txreq -url "/recursive_indirect"
+	rxresp
+	expect resp.status == 503
+} -start
+client c4 {
+	txreq -url "/rollback"
+	rxresp
+	expect resp.status == 200
+} -start
+client c5 {
+	txreq -url "/checkwrong"
+	rxresp
+	expect resp.status == 500
+	expect resp.body == {Dynamic call to "sub bar{}" not allowed from here}
+} -start
+client c6 {
+	txreq -url "/priv_top"
+	rxresp
+	expect resp.status == 503
+} -start
+
+varnish v1 -errvcl {Impossible Subroutine('<vcl.inline>' Line 8 Pos 13)} {
+	import std;
+	import debug;
+
+	backend dummy None;
+
+	sub impossible {
+		set req.http.impossible = beresp.reason;
+	}
+	sub vcl_recv {
+		if (req.url == "/impossible") {
+			debug.call(impossible);
+		}
+	}
+}
+
+client c1 -wait
+client c2 -wait
+logexpect l2 -wait
+client c3 -wait
+logexpect l3 -wait
+client c4 -wait
+client c5 -wait
+client c6 -wait
--- a/bin/varnishtest/tests/m00054.vtc
+++ b/bin/varnishtest/tests/m00054.vtc
@@ -1,0 +1,36 @@
+varnishtest "VCL_SUB wrong vmod behavior"
+
+varnish v1 -arg "-p feature=+no_coredump" -vcl {
+	import debug;
+
+	backend dummy None;
+
+	sub foo {
+		set resp.http.it = "works";
+	}
+
+	sub vcl_init {
+		debug.bad_memory(foo);
+	}
+} -start
+
+varnish v1 -vcl {
+	import debug;
+	backend dummy None;
+
+	sub vcl_recv {
+		debug.call(debug.total_recall());
+	}
+
+}
+
+client c1 {
+	txreq -url "/foo"
+	expect_close
+} -run
+
+varnish v1 -wait-stopped
+varnish v1 -cliexpect "Assert error in VPI_Call_Check" "panic.show"
+varnish v1 -cliok "panic.clear"
+varnish v1 -expect MGT.child_panic == 1
+varnish v1 -expectexit 0x40
--- a/bin/varnishtest/tests/v00020.vtc
+++ b/bin/varnishtest/tests/v00020.vtc
@@ -246,6 +246,17 @@ varnish v1 -errvcl {Not available in subroutine 'vcl_recv'.} {
 	}
 }
 
+varnish v1 -cliok "param.set vcc_err_unref off"
+
+varnish v1 -errvcl {Impossible Subroutine} {
+	backend b { .host = "127.0.0.1"; }
+	sub foo {
+		set req.http.foo = 100 + beresp.status;
+	}
+}
+
+varnish v1 -cliok "param.set vcc_err_unref on"
+
 varnish v1 -errvcl {
 ('<vcl.inline>' Line 4 Pos 44) -- (Pos 49)
         sub foo { set req.http.foo = 100 + beresp.status; }
--- a/bin/varnishtest/tests/v00021.vtc
+++ b/bin/varnishtest/tests/v00021.vtc
@@ -49,11 +49,6 @@ Subroutine recurses on
         sub foo { call foo; }
 -----------------------###---
 
-
-...called from "vcl_recv"
-('<vcl.inline>' Line 6 Pos 29)
-        sub vcl_recv { call foo; }
-----------------------------###---
 } {
 	backend b { .host = "${localhost}"; }
 
@@ -62,16 +57,9 @@ Subroutine recurses on
 }
 
 varnish v1 -errvcl {
-Subroutine recurses on
-('<vcl.inline>' Line 6 Pos 13)
-        sub foo { call bar; }
-------------###--------------
-
-
-...called from "bar"
-('<vcl.inline>' Line 5 Pos 24)
+('<vcl.inline>' Line 5 Pos 13)
         sub bar { call foo; }
------------------------###---
+------------###--------------
 
 
 ...called from "foo"
@@ -80,10 +68,11 @@ Subroutine recurses on
 -----------------------###---
 
 
-...called from "vcl_recv"
-('<vcl.inline>' Line 7 Pos 29)
-        sub vcl_recv { call foo; }
-----------------------------###---
+...called from "bar"
+('<vcl.inline>' Line 5 Pos 24)
+        sub bar { call foo; }
+-----------------------###---
+
 } {
 	backend b { .host = "${localhost}"; }
 
--- a/doc/sphinx/reference/vmod.rst
+++ b/doc/sphinx/reference/vmod.rst
@@ -460,6 +460,26 @@ TIME
 
 	An absolute time, as in 1284401161.
 
+VCL_SUB
+	C-type: ``const struct vcl_sub *``
+
+	Opaque handle on a VCL subroutine.
+
+	References to subroutines can be passed into VMODs as
+	arguments and called later through ``VRT_call()``. The scope
+	strictly is the VCL: vmods must ensure that ``VCL_SUB``
+	references never be called from a different VCL.
+
+	``VRT_call()`` fails the VCL for recursive calls and when the
+	``VCL_SUB`` can not be called from the current context
+	(e.g. calling a subroutine accessing ``req`` from the backend
+	side).
+
+	``VRT_check_call()`` can be used to check if a ``VRT_call()``
+	would succeed in order to avoid the potential VCL failure.  It
+	returns ``NULL`` if ``VRT_call()`` would make the call or an
+	error string why not.
+
 VOID
 	C-type: ``void``
 
--- a/include/vcc_interface.h
+++ b/include/vcc_interface.h
@@ -81,3 +81,15 @@ struct vpi_ii {
 
 void VPI_re_init(struct vre **, const char *);
 void VPI_re_fini(struct vre *);
+
+/* VCL_SUB type */
+
+struct vcl_sub {
+	unsigned		magic;
+#define VCL_SUB_MAGIC		0x12c1750b
+	const unsigned		methods;	// ok &= ctx->method
+	const char * const	name;
+	const struct VCL_conf	*vcl_conf;
+	vcl_func_f		*func;
+	unsigned		n;
+};
--- a/include/vcc_interface.h
+++ b/include/vcc_interface.h
@@ -92,4 +92,6 @@ struct vcl_sub {
 	const struct VCL_conf	*vcl_conf;
 	vcl_func_f		*func;
 	unsigned		n;
+	unsigned		nref;
+	unsigned		called;
 };
--- a/include/vcc_interface.h
+++ b/include/vcc_interface.h
@@ -95,3 +95,8 @@ struct vcl_sub {
 	unsigned		nref;
 	unsigned		called;
 };
+
+enum vcl_func_fail_e VPI_Call_Check(VRT_CTX, const struct VCL_conf *conf,
+    unsigned methods, unsigned n);
+void VPI_Call_Begin(VRT_CTX, unsigned n);
+void VPI_Call_End(VRT_CTX, unsigned n);
--- a/lib/libvcc/vcc_action.c
+++ b/lib/libvcc/vcc_action.c
@@ -53,7 +53,7 @@ vcc_act_call(struct vcc *tl, struct token *t, struct symbol *sym)
 	if (sym != NULL) {
 		vcc_AddCall(tl, t0, sym);
 		VCC_GlobalSymbol(sym, SUB, "VGC_function");
-		Fb(tl, 1, "%s(ctx);\n", sym->rname);
+		Fb(tl, 1, "%s(ctx);\n", sym->lname);
 		SkipToken(tl, ';');
 	}
 }
--- a/lib/libvcc/vcc_action.c
+++ b/lib/libvcc/vcc_action.c
@@ -315,7 +315,7 @@ vcc_act_return_vcl(struct vcc *tl)
 static void v_matchproto_(sym_act_f)
 vcc_act_return(struct vcc *tl, struct token *t, struct symbol *sym)
 {
-	unsigned hand;
+	unsigned hand, mask;
 	const char *h;
 
 	(void)t;
@@ -335,6 +335,7 @@ vcc_act_return(struct vcc *tl, struct token *t, struct symbol *sym)
 		if (vcc_IdIs(tl->t, #l)) {		\
 			hand = VCL_RET_ ## U;		\
 			h = #U;				\
+			mask = B;			\
 		}
 #include "tbl/vcl_returns.h"
 	if (h == NULL) {
@@ -344,7 +345,7 @@ vcc_act_return(struct vcc *tl, struct token *t, struct symbol *sym)
 	}
 	assert(hand < VCL_RET_MAX);
 
-	vcc_ProcAction(tl->curproc, hand, tl->t);
+	vcc_ProcAction(tl->curproc, hand, mask, tl->t);
 	vcc_NextToken(tl);
 	if (tl->t->tok == '(') {
 		if (hand == VCL_RET_SYNTH || hand == VCL_RET_ERROR)
--- a/lib/libvcc/vcc_action.c
+++ b/lib/libvcc/vcc_action.c
@@ -53,7 +53,8 @@ vcc_act_call(struct vcc *tl, struct token *t, struct symbol *sym)
 	if (sym != NULL) {
 		vcc_AddCall(tl, t0, sym);
 		VCC_GlobalSymbol(sym, SUB, "VGC_function");
-		Fb(tl, 1, "%s(ctx);\n", sym->lname);
+
+		Fb(tl, 1, "%s(ctx, VSUB_STATIC, NULL);\n", sym->lname);
 		SkipToken(tl, ';');
 	}
 }
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -176,6 +176,7 @@ vcc_EmitProc(struct vcc *tl, struct proc *p)
 	AZ(VSB_finish(p->cname));
 	AZ(VSB_finish(p->prologue));
 	AZ(VSB_finish(p->body));
+	AN(p->sym);
 
 	Fh(tl, 1, "vcl_func_f %s;\n", VSB_data(p->cname));
 	Fh(tl, 1, "const struct vcl_sub sub_%s[1] = {{\n",
@@ -185,7 +186,9 @@ vcc_EmitProc(struct vcc *tl, struct proc *p)
 	Fh(tl, 1, "\t.name\t\t= \"%.*s\",\n", PF(p->name));
 	Fh(tl, 1, "\t.vcl_conf\t= &VCL_conf,\n");
 	Fh(tl, 1, "\t.func\t\t= %s,\n", VSB_data(p->cname));
-	Fh(tl, 1, "\t.n\t\t= %d\n", tl->nsub++);
+	Fh(tl, 1, "\t.n\t\t= %d,\n", tl->nsub++);
+	Fh(tl, 1, "\t.nref\t\t= %d,\n", p->sym->nref);
+	Fh(tl, 1, "\t.called\t\t= %d\n", p->called);
 	Fh(tl, 1, "}};\n");
 	/*
 	 * TODO: v_dont_optimize for custom subs called from vcl_init/fini only
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -161,6 +161,7 @@ vcc_NewProc(struct vcc *tl, struct symbol *sym)
 	p->cname = VSB_new_auto();
 	AN(p->cname);
 	sym->proc = p;
+	p->sym = sym;
 	return (p);
 }
 
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -171,8 +171,10 @@ static void
 vcc_EmitProc(struct vcc *tl, struct proc *p)
 {
 	struct vsb *vsbm;
-	unsigned mask;
+	unsigned mask, nsub;
 	const char *maskcmp;
+	const char *cc_adv;
+	int dyn = (p->sym->nref > p->called);
 
 	AN(p->okmask);
 	AZ(VSB_finish(p->cname));
@@ -188,6 +190,13 @@ vcc_EmitProc(struct vcc *tl, struct proc *p)
 		maskcmp = "&";
 	}
 
+	if ((mask & VCL_MET_TASK_H) && (mask & ~VCL_MET_TASK_H) == 0)
+		cc_adv = "v_dont_optimize ";
+	else
+		cc_adv = "";
+
+	nsub = tl->nsub++;
+
 	Fh(tl, 1, "vcl_func_f %s;\n", VSB_data(p->cname));
 	Fh(tl, 1, "const struct vcl_sub sub_%s[1] = {{\n",
 	   VSB_data(p->cname));
@@ -196,14 +205,23 @@ vcc_EmitProc(struct vcc *tl, struct proc *p)
 	Fh(tl, 1, "\t.name\t\t= \"%.*s\",\n", PF(p->name));
 	Fh(tl, 1, "\t.vcl_conf\t= &VCL_conf,\n");
 	Fh(tl, 1, "\t.func\t\t= %s,\n", VSB_data(p->cname));
-	Fh(tl, 1, "\t.n\t\t= %d,\n", tl->nsub++);
+	Fh(tl, 1, "\t.n\t\t= %d,\n", nsub);
 	Fh(tl, 1, "\t.nref\t\t= %d,\n", p->sym->nref);
 	Fh(tl, 1, "\t.called\t\t= %d\n", p->called);
 	Fh(tl, 1, "}};\n");
-	Fc(tl, 1, "\nvoid %sv_matchproto_(vcl_func_f)\n",
-	   ((mask & VCL_MET_TASK_H) && (mask & ~VCL_MET_TASK_H) == 0) ?
-	   "v_dont_optimize " : "");
-	Fc(tl, 1, "%s(VRT_CTX)\n{\n", VSB_data(p->cname));
+
+	if (dyn) {
+		Fc(tl, 1, "\nstatic inline void %s\n", cc_adv);
+		Fc(tl, 1, "%s_checked(VRT_CTX)\n{\n", VSB_data(p->cname));
+	} else {
+		Fc(tl, 1, "\nvoid %sv_matchproto_(vcl_func_f)\n", cc_adv);
+		Fc(tl, 1, "%s(VRT_CTX, enum vcl_func_call_e call,\n",
+		    VSB_data(p->cname));
+		Fc(tl, 1, "    enum vcl_func_fail_e *failp)\n{\n");
+		Fc(tl, 1, "  assert(call == VSUB_STATIC);\n");
+		Fc(tl, 1, "  assert(failp == NULL);\n");
+	}
+
 	vsbm = VSB_new_auto();
 	AN(vsbm);
 	vcc_vcl_met2c(vsbm, mask);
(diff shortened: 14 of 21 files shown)


slimhazard:
> In existing code, we check if used objects and returned actions are allowed within the respective calling context.
> 
> To build the methods bitmask, we now begin with an all-one bitfield and clear the bits of methods which would not be allowed for any used object or returned action.

Interesting, we've talked about this from time to time, and I had always a pictured a VCC declaration of the SUB in which the VCC author has to declare the VCL subs within which it may be used.

But if I'm reading this right, the idea would be to infer those contexts from the contents mentioned in the SUB's definition. For example, if ``req`` is mentioned, it has to be client context; if ``resp`` is mentioned, it can only be ``vcl_deliver`` or ``vcl_synth``; and so forth. Is that right?

If so, it sounds much easier on the VCC author, and more likely to ensure correctness (the VCC declaration could have been inconsistent, or hard to get right).

nigoroll:
> the idea would be to infer those contexts from the contents mentioned in the SUB's definition. [...] Is that right?

correct

andrewwiik:
@nigoroll so I don't understand  the recursive function, it never stops, VCL has built in logic to PREVENT this from happening, why circumvent that? Talking about the VTC just to be clear.

Dridi:
Not looking at the code, I will only share my opinion in general at this point.

First, I think a rename task force is in order in this area. In the `VCL_MET_*` bits, `MET` stands for `method` and a good chunk of libvcc refers to VCL subroutines as methods. This is probably a historical thing that we carried over because we never really made significant changes in this area to warrant such a cleanup, but between this and #3158 I think it's time to tidy things up.

Next, as much as I like the idea I wouldn't want to expose `struct vcl_sub` (and because I feel really nitpicky I would call it `struct vrt_sub`). So it should start opaque (and it doesn't prevent from exposing it afterwards) and be backed by a function:

```c
// vrt.h
struct vrt_sub;
int VRT_call(VRT_CTX, struct vrt_sub *);

// cache_vrt.c
int
VRT_call(VRT_CTX, struct vrt_sub *s)
{

    CHECK_OBJ_NOTNULL(ctx, ...);
    CHECK_OBJ_NOTNULL(s, ...);
    AN(sub->methods);
    if (!(sub->methods & ctx->method))
        return (-1);
    AN(sub->func);
    sub->func(ctx);
    return (0);
}
```

Then the question is, should `VRT_call` fail the transaction? Details for later. But I don't think it's wise to engage with VMOD-made VCL subs right off the bat. Even if we do, we can also keep the struct opaque, expose the function signature and provide a function to build a sub from a function pointer and a bit mask.

Then we run into the bit mask inference problem. If my sub only touches `bereq`, is its scope "backend" or `vcl_pipe`? Same for `obj`, is it meant for `vcl_hit`, `vcl_deliver` or both?

I think that in the spirit of VIP4 we should rather make this opt-in. If you want your sub to be callable from a VMOD, then you need a different syntax and a different type (let's call it `CALLABLE` for the sake of the argument). Then once we have a callable sub, this symbol might be used by a VMOD and for the `call` action libvcc uses it transparently (and ideally with the same mechanic for normal and callable subs).

We can imagine a syntax like:

```vcl
callable $scope (, $scope)*
sub $name {
    ...
}
```

It forces you to declare the intended scope, but it allows a sanity check where a given piece of code may compile differently depending on where it is called from (eg. `vcl_pipe` vs `vcl_backend_fetch`).

```vcl
callable pipe
sub something {
    set bereq.http.connection = "close";
}
```

Another example:

```vcl
callable synth, deliver
sub something-else {
    unset resp.http.x-varnish;
}
```

Let's also not forget about VIP3 (and to some extent VIP2), it could be conceivable to "import" subroutines from other libraries. I think this should really become prominent, and in particular, move built-in subroutine logic into callable subroutines. And similar to the `_type` namespace from #3158, we could call this import `_builtin`. The `_builtin` namespace would always be implicitly imported, just like the built-in VCL is always implicitly appended.

This way the built-in VCL could end up like this:

```vcl
sub vcl_recv {
    call _builtin.recv_lower_host;
    call _builtin.recv_h2_prism;
    call _builtin.recv_rfc_methods;
    call _builtin.recv_get_methods;
    call _builtin.recv_authorization;
    call _builtin.recv_cookie;
    return (hash);
}
```

And finally we could put the built-in logic in the middle of `vcl_recv` without resorting to copy-pasta, cherry-picking only the parts we need, or by default, falling back to all of them if we don't return from `vcl_recv`, as usual.

Apologies for this long message, but since this is an RFC I wanted to take the time to be as comprehensive as possible, and in particular cross-reference existing VIPs to make sure we move in a direction that takes them into account. I also disagreed with the implementation details, because I think we should avoid exposing too much, and we should avoid an implicit syntax. For the inference we'd probably run into problems during the implementation anyway.

rezan:
I reallylime this idea. Basically this is a callback and this opens the door to `for each` loops where the callback acts as the body. This would be an excellent addition.

However, as always, I think we should keep it simple and to the point. What you have in this PR is really all the syntax we need.

Dridi:
> Basically this is a callback and this opens the door to for each loops where the callback acts as the body. This would be an excellent addition.

And then, anonymous subs :D

```
import foreach;

foreach.header(req.http.cookie, sub {
    # do stuff
});

**Comment by slimhazard** [comment]
+1 to @Dridi's idea of a VRT function to call the SUB, actually I implicitly assumed that the ``.call`` method in @nigoroll's VMOD debug patch was an illustration of what was meant to end up in VRT. Varnish may crash if this isn't done right, so IMO it has to be encapsulated for VMOD developers.

@andrewwiik I believe that may answer your question -- the call function for the SUB checks for recursion and fails the call if recursion is detected. I assume that a ``VRT_call()`` implementation would do that as well. Notice that the vtc verifies status 503 for that case.

The consequence is that the recursion check happens at runtime -- I'm not sure if the compile-time check for recursion, as we have now in VCC, will be possible for the SUB type.

**Comment by Dridi** [comment]
> I assume that a VRT_call() implementation would do that as well.

At this point all bets are off. If you use a vmod to do sub laundering there's no keeping track of that at compile time.

**Comment by rezan** [comment]
I don't think we need the `VCL_MET_*` bits. If you use a `sub` as a variable/callback, we can do the checks at compile time to make sure its being used in a valid context. The same thing is done today when you use `call`. Right?

**Comment by Dridi** [comment]
> The same thing is done today when you use `call`. Right?

In the context of this discussion, there's @nigoroll's vmod_dispatch. You could see it as kvstore growing support for subs: some sort of hash map where you register subs and then call them by name (for example a domain name).

So no, in the context of this RFC we need to get the bit mask somehow. @nigoroll suggested inference, and I commented on how brittle it sounds to me.

**Comment by rezan** [comment]
I would have to see an example of this.

Im almost positive the compile time checks work on sub usage. If you need something more advanced, just do the usual runtime MET check.

**Comment by Dridi** [comment]
https://code.uplex.de/uplex-varnish/libvmod-dispatch/blob/master/src/vmod_dispatch.vcc#L64-80

**Comment by Dridi** [comment]
I can have a deeper look later, I doubt we can do anything without embedding the bit mask with the function pointer. And maybe libvcc already does TheRightThing(tm), but we can't overlook VIPs 2, 3 and 4 if we turn subs into VMOD-able symbols.

**Comment by rezan** [comment]
Sorry, may have edited my comment above too late. The bitmask and `VCL_MET` are both known at compile time, so checking it at runtime doesn't make sense. Look into how `call` figures out if the `sub` is valid for the client/backend scope.

VIP2 is definitely interesting. However, functions would be their own thing, like `function`. `sub` is a different animal. It would be messy to overload `sub` with functions. This PR can be implemented now with `sub` and when functions come along, they wont break this and can be supported later.

**Comment by slimhazard** [comment]
@rezan if the SUB call is done from a VMOD, then in general you can't know at compile time where it will be called, because the VMOD invocation, presumably via something like ``VRT_call()``, doesn't necessarily make the SUB apparent from the syntax. The SUB that gets called may not be known at all until runtime.

This is what we have happening with VMOD dispatch. The VMOD sets up an indexed table of subs (internally it's just an array), and the SUB to be called is invoked via the index: ``myobj.call(4711)``. We invariably use that together with the VMODs re2 or selector, which are used to return the index:
 ```vcl
myobj.call(myset.which());
```
The subroutines are assigned to the dispatch object in ``vcl_init``. So there's just no way to tell from the syntax where the subroutine is invoked by the VMOD. This is where the VMOD is risky (aside from the black magic of getting the C function that corresponds to the VCL subroutine, which requires knowledge of how VCC works) -- *we* have to make sure that the subroutine is called in such a way as to be legal. Get it wrong and Varnish crashes. VRT support would put the VMOD on safer ground.

VMOD dispatch is the example here, but generally, if a VMOD can encapsulate a SUB invocation, then there's no assurance that the context of the invocation can be inferred the way VCC does it. To make it safe (and not live with the risks we have with VMOD dispatch currently), there's no way around the runtime check.

nigoroll:
Thank you for the good feedback everyone. I will try to summarize a bit and answer some aspects:

* compile time / runtime failures

As noted by others, the compile time checks VCC can do on scope and recursion are not possible with sub references, so if we want this flexibility, we need to accept more possible runtime errors. But at the same time I think we should not lower the standards for the compile time checks.

* Scope inference

At this point, I am not convinced by @Dridi's argument and think that we can and should be able to keep the Scope inference for subs in place. For example, with `$Scope` declared vmod functions/methods, we can just extend the existing code, which checks possible scopes for built in variables (`beresp` and the like). We really should spare the VCL author the responsibility to define scopes and just make possible what is possible.

On one detail, @Dridi:

> If my sub only touches bereq, is its scope "backend" or vcl_pipe

Please look at the code and maybe re-read my explanation in the initial note. It is both. The `okmask` defines the allowed scopes, and finally gets copied into the `methods` member of the `struct vcl_sub`.

Also I think it does not matter if some object behaves differently in different Scopes, we only need to check if it is accessible.

* compatibility with VIPs

While I appreciate the pointers and agree that we need to keep them in mind, I do think that this suggestion is in line.

* `VRT_call()`

@slimhazard understood me right that `debug.call()` was just a poc demo for how some VRT interface could look like. And yes, the struct would be opaque to vmods.

* @Dridi's `_builtin`

great idea, and no problem hijacking this RFC, but still unrelated :)

Dridi:
> On one detail, @Dridi:
>
> > If my sub only touches bereq, is its scope "backend" or vcl_pipe
>
> Please look at the code and maybe re-read my explanation in the initial note. It is both. The > `okmask` defines the allowed scopes, and finally gets copied into the `methods` member of the `struct vcl_sub`.

I looked a bit and it should have been obvious to me that VRT would be in charge of dealing with scope differences, so inference should work.

Currently of course, it doesn't, because the recursive check is lazy and based on actual calls.

This test passes as a result:

```
varnishtest "current scope inference"

varnish v1 -cliok "param.set vcc_err_unref off"
varnish v1 -vcl {
        backend be none;
        sub unref {
                set req.http.impossible = beresp.reason;
        }
}
```

I'm curious, does your PoC catch that?

So we need some infrastructure to allow scope discovery, and it should work the way you describe your implementation, and as a result VIP4 is no longer on the table.

As I was looking in this area I introduced an unfortunate conflict: https://github.com/varnishcache/varnish-cache/commit/db0baa744db9b71f5f14fba4b2e3b0c2e9d5a807, but thank you for leading me in this direction :)

I will look at the code at some point.

slimhazard:
@Dridi's example brings up something important -- the mention of a SUB should count as a reference, so that you don't have to turn off ``vcc_err_unref``.

With VMOD dispatch we either have to do that, or do something really stupid like:

```vcl
sub vcl_recv {
    if (false) {
        call sub_that_only_gets_invoked_via_vmod;
    }
}
```

Which I usually do, because I'd rather have ``vcc_err_unref=on``. (Hoping that gcc optimizes that code away, but I've never actually checked.)

nigoroll:
@Dridi I added your example. When you look at the generated C ...
```
vcl_func_f VGC_function_impossible;
const struct vcl_sub sub_VGC_function_impossible[1] = {{
        .methods        = 0x0,
        .name           = "impossible",
        .func           = VGC_function_impossible
}};
```
you will notice that no meth^H^H^H^Hscopes may call this sub.

@slimhazard a SUB in my poc is a reference, so you do not need that kind of hack

nigoroll:
@Dridi also
> So we need some infrastructure to allow scope discovery, and it should work the way you describe your implementation, and as a result VIP4 is no longer on the table.

both: the infrastructure already exists yet we do need VIP4 - but not for this.

And yes, you would actually spare some of the discussion if you looked at the code :)

nigoroll:
btw: we could actually handle the _uncallable_ special case and error out at compile time.

hermunn:
After just skimming the code I could not help sharing my impression of the discussion and ideas in this ticket.

It seems to me that it is a good idea to think of the VCL subs as *lambdas* which implicitly bind (by reference) to various VCL objects, and that the compiler should see this automatically and produce compile time errors in as many cases as possible. The VMOD author should not have to do too much in terms of checking what the current sub is.

I will switch out the word *sub* for *lambda* for the subs users write, since I will now describe what I feel is a natural feature set for what this PR is a bases for. The word *sub* is the reserved for the `sub vcl_` type which return the next state in the state machine.

When the compiler encounters a lambda definition in in the VCL, it will accumulate a set of flags for the different types of variables the lambda uses. For example, if `req.http` is used as an *rvalue* (right side of the `=`), then the compiler will remember that, and if it is used as a *lvalue*, then that will be remembered. If the lambda calls a (previously defined) lambda, then it will acquire all the flags for that lambda. Ideally, VMOD functions should also provide such information (which variables does the function need to work) in their `.vcc` files along with additional flags on the form "not callable from vcl_synth".

If you try to pass a lambda to a VMOD expecting, then it can only be done if you are in a sub where the variables are properly bound. So if the lambda a uses a `req` variable, then this lambda can only be called and/or passed to a VMOD if the `sub vcl_` has access to the variable (read or write). An obvious extension of this is to allow bans which uses lambdas, and that such lambdas will be lurker friendly only if they stay away from `req.` (ie. they only bind to `obj.`), and this will be automatic from the implementation of the lambda.

The lambdas should be allowed to return values, so that you can make a "foreach header with short circut or" VMOD.

rezan:
Hmm... I'm still not understanding why the scope/context can't be checked at compile time. It works for `call`.

```
sub my_cb {
    set resp.http.foo = "bar";
}

sub vcl_recv {
    header.foreach(my_cb);
}
```

In the above example, `my_cb` is called in `vcl_recv`. Vcc checks the sub and then generates an error because `resp` is not available in `vcl_recv`. Basically, vcc guarantees the `sub` will always be valid when passed into a vmod. No need to check this at runtime. What am I missing?

slimhazard:
@rezan you're assuming that the SUB object is always named in the VCL source at the place where it's invoked, in your case ``my_cb`` mentioned explicitly in ``vcl_recv``. But the signature for a VMOD method or function that invokes ``VRT_call()`` doesn't have to require that.

```vcl
sub vcl_recv {
    # Note: no SUB named here
    myobj.invoke_some_sub_internally_with_vrt_call();
}
```
Further up, @Dridi posted a link to some lines in the VMOD dispatch docs that illustrate such a case. My own account of VMOD dispatch a few comments also explained that, I'm not sure what else I can add.

Dridi:
> btw: we could actually handle the uncallable special case and error out at compile time.

Agreed, this is what I had in mind with my test case.

> The lambdas should be allowed to return values

I disagree if the lambda here is a sub. We have VIP2 for typed functions that may return a value, so assuming the syntax from VIP2:

```
def STRING
foobar(INTEGER a1, BACKEND a2, STRING_LIST a3) {
        ...
}
```

Then yes, you could imagine a lambda syntax, although that could become ugly. For the sub case I would really call them anonymous subroutines.

@hermunn the problem with your proposal and this RFC is that @nigoroll wants the ability to stash a subroutine and call it later at some point. If your lambda is able to capture its surrounding environment then you definitely don't want to pass it to VMOD space.

rezan:
My last comment on this.

If you want to call a vmod function from vmod code, aka call a C function from C code, then you don't need this to accomplish that.

Regarding vmod dispatch, I looked at it a bit more deeply. Its actually unsafe and furthers why we need vcc checking here. Vmod dispatch is aliasing function pointers behind strings and then handing those strings to the user. This bypasses the vcc checks for call safety. This is true for any kind of aliasing in C, you bypass the standard safety checks. Just use the subs normally and now you have compile time safety. This is why aliasing needs to be used carefully in C and I guess now VCL. So using this (anti) pattern as a basis of a new feature isnt the best way forward.

Dridi:
@nigoroll believe it or not, I didn't have time to review your code, but I think I finally managed to wrap my head around the existing one. The result was #3167, which hopefully would play well with this PoC, though they would certainly conflict.

nigoroll:
@hermunn what happens in existing code and what I am suggesting to make use of is similar to what you describe, but not exactly the same:

For the existing use cases, the scope (e.g. `vcl_recv`) is always known. VCC (recursively) goes through all _built in subs_ and registered calls of any _custom sub_ (lambda should really denote something else) and checks if all the operations are valid (in that read/write of the respective object is allowed).

I think that for expansions like vmod `$Scope`, we can re-use this mechanism.

The addition in this PR is to, as the above mentioned checks are conducted, also record the maximum possible set of scopes valid for each _sub_.

So I think that both the existing implementation as well as this suggested additional use basically match your ideas, just that the details differ slightly. For example, it is not the `req` access that is being checked directly, but rather the fact that `req` is available from the `vcl_recv` scope.

slimhazard:
The details will undergo the usual scrutiny, but for the feature in general I am all kinds of +1. +2 if I could.

@nigoroll maybe I'm being overeager and premature, but perhaps this PR could go ahead and implement the VRT part, and then VMOD debug could demonstrate its use. (Well, I suppose a round of bugwash would be constructive, so I'll try my best to calm down and be patient.)

nigoroll:
@slimhazard yes, I will do this, but would like to hear @bsdphk's opinion first in order to make good use of my time.

nigoroll:
I have updated the PR based on the previous feedback and moved the prototype to VRT as VRT_call().
The recursion check is not hardened against rollbacks, see commit message of the "Add VRT_call() ..." commit for details.

nigoroll:
FTR: has been discussed between @Dridi / @bsdphk / @nigoroll :

@bsdphk wants to explore alternative ways of implementing a vcl sub type which might allow for more compile time checks.

nigoroll:
During yesterday's pow-wow, @bsdphk asked what would happen if a vmod kept a `VC_SUB` across VCLs and called it from the context of a different VCL. Sure the answer would be *bad things*, so I have now added a check guarding against this (by panicking varnish and thus pointing to the wrong behavior).

bin/varnishtest/tests/bad_call.vtc
```diff
@@ -0,0 +1,34 @@
...
+	import debug;
+
+	backend dummy None;
+
+	sub foo {
+		set resp.http.it = "works";
+	}
+
+	sub vcl_init {
+		debug.bad_memory(foo);
```
rezan: Shouldn't this be a compile error? You are invoking foo which touches resp in vcl_init?

nigoroll: no, because it is not a `call`.
If that was an error, sub references were impossible.

rezan:
So I am -1 on this just because this feature moves too far into unsafe function aliasing. Basically, if I pass a function to a VMOD, the common use case is a callback. In this scenario, we want compile time checks to make sure our sub is valid and legal. As designed, this PR leaks away those checks and replaces it with a generic VRT_fail(). 

I think skipping the sub checks should be behind a flag, like: `vcc_err_subcheck=false`. This way, we can maintain compile time safety for callbacks, and if you want to do sub aliasing, flip the flag and off you go.

Thoughts?

slimhazard:
Compile-time checks might be possible with a syntax that indicates where you intend to call the SUB. Something like:
```
sub vcl_init {
    myvmod.addsub(sub=mysub{recv, pass, hit});
}
```
And then ``VRT_call()`` will permit invocation of ``mysub`` in ``recv``, ``pass`` and ``hit``. There would still have to be ``VRT_fail()`` in case the sub is not in fact invoked where it's permitted.

nigoroll:
Re @rezan 

> we want compile time checks to make sure our sub is valid and legal.

This is the case to the extent possible: Each sub has a mask attached to it with legal call contexts.

> As designed, this PR leaks away those checks and replaces it with a generic VRT_fail().

Two comments:

* no existing compile-time checks are removed or replaced
* for `VRT_call()`, a `VRT_fail()` is used where VCC would otherwise fail at compile time

> I think skipping the sub checks should be behind a flag, like: vcc_err_subcheck=false.

That parameter name suggests to me that you might have misunderstood the above, that no existing checks are reduced.


Re @slimhazard 

> ```
> sub vcl_init {
>     myvmod.addsub(sub=mysub{recv, pass, hit});
> }
> ```

We could certainly have something like this to fail earlier. Yet this would only be declarative, it would just be an additional check: We could not prevent a vmod from making the call from a wrong context and the `VRT_fail` would still happen.

rezan:
I think this example really illustrates why removing compile time checking puts VCL writers in a bad place.

Lets say I have a basic JSON iterator, the `.vcc` definition:

```
$Function VOID for_each(SUB callback)
```

Now lets say I define a callback:

``` vcl
sub json_print {
    std.log(req.http.xj-name + ": " + req.http.xj-value);
}
```

And now I use this:

```
import json

sub vcl_backend_response {
    json.for_each(json_print);
}
```

VCL compiles. When tested, you get 503s and an error message which says `not allowed here`. If we did standard compile time checking, we would have gotten this error:

```
Message from VCC-compiler:
('/tmp/vc/test.vcl' Line 12 Pos 13) -- (Pos 15)
    std.log(req.http.xj-name + ": " + req.http.xj-value);
------------###------------------------------------------

Not available from subroutine 'vcl_backend_response'.

...in subroutine "json_print"
('/tmp/vc/test.vcl' Line 10 Pos 5)
sub json_print
----##########
```

My point here is that callbacks and iterators require the use of headers to pass information back and forth. By stripping away compile time header checks in the subs, we introduce all kinds of new surface area for hard to debug errors. That's why I feel this PR, as designed, is dangerous for VCL writers. The callback and iterator usecase is strong, so we need to make sure we design this feature accordingly. Function aliasing should not be a 1st class VCL feature in my opinion.

nigoroll:
@rezan I suggest we consider the question of how to pass arguments out of scope. Besides headers, you could also use task scoped variables or we might add functions (with arguments) to VCL in the future.

You have now repeated a second time your point that you prefer compile time checks. We do agree on this, I do prefer compile time checks too. I would prefer if we could discuss a single argument just once and I can not see any new insights in your latest comment.

I still _oppose_ your notion of "stripping away" compile time checks. This PR does *not strip away anything*. In particular, when `VCL_SUB`s are not used, there is no difference, and use of this feature is of course up to VMOD/VCL authors.

We simply cannot provide compile time checks for uses of sub references which we do not control. If we want them to happen in VMODs, we simply can not provide compile time checks because we do not know about the vmod's semantics. *).

I cannot follow your argument on "hard to debug errors". We introduced `VRT_fail()` a long time ago to handle run time errors and checking for them should be considered good practice. In particular, I cannot follow the argument that the error messages `recursive call to ...`, `not allowed here` and `no rollback during dynamic sub calls` would leave any doubt.

You say "we need to make sure we design this feature accordingly" and I agree, this is what I intend to do. Can you please share your concrete proposals?

----

*) In principle, we could add VCC-time callbacks for vmods which could allow syntactical or even limited semantic checks. If we had these, vmods could generate compile time errors for wrong uses. _But_ even in this case we would still need to `VRT_fail()` for the case that a vmod failed to properly implement the compile time check.

rezan:
> Can you please share your concrete proposals?

Of course. I would recommend 2 things:

* Have compile time checking of SUB variables be on by default. Add a flag, like `vcc_err_subcheck` which moves the check from compile time to runtime VRT_fail().
* Beef up the runtime error message. Right now it just says `not allowed here`. It should somewhat match the compile time error which says you are referencing X in an incorrect context.

nigoroll:
Thank you @rezan, this is the kind of feedback I can work with

> * Have compile time checking of SUB variables be on by default. Add a flag, like `vcc_err_subcheck` which moves the check from compile time to runtime VRT_fail().

To rephrase your suggestion, you are saying that only the _callback_ use case should be allowed by default, as in your `json.for_each(json_print);`, correct? In this case, you would demand that the vmod call `json_print` from exactly that context only.

While we could treat this just like a `call` statement, we have no means of preventing a vmod from keeping a reference and make the call from a different context.

So while this kind of check would prevent other use cases but the _callback_ construct, it would not avoid any of the other checks I have implemented.

So, my answer here has two parts:

* We certainly could enable the extended use case only by a parameter. I would suggest `vcc_allow_dynamic_sub`.
* But that would not make the VCL any safer

> * Beef up the runtime error message. Right now it just says `not allowed here`. It should somewhat match the compile time error which says you are referencing X in an incorrect context.

I will have a look how the error message can be improved.

nigoroll:
@rezan with the last commit, examples for the error messages from `call.vtc` are:
* `dynamic call to "sub impossible{}" not allowed from this context`
* `recursive dynamic call to "sub recursive{}"`

Including the sub name in the `no rollback during dynamic sub calls` message would require allocating workspace, which I would prefer to avoid.

rezan:
> you are saying that only the callback use case should be allowed by default

I don't see how we can have both compile time checks on callbacks and to allow aliasing at the same time. Unless we make a second type, like SUB_ALIAS, and that type has no compile time checks? I have callback and iterator usecases for at least 5 VMODs, so for me, this is the priority usecase.

> we have no means of preventing a vmod from keeping a reference and make the call from a different context

That's fine and keeping the runtime check around is perfect valid.

slimhazard:
Production experience with something very similar to this suggests to me that the extra parameter and a limitation of the permitted usage are overkill. And I think @rezan's example actually undermines the argument.

@rezan, note that the failure in your example is guaranteed, on the first and every test -- a prominent, fail-fast error with an explanation in the log. In real practice, VCL authors are only at genuine risk if they deploy VCL as production code without any testing at all. If they're doing that, then this is likely not they're only problem.

The mechanism we have been using, for years now, doesn't even have the runtime check -- Varnish will crash if you get it wrong. It would certainly be better with the runtime check, which is one of the reasons why I am +1 for the PR. But I for one have never experienced the error actually happening.

I have use cases for the runtime SUB call, which have amply demonstrated their usefulness and effectiveness in real projects. In contrast, the iterator-like use case is not a priority for me.

I just don't agree that deferring the checks to runtime is some kind of unpardonable sin, a rubicon that must never be crossed. And this is more than just a gut feeling, it's confirmed by extensive project experience.

slimhazard:
@nigoroll I think that the bit of workspace, rarely more than a dozen characters or so for the sub name, is an acceptable price to pay for an informative error message.

nigoroll:
Re @rezan 

Your response makes be wonder if you might have missed the most relevant response to that bit of the discussion:

> But that would not make the VCL any safer

Whether you add a compile time check or not, whether you enable the aliasing through a parameter or whether you add a second type, one thing does not change: We can not in any way prevent a VMOD to keep a reference and make the call some place else.

That said, if there was consensus to have the "treat VCL_SUB argument like a `call`" semantics, combined with a parameter, I will be happy to add that.

Re @slimhazard 

We would require at least two pointers worth of workspace for each `VRT_call()`, you are probably right that hundreds or thousands of `VRT_call()`s are unlikely.

I'll write the code.

rezan:
I do firmly believe that compile time errors are safer than runtime errors. Runtime errors lead to downtime, which is basically a system failure.

nigoroll:
From an IRC discussion with @rezan , I hope to have finally understood his point. We now agree on the following:
* If a `VCL_SUB` is passed in `vcl_init{}`, it will have _alias_ semantics, the reference will be passed without additional checks.
* If a `VCL_SUB` is passed elsewhere, it will have _callback_ semantics and be compile-time checked as if it was `call`ed at that place.

In particular, if a `VCL_SUB` is `VRT_call()`ed in `vcl_init{}`, any _runtime_ `VRT_fail()` de-facto turns into a load error, so the net result is similar to a failing compile time check.

slimhazard:
I'm ok with the two bullet-pointed suggestions. But I'm not sure I understand whether this next part changes that or not:

> In particular, if a VCL_SUB is VRT_call()ed in vcl_init{}, any runtime VRT_fail() de-facto turns into a load error, so the net result is similar to a failing compile time check.

Makes sense. But what I have in mind is that a SUB object is passed to a VMOD in ``vcl_init``, and then ``VRT_call()``ed elsewhere. It seems to me that it would then be subject to the runtime check. Have I understood correctly?

nigoroll:
@slimhazard you have understood correctly.

nigoroll:
@slimhazard An example for the rollback error message is now `Rollback during dynamic call to "sub rollback{}"`

I have also slightly polished the other error messages respecting @Dridi 's rightful criticism of zejerman's angst of Capitalized Letters.

slimhazard:
Where does rollback fit into this? We've talked about a compile-time error, load-time error, or ``VRT_fail()``. Is there a rollback after VCL failure? If so, why especially for this? Since AFAIK we don't invoke rollback generally for ``VRT_fail()``.

slimhazard:
I think it's worth saying that this seems to be a pretty good solution, at least as I understand the two bullet points above. I wouldn't have thought of handling the two cases differently, and it makes a lot of sense. This may have been a strenuous discussion (at times it felt that way to me, anyway). But now I think it's a good thing we did it.

include/vrt.h
```diff
@@ -61,6 +61,7 @@
  *	VRT_l_beresp_body() changed
  *	VRT_Format_Proxy() added	// transitional interface
  *	VRT_AllocStrandsWS() added
+ *	VRT_call() added
```
slimhazard: This will need to be moved up to after 2020-03-16.

nigoroll:
First of all, a quick update: Working on the compile time checks, I noticed that we missed something like this:
```
debug.call(somevmod.returning_a_sub());
```
That is to say, an argument does not need to be constant but rather could be the return value from another vmod.

Right now I tent towards simply failing for this case and _require_ any `SUB` argument to be constant at compile time. As I have not written the code yet, now is the time to comment on this if you think it would be an issue.

Re @slimhazard 

A rollback would destroy the `PRIV_TASK` state we need for recursion detection, so it is forbidden for the duration of a `VRT_call()`. See the explanation in the commit message of https://github.com/varnishcache/varnish-cache/pull/3163/commits/ab29419a8881b74ce621456cf9538b2968673e7e

include/vrt.h
nigoroll: yes, I stopped basing this branch for now in order to help reviewers. Substantial squashing will probably happen before any final merge, and I will keep this conversation open as a reminder

nigoroll:
FTR: Besides limiting sub references to immediate (compile time) sub references, I also need to adjust the patch to ed36b6389f9dbf6962e726f0df34d59132d9241a which is something for another day

nigoroll:
Good idea by @bsdphk  during pow-wow:
The recursion check could be made more efficient and simplified by:
* VCC assigning an integer to all subs if VCL_SUBs are present
* VRT init'ing a bitmask
Then VRT_call can check/set that bitmask for recursion detection

nigoroll:
the last commit of the last force-push contains an implementation of the idea which was mentioned here: https://github.com/varnishcache/varnish-cache/pull/3163#issuecomment-612022390 I have kept it as a patch on the previous implementation to hopefully make reviews easier, but a final merge would require some squashing.

I think the code of this PR is ready now for merge and would appreciate reviews.

----

I also pondered more the idea discussed with @rezan noted here: https://github.com/varnishcache/varnish-cache/pull/3163#issuecomment-602728910

>     * If a `VCL_SUB` is passed elsewhere, it will have _callback_ semantics and be compile-time checked as if it was `call`ed at that place.

The problem here is that this makes sense only for immediate references like `myvmod.function(subname)`. If a vmod was to return a sub reference like `myvmod.function(another.returning_sub())`, we could not compile-time check the VCL sub.

One might want to call for still having the former check, but that would complicate VCC quite a bit, only to add a compile time check for a very specific case. Also, we would prevent storing sub references in variables. Consider this example

```
sub foo {
  set resp.http.is = "foo";
}

sub bar {
  set resp.http.is = "var";
}

sub vcl_init {
  new subvar = taskvar.sub();
}

sub vcl_recv {
  if (req.http.foo) {
    subvar.set(foo);
  } else {
    subvar.set(bar);
  }
}

sub vcl_deliver {
  mystuff.calling(subvar.get());
}
```

So, after having gone full circle with your suggestion, I realized that, when we discussed it, we did only consider the case of an immediate sub value and that implementing that specific case would render other sensible use cases impossible.

To summarize, I do not think that we can have any compile time call context checking for sub references. With this PR, you get run time call context checking, but that is, I think, something you need to accept if you want to use sub references.

lib/libvmod_vtc/vmod_vtc.vcc
```diff
@@ -169,6 +169,20 @@ version ``v1``).
 Candidate for moving into vmod_proxy, but there were concerns about
 the interface design
 
+$Function DURATION benchmark(SUB, INT total=1000, INT leg=10)
+
+Benchmark a VCL sub and return the average runtime.
+
```
bsdphk, line 175: Is this generally useful enough that it ought to live in `vmod_std`, or is it too dangerous for that ?

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -444,6 +449,15 @@ vcl_call_method(struct worker *wrk, struct req *req, struct busyobj *bo,
 	assert(ctx.now != 0);
 	ctx.specific = specific;
 	ctx.method = method;
+	if (track_call > 0) {
```
bsdphk, line 452: When can this ever be zero ?

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -472,8 +488,40 @@ VCL_##func##_method(struct vcl *vcl, struct worker *wrk,		\
...
+	struct vbitmap *vbm;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(sub, VCL_SUB_MAGIC);
+	CHECK_OBJ_NOTNULL(ctx->vcl, VCL_MAGIC);
+	assert(sub->vcl_conf == ctx->vcl->conf);
+
+	if ((sub->methods & ctx->method) == 0) {
+		VRT_fail(ctx, "Dynamic call to \"sub %s{}\" not allowed "
+		     "from this context", sub->name);
```
bsdphk, line 512: I think "context" is not very helpful here.

We can stash a const char * with the name of the method we started from in VRT_CTX, so that me message could say:
    ... starting from vcl_recv{}
instead of "from this context".

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -472,8 +488,40 @@ VCL_##func##_method(struct vcl *vcl, struct worker *wrk,		\
...
+	vbm = ctx->called;
+	AN(vbm);
+
+	if (vbit_test(vbm, sub->n)) {
+		VRT_fail(ctx, "Recursive dynamic call to \"sub %s{}\"",
+		    sub->name);
+		return;
+	}
+	vbit_set(vbm, sub->n);
+	sub->func(ctx);
```
bsdphk, line 525: AN(sub->func) please.

lib/libvcc/generate.py
```diff
@@ -654,6 +653,7 @@ def tbl40(a, b):
 	const struct vpi_ref	*ref;
 
 	int			nsrc;
+	unsigned		nsub;
 	const char		**srcname;
```
bsdphk, line 657: If VCC also produced a `const char *subname[nsub]`, could that improve any error messages ?

lib/libvmod_vtc/vmod_vtc.vcc
nigoroll: I would think benchmarking was a test function and nothing for production, thus the `vtc` vmod

bin/varnishd/cache/cache_vrt_vcl.c
nigoroll: it gets set to zero if there are no references to subs produced
https://github.com/varnishcache/varnish-cache/pull/3163/files#diff-c974b53a6559894584abbf06417cc5e4R507

bin/varnishd/cache/cache_vrt_vcl.c
nigoroll: we already have the method in `(struct vrt_ctx).method`. As we need the name only for the error message, we could have a generated function which returns the name for the method bitfield, asserting that only one bit is set.
Yet the error message is not emit in isolation, VSL already contains info on the called builtin sub.
Do you still think that we should have the method again in the fail message?

bin/varnishd/cache/cache_vrt_vcl.c
nigoroll: +1 thx

lib/libvcc/generate.py
nigoroll: see above.
having the name of all the (custom) subs available would also require to modify the VRT_CTX in the preamble of the generated C functions. Or we would need to function-wrap the C code for VCL `call`.
In my mind, this is overkill.

bin/varnishd/cache/cache_vrt_vcl.c
nigoroll: maybe rephrase as _now allowed from here_?

lib/libvcc/generate.py
nigoroll: and I forgot to mention that `VRT_CTX` is const.

nigoroll:
FYI: Up to now, this PR's branch still showed the evolution in the course of the discussion. I have copied that to https://github.com/nigoroll/varnish-cache/tree/rfc_call_20201016 and will condense this PR.

nigoroll:
Boiled down to three commits with reworked messages. I hope this will make the next round of reviews easier for reviewers.

Dridi (changes requested):
Thank you for reducing the changes to three logical commits.

I felt a bit uncomfortable initially that we would treat static and dynamic calls differently, and was wondering whether static calls could be generated somewhat like this:

```c
AZ(VRT_call(...));
```

The question I was asking myself was whether the return value would be tainted by a faulty dynamic call, and whether it would be possible to avoid the conditional case of sub tracking and always do it instead. That's when it occurred to me that in its current state a static call would not track the sub and allow it to recursively call itself at most once:

```diff
--- i/bin/varnishtest/tests/m00053.vtc
+++ w/bin/varnishtest/tests/m00053.vtc
@@ -21,6 +21,12 @@ varnish v1 -vcl {
 	sub recursive {
 		debug.call(recursive);
 	}
+	sub recursive_indirect {
+		if (!req.http.recall) {
+			set req.http.recall = "true";
+			debug.call(recursive_indirect);
+		}
+	}
 	sub rollback {
 		std.rollback(req);
 	}
@@ -35,6 +41,9 @@ varnish v1 -vcl {
 		if (req.url == "/recursive") {
 			debug.call(recursive);
 		}
+		if (req.url == "recursive_indirect") {
+			call recursive_indirect;
+		}
 		return (synth(200));
 	}
 	sub vcl_synth {
@@ -84,17 +93,22 @@ client c1 {
 	rxresp
 	expect resp.status == 503
 } -run
-client c1 {
+client c2 {
 	txreq -url "/recursive"
 	rxresp
 	expect resp.status == 503
 } -run
-client c1 {
+client c3 {
+	txreq -url "/recursive_indirect"
+	rxresp
+	expect resp.status == 503
+} -run
+client c4 {
 	txreq -url "/impossible"
 	rxresp
 	expect resp.status == 503
 } -run
-client c1 {
+client c5 {
 	txreq -url "/rollback"
 	rxresp
 	expect resp.status == 200
```

You may note that I guarded the recursive call to only make it once, we can remove that ceremony, no strong feelings. It was convenient for me to verify single vs multiple recursion in this case by simply commenting one line out. It doesn't have to remain in the final test case.

You may note that I renamed subsequent clients to avoid reusing `c1`. I find this much more convenient to inspect test logs.

I know I have my share of responsibility in the impossible sub handling, but if the sub ends up with zero bit mask I think it shouldn't even pass the compilation step, regardless of static vs dynamic calls.

lib/libvmod_vtc/vmod_vtc.vcc
```diff
@@ -169,6 +169,20 @@ version ``v1``).
 Candidate for moving into vmod_proxy, but there were concerns about
 the interface design
 
+$Function DURATION benchmark(SUB, INT total=1000, INT leg=10)
```
slimhazard, line 172: Just a naming comment, but IMO "leg" is a strange name here. Initially I thought it might be an obscure abbreviation for something like "legacy" or "legal", or maybe an equally obscure allusion to arms and legs. The docs didn't help me much in understanding (more about that further down), I had to read the code to get the idea.

Admittedly it's hard to think of compelling names for the parameters. What's happening here is inner iterations without rollbacks and outer iterations with rollbacks. But `inner_iters_without_rollback` and `outer_iters_with_rollback` are, well ...

lib/libvmod_vtc/vmod_vtc.c
```diff
@@ -372,3 +372,38 @@ vmod_proxy_header(VRT_CTX, VCL_ENUM venum, VCL_IP client, VCL_IP server,
...
+	vtim_mono t_tot = 0;
+	vtim_mono b, e;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+
+	where = ctx->http_req ? ctx->http_req : ctx->http_bereq;
+
+	while (count < total) {
+		if (WS_Overflowed(ctx->ws))
+			break;
```
slimhazard, line 394: Shouldn't the method fail here, or at least somehow make it recognizable that not all of the requested iterations were executed? Suppose that workspace overflows after very few iterations, even only one, or none at all. The the benchmark result does not have the benefit of averaging, and could be very non-representative. How could you know that happened?

Maybe write a message to the log here?
Maybe a summary for the log, including the number of iterations, in the successful case (no overflows)?

lib/libvmod_vtc/vmod_vtc.vcc
```diff
@@ -169,6 +169,20 @@ version ``v1``).
...
+Initially and after every *leg* invocations, the task is rolled back
+in order to avoid workspace overflows, unless the workspace is already
+overflowed, in which case the benchmark is aborted. This implies that
+**the sub to be benchmarked can not use any VCL-generated data or
+PRIV_TASK state**.
+
+The run time is averaged over *total* runs. The actual total may be up
+to *leg* - 1 more, in other words, there are always *leg* calls
+executed before *total* is checked.
+
```
slimhazard, line 185: Pretty sure I wouldn't have understood this without reading the code. The point here is that rollbacks are probably necessary if you repeatedly call a VCL sub, which I probably wouldn't have thought about at first, but it's easy to see once it's pointed out. I suggest using the "inner" vs "outer" terminology to explain the difference -- inner iterations without rollbacks, and a rollback for each outer iteration.

The second paragraph threw me off, first because I initially thought it would be *leg* * *total* results (maybe say something like "each inner iteration counts towards the *total*"). And it seems strange that could ever be less than *total* runs, until you realize that the benchmark might have been aborted due to WS overflow, except you currently can't tell whether that happened (see the comment above).

slimhazard (commented):
Very happy about `VRT_call()`. I have some nitpicking comments about naming and docs for `vtc.benchmark()`, but none of that is a showstopper, and it's not the focus of the PR.

@nigoroll do IUC that it would be straightforward to add a `BOOL VRT_check_call(ctx, sub)`, to check the bitmask of a SUB for whether it's legal to call in the current context? That would give a VRT client a way to avoid VCL failure on those grounds.

I assume that a similar check for the recursion case is not realistic.

`VRT_call()` is LGTM and I can't wait to use it, the rest are considerations on the side. I'd like to have `VRT_check_call()`, but that could come after the PR.

Dridi:
Regarding my last comment, the "at most once" recursion when mixing a static call with a dynamic call is done on purpose. The goal being to only prevent infinite recursion and avoid the dynamic overhead for static calls. The `c3` client from my previous diff should reflect that by expecting a 200 response to avoid unexpected change of behavior down the line.

nigoroll:
Re @Dridi 

> That's when it occurred to me that in its current state a static call would not track the sub and allow it to recursively call itself at most once.

Right. I wanted to avoid the tracking overhead for the purely static case, to the benefit of users who do not need dynamic calls.  That said, I have not gathered any performance data on how relevant the overhead actually is and we might decide to unify this later.

As discussed on IRC, it was a conscious decision to only check recursions for dynamic calls. From the commit message of 778d37719d1704303cd4aa7d613439f1fa34773d:

> We only conduct this check for dynamic calls through VRT_call(), avoiding additional overhead to static calls, because VCC already guarantees that static calls are loop free - so loops can only be introduced through VRT_call().

I have used your VTC suggestion without the "call only once" semantics. Detecting the second recursive call still prevents an infinite recursion. I see a point in wanting to detect mixed static / dynamic call recursions the first time, but that would come at a cost, which I would not want to impose when no dynamic calls are involved at all.

For the VTC, I have also changed the clients to run in parallel, which, I think, is a good idea to keep test runtimes low in general (plus we get a tiny bit more exposure to concurrency in the test cases).

> I know I have my share of responsibility in the impossible sub handling, but if the sub ends up with zero bit mask I think it shouldn't even pass the compilation step, regardless of static vs dynamic calls.

I have now added a compile time check for impossible subs. Thank you for pointing this out.

updated the pr (force push)

lib/libvmod_vtc/vmod_vtc.vcc
nigoroll: I had asked [my advisor the lion](https://dict.leo.org/german-english/Teilstrecke), if anyone has a better suggestion, please let me know. But two from the previous comment, are, well ... :)

lib/libvmod_vtc/vmod_vtc.c
nigoroll: true, thank you. i have added a `VRT_fail()`

lib/libvmod_vtc/vmod_vtc.vcc
nigoroll: addressed

doc/changes.rst
```diff
@@ -56,6 +56,16 @@ Varnish Cache Next (2021-03-15)
 * All shard ``Error`` and ``Notice`` messages now use the unified
   prefix ``vmod_directors: shard %s``.
 
+* The ``VCL_SUB`` data type is now supported for VMODs to save
+  references to subroutines to be called later using
+  ``VRT_call()``. This will fail the VCL for calls from a wrong
+  context (e.gl. calling a subroutine accessing ``req`` from the
```
Dridi, line 62: "e.gl." typo

Dridi (approved):
I think I'm down to only complaining about a typo.

nigoroll:
Re @slimhazard 
> `BOOL VRT_check_call(ctx, sub)`

I made that `STRING` returning the error message, if any.

doc/changes.rst
nigoroll: thx

doc/changes.rst
```diff
@@ -56,6 +56,23 @@ Varnish Cache Next (2021-03-15)
...
+  context (e.g. calling a subroutine accessing ``req`` from the
+  backend side) and for recursive calls.
+
+  Note that, for performance reasons, recursive call detection only
+  happens for the second dynamic call. This implementation detail is
+  subject to change and must not be relied on.
+
+* ``VRT_check_call()`` can be used to check if a ``VRT_call()`` would
+  succeed in order to avoid the potential VCL failure in case it would
+  not.
```
Dridi, line 71: This is only true for the direct call, not subsequent dynamic calls, right?

nigoroll: yes

lib/libvmod_vtc/vmod_vtc.vcc
slimhazard: If we're looking to the lion, there are two suggestions there that IMO are better: stages, sections. If it were my choice, I'd go with stages. BTW I think plural would be better in any case -- it's a number of iterations, and more than one will almost always be chosen.

#bikeshed except that IMNSHO "leg" is the worst of these choices.

slimhazard:
It's all LGTM now, except for my bikeshedding about the parameter name "leg", and a lot of tests appear to be failing.

nigoroll:
re @slimhazard I think the causes for cci failing are outside this patch and IIUC someone had changed some parameters recently. Let's see how things look after a rebase

doc/changes.rst
```diff
@@ -56,6 +56,23 @@ Varnish Cache Next (2021-03-15)
 * All shard ``Error`` and ``Notice`` messages now use the unified
   prefix ``vmod_directors: shard %s``.
 
+* The ``VCL_SUB`` data type is now supported for VMODs to save
+  references to subroutines to be called later using
+  ``VRT_call()``. This will fail the VCL for calls from a wrong
```
bsdphk, line 61: Better:  "Calls from a wrong context will fail (..."

vmod/vmod_vtc.c
```diff
@@ -372,3 +372,43 @@ vmod_proxy_header(VRT_CTX, VCL_ENUM venum, VCL_IP client, VCL_IP server,
 	return (VRT_blob(ctx, "proxy_header", h, l,
 	    BLOB_VMOD_PROXY_HEADER_TYPE));
 }
+
+/*--------------------------------------------------------------------*/
+
+VCL_DURATION v_matchproto_(td_vtc_benchmark)
+vmod_benchmark(VRT_CTX, VCL_SUB sub, VCL_INT total, VCL_INT leg)
+{
```
bsdphk, line 380: Please leave this out of this PR.

We probably want something like this, but it needs to report proper statistics (min/max/avg/stddev and probably median)

bin/varnishtest/tests/v00018.vtc
```diff
@@ -9,7 +9,7 @@ varnish v1 -vcl {
 
 varnish v1 -errvcl {Variable is read only.} {
 	backend b { .host = "127.0.0.1"; }
-	sub vcl_miss { set now = 1000; }
+	sub vcl_miss { set now += 1s; }
 }
```
bsdphk, line 13: This looks like something that should just have been committed ?

bsdphk (dismissed):

doc/changes.rst
nigoroll: yes, thank you, fixed with next force push

bin/varnishtest/tests/v00018.vtc
nigoroll: done: cfed0152c0b846285b18c897e96f1e4e4b56a98d

vmod/vmod_vtc.c
nigoroll: removed in the next force-push

nigoroll:
@bsdphk thank you for your review, I have addressed all points and force-pushed

Dridi (approved):
Document the SUB type in `doc/sphinx/reference/vmod.rst` and I'm good.

nigoroll:
docs added

rezan:
Just to recap our takeaway, me and Andrew are currently working on a flag to switch back on the compile time VCC sub checks. This will supersede the previous compromise where existing VCC checks would apply everywhere but `vcl_init`.

Im hoping this will be very non controversial given that the functionality the PR introduces will now go in unchanged. We also have a handful of similar VCC flags which switch on/off other kinds of VCL error checking, so this fits right in.

As for why, this is a priority backport for us (looping usecase), but we cannot backport something which drops the VCC checks as proposed. Subs are frequently used and hitting the existing VCC errors is needed safety. So this flag lets us have the feature, skip the unsafety, and maintain VCC feature stability. So im all for it.

Dridi:
> Just to recap our takeaway, me and Andrew are currently working on a flag to switch back on the compile time VCC sub checks. This will supersede the previous compromise where existing VCC checks would apply everywhere but `vcl_init`.
> 
> Im hoping this will be very non controversial given that the functionality the PR introduces will now go in unchanged. We also have a handful of similar VCC flags which switch on/off other kinds of VCL error checking, so this fits right in.

Of course it's going to be controversial when you make specious statements.

> As for why, this is a priority backport for us (looping usecase), but we cannot backport something which drops the VCC checks as proposed. Subs are frequently used and hitting the existing VCC errors is needed safety.

So on  one hand we have one use case, let's call it @nigoroll's use case:

```
$Function VOID register_callback(STRING, SUB)
$Function VOID dynamic_dispatch(STRING)
```

Basically, register a subroutine, for example (but not necessarily) `vcl_init` and effectively call it from a different sub. This dynamic dispatch does not drop VCC checks for the existing `call` action that works with static dispatch, it simply cannot verify anything if the call site is different from the register site.

On the other hand we have another use case, let's call it @rezan's use case:

```
$Function VOID call_right_now(SUB)
```

It works like a static `call`, with a twist, but because it uses a dynamic facility it can't benefit from static checks.

Now @rezan's suggesting to have a flag, which I assume is a boolean parameter, let's call it `vcc_check_sub`. The problem is that this flag makes the two use cases incompatible. You can no longer have a VCL using dynamic dispatch if you also want to statically enforce for-each types of functions. This kind of flag would be counter-productive.

My suggestion during bugwash was to instead have a different VCL type:

```
$Function VOID register_callback(STRING, SUB)
$Function VOID dynamic_dispatch(STRING)
$Function VOID call_right_now(STATIC_SUB) # note the different type here
```

If we have a different type, then we can enforce at compile time that:

- only a sub identifier is expected
- it is compatible in scope

Everyone wins, and there's no compromise to make between the two use cases.

> So this flag lets us have the feature, skip the unsafety, and maintain VCC feature stability. So im all for it.

And I'm all for NOT having that flag. The `VCL_STATIC_SUB` type shouldn't be too hard to add on top of this patch series now that we have all the infrastructure to pass subs to VMODs.

andrewwiik:
Let's make this really really simple because I do not have the patience to explain this in detail, we can tell if the argument value is a static sub because the symbol struct "kind" field will be sym_FUNC if it is not then we enforce our static check.

andrewwiik:
TL;DR I can do it, I am doing it right now, it will not add a lot of code most likely.

andrewwiik:
@nigoroll I would like a test case that covers your dynamic_dispatch use-case.

nigoroll:
Re @Dridi : I would not oppose a switch turning off the meat of this feature, as long as that can remain _on_ ;)
That said, I think your proposal could be a good way forward.

Re @rezan @andrewwiik, as your proposal is about something else, I suggest you continue it in a different PR.

andrewwiik:
@nigoroll I disagree, our propsoal relates directly to this.

andrewwiik:
@nigoroll would this demonstrate your use-case for dynamic dispatch? https://pastebin.varnish-software.com/388747fe23d0c2bee265f7e13e2b11b9c5e0a1278ab00b6f06a1980f005c1268

andrewwiik:
Okay I full understand this now, @Dridi's idea is not bad but I think dynamic_dispatch is the edge case so I will amend:

```
$Function VOID register_callback(STRING, DYNAMIC_SUB)
$Function VOID dynamic_dispatch(STRING)
$Function VOID call_right_now(SUB) # note the different type here
```

andrewwiik:
(and yes he is right it would not be a ton of work)

nigoroll:
@andrewwiik I have spent a significant amount of time and effort with the feedback made on this PR, and its current state is the result of careful consideration. I have laid out the reasons why the `STATIC_SUB` case is incompatible with the original use case of this PR [here](https://github.com/varnishcache/varnish-cache/pull/3163#issuecomment-669084100). I am fine with another proposal that builds upon this PR, but, again, it is a different use case.
Please implement your suggestion and open another PR when you are ready.

nigoroll:
> @nigoroll would this demonstrate your use-case for dynamic dispatch?

can you please read the very first paragraph of this PR?

Dridi:
> Okay I full understand this now, @Dridi's idea is not bad but I think dynamic_dispatch is the edge case so I will amend:
> 
> ```
> $Function VOID register_callback(STRING, DYNAMIC_SUB)
> $Function VOID dynamic_dispatch(STRING)
> $Function VOID call_right_now(SUB) # note the different type here
> ```

I consider this progress if we now disagree on the paint color :)

I having second thoughts on the two VCL types sharing the same C type. Maybe the dynamic one should have the opaque `const struct vcl_sub *` type while the static one has a ready-to-call `vcl_func_f` type (no VRT detour) with the caveat in both cases that it's the VMOD's job to check whether the sub returned a transition.

This way you couldn't confuse one with the other without a compilation error.

nigoroll:
> Maybe the dynamic one should have the opaque `const struct vcl_sub *` type while the static one has a ready-to-call `vcl_func_f` type (no VRT detour)

I am not sure if that would be a good idea. Once we pass a function pointer into a vmod, it can do whatever it wants with it, including calls from forbidden places. The intention behind the opaque `const struct vcl_sub *` is to make it clear that such direct calls should not happen. That is to say, other `VCL_SUB` use cases should still be scrutinized at runtime as implemened with this PR (assert on same VCL, context and recursion check).

What is being discussed by @rezan and @andrewwiik is a an interface to `VCL_SUB` with additional limitations. As I have explained before, I have undestood their motivation and use cases, but I do not see how this can be implemented best. It is clear that there would be additional compile time checks, but I do not understand how one would want to prevent a vmod from doing things like stashing away a `VCL_SUB` and calling it from a different place.

So, to summarize, let's see what the proposal will be, but whatever it might be, I think the "VRT detour" has to stay.

rezan:
VCC flag is here: https://github.com/varnishcache/varnish-cache/commit/8a01b782acc942a645812e94752f14e87c2c4636

We can go ahead and implement Dridi's idea of splitting the `SUB` type into 2 different types, one dynamic and runtime checked and one static and compile time checked, but it will change around this PR a bit (ie breaking changes). Its best that this is done as one push. I do think this is the best solution to the problem.

nigoroll:
@rezan please open a PR.

rezan (dismissed):
Would like to hold this until we can get our split type enhancement completed and merge everything together. This enhancement will break the existing PR since we will have to change the type names around and we seem to be in agreement that this is the optimal solution. This PR has been on hold for over a year, so im really just asking for 1 more week to get the final piece in. Otherwise im concerned that the enhancement will slip pass the current merge window.

rezan:
Just want to give an update since I think we are pretty close to the finish line here. We went ahead and did a quick patch on Friday to split the `VCL_SUB` type into two different types and it came out really nice and clean. Infact, I would say this is less of a compromise and more like the proper way to design this feature because we now have a nice and logical progression from `sub`/`call` => `VCL_SUB` => `VCL_SUB_DYNAMIC` (this is what we are calling the type introduced in this PR). `sub` and `VCL_SUB` are pretty much identical in implementation and behavior and `VCL_SUB_DYNAMIC` simply comes in as a new type with runtime (VRT) checks, which are then shared back to `VCL_SUB` since they both share their core definition. Very elegant.

We actually have a few more interesting ideas we want to implement and we will do that Monday and we plan to have the final PR ready on Tuesday.

nigoroll:
I have rebased and added commits to keep up with recent events

* 0809bdc0c5305d9668f89f74cd5d58a5b718dfd2 to stabilize a test case
* c843d91705ca4a23971538053168fe8c28a74334 to test `PRIV_TOP`
* afb355be9dcaa59c5923492aaeaca25fd1e954d7 to make use of the recently added `VCL_MET_TASK_ALL`.

nigoroll:
FTR: My original plan was to suggest to basically rename `debug.call()` into `std.call()` and make that the recommended way of using VCL_SUBs: `VRT_call()` needs to exist, but the recommended way of using this feature, in my mind, should be to have VMODs return VCL_SUB and make the call explicit in VCL.
During a discussion today, @rezan suggested to teach VCC to support `call vmod.returns_vcl_sub()` instead of `std.call(vmod.returns_vcl_sub())`. I think this should be pretty straight forward in VCC (it would issue a VRT_call() instead of the C function call) and I am :+1: for this idea.
That is, after this PR has been settled.

nigoroll:
On the previous note, @bsdphk also suggested that VCC could `VRT_call()` wrap any function which is referenced as a `VCL_SUB` argument anywhere. The benefit would be that recursion detection would kick in immediately and not possibly delayed, but we would need to take the additional performance hit.
My take on this is that it does not matter so much when we stop the recursion, the important bit is that we stop it at all.
But I am open to weighing this differently.

bsdphk:
VCC can tell conclusively that VCL_SUB arguments have been passed to VMODs but has no idea what happens to them past that.

In VCLs where VCL_SUB arguments are used, the VCC compiled calls must participate in the run-time recursion check:

    sub foo2 {
        vmod.call_stashed();
    }

    sub foo1 {
         call foo2;
    }

    sub vcl_recv {
        vmod.stash(foo1);
        call foo1;
    }

Since VCC code generation happens in the first (and only) pass, the only way to ensure that, is to always wrap VCC compiled call in VRT_call() and have the latter examine the VCL_conf if the bitmap operations are required.

However, *that* throws a serious spanner in the works when it comes to VIP2 adding typed arguments to SUBs (https://github.com/varnishcache/varnish-cache/wiki/VIP2%3A-VCL-typed-functions), even more so if we want to have C-compiler type-checking.

That makes the entire concept of a unified `VRT_call()` function problematic, and while I managed to prototype a CPP-macro-nightmare version, I found a much simpler and much better solution:

When VCC compiles SUB's, it should output a "prologue" which calls a VPI function, to ask if it was OK to call this SUB at this point in the first place.

The VPI function checks a bit in a bitmap in VRT_CTX, like the `VRT_call()` proposed in this PR, the bit# being allocated at VCC compile time and passed to the VPI function as an argument from the compiled VCL_SUB.

That way the VGC C-function will be able to take arguments and return values for VIP2 in the future, and getting those properly scrutinized (and inlined!) by the C-compiler.

Now `VRT_call()` only needs to validate the VCL_SUB belongs to this VCL and then it can just call the actual function pointer.

That still leaves the question of how VIP2 plays with VMODs.

The VMOD would obviously need to specify what "prototype" it expects for the passed VCL_SUB at `.vcc` compile time, so the solution to that will probably be to have the `vmodtool.py` produce a wrapper function with a proper C-prototype, to be compiled into the VMOD, which bails if the "signature" of the passed VCL_SUB does not match the expected signature, before it calls the `TRUSTME()` wrapped function pointer.

PS: The bitmap in the VRT_CTX needs to be NULL'ed on `return(vcl)` processing, because the new VCL might need more slots than the old VCL.

nigoroll:
After 101 discussion with @bsdphk, my takeaway is this
* ensure we do not break plans for VIP2
* he does not accept my suggestion for the delayed recursion check
* I want a solution which only impacts subroutines which actually _have_ SUB references, and we have ideas how to get there

Regarding

> PS: The bitmap in the VRT_CTX needs to be NULL'ed ...

This should already be the case because it gets created in `vcl_call_method()`

nigoroll:
note of an idea @Dridi mentioned during pow-wow: VCC could warn about indirect dynamic calls

rezan:
So are we still planning on having an option for compile time checks? I guess if I were to list out my requirements:

* Have a VMOD take in a `VCL_SUB` and have it be compile time checked the same way a `call` would be checked. Basically, is the `sub` valid for the context its being (immediately) used in.
* Call the `VCL_SUB` inside of the VMOD multiple times in succession. Basically a VMOD driven `for each` loop.

If these 2 things are possible, then I would be very happy.

nigoroll:
I would like add to the case from https://github.com/varnishcache/varnish-cache/pull/3163#issuecomment-769718143 a point which @bsdphk has reminded me of during an IRC conversation (thank you!):

If we want to catch recursions the first time, we need to
* always runtime-check those subs which are used as a SUB argument anywhere
* and additionally those subs which are statically called from them using the `call` statement (we could call this the Transitive Closure)

to illustrate, a modified example of the above is:

```
sub foo2 {
    call foo1;
}

sub foo1 {
     vmod.call_stashed();
}

sub vcl_recv {
    vmod.stash(foo2);
    call foo1;
}
```

By the first rule, only `foo2` would be runtime checked, but because `foo1` is reachable via `foo2`, we need to runtime-check it too.

Some additional remarks:
* This is better than runtime-checking all subs: Calltrees which are not reachable via referenced SUBs (= Calltress which have an empty intersection with referenced SUBs' transitive closure) do not need to be runtime-checked
* _If we knew_ whether or not a sub might make a dynamic call, we could reduce runtime checking further. Yet that would contradict with the existence of `VRT_call()`, or however it is going to be named in the end: Calls would always need to be explicit from VCL. This would likely be fine for the @slimhazard / dispatch use case but definitely break the @rezan / foreach use case. So `VRT_call()` has to stay and we need to runtime-check based on reachability

nigoroll:
With the last force-push, I have addressed the [points raised by phk on Friday](https://github.com/varnishcache/varnish-cache/pull/3163#issuecomment-769718143):
* Recursions are now detected the first time
* The dynamic call checks have been moved to VGC with help of three VPI functions.
* So 'VRT_call()` is now not strictly needed, the actual VGC functions _could_ be called directly (and still are for VCL `call` actions). This avoids any foreclosure of [VIP2 - typed functions](https://github.com/varnishcache/varnish-cache/wiki/VIP2:-VCL-typed-functions)
* With respect to [VIP2 - restriction of VMOD function call sites](https://github.com/varnishcache/varnish-cache/wiki/VIP4:-Restrict-VMOD-function-call-sites), I do see how any commit from this PR added complications. On the contrary, @Dridi showed in 6d49b18f1a7ad612b3c9ba6b3a8a5704fe84ea61 how it can be done in VCC and this PR needed zero adjustment to it

Advise to reviewers:

I have kept all previous commits, but rebased onto #3512. Other than that, you should be able to build on your previous reviews starting at 6d3d13b93cecf6a4cbd2bfbdebaed478364a5786

The first series of commits contains rather trivial prep work building up the two main commits:
* 9ad5e6e036aecf61cc396103773940ec5f19094c is the VPI interface to check the recursion bitmask (also integrating the other runtime checks)
* ca173844ac56e7b434714b91aa7245547220c15a is the meat of the move to VGC

I recommend to go through the commits in commit order (not date order) and consider the individual commit messages.

nigoroll:
I also wanted to note that this commit series should be squashed before merge, I just did not want to make reviewers' lives harder than necessary.

nigoroll:
Re-submitted the last comment in order to bring it back to the default view after yet another force-push to reflect recent clean-up work on our includes:

With the last force-push, I have addressed the [points raised by @bsdphk  on 2021-01-29](https://github.com/varnishcache/varnish-cache/pull/3163#issuecomment-769718143):
* Recursions are now detected the first time
* The dynamic call checks have been moved to VGC with help of three VPI functions.
* So `VRT_call()` is now not strictly needed, the actual VGC functions _could_ be called directly (and still are for VCL `call` actions). This avoids any foreclosure of [VIP2 - typed functions](https://github.com/varnishcache/varnish-cache/wiki/VIP2:-VCL-typed-functions)
* With respect to [VIP4 - restriction of VMOD function call sites](https://github.com/varnishcache/varnish-cache/wiki/VIP4:-Restrict-VMOD-function-call-sites), I do not see how any commit from this PR added complications. On the contrary, @Dridi showed in 6d49b18f1a7ad612b3c9ba6b3a8a5704fe84ea61 how it can be done in VCC and this PR needed zero adjustment to it

Advise to reviewers:

I have kept all previous commits, but rebased onto #3512. Other than that, you should be able to build on your previous reviews starting at 845b873fde338b0ae0f623d0c81c00a56a2d3e7a

The first series of commits contains rather trivial prep work building up the two main commits:
* 44678c138ec0c0121e0db5b8976a755c39f355a7 is the VPI interface to check the recursion bitmask (also integrating the other runtime checks)
* c2b1298c615755970537106f7af92241de60a6d4 is the meat of the move to VGC

I recommend to go through the commits in commit order (not date order) and consider the individual commit messages.

nigoroll:
pow-wow: phk is ok now, I will squash/polish and then @Dridi wants to have a final look.

nigoroll:
I have now force-pushed the squash/polish, restructuring most commits into what I think is a more sensible order after the last rework to move checks to VGC and tighten the recursion check, which required more groundwork in vcc land.
The result of this rework is identical to the previous state of this PR, except for a changelog fix which was previously overlooked. I have saved the previous state [here](https://github.com/nigoroll/varnish-cache/tree/rfc_call_pre-squash_20210206) for reference.
I would ask @Dridi to conduct his final (I hope) review.

lib/libvcc/vcc_symb.c
```diff
@@ -471,6 +471,14 @@ VCC_GlobalSymbol(struct symbol *sym, vcc_type_t type, const char *pfx)
...
+	sym->lname = strdup(VSB_data(vsb));
+	if (type == SUB) {
+		VSB_destroy(&vsb);
+		vsb = VSB_new_auto();
+		AN(vsb);
+		VSB_printf(vsb, "sub_%s", sym->lname);
+		AZ(VSB_finish(vsb));
+	}
 	sym->rname = strdup(VSB_data(vsb));
 	AN(sym->rname);
```
Dridi, line 483: We are missing an `AN(sym->lname)` here, but we could also replace both `strdup()` calls with the `REPLACE()` macro.

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -517,3 +517,51 @@ VCL_##func##_method(struct vcl *vcl, struct worker *wrk,		\
...
+
+VCL_VOID
+VRT_call(VRT_CTX, VCL_SUB sub)
+{
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(sub, VCL_SUB_MAGIC);
+
+	AN(sub->func);
+	sub->func(ctx, VSUB_DYNAMIC, NULL);
```
Dridi, line 566: Should we check the context handling to avoid making progress somewhere:tm: after a return action?

Dridi (approved):
Many thanks for cleaning up the commit history, very straightforward to follow.

I have remarks, but feel free to address them before or after merging at your convenience,

bin/varnishd/cache/cache_vrt_vcl.c
nigoroll: see 1:1 IRC: I do not understand what you mean.

lib/libvcc/vcc_symb.c
nigoroll: lname/rname are `const`, so `REPLACE()` is not ideal. But The `AN()` is missing, thank you

PR #3167, opened 2019-12-18
Defer the illegal write check a bit

Dridi:
After the initial discussion from #3163, and looking more closely at how
variable access is handled in subroutines I noticed a discrepancy.

Setting a read only variable like `obj.ttl` in `vcl_recv` would result in
a misleading error message explaining that it is read only, instead of
simply not available.

This change defers the illegal write check, registering unconditionally
that the symbol was used in a set action. As a result we always get the
correct error message but depending on whether this is happening in a
vcl_ or custom subroutine we either get "in method" or "from method" in
the error message. A minor discrepancy probably worth getting rid of the
prior inconsistency.

This is covered by the v21 test case.


nigoroll:
merging with #3163 looks pretty easy: https://github.com/nigoroll/varnish-cache/tree/illegal_write%2Brfc_call

bsdphk:
looks like an improvement.

PR #3171, opened 2019-12-19
Expose the master and worker PIDs via the CLI

Dridi:
I felt a bit embarrassed by this email from @carlosabalde:

https://varnish-cache.org/lists/pipermail/varnish-misc/2019-December/026767.html


Dridi:
Bugwash conclusion: move the command implementation to `mgt_child.c`, then OK.

PR #3175, opened 2019-12-24
Better weeklies, fix alpine builds

gquintard:
if version is trunk, change it to the current date and add a flag file
so that packages can be suffixed with -weekly/.weekly depending on the
platform

also, drop the number splitting/assembling dance, just use the upstream version

Alpine has some very strict versioning conventions, and I didn't want to change the current scheme for debian/rhel, but we could just drop the weekly suffix for them too

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,24 +105,13 @@ jobs:
       - run:
           name: Update changelog version
           command: |
-            VERSION=$(./configure --version | awk 'NR == 1 {print $NF}')
-
-            # VERSION looks like 5.2.1 or 5.2.0-rc1
-            MAJOR=${VERSION%.*}		# 5.2
-            MINOR=${VERSION##*.}	# 1 or 0-rc1
-            MINOR=${MINOR%%-*}		# 1 or 0
-            RELEASE=${VERSION#*-}	# 5.2.1 or rc1
-            RELEASE=${RELEASE#$VERSION}	# '' or rc1
-
-            # Take version override set on Jenkins builds into account.
-            if [ "$VERSION" = "trunk" ]; then
-            	DEBVERSION=`date "+%Y%m%d"`-weekly~<< parameters.release >>
+            if [ -e .is_weekly ]; then
+                WEEKLY='-weekly'
             else
-            	DEBVERSION="$MAJOR.$MINOR"-1~<< parameters.release >>
+                WEEKLY=
             fi
-
-            sed -i -e "s|@SECTION@|varnish-$MAJOR|" "debian/control"
-            sed -i -e "s|@VERSION@|$DEBVERSION|"  "debian/changelog"
+            VERSION=$(./configure --version | awk 'NR == 1 {print $NF}')$WEEKLY~<< parameters.release >>
+            sed -i -e "s|@VERSION@|$VERSION|"  "debian/changelog"
       - run:
           name: Install Build-Depends packages
           command: |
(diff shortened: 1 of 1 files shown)


Dridi:
I think ec464b943a23107b36a74c06e6f7b16a837eebb9 has nothing to do in this pull request, please squash the rest in a single commit.

gquintard:
This commit actually reduce the difference between master. The first patch
did a bit too much (remove useless macro definitions, but also reordered
arguments)

Dridi:
Can we stick to alpine changes in this pull request? Also I'm not sure we should change `build_alpine` to `build_apks`: the former is explicit and the latter could mean anything supporting that package manager (or confuse android users).

While at it, since Alpine Linux is not an OS we formally support, is it possible not to fail the build for that job?

gquintard:
> Can we stick to alpine changes in this pull request?

I removed the argument changes in `build_rpms`, but I feel they could have gotten in without issue

> Also I'm not sure we should change build_alpine to build_apks: the former is explicit and the latter could mean anything supporting that package manager (or confuse android users).

we have `build_rpms` and `build_debs`, I was looking for consistency

> While at it, since Alpine Linux is not an OS we formally support, is it possible not to fail the build for that job?

I can make the build non-fatal, but it will then always we green.
It might be time to consider official support for it, or at least do what we can for the tests to pass. Lack of CI means an uphill battle to have builds passing consistently. I regularly need to backport downstream patches because we don't care enough about that platform.

Dridi:
> I removed the argument changes in `build_rpms`, but I feel they could have gotten in without issue

If that's the case, please push those changes directly.

> we have `build_rpms` and `build_debs`, I was looking for consistency

Yes but they are shared by OS builds so in the CircleCI jobs matrix we consistently see `build_${OS}`.

> I can make the build non-fatal, but it will then always we green.

That sucks. At least on TravisCI when the osx job fails it's red, but the overall build is green.

> It might be time to consider official support for it, or at least do what we can for the tests to pass. Lack of CI means an uphill battle to have builds passing consistently. I regularly need to backport downstream patches because we don't care enough about that platform.

I'm not calling that shot, but how much have you engaged with the Alpine folks? It's great that Varnish is part of the Alpine package collection, but it would be even greater if they tried to upstream their patches.

But I agree that a passing CI is better, and I suppose my frustration comes from the fact that ever since we updated to the latest release I got continuous spam from CircleCI, and this PR would fix that.

Do we agree with "all of the above"? If so please remove RPM stuff from this pull request, make the requested changes and squash the result in a merge-able commit.

gquintard:
> If that's the case, please push those changes directly.

noted

> Yes but they are shared by OS builds so in the CircleCI jobs matrix we consistently see build_${OS}.

gotcha, I hanged the job name in yaml to `build_apks`, but the interface will show `build_alpine`. Works?

> That sucks. At least on TravisCI when the osx job fails it's red, but the overall build is green.

yuuuup, agreed

> I'm not calling that shot, but how much have you engaged with the Alpine folks? It's great that Varnish is part of the Alpine package collection, but it would be even greater if they tried to upstream their patches.

Not much, I encouraged them to upstream every time I backported the patches, never got a response.

> But I agree that a passing CI is better, and I suppose my frustration comes from the fact that ever since we updated to the latest release I got continuous spam from CircleCI, and this PR would fix that.

Next time, open a ticket and assign it to me, this case was a lack of reactivity from me.

> Do we agree with "all of the above"? If so please remove RPM stuff from this pull request, make the requested changes and squash the result in a merge-able commit.

All good for me, though some RPM stuff needs to stay as they are required for the proper weekly handling, which in turns give us the alpine fix. Code is pushed, I'll merge tonight if okay with you.

PR #3176, opened 2019-12-26
Stop pretending we are cool with cooling backends

Dridi:
As I was trying to add test coverage for the lines this patch removes I
did some research and realized this was actually dead code.

Consider the state machine from 57a5b9c78f66:

      .----------W>---------------.
      |              .--W>---.    v
    init ---> cold --|       |-- warm --.
              ^  ^   '---<C--'    |     |
              |  |                |     |
              |  '--- cooling <C--'     |
              |         ^               |
              |         C               |
              |         |    .---<---.  |
              '--<C-- busy --|       |--'
                             '--->---'

It says:

> The transitions marked with a 'W' or a 'C' are the ones dispatching
> WARM and COLD events respectively.
>
> Since the "busy" state is considered warm, VMODs should operate as
> such and backend creation remains possible.

In other words, when we first ran into backend creation problems with
regards to VCL temperature, it was because signals would be sent whether
the VCL would still have ongoing transactions or not, and this was fixed
by the introduction of the busy state. It was not the intention _then_
to allow backend creation in the COOLING state.

The result is that a COLD event is only dispatched when the VCL can
effectively cool down, so any VMOD creating backends in the event
function must make sure to only do it for INIT or WARM events, and any
VMOD doing asynchronous backend creation (outside of a formal VCL task)
should have some form of synchronization between its event function and
the thread creating backends, because once a COLD event is dispatched we
are effectively forbidden from doing any "warm" operation.

The COOLING state can only be reached when a VMOD holds a VCL reference
and could not otherwise transition immediately to the COLD state. This
prevents varnishd from pulling the rug under the VMOD's feet.

It is possible to add a director in the cooling state and get away with
it, but if the director in question is a VBE backend we run into two
problems: vbe_destroy expects to run in the CLI thread, and lifting the
constraint would then let the backend try to remove it self from a list
it never entered. This is a guaranteed panic, pick your flavor...

Instead of "fixing" the code as described above to tolerate a backend
creation that is guaranteed to fail in the COOLING state, we should
rather enforce that COLD events mean the end of resource creation until
the next WARM event, and that VMODs should behave accordingly.

### DIFF
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -160,10 +160,6 @@ VRT_AddDirector(VRT_CTX, const struct vdi_methods *m, void *priv,
 	vcl = ctx->vcl;
 	CHECK_OBJ_NOTNULL(vcl, VCL_MAGIC);
 
-	// opportunistic, re-checked again under lock
-	if (vcl->temp == VCL_TEMP_COOLING)
-		return (NULL);
-
 	ALLOC_OBJ(d, DIRECTOR_MAGIC);
 	AN(d);
 	ALLOC_OBJ(vdir, VCLDIR_MAGIC);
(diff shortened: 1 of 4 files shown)


Dridi:
In order to add test coverage I ended up changing this condition:

https://github.com/varnishcache/varnish-cache/blob/a453994978df8c11a7dbf7e188a8d0079bd5f71a/bin/varnishd/cache/cache_vrt_vcl.c#L163-L165

I would only return if the `VTC_MODE` debug flag was not set (I'm just saying, if someone wants to go through the trouble of testing it themselves).

See the dead code currently visible here:

https://varnish-cache.org/gcov/bin/varnishd/cache/cache_vrt_vcl.c.html#200

Hopefully it will go away with this patch :)

Dridi:
Rebased against master, I added test coverage on top of f9b7f1818 for the COOLING state.

Dridi:
I did more research and overlooked a critical flaw in my reasoning: we change the temperature before we send events, so we actually need to fix the undo code path.

Dridi:
Unfortunately it took me several attempts to solve this problem, but I believe I'm done.

For a summary:

```
git diff f9b7f1818..00fa3c754 -- bin/varnishd/cache/cache_backend.c bin/varnishd/cache/cache_vrt_vcl.c bin/varnishtest/tests/v00063.vtc lib/libvmod_debug/vmod.vcc lib/libvmod_debug/vmod_debug.c
```

PR #3178, opened 2020-01-02
Reshape the HTTP/1 send loop

Dridi:
There is now a single `writev` call site and as documented we stop as soon
as we hit either of `send_timeout` or `idle_send_timeout`. Previously, we
would keep trying to send after reporting `idle_send_timeout` contrary to
what the documentation said, making it ineffective for HTTP/1 sessions.

### DIFF
--- a/bin/varnishd/http1/cache_http1.h
+++ b/bin/varnishd/http1/cache_http1.h
@@ -60,6 +60,6 @@ void V1L_Chunked(const struct worker *w);
 void V1L_EndChunk(const struct worker *w);
 void V1L_Open(struct worker *, struct ws *, int *fd, struct vsl_log *,
     vtim_real deadline, unsigned niov);
-unsigned V1L_Flush(const struct worker *w);
-unsigned V1L_Close(struct worker *w, uint64_t *cnt);
+enum sess_close V1L_Flush(const struct worker *w);
+enum sess_close V1L_Close(struct worker *w, uint64_t *cnt);
 size_t V1L_Write(const struct worker *w, const void *ptr, ssize_t len);
--- a/bin/varnishd/http1/cache_http1_deliver.c
+++ b/bin/varnishd/http1/cache_http1_deliver.c
@@ -85,7 +85,7 @@ void v_matchproto_(vtr_deliver_f)
 V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
 {
 	int err = 0, chunked = 0;
-	unsigned u;
+	enum sess_close sc;
 	uint64_t hdrbytes, bytes;
 
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
@@ -152,7 +152,7 @@ V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
 			V1L_EndChunk(req->wrk);
 	}
 
-	u = V1L_Close(req->wrk, &bytes);
+	sc = V1L_Close(req->wrk, &bytes);
 	AZ(req->wrk->v1l);
 
 	/* Bytes accounting */
@@ -163,7 +163,9 @@ V1D_Deliver(struct req *req, struct boc *boc, int sendbody)
 		req->acct.resp_bodybytes += bytes - hdrbytes;
 	}
 
-	if ((u || err) && req->sp->fd >= 0)
-		Req_Fail(req, SC_REM_CLOSE);
+	if (sc == SC_NULL && err && req->sp->fd >= 0)
+		sc = SC_REM_CLOSE;
+	if (sc != SC_NULL)
+		Req_Fail(req, sc);
 	VDP_close(req);
 }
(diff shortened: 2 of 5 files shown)


Dridi:
Some context here: I tried to stabilize s10 in df9f348975092878d54cb9ccd679e86eed5e34ae after noticing that it often made it in the VTEST headlines.

On FreeBSD it would sometimes happen that varnishd would hit `idle_send_timeout` after the backend closed its connection, resulting in a broken pipe instead. With df9f348975092878d54cb9ccd679e86eed5e34ae we went from a somewhat frequent test failure to a systematic one: currently none of the FreeBSD instances are succeeding.

Why does it fail only on FreeBSD? To trigger the timeouts s10 reduces the send buffer of the `varnishd` session and the receive buffer of the `varnishtest` client. We ask for a 128 bytes buffer and get exactly that on FreeBSD, but on Linux we get around 4kB. So from there Linux boxes end up receiving the complete response in 4kB chunks while FreeBSD ends up orders of magnitude slower and hits the 1m VTC timeout.

We skip the test on SunOS for now because of additional constraints we can't yet satisfy.

So in its current form the test case does not catch the false-negative failure, because we need the client to be `non_fatal` since trying to resume the fetch is a guaranteed failure.

Dridi:
FYI I'm working on ensuring s10 does catch all failure conditions, it currently can't. Such a fix will be more complicated after a first-look estimate, this patch is rather trivial in comparison to what I'm cooking.

Dridi:
The additional patch (d245305daf155618c2d380636ad59eae88714ef7) I was working on ended up not as complicated as I anticipated, so I'm feeling confident to add it to this pull request.

Unless we want to open a can of `SC_TX_TIMEOUT` worms this pull request is complete and should only need a single rebase to squash b0ed6bc8b6abdd82ad29b775ca639a1db85b889c before a merge, barring further review change requests.

@bsdphk would like to see this patch series get some serious live traffic outside the test suite before that happens.

Dridi:
Out of curiosity I tried on FreeBSD 12.1-STABLE and it doesn't work, we never run into `EWOULDBLOCK` and for some reason the send timeout never kicks in. I will revisit this on Monday.

nigoroll:
@Dridi I can help with live traffic but would wait for your next iteration first.

Dridi:
The next iteration is ready for a review.

Dridi:
I forgot to mention, one patch is easier to review if we ignore whitespace: https://github.com/varnishcache/varnish-cache/pull/3178/commits/8c921c43a0ffb7ea51f67f43d71111bc360f6f6a?w=1

nigoroll:
Um, yes, the timeout names are so, so bad. And yes, we are in the middle of the process of improving them, but they are bad: When I looked at this initially I did not see that already the description points to a breaking change:

> as documented we stop as soon as we hit either of `send_timeout` or `idle_send_timeout`

I do not think this is what the documentation says:

* [idle_send_timeout](http://varnish-cache.org/docs/trunk/reference/varnishd.html#ref-param-idle-send-timeout)

  Send timeout for individual pieces of data on client connections. May get extended if `send_timeout` applies.

* [send_timeout](http://varnish-cache.org/docs/trunk/reference/varnishd.html#send-timeout)

  When `idle_send_timeout` is hit while sending an HTTP1 response, the timeout is extended unless the total time already taken for sending the response in its entirety exceeds this many seconds.

@Dridi I suppose we agree that `idle_send_timeout` is pretty useless from a user perspective (what benefit would users have from controlling how long an individual `writev()` may take?), so I actually would agree that your suggestion makes a lot of sense.

But I would think that as we want to streamline the timeout names, that would be a better opportunity to change the semantics.

So I would suggest that you reduce your patch to just the polishing (which I like) and add a marker that we want to replace the current `idle_send_timeout` with something more useful.

Dridi:
I was convinced I read that `idle_send_timeout` would abort the transaction, but nope.

Thanks to you I was able to better understand why s10 was so unreliable on FreeBSD and hopefully fixed that for good in bcba964921106fb2267913208141e3b94c65d766.

FreeBSD clearly behaves as documented, a partial write implies a send timeout on the socket. The Linux documentation says the same thing, so the code looks sound as it is, see 9a7dc49bdf14c05de14ad8502ce38bd1a149d805.

I finally wrapped my head around `idle_send_timeout`. It's definitely not useless, and defines how frequently we get the occasion to check whether we went past the `send_timeout` deadline. The documentation isn't a bit confusing, but with the default values it basically means that for the 10 minutes of allowed delivery we get roughly 10 chances to check whether `send_timeout`.

In other words it's a trade off of how often we check at the expense of context switches, and the opportunity to pace the delivery as suggested by the XXX comment, which we currently don't. Once we have that then idle_send_timeout brings much more value.

nigoroll:
I think the `SC_TX_ERROR` handling was a good idea

Dridi:
#3185

Issue #3181, opened 2020-01-06
RFC: aux_path parameter


nigoroll:
This ticket is to ask for opinions about whether or not an `aux_path` parameter (other names suggested so far: `cache_path`, `search_path`) would be considered worth the implementation effort.

* `aux_path` would define a search path paralleling `vcl_path` and `vmod_path`, but for auxiliary use cases like files with synthetic response pages.
* a default `aux_path` could be `$PREFIX/etc/varnish/aux:$PREFIX/share/varnish/aux`
* `aux_path` would be available to vmods, but with the currently envisioned simplified implementation, only during vcl events and vmod initialization (iow from CLI context)
  * this would imply that changes to `aux_path` would only become effective for newly loaded VCLs
* the primary use case for `aux_path` would be a default search path for vmods handling files, see for example [vmod_file](https://code.uplex.de/uplex-varnish/libvmod-file/blob/master/README.rst)

(subjective) summary of previous discussion:
* `vcl_path` cannot easily be made available in the worker process because it gets changed in the master process and the respective memory region is not shared. (see https://github.com/varnishcache/varnish-cache/issues/3172#issuecomment-571168935 )
* PRO: a new parameter may simplify configuration of multiple instances using the same VCL
* CON: instead of a parameter, a per-instance include can be used
* CON: could be problematic if we were to confine the worker process to some subtree (think: chroot(), namespaces)

Please provide feedback how important/useful such a parameter would be for you.

slimhazard:
`std.fileread()` could also take advantage of `aux_path`. (VMOD file can be thought of as "`std.fileread()` except you can change the file".)

IMO Varnish can and should provide the tools that allow admins to decide how they can restrict the worker process and also provide access to the file system. For example, `aux_path` could be set so that it is accessible within a chroot jail.

Dridi:
What should `std.fileread()` do with a relative path today? I don't think accessing the working directory makes sense. As per the bugwash discussion, I'd rather have something like a `VRT_SearchPath()` function for relative paths lookups. I don't know yet what default value we would want or how we would manage the "auxiliary" patch, I need to ponder this.

But if this is hidden behind the VRT curtain, can't we make it update-able via the CLI? The master already forwards CLI commands to the child, we can maybe have a syntax for "private" commands that would never be forwarded?

```
@aux_path <string>
```

The manager could know NOT to forward such a command (starting with `@`):

```
$ varnishadm @aux_path "/something"
Unknown request.
[...]

Issue #3172, opened 2019-12-20
RFC: vmod read-access to parameters


nigoroll:
A very long time ago, we already discussed if parameters should be part of the API for vmods to use. The answer at the time was *no* and I _think_ it was 9699424ccd9f7fc51a8d1c651889433fdd200ce5 which put the final nail on the coffin to discourage vmod access to `struct param`, unless you were willing to leave `$ABI vrt` and compile requiring access to varnish internal headers.

The example at the time was access to the `pcre_match_limit` and `pcre_match_limit_recursion` which would be useful as defaults for [vmod_re](https://code.uplex.de/uplex-varnish/libvmod-re/blob/master/README.rst).

Since then, varnish has grown additional parameters and we would now have a use case where for a new [vmod_file](https://code.uplex.de/uplex-varnish/libvmod-file/blob/master/README.rst), access to `vcl_path` would be the sensible thing to set the default *path*.

So the topic is back on the table with the question: Should we make _some_ of the varnish parameters part of VRT? IMHO, at least the following could/would be useful and safe for vmods to access:

* ``*_timeout``, ``timeout_*``
* ``*_support``
* ``pcre_match_limit``
* ``pcre_match_limit_recursion``
* ``thread_pool_stack``
* ``vcl_path``

slimhazard:
As I recall, the issue was that VMODs should not be able to modify the params, and I would agree. We have the CLI with `param.set` as the public interface for that, it should be enough.

If that's the issue, then why not just have read-only VRT access to all of the params? Rather than try to decide which of them to provide.

I would add that the API feels a bit inconsistent in some places, because it exposes some capabilities without a way to know about the params that might affect them. A good example is VRE -- `VRE_exec()` has an argument for the `pcre_*` params, and users may have changed those values for a reason. But a VMOD using VRE has a take a guess, or ask users to set the params a second time. The natural choice for `VRE_exec()` would be to default to whatever the current `pcre_*` settings are, but a VMOD can't do that and stay within VRT.

nigoroll:
bugwash:
* `VRT_param_${type}(VCL_STRING name)`
* all parameters exposed
* return `const * const` for pointer types
* vmods have to behave and we are aware that vmods which intend not to, can do so anyways

bsdphk:
The issue with changing parameters from anywhere in the worker process, is that the master-copy in the mgt process would not have them, so they would be lost in a restart.

bsdphk:
[...]

PR #3182, opened 2020-01-07
Handle malformed paths for param.set

Dridi:
The bar is not very high, it needs to either be an empty string or a
list of absolute directories separated by colon characters. But that
validation is pretty strict.

---

When we discussed #3172 during last bugwash the question of validating path parameters came up. In this patch series I first trimmed some fat, then implemented a absolute path validation, and plugged everything into parameter handling and libvcc.

A nice side effect to this change is that we only ever build VCL and VMOD paths once per change, including once for the initial value. So between the single allocation for the data structure and no longer rebuilding both paths whenever we load a VCL it's a bit of churn that goes away, admittedly not in a critical part of the code base.

Another interesting result of this change is that we now have a single call site for `VFIL_setpath()`, and that could still be the case if #3181 reached consensus.

I deliberately left "path escape" out of the scope. By that I mean using `"../"` directory components to attempt reading files outside of the designated path. There was no clear "we should do that" feeling from the bugwash, contrary to the absolute paths question.

There might be other means to escape the path, like creating symbolic links, I don't know whether we could really do something about that. And next thing you know someone's gonna yell that we broke their `../` use case.

### DIFF
--- a/bin/varnishd/mgt/mgt.h
+++ b/bin/varnishd/mgt/mgt.h
@@ -217,8 +217,8 @@ void mgt_vcl_startup(struct cli *, const char *vclsrc, const char *origin,
 int mgt_push_vcls(struct cli *, unsigned *status, char **p);
 int mgt_has_vcl(void);
 extern char *mgt_cc_cmd;
-extern const char *mgt_vcl_path;
-extern const char *mgt_vmod_path;
+extern struct vfil_path *mgt_vcl_path;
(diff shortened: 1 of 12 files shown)


nigoroll:
bugwash:
* we should not enforce absolute paths
* @Dridi and @nigoroll see a cleanup benefit
* no clear vote from @bsdphk yet, wants to take it offline.

PR #3184, opened 2020-01-09
import varnish-modules/cookie in-tree

gquintard:
Following up on the whole "import essential vmods in-tree"  and specifically https://github.com/varnishcache/varnish-cache/pull/3055#issuecomment-529339906, here's `vmod_cookie`


lib/libvmod_cookie/vmod.vcc
```diff
@@ -0,0 +1,248 @@
...
+get/set/delete functions are available. A filter_except() method removes all
+but a set comma-separated list of cookies. A filter() method removes a comma-
+separated list of cookies.
+
+Regular expressions can be used for either selecting cookies, deleting matching
+cookies and deleting non-matching cookie names.
+
+A convenience function for formatting the Set-Cookie Expires date field
+is also included. If there are multiple Set-Cookie headers vmod-header
+should be used.
```
Dridi: I don't think an in-tree VMOD should reference an out-of-tree VMOD to deal with its shortcomings.

As much as I liked vmod_cookie to the point where I took inspiration from it for vmod_querystring, I ended up rewriting it from scratch (designing a new API taking advantage of new features from Varnish 4.1) to offer something both easier and more efficient to use.

Of course "easier" is subjective, but it resulted in something easier to implement, document and maintain (I'm looking at you, `filter_except`, and you know why).

I'm not suggesting we copy vmod_querysting to implement vmod_cookie but I'd rather start from a new VCC file. Once in tree, we can then rename vmod_cookie to something else like vmod_legacycookie in the varnish-modules project and end-users can keep using it with minimal changes in their existing VCL:

```diff
-import cookie;
+import legacycookie as cookie;
```

Also, a mistake happened at some point: `vmod.c` is a copy of `vmod.vcc`.

This module comes with a ton of tests and the current namespace we have makes it increasingly harder to figure where to find existing or put new VMOD coverage. With this many new test cases I'd rather start a new namespace for test cases dedicated to VMOD functionality coverage. We can introduce that change with vmod_cookie and then move other in-tree tests for the rest of our VMODs.

Of course it doesn't mean we can't use VMODs in test cases from the current namespace.

An exception I would make is vmod_vtc and vmod_debug that are only meant to support the test suite.

As we are increasing the number of in-tree modules I'd also like to stop using `vmod.vcc` in favor of `vmod_*.vcc`.

gquintard:
- [x] `vmod.c` nuked
- [x] don't mention `vmod_header`
- [ ] decide if we should have a `bin/varnishtest/tests/vmods` directory to isolate vmods
- [ ] trim/edit the API, maybe

I have no preference on the API, I don't think we should change it to much as the goal is to help packaging and users, so if the VCL can be immediately compatible, I'd prefer that.

gquintard:
No feedback, so ok?

gquintard:
I'll commit later today if no one yells

gquintard:
just saw Nils' comment on IRC. I'll move the cookie tests to their own directory

nigoroll:
@gquintard I would rather wait for OK from @Dridi or @bsdphk

gquintard:
@nigoroll, both are +0 on this one

nigoroll:
@gquintard see [how we handle PRs](http://varnish-cache.org/organization/index.html#organization-maintainers):

> if the maintainers who are for the change outnumber the ones against by at least two, the pull request can be merged or the change implemented.

I think we should have another +1 on this.

gquintard:
Fair enough. I need to isolate the vtcs in their own directory, hopefully,
that will win Dridi over :-)

gquintard:
@Dridi, `vtc`s are now in `bin/varnishtest/vmodtests/cookie/`

Dridi:
We finally discussed the matter of VMODs test suites with @nigoroll and @bsdphk and we have an outcome:

- we want to co-locate the vmod-specific vtc files with the vmod sources
  - `lib/libvmod_<name>/tests/<name>_[id]%04d.vtc`
  - example: `lib/libvmod_cookie/tests/cookie_b0001.vtc` (first basic vmod_cookie test)
  - same nomenclature as the current test suite, 4 digits instead of 5
- but we want to run them as part of the global test-run
- we want to generate `bin/varnishtest/{vmods,programs}.h`
- we want the gcov survey to tell us what the vtcs actually test

I will take care of setting up everything except the last point that @bsdphk will handle.

Can you please update your PR to match those requirements? We can merge it once we have the bits and pieces needed (everything except the last item).

gquintard:
> same nomenclature as the current test suite, 4 digits instead of 5

isn't that an issue for regressions?

I'll update the PR

Dridi:
I guess we could stick to 5 digits, no harm in my opinion.

gquintard:
@Dridi , we should be ok

Dridi:
I already pushed most of the changes needed before merging this, however we agreed with @gquintard that I would make the final adjustments myself to save ourselves additional review rounds (for example whitespace nitpicks in the test suite).

It should be over sometime today.

Dridi:
To give an idea of the things that wouldn't have passed my review:

- missing SPDX tags
- code style violations
- whitespace OCD
- broken RST and manual style
- deprecated function

There may be other issues, this isn't meant to be exhaustive, but it might take me some time. Since I'm going to push directly I encourage interested parties to review my own changes afterwards.

PR #3187, opened 2020-01-10
Don't ignore OA_GZIPBITS if there is a boc

Dridi:
Under load, client c4 from g00005.vtc may fail with a 200 response
instead of the expected 206 partial response.

There is a window during which we might still see a boc, but because
c4 sets `beresp.do_stream` to `false`, the fetch has to be over. To close
this race we can instead reference the boc as suggested in #2904 and
keep track of the boc state.

---

With enough load c4 explodes, spotted via vtest.

### DIFF
--- a/bin/varnishd/cache/cache_gzip.c
+++ b/bin/varnishd/cache/cache_gzip.c
@@ -42,6 +42,7 @@
 
 #include "cache_varnishd.h"
 #include "cache_filter.h"
+#include "cache_objhead.h"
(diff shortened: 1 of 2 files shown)


Dridi:
Bugwash outcome: I will also follow up with a pull request to grab the state in a single operation.

Dridi:
After a quick inspection, I didn't find a common pattern in the code base that would mandate the bugwash suggestion. Until proven wrong, that subsequent pull request will not happen.

PR #3193, opened 2020-01-20
Allow globbing patterns when including VCL files

gquintard:
There are a few use cases for this, but the main one is dynamic
setups where each domain is handled by a different VCL, so one can just
drop a file in a directory and reload to take it into account. A second
example are VCL that will provide default logging/logic that the admins
want to be picked up automatically without effort from the VCL writers.

### DIFF
--- a/lib/libvcc/vcc_compile.c
+++ b/lib/libvcc/vcc_compile.c
@@ -52,7 +52,9 @@
 
 #include "config.h"
 
+#include <errno.h>
 #include <fcntl.h>
+#include <glob.h>
 #include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
@@ -540,10 +542,12 @@ vcc_file_source(const struct vcc *tl, const char *fn)
 static void
 vcc_resolve_includes(struct vcc *tl)
 {
-	struct token *t, *t1, *t2;
+	struct token *t, *t1, *t2, *tt;
 	struct source *sp;
 	struct vsb *vsb;
 	const char *p;
+	glob_t glb;
+	char **s;
 
 	VTAILQ_FOREACH(t, &tl->tokens, list) {
 		if (t->tok != ID || !vcc_IdIs(t, "include"))
@@ -567,6 +571,8 @@ vcc_resolve_includes(struct vcc *tl)
 			return;
 		}
 
+		vsb = VSB_new_auto();
+		AN(vsb);
 		if (t1->dec[0] == '.' && t1->dec[1] == '/') {
 			/*
 			 * Nested include filenames, starting with "./" are
@@ -580,26 +586,53 @@ vcc_resolve_includes(struct vcc *tl)
 				vcc_ErrWhere(tl, t1);
 				return;
 			}
-			vsb = VSB_new_auto();
-			AN(vsb);
 			p = strrchr(t1->src->name, '/');
 			AN(p);
 			VSB_bcat(vsb, t1->src->name, p - t1->src->name);
 			VSB_cat(vsb, t1->dec + 1);
-			AZ(VSB_finish(vsb));
-			sp = vcc_file_source(tl, VSB_data(vsb));
-			VSB_destroy(&vsb);
+		} else
+			VSB_cat(vsb, t1->dec);
+
+		AZ(VSB_finish(vsb));
+
+		/* only try to glob if the path starts with "./" or "/", and if 
+		 * it contains a "*"
+		 */
+		if (((t1->dec[0] == '.' && t1->dec[1] == '/') ||
+		    t1->dec[0] == '/') && strchr(VSB_data(vsb), '*')) {
+			if (glob(VSB_data(vsb), 0, NULL, &glb)) {
+				VSB_printf(tl->sb,
+				    "Globbing pattern matched no file (%s)\n",
+				    VSB_data(vsb));
+				vcc_ErrWhere(tl, t1);
+				return;
+			}
+			tt = tl->t;
+			for (s = glb.gl_pathv; *s; s++) {
+				tl->t = tt;
+				sp = vcc_file_source(tl, *s);
+				if (sp == NULL) {
+					vcc_ErrWhere(tl, t1);
+					return;
+				}
+				VTAILQ_INSERT_TAIL(&tl->sources, sp, list);
+				sp->idx = tl->nsources++;
+				tl->t = t2;
+				vcc_Lexer(tl, sp);
+			}
+			globfree(&glb);
 		} else {
-			sp = vcc_file_source(tl, t1->dec);
+			sp = vcc_file_source(tl, VSB_data(vsb));
+			if (sp == NULL) {
+				vcc_ErrWhere(tl, t1);
+				return;
+			}
+			VTAILQ_INSERT_TAIL(&tl->sources, sp, list);
+			sp->idx = tl->nsources++;
+			tl->t = t2;
+			vcc_Lexer(tl, sp);
 		}
-		if (sp == NULL) {
-			vcc_ErrWhere(tl, t1);
-			return;
-		}
-		VTAILQ_INSERT_TAIL(&tl->sources, sp, list);
-		sp->idx = tl->nsources++;
-		tl->t = t2;
-		vcc_Lexer(tl, sp);
+		VSB_destroy(&vsb);
 
 		VTAILQ_REMOVE(&tl->tokens, t, list);
 		VTAILQ_REMOVE(&tl->tokens, t1, list);
--- a/bin/varnishtest/tests/b00072.vtc
+++ b/bin/varnishtest/tests/b00072.vtc
@@ -1,0 +1,44 @@
+varnishtest "Test globbing includes in VCL"
+
+server s1 {
+	rxreq
+	txresp
+} -start
+
+varnish v1 -errvcl "Globbing pattern matched no file (${tmpdir}/*.vcl)" {
+	vcl 4.0;
+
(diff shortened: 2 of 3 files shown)


gquintard:
I won't be there for intwash, but I'll dutifully read the IRC backlog

lib/libvcc/vcc_compile.c
```diff
@@ -580,26 +586,53 @@ vcc_resolve_includes(struct vcc *tl)
...
+		 * it contains a "*"
+		 */
+		if (((t1->dec[0] == '.' && t1->dec[1] == '/') ||
+		    t1->dec[0] == '/') && strchr(VSB_data(vsb), '*')) {
+			if (glob(VSB_data(vsb), 0, NULL, &glb)) {
+				VSB_printf(tl->sb,
+				    "Globbing pattern matched no file (%s)\n",
+				    VSB_data(vsb));
+				vcc_ErrWhere(tl, t1);
+				return;
```
nigoroll: for this and subsequent `return` statements, the VSB leaks (iow, a `VSB_destroy()` is missing)

lib/libvcc/vcc_compile.c
```diff
@@ -580,26 +586,53 @@ vcc_resolve_includes(struct vcc *tl)
...
+
+		/* only try to glob if the path starts with "./" or "/", and if 
+		 * it contains a "*"
+		 */
+		if (((t1->dec[0] == '.' && t1->dec[1] == '/') ||
+		    t1->dec[0] == '/') && strchr(VSB_data(vsb), '*')) {
+			if (glob(VSB_data(vsb), 0, NULL, &glb)) {
+				VSB_printf(tl->sb,
+				    "Globbing pattern matched no file (%s)\n",
+				    VSB_data(vsb));
```
nigoroll: assert that the `glob()` return is `GLOB_NOMATCH` ?

lib/libvcc/vcc_compile.c
```diff
@@ -580,26 +586,53 @@ vcc_resolve_includes(struct vcc *tl)
...
+				vcc_Lexer(tl, sp);
+			}
+			globfree(&glb);
 		} else {
-			sp = vcc_file_source(tl, t1->dec);
-		}
-		if (sp == NULL) {
-			vcc_ErrWhere(tl, t1);
-			return;
+			sp = vcc_file_source(tl, VSB_data(vsb));
```
nigoroll: maybe we can avoid some duplication here, for example by using a local `glob_t` when there is no glob expansion.

lib/libvcc/vcc_compile.c
nigoroll: and wouldn't the right way be to attempt including the pathname with the literal wildcard?

Dridi:
I don't see a reason to restrict glob support to includes starting with either `/` or `./`.

We either need a parameter to enable this, a different syntax (eg. `include pattern "<glob>";`) or it applies to all includes. I think this should also be a feature for `struct vfil_path` either at creation or search time, instead of something bolted in libvcc.

Possibly a new function like `VFIL_searchglob` that doesn't stop at the first match (but stops at the first path entry matching at least one file).

rezan:
I like this feature, the use case you point out is spot on. However, just wondering why no relative path support?

Dridi:
In case this is discuted during intwash, my take on it would be:

- not implemented with a parameter (either a VCL keyword or all in)
- no arbitrary limitation on file names
- implemented as a new `VFIL_` function (doesn't leak into `vmod_path` handling)

Interesting question: what if we have glob material in `vcl_path`?

gquintard:
@nigoroll thanks for the review, I'll fix the issues you pointed at.

For all three of you wondering about the limitations on the the path, @Dridi got it on his last question:

> what if we have glob material in vcl_path?

The current approach only globs if there's no question about the directory being globbed. I feared that adding a directory to `vcl_path` could silently add files to the globbing results. But if we are fine with that, no limitation needed.

lib/libvcc/vcc_compile.c
gquintard: > and wouldn't the right way be to attempt including the pathname with the literal wildcard?

sorry, I didn't catch your meaning here

nigoroll: like shell attempts to use the file with the literal asterisk when the glob expansion fails

```slink@haggis:~$ ls *guillaume*
ls: cannot access '*guillaume*': No such file or directory
slink@haggis:~$ touch *guillaume*
slink@haggis:~$ ls *guillaume*
*guillaume*
```

(and no, I do not mean to `touch` you no matter how many stars around you ;) )

gquintard: ah, good point!

gquintard: actually, no that's unnecessary since the string `*guillaume*` is a match for the globbing pattern `*guillaume*`. If globbing failed, there's nothing more we can do. Or did I miss something?

nigoroll: I see your point @gquintard , my example was bad.
I guess `[ab].vcl` is a better one, this should, upon failure to expand, attempt to open a file by that literal name.

gquintard: right, `*` isn't the only possible pattern. I'll adjust

bsdphk:
Bugwash was not too happy about having glob's in paths, that's a security think.

No conclusion either way on glob's in include.

Dridi:
We finally discussed the matter of VMODs test suites with @nigoroll and @bsdphk and we have an outcome:

- libvcc shouldn't duplicate `vfil_searchpath`'s job
- existing VCL shouldn't suddenly resolve globs
  - new syntax: `include glob "<literal>";`
- new `VSB_QUOTE_GLOB` quoting
- new `VFIL_searchpathglob()` function
  - quotes the path component, concatenates the subject string as is
  - iterates over all entries for the first matching path component

nigoroll:
@gquintard is this still hot on your end?

Dridi:
I think this one is awaiting me, I will revisit this.

bsdphk:
Note:  Maybe the output of glob should be sorted for deterministic behaviour ?

nigoroll:
> sorted for deterministic behaviour ?
plus this is what shells do

Dridi:
@gquintard can you give a try to trunk before next bugwash? See https://github.com/varnishcache/varnish-cache/commit/195929e4aa29f46c49038b3982b842a4d647a1d5 for the details.

gquintard:
looks good to me, I've converted the vtc from this PR to align with `master`, I'll be off next week, so feel free to push it yourself if you think it makes sense:

```
varnishtest "Test globbing includes in VCL"

server s1 {
	rxreq
	txresp
} -start

varnish v1 -errvcl "glob pattern matched no files." {
	vcl 4.0;

	include +glob "${tmpdir}/*.vcl";
}

shell {
	echo 'sub vcl_deliver { set resp.http.foo = "foo"; }' > ${tmpdir}/sub_foo.vcl
	echo 'sub vcl_deliver { set resp.http.bar = "bar"; }' > ${tmpdir}/sub_bar.vcl
	echo 'vcl 4.0; backend default { .host = "0:0"; } include +glob "./sub_*.vcl";' > ${tmpdir}/top.vcl
}

varnish v1 -vcl+backend {
include +glob "${tmpdir}/sub_*.vcl";
} -start

client c1 {
	txreq
	rxresp
	expect resp.http.foo == foo
	expect resp.http.bar == bar
} -run

varnish v1 -errvcl {needs absolute filename of including file.} {
	include +glob "./sub_*.vcl";
	backend default none;
}

varnish v1 -cliok "vcl.load foo ${tmpdir}/top.vcl"
varnish v1 -cliok "vcl.use foo"

client c1 {
	txreq
	rxresp
	expect resp.http.foo == foo
	expect resp.http.bar == bar
} -run
```

rezan:
> varnish v1 -errvcl "glob pattern matched no files." {

Should this be an error? I can envision use-cases where we have a VCL based plugin or extension directory and it defaults to empty, but we want Varnish to continue loading.

Dridi:
FWIW now that we have a `+glob` flag you could imagine more flags, like one to treat not matching file as including nothing.

bsdphk:
I have rescued the VTC into `c00053`.

Dridi:
Reopening to get a chance to discuss a potential flag to ignore missing files. Including from a drop-in directory is a common pattern.

bsdphk:
That is a new feature request, and the most efficient way would be to open a new PR which implements it.

HOSTED-POWER:
In which version is this available? I tried 6.6.x inside default.vcl:
```

include +glob "/etc/varnish/conf.d/*.vcl";
--------#---------------------------------

Running VCC-compiler failed, exited with 2
VCL compilation failed
```

gquintard:
it was merged in 195929e4aa29f46c49038b3982b842a4d647a1d5 last april, so after `6.6` was released, so it'll be in the next version

If you need it right away, it's in the weeklies: https://packagecloud.io/varnishcache/varnish-weekly

HOSTED-POWER:
Ok I think this is a long time missing feature !! :)

I assume it loads the file in order? So 001_xx will be loaded first, then 002_xx etc , right?

Any ETA on a new release version of Varnish with this included? :)

gquintard:
yes, files are sorted to help with reproducibility.

regarding release schedule: https://www.mail-archive.com/varnish-misc@varnish-cache.org/msg08486.html

HOSTED-POWER:
Ok great, that is pretty soon, thanks a lot for all prompt feedback :D

PR #3195, opened 2020-01-20
Add 'arch' to the jobs, so that it tests both amd64 and arm64 on Linux

martin-g:
Hello,

With this Pull Request I'd like to request to add arm64 architecture to the build matrix at TravisCI.

ARM64 is used more and more for servers too and it would be good if Varnish Cache is tested regularly.
I've ran the build and tests both on one of our arm64 machines and at [TravisCI](https://travis-ci.org/martin-g/varnish-cache/builds/639491964) and at the moment both are fine!

amd64 is the default arch at TravisCI so it is preserved!

At https://github.com/varnishcache/varnish-cache/issues?utf8=%E2%9C%93&q=is%3Aissue+arm I've seen that there were some problems on ARM based architectures in the past. Hopefully with this change the ARM specific problems will stay away in the future too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ jobs:
       stage: test
       os: linux
       dist: xenial
+      arch: amd64
       compiler: clang
       addons:
         apt:
@@ -19,6 +20,7 @@ jobs:
             - python3-docutils
             - python3-sphinx
             - libunwind-dev
+            - libpcre3-dev
       before_script:
         - ./autogen.sh
         - ./configure --with-unwind
@@ -30,7 +32,15 @@ jobs:
             make -j3 check VERBOSE=1
           fi
     - <<: *test-linux
+      arch: arm64
+    - <<: *test-linux
       compiler: gcc
+      before_script:
+        - ./autogen.sh
+        - ./configure
+    - <<: *test-linux
+      compiler: gcc
+      arch: arm64
       before_script:
         - ./autogen.sh
         - ./configure


gquintard:
Hi,

thank you for looking into this. Is there a reason for not just going with

```
arch:
  - amd64
  - arm64
```

or something similar?

martin-g:
This was my initial version of the change but it led to 2 more build entries in TravisCI UI which were failing because `./autoconf` was not executed for them.
Here is one of those failing builds: https://travis-ci.org/martin-g/varnish-cache/builds/639482129

Dridi:
I don't understand why run `./autogen.sh && ./configure` instead of `./autogen.des` in the first place, maybe @fgsch will know better.

martin-g:
`autogen.des` is being used for CircleCI: https://github.com/varnishcache/varnish-cache/blob/e5e545f9fe14b4bfd4003c26403d80645c73385a/.circleci/config.yml#L98

nigoroll:
just FYI @martin-g we also got arm boxes in [vtest](http://varnish-cache.org/vtest/). But this does not make the addition to travis any less useful.

martin-g:
Thank, @nigoroll ! I didn't know about this!

martin-g:
I've just tried to run the vtests on one of our ARM machine :

```
make varnishtest
cc \
	-O2 -s -Wall -Werror \
	-o varnishtest \
	-DVTEST_WITH_VTC_VARNISH \
	-DVTEST_WITH_VTC_LOGEXPECT \
	-Isrc -Ilib -I/usr/local/include -pthread \
	-I/home/ubuntu/git/varnish/varnish-cache/include \
	src/*.c lib/*.c \
	-L/usr/local/lib -lm -lpcre -lz \
	-L/home/ubuntu/git/varnish/varnish-cache/lib/libvarnishapi/.libs \
	-Wl,--rpath,/home/ubuntu/git/varnish/varnish-cache/lib/libvarnishapi/.libs \
	-lvarnishapi
src/vtc_main.c: In function ‘cleaner_setup’:
src/vtc_main.c:232:3: error: ignoring return value of ‘nice’, declared with attribute warn_unused_result [-Werror=unused-result]
   (void)nice(1);  /* Not important */
   ^~~~~~~~~~~~~
cc1: all warnings being treated as errors
Makefile:51: recipe for target 'varnishtest' failed
make: *** [varnishtest] Error 1
```

Do I need to do something in addition ?
`/home/ubuntu/git/varnish/varnish-cache` is a fresh clone of this repo.

fgsch:
> I don't understand why run `./autogen.sh && ./configure` instead of `./autogen.des` in the first place, maybe @fgsch will know better.

It does more things than simply autogen.sh && ./configure (and it might have more in the future).
I'd rather be explicit about how and what we pass.

fgsch:
Looks good to me. Non-blocker question: do you really to add libpcre3-dev?

martin-g:
Without libpcre3-dev it failed: https://travis-ci.org/martin-g/varnish-cache/jobs/639166575

fgsch:
> Without libpcre3-dev it failed: https://travis-ci.org/martin-g/varnish-cache/jobs/639166575

Thanks, that's what I thought; the images have different packages installed by default.

Dridi:
> It does more things than simply autogen.sh && ./configure (and it might have more in the future).
I'd rather be explicit about how and what we pass.

If you wish to be explicit you can always pass `configure` options to the `autogen.des` script, and currently with a plain `./configure` we don't have coverage for the persistent storage or developer warnings on Travis CI.

fgsch:
> If you wish to be explicit you can always pass `configure` options to the `autogen.des` script, and currently with a plain `./configure` we don't have coverage for the persistent storage or developer warnings on Travis CI.

I disagree. You might override some of the options set by autogen.des, but you won't be explicit unless you copy all the arguments that this script uses. 
Also by not using autogen.des you don't have to worry about future changes to this script.

Persistent storage is enabled for the sanitizers job. Not enabling in the rest was by design.

nigoroll:
the addition of the arm hosts already improved my PRs :)

Issue #3194, opened 2020-01-20
WS_Reset() and overflown workspaces


nigoroll:
noticed here: https://github.com/varnishcache/varnish-cache/commit/b12af6813d3a3ab9a6579c583d17fdd29ac1adfb

We use workspace overflows to signal to bail out for example after a failing `VRT_SetHdr()`. This is a guarantee that if some serious issue occurred during processing, we rather send an error downstream than an incomplete response or the result of incomplete processing.
We use the `WS_Snapshot() ...  WS_Reset()` pattern as some kind of second order workspace allocation where the called code itself uses `WS_Reserve()`.
This combination is problematic because `WS_Reset()` calls `ws_ClearOverflow(ws)`, potentially clearing the overflow bit from a previous relevant failure.

I think we should simply forbid `WS_Snapshot()` on overflown workspaces, mandating prior calls of `WS_Overflowed()` for the above mentioned pattern.

Feedback welcome

nigoroll:
this patch to `c00071.vtc` demos the issue. Notice that the vtc already failed in the wrong place (while pushing processors rather than in the `regsub()`), which remained unnoticed.

```
diff --git a/bin/varnishtest/tests/c00071.vtc b/bin/varnishtest/tests/c00071.vtc
index 7e3c156ee..d4c17de1b 100644
--- a/bin/varnishtest/tests/c00071.vtc
+++ b/bin/varnishtest/tests/c00071.vtc
@@ -13,6 +13,7 @@ server s1 {
 
 varnish v1 -vcl+backend {
        import vtc;
+       import std;
        sub vcl_deliver {
                vtc.workspace_alloc(client, -192);
 
@@ -21,7 +22,8 @@ varnish v1 -vcl+backend {
                        vtc.workspace_alloc(client, -10);
                }
                else if (req.url ~ "/baz") {
-                       set resp.http.x-foo = regsub(req.url, "baz", "baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaz");
+                       set resp.http.x-foo = regsub(req.url, "baz", "baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaz");
+                       std.log("dummy");
                }
[...]

PR #3196, opened 2020-01-20
Add a strands interface to VSL and use it for std.log

nigoroll:
Since 8baf4a690fdb699923dd8438d0088f34fcb24365 we lost information about an overflowed workspace by calling `std.log()`, which resulted from use of the workspace for construction of a contiguous string from the argument constituents.

Since then, we have changed the interface to STRANDS, but this issue remained.

We now solve the case for real by pushing the string concatenation down to VSL: New versions of `VSL` and `VSLb` (coded by example of `VSLv()` and `VSLbt()`) take a strands argument and create a log record without additional copy overhead.

These solve #3194 for `std.log()`, make logging more efficient and, in particular, allow use of `std.log()` in low workspace conditions (because they do not require any).

Also improve test coverage for `std.log()`

Ref #3194


nigoroll:
@bsdphk approved during bugwash

PR #3197, opened 2020-01-20
address most other overflow vs. WS_Reset() cases

nigoroll:
Reviewers please read #3194 first

We add checks for cases where a `WS_Reset()` could reset the overflowed workspace marker to ensure that the workspace is not already overflowed before using it.

See also #3196 for `std.log()`

This concludes the fix for #3194 together with the above (except for one h2 case, maybe)


nigoroll:
bugwash: prep alternative patch to mark for `WS_Reset()` not to clear the overflow bit

nigoroll:
closing in favor of #3202 which is the better idea, after all

PR #3198, opened 2020-01-22
reduce vcl_mtx contention with an implicing VCL_Rel() in vcl_get()

nigoroll:
Both callers of `vcl_get()`, `VPI_vcl_select()` and `VCL_Refresh()`, potentially have to unref another vcl first.
    
For `VCL_Refresh()`, this happens when the active vcl changes and the previously active vcl is cached in `(struct worker).vcl`, which can be assumed to always be the case on a busy system.

For `VPI_vcl_select()`, this happens when switching vcls at `esi_level > 0`.

To summarize, this patch will primarily reduce contention on the `vcl_mtx` after a `vcl.use` CLI command on a busy system and for `return(use(vcl))` with ESI.

I added a trivial `AZ(*vcc)` patch and ran the checks before the actual patch to ensure that really the previous use of `vcl_get()` was only ever with `vcc` pointing to NULL (that is trivially clear from the caller code, but anyway...).

### DIFF
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -259,10 +259,21 @@ VCL_Panic(struct vsb *vsb, const char *nm, const struct vcl *vcl)
 void
 vcl_get(struct vcl **vcc, struct vcl *vcl)
 {
+	struct vcl *old;
+
 	AN(vcc);
-	AZ(*vcc);
+
+	old = *vcc;
+	*vcc = NULL;
+
+	CHECK_OBJ_ORNULL(old, VCL_MAGIC);
 
 	Lck_Lock(&vcl_mtx);
+	if (old != NULL) {
+		assert(old->busy > 0);
+		old->busy--;
+	}
+
 	if (vcl == NULL)
 		vcl = vcl_active; /* Sample vcl_active under lock to avoid
 				   * race */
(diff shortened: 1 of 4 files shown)


bin/varnishd/cache/cache_vcl.c
```diff
@@ -257,9 +257,23 @@ VCL_Panic(struct vsb *vsb, const char *nm, const struct vcl *vcl)
...
 {
+	struct vcl *old;
+
 	AN(vcc);
 
+	old = *vcc;
+	*vcc = NULL;
+
+	if (old != NULL)
+		CHECK_OBJ(old, VCL_MAGIC);
```
fgsch: use CHECK_OBJORNULL()

nigoroll: right, thank you. The `if`-block used to contain more, this was a residue of a previous incarnation

nigoroll:
I have just added a [third commit](https://github.com/varnishcache/varnish-cache/pull/3198/commits/9ed78d85512e1a8903535bff534334a22050e05f) to this PR which I only now realized was possible:
We can move assertions on the label integrity outside the critical region.

nigoroll:
Apologies for the back-and-forth, I _was_ right initially and wrong now. The additional commit is removed again because `active_vcl` can be a label.

nigoroll:
bugwash:
* remove "keep in sync" comments
* rename `vcl_get()` -> `VCL_Update()`

nigoroll:
FTR, an unrelated patch momentarily snug into this PR. removed again with a second force push after merge

PR #3201, opened 2020-01-27
respect `send_timeout` for "dripping" http1 writes

### LINKED ISSUE #3189: http/2 send_timeout triggered but not in http/1
Using http/1 and `beresp.do_stream = true`, we can take as long as we want to send data,
but after switching to http/2, the send_timeout is triggered and connection closes.

## Expected Behavior
Either connection should close using http/1 to honor send_timeout configuration, or it should not close in http/2.

## Current Behavior
Connection is closed after send_timeout using http/2, but not http/1, both with `beresp.do_stream = true`, `beresp.do_gzip = false` and `beresp.ttl = 0s;`

## Possible Solution
Improve docs to explain http/2 behaviour
Ability to configure send_timeout in VCL will solve problem (#2983)

## Steps to Reproduce
1. have a backend which send data regularly, taking a high time, for exemple, we sent a dot every second for 700seconds (to reach the default 600s timeout)
2. use http/2 to call the backend. we used haproxy in tcp mode for ssl termination
3. wait until the timeout, connection get closed.
4. using http/1, connection is not closed

## Context
We're sending a file to client, generating it on the fly.  Right now we increased timeout to 4 hours so it's working, but we might be keeping idle connections for naught.
I must stress out that connection is not closed using http/1, so we're sending enough data to keep connection alive

## Your Environment
* Version used: 6.3.1
* Operating System and version: docker image
* Source of binary packages used (if any) "official" docker image 6.3.1
Running behind a haproxy 7.5 in proxy mode

nigoroll:
problem
---

Previously, we only checked `v1l->deadline` (which gets initialized from the `send_timeout` session attribute or parameter) only for short writes, so for successful "dripping" http1 writes (streaming from a backend busy object with small chunks), we did not respect the timeout.

solution
---

This patch restructures `V1L_Flush()` to always check the deadline before a write by turning a `while() { ... }` into a `do { ... } while` with the same continuation criteria: `while (i != v1l->liov)` is turned into `if (i == v1l->liov) break;` and `while (i > 0 || errno == EWOULDBLOCK)` is kept to retry short writes.

This also reduces the `writev()` call sites to one.

Fixes #3189

(potentially breaking) changes
---

After (if) the patch is accepted, we should put a prominent warning into the release docs to re-check `send_timeout`.

This is particularly important as `send_timeout` is now enforced also for sending request bodies to the backend (see s00011.vtc)

### DIFF
--- a/bin/varnishd/http1/cache_http1_fetch.c
+++ b/bin/varnishd/http1/cache_http1_fetch.c
@@ -97,9 +97,8 @@ V1F_SendReq(struct worker *wrk, struct busyobj *bo, uint64_t *ctr_hdrbytes,
 	}
 
 	VTCP_blocking(*htc->rfd);	/* XXX: we should timeout instead */
-	/* XXX: what is the right timeout ? Isn't send_timeout client side? */
-	V1L_Open(wrk, wrk->aws, htc->rfd, bo->vsl,
-	    bo->t_prev + cache_param->send_timeout, 0);
+	/* XXX: need a send_timeout for the backend side */
+	V1L_Open(wrk, wrk->aws, htc->rfd, bo->vsl, nan(""), 0);
 	hdrbytes = HTTP1_Write(wrk, hp, HTTP1_Req);
 
 	/* Deal with any message-body the request might (still) have */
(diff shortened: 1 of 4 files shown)


nigoroll:
fixed a missed initialization glitch. force pushed

nigoroll:
bugwash:
* @bsdphk ok
* wait for @Dridi review tomorrow, then merge for 6.4

Dridi:
I requested changes and approved offline, thanks!

PR #3202, opened 2020-01-27
An overflowed workspace must remain overflowed after WS_Reset()

### LINKED ISSUE #3194: WS_Reset() and overflown workspaces

nigoroll:
This is an alternative approach to #3197 on how to fix #3194 as discussed during today's bugwash:
   
`WS_Snapshot()` now returns a magic value which gets recognized by `WS_Reset()` to ensure that the overflowed marker gets preserved or even restored. In this case, the workspace does not actually get reset.

background:

We use workspace overflows to signal to bail out for example after a failing `VRT_SetHdr()`. This is a guarantee that if some serious issue occurred during processing, we rather send an error downstream than an incomplete response or the result of incomplete processing.

We use the `WS_Snapshot() ...  WS_Reset()` pattern as some kind of second order workspace allocation where the called code itself uses `WS_Reserve()`.

With this usage pattern, `WS_Reset()` called `ws_ClearOverflow(ws)`, potentially clearing the overflow bit from a previous relevant failure.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -110,8 +110,8 @@ void Bereq_Rollback(struct busyobj *bo)
 	VCL_TaskLeave(bo->privs);
 	VCL_TaskEnter(bo->privs);
 	HTTP_Clone(bo->bereq, bo->bereq0);
-	WS_Reset(bo->bereq->ws, bo->ws_bo);
-	WS_Reset(bo->ws, bo->ws_bo);
+	WS_Rollback(bo->bereq->ws, bo->ws_bo);
+	WS_Rollback(bo->ws, bo->ws_bo);
 }
 
 /*--------------------------------------------------------------------
--- a/bin/varnishd/cache/cache_req.c
+++ b/bin/varnishd/cache/cache_req.c
@@ -200,7 +200,7 @@ Req_Rollback(struct req *req)
 	HTTP_Clone(req->http, req->http0);
 	if (WS_Overflowed(req->ws))
 		req->wrk->stats->ws_client_overflow++;
-	WS_Reset(req->ws, req->ws_req);
+	WS_Rollback(req->ws, req->ws_req);
 }
 
 /*----------------------------------------------------------------------
@@ -251,7 +251,7 @@ Req_Cleanup(struct sess *sp, struct worker *wrk, struct req *req)
 	if (WS_Overflowed(req->ws))
 		wrk->stats->ws_client_overflow++;
 
-	WS_Reset(req->ws, 0);
+	WS_Rollback(req->ws, 0);
 }
 
 /*----------------------------------------------------------------------
--- a/bin/varnishd/cache/cache_varnishd.h
+++ b/bin/varnishd/cache/cache_varnishd.h
@@ -463,6 +463,9 @@ void VMOD_Panic(struct vsb *);
 /* cache_wrk.c */
 void WRK_Init(void);
 
+/* cache_ws.c */
+void WS_Rollback(struct ws *, uintptr_t);
+
 /* http1/cache_http1_pipe.c */
 void V1P_Init(void);
 
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -163,7 +163,7 @@ VCL_Rel_CliCtx(struct vrt_ctx **ctx)
 	if (ctx_cli.vsl)
 		VSL_Flush(ctx_cli.vsl, 0);
 	WS_Assert(ctx_cli.ws);
-	WS_Reset(&ws_cli, ws_snapshot_cli);
+	WS_Rollback(&ws_cli, ws_snapshot_cli);
 	INIT_OBJ(*ctx, VRT_CTX_MAGIC);
 	*ctx = NULL;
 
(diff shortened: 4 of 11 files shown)


nigoroll:
@Dridi you are right, `ws_ClearOverflow()` is private to `cache_ws.c` and `WS_Reset()` is the only caller.

I will be happy to add a test case once we have decided on how to continue.

Dridi:
I'm fine with this approach, FWIW, with somewhat transparent error handling. The magic `snap_overflowed` cookie is not exposed, but we already have have `WS_Overflowed()` so I really don't see a problem.

I would approve this PR if that's the API design that reaches consensus (modulo the suggestions from my previous comment).

nigoroll:
> I would approve this PR if that's the API design that reaches consensus

@Dridi I don't think this is compatible with our modus operandi. I need both your and phks approval, so you can either vote +1, -1 or go with whatever phk votes.

nigoroll:
> I would approve this PR if that's the API design that reaches consensus

@Dridi I don't think this is compatible with our modus operandi. I need both your and phks approval, so you can either vote +1, -1 or go with whatever phk votes.

nigoroll:
sorry @Dridi I thought I had integrated your feedback except for the improved vtc, which I would like to do after merge. What am I missing?

nigoroll:
@Dridi nevermind, I'll write the vtc

nigoroll:
vtc done.
Anything else @Dridi ?

Dridi:
I think it's all good, thanks!

I'm wondering whether we could land here with an overflowed workspace:

https://github.com/varnishcache/varnish-cache/blob/f75bce17edbda8c30b083390210463dd0a83fbc9/bin/varnishd/cache/cache_vrt_vcl.c#L448-L463

I'd need to spend extra time reviewing existing usage of `WS_Snapshot`. For example, can we get in a situation where we have a nested snapshot in VMOD space that could inadvertently clear an overflow?

- `WS_Snapshot(ws) = ok`
  - [...]
  - `WS_Snapshot(ws) = overflow`
  - [...]
  - `WS_Reset(ws, overflow)`
  - [...]
- `WS_Reset(ws, ok)` (accidental `ws_ClearOverflow`)

That's not something we could solve with the current API, unless we decide that only `WS_Reset(ws, 0)` clears the workspace and that `WS_Snapshot(ws)` always returns the front pointer.

nigoroll:
> I think it's all good, thanks!

wouls you adjust your review status?

> I'm wondering whether we could land here with an overflowed workspace:
> 
> https://github.com/varnishcache/varnish-cache/blob/f75bce17edbda8c30b083390210463dd0a83fbc9/bin/varnishd/cache/cache_vrt_vcl.c#L448-L463

Yes. We do not abort on an overflown aws at the moment and I guess we should. But that is largely orthogonal to this ticket.

> I'd need to spend extra time reviewing existing usage of `WS_Snapshot`. For example, can we get in a situation where we have a nested snapshot in VMOD space that could inadvertently clear an overflow?

yes. And there are cases where we explicitly want to achieve this, for example in `Req_Rollback()` and when we clean up requests for reuse.

The good thing about this patch is simplicity, but I agree that actually it would be more correct follow the approach in #3194 

Or have more checks for overflows, not only at the end of the request.

nigoroll:
from discussion with @Dridi and @bsdphk : We should reserve `WS_Reset(x, 0)` to varnishd code and make that the only way to reset the overflowed marker

bsdphk:
There's an if() in h2_deliver() which should be removed once this is fixed.

nigoroll:
> from discussion with @Dridi and @bsdphk : We should reserve `WS_Reset(x, 0)` to varnishd code by making it a separate function and make that the only way to reset the overflowed marker
> std.rollback will continue to clear the overflow

this is a much better idea, which I have implemented now:

We now avoid any other unintended clears of the overflow bit splitting two functions:

* `WS_Rollback()` is now what `WS_Reset()` used to be: It clears overflows and accepts the zero cookie for a reset-to-start

  It is only intended for use within `varnishd` and is thus declared in `cache_varnishd.h`

* `WS_Reset()` does not touch the overflow bit any longer, ensuring that a once-overflowed workspace stays overflowed

With this in place, the magic `snap_overflowed` cookie would not be needed any more, but it still serves two good purposes:

- better debugging and
- as a safety measure against passing a cookie from an already overflowed workspace to `WS_Rollback()`

nigoroll:
OK'ed by @bsdphk during bugwas.

@Dridi you asked for a final post-squash review, please have a look ASAP so we can merge for 6.4

PR #3204, opened 2020-01-30
[cci] centos:8 support

gquintard:
@slimhazard , here you go :-)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,38 +7,13 @@ parameters:
   pkg-commit:
     type: string
     default: "weekly"
-commands:
-  debian_install_build_deps:
-    description: Install build dependencies
+jobs:
+  dist:
+    docker:
+      - image: centos:7
     steps:
       - run:
-          name: Install build dependencies
-          command: |
-            export DEBIAN_FRONTEND=noninteractive
-            export DEBCONF_NONINTERACTIVE_SEEN=true
-            apt-get update
-            apt-get install -y \
-                autoconf \
-                automake \
-                build-essential \
-                ca-certificates \
-                git \
-                graphviz \
-                libconfig-dev \
-                libedit-dev \
-                libjemalloc-dev \
-                libncurses-dev \
-                libpcre3-dev \
-                libtool \
-                libunwind-dev \
-                pkg-config \
-                python3-sphinx \
-                sudo
-  centos_install_build_deps:
-    description: Install build dependencies
-    steps:
-      - run:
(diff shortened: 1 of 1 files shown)


nigoroll:
isn't this something which can be merged straight away?

.circleci/config.yml
```diff
@@ -146,7 +92,70 @@ jobs:
...
+                yum install -y \
+                    automake \
+                    jemalloc-devel \
+                    libconfig-devel \
+                    libcurl-devel \
+                    libedit-devel \
+                    libtool \
+                    libunwind-devel \
+                    make \
+                    openssh-clients \
```
nigoroll: I think we do not need
* libconfig
* libcurl
* openssh-client

slimhazard:
I'm assuming that in the centos8 build, installing epel-release will result in the EPEL repo pointing to el8 packages, in particular jemalloc 5.2.1. And in the end ldd varnishd will show libjemalloc.so.2 (rather than libjemalloc.so.1 as currently for el7).

@gquintard do you happen to know if that's correct?

Asking because currently I'm working on getting jemalloc 5.2.1 to install on centos7, and am hoping for a beautiful centos8 future in which that all Just Works.

.circleci/config.yml
gquintard: indeed

gquintard:
@slimhazard, correct:

```
# rpm -qpR rpms/x86_64/varnish-20200203.weekly-0.0.el8.x86_64.rpm | grep jemalloc
libjemalloc.so.2()(64bit)
```

```
# ldd /usr/sbin/varnishd | grep jemalloc
	libjemalloc.so.2 => /lib64/libjemalloc.so.2 (0x00007f5fb7ff1000)
```

gquintard:
looks like this got in before I could clean the PR, closing

nigoroll:
@gquintard I do not think anyone has merged your PR.
I just updated the docs in f5dcaeabba2525c5609b66001f4743ba2fbbf7f8 / 0e192d6fa1691ebe9a4a3dcb600579cd43a0c1ba while I was working on a CentOS 8 installation anyway

gquintard:
https://github.com/varnishcache/varnish-cache/commit/30453b94901d3ccaa2230dc0a9a477fbcb7c92af appears to be in master though

nigoroll:
oh right. Wasn't that you?

gquintard:
I don't think so, but that's a possibility

PR #3206, opened 2020-01-31
Escape regex special characters

Dridi:
The goal is to allow building a regular expression that is partly made of user input that is not regex material, like for example a value taken from a req or beresp header.

Conceptualy, in order to create a regular expression like this:

    "<maybe regex material>" + VRT_re_quote("<exact match>") + "<maybe more regex stuff>"

The patch series is incomplete, there is no test coverage. I know @rezan would like to expose this in vmod_std but without functions to deal with dynamic regular expressions I'm not sure how helpful that would be. So I'd be more inclined to add coverage via vmod_debug. So until I know whether this should be exposed in vmod_std, I'm deferring the test cases.

See the individual commits for the implementation details, it is unfortunate that PCRE doesn't offer that kind of feature. The function names are inspired by PHP's `preg_quote()` function.


bsdphk:
If we had had vmods back then, we would not have had regexps in the VCL language, but in a vmod_regexp instead.

I think we should still seriously consider concentrating regexp-improvements in an in-tree regexp-vmod.

nigoroll:
Why have all these complications when there is [`\Q` and `\E`](https://pcre.org/original/doc/html/pcrepattern.html#SEC5) ?

Dridi:
@nigoroll interesting, but you still need a way to quote the subject string in case it contains `\E` (for example adding a `\Q` right after), which is arguably simpler.

nigoroll:
@Dridi correct, if you can't trust it.

Dridi:
> @Dridi correct, if you can't trust it.

Whether it comes from an expression or a string literal, I wouldn't trust user input for this purpose, which is the goal of this VRE/VRT addition.

nigoroll:
@Dridi to clarify:

I think for the literal bit the following should work (VCL):
```
"\Q" + regsuball("lit\Eral", "\E", "\E\\E\Q") + "\E"
```

For clarity, this could be wrapped in a vmod function.

Dridi:
> For clarity, this could be wrapped in a vmod function.

On this topic, see my comment from the pull request description:

> I know @rezan would like to expose this in vmod_std but without functions to deal with dynamic regular expressions I'm not sure how helpful that would be.

I'd rather make this available via VRE, because it is our thin abstraction layer above PCRE so using a PCRE-specific `\Q...\E` enclosure should be handled there.

nigoroll:
I did a minimal amount of digging:

pcre actually implements '\Q...\E' in the compile code: https://github.com/svn2github/pcre/blob/10e4adbfe9aab5869079fba5edc0ce56fa1b078e/pcre_compile.c#L4654

Yet quoting all meta characters even goes back to the good old perl times, for example:
* https://perldoc.perl.org/functions/quotemeta.html
* https://github.com/svn2github/pcre/blob/10e4adbfe9aab5869079fba5edc0ce56fa1b078e/pcrecpp.cc#L526

So while I still think using '\Q...\E' with `\E` escaping would be the simplest to implement, most efficient and hopefully most robust option as long as we ware using PCRE, I surrender my opposition to quoting, but I still fear we could get it wrong.

nigoroll:
FTR preserving a nice idea by @Dridi - the following version of https://github.com/varnishcache/varnish-cache/pull/3206#issuecomment-581416912 is simpler and while I initially thought it was confusing, it actually looks safer because the `\` from the literal `\E` is part of the quoted region:

```
"\Q" + regsuball("lit\Eral", "\E", "\\EE\Q") + "\E"
```

nigoroll:
pow-wow: to be replaced with a new PR by @Dridi

Dridi:
#3345

PR #3209, opened 2020-02-05
varnishstat json schema change

### LINKED ISSUE #3186: RFC - "varnishstat -j"'s timestamp is a pain for JSON parsers
The `.timestamp` field is a string in the `varnishstat -j` output, whereas all the others are objects that all follow the same structure.
This makes writing parsers a pain because `timestamp` always needs to be special-cased for little to no extra value.

I'd like to see it gone for the next major, and if people are ok with that, there's the question of the replacement:
1. none: sit on this ticket and once we know that the next release will be `7.0`, just retire `.timestamp`.
2. offer a `.time` attribute that follows the common structure

I'm personally in favor of option 1, once the data has been de-serialized, adding your own timestamp is super easy. Also, `varnishstat -1` doesn't have a `timestamp` field.

gquintard:
fixes #3186

This changes the format in only two ways:
- counters go into their own `counters` sub-object
- new `version` field, set to 1 (since most lenient parser will say it's 0 if the field is absent)

However, it will break parsers downstream, it maybe an issue if the next release isn't a major one. I'm happy to communicate with the downstream projects I know/can reach to ease the migration.

### DIFF
--- a/bin/varnishstat/varnishstat.c
+++ b/bin/varnishstat/varnishstat.c
@@ -134,12 +134,10 @@ do_json(struct vsm *vsm, struct vsc *vsc)
 	int jp;
 
 	jp = 1;
-
-	printf("{\n");
 	now = time(NULL);
 
 	(void)strftime(time_stamp, 20, "%Y-%m-%dT%H:%M:%S", localtime(&now));
-	printf("  \"timestamp\": \"%s\",\n", time_stamp);
+	printf("{\n  \"timestamp\": \"%s\",\n", time_stamp);
 	(void)VSC_Iter(vsc, vsm, do_json_cb, &jp);
 	printf("\n}\n");
 }
--- a/bin/varnishstat/varnishstat.c
+++ b/bin/varnishstat/varnishstat.c
@@ -114,15 +114,13 @@ do_json_cb(void *priv, const struct VSC_point * const pt)
 	else
 		printf(",\n");
 
-	printf("  \"");
-	/* build the JSON key name.  */
-	printf("%s\": {\n", pt->name);
-	printf("    \"description\": \"%s\",\n", pt->sdesc);
+	printf("    \"%s\": {\n", pt->name);
+	printf("      \"description\": \"%s\",\n", pt->sdesc);
 
-	printf("    \"flag\": \"%c\", ", pt->semantics);
+	printf("      \"flag\": \"%c\", ", pt->semantics);
 	printf("\"format\": \"%c\",\n", pt->format);
-	printf("    \"value\": %ju", (uintmax_t)val);
-	printf("\n  }");
+	printf("      \"value\": %ju", (uintmax_t)val);
+	printf("\n    }");
 	return (0);
 }
 
@@ -137,9 +135,12 @@ do_json(struct vsm *vsm, struct vsc *vsc)
 	now = time(NULL);
 
 	(void)strftime(time_stamp, 20, "%Y-%m-%dT%H:%M:%S", localtime(&now));
-	printf("{\n  \"timestamp\": \"%s\",\n", time_stamp);
+	printf("{\n"
+	    "  \"version\": 1,\n"
+	    "  \"timestamp\": \"%s\",\n"
+	    "  \"counters\": {\n", time_stamp);
 	(void)VSC_Iter(vsc, vsm, do_json_cb, &jp);
-	printf("\n}\n");
+	printf("\n  }\n}\n");
 }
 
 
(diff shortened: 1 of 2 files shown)


nigoroll:
have not tried the patch, but as simple as-is, LGTM.
Should we improve vtc coverage?

bin/varnishstat/varnishstat.c
```diff
@@ -136,9 +134,12 @@ do_json(struct vsm *vsm, struct vsc *vsc)
 	now = time(NULL);
 
 	(void)strftime(time_stamp, 20, "%Y-%m-%dT%H:%M:%S", localtime(&now));
-	printf("{\n  \"timestamp\": \"%s\",\n", time_stamp);
+	printf("{\n"
+	    "  \"version\": 1,\n"
```
Dridi: I'd be tempted to start right away with version 2.

bin/varnishstat/varnishstat.c
```diff
@@ -113,15 +113,13 @@ do_json_cb(void *priv, const struct VSC_point * const pt)
...
+	printf("    \"%s\": {\n", pt->name);
+	printf("      \"description\": \"%s\",\n", pt->sdesc);
 
-	printf("    \"flag\": \"%c\", ", pt->semantics);
+	printf("      \"flag\": \"%c\", ", pt->semantics);
 	printf("\"format\": \"%c\",\n", pt->format);
-	printf("    \"value\": %ju", (uintmax_t)val);
-	printf("\n  }");
+	printf("      \"value\": %ju", (uintmax_t)val);
+	printf("\n    }");
```
Dridi: We should probably change this sequence of `printf` statements into a single one like you did line 137.

bin/varnishstat/varnishstat.c
gquintard: if you want, but as said, most parser will consider that a missing `version` field means `version=0`, so I found that relatively pleasing.

But we are in bikeshedding territory, so I don't care tooo much

bin/varnishstat/varnishstat.c
gquintard: not as easy to do since there are arguments to format?

gquintard:
@nigoroll I can do that. Is that what the "action required" label is for?

nigoroll:
@gquintard no, because this is a breaking change, we need to decide if we go for a major version or merge after release

nigoroll:
bugwash:
* merge this after the march release soak in
* decide on the version modus operandi:
  * support multiple versions to switch with `-J%d` argument (0 default = latest)
  * or support only latest and expect clients to parse multiple

nigoroll:
maintainer pow-wow decision:
* support only latest and expect clients to parse multiple

@gquintard anything left to be done with that regard?

gquintard:
@nigoroll , only one thing: a vtvc. I'll get it ready for the next bugwash

gquintard:
@nigoroll , vtc in, I think we are ok

bin/varnishstat/varnishstat.c
Dridi: Then let's make it obvious to actual end users that this is a new version and bump it to 2.

nigoroll: My vote goes to zero-based: The current format is version NIL ≍ 0, so the next version is 1.

bin/varnishstat/varnishstat.c
nigoroll: @Dridi can we consider this feedback addressed?

Dridi: I will do that myself once merged.

bin/varnishstat/varnishstat.c
nigoroll: bugwash decision: @gquintard you get to break the tie here. Please decide on the initial version number, then the PR is ready for merge in april.

bsdphk:
Nobody in the bugwash seems to care.

@gquintard to do as he pleases.

nigoroll:
Just to be sure this does not get overlooked: This is to be merged April 1st at the earliest, after the post-release freeze.

bin/varnishstat/varnishstat.c
gquintard: My opinion has not changed, and is reflected by the current code. I don't care enough veto a version 2 though, so @Dridi if you want to update the PR, please go ahead

nigoroll: d-oh @gquintard I appreciate your kindness, but we tried to break the tie... Now If you do not make a decision, aren't we back where we started?

gquintard: @nigoroll : my position is that we should start at `1`, as the code currently says, meaning I tip the scales in your favor. Of did I misread the situation?

nigoroll: no, you did not.

PR #3212, opened 2020-02-08
A more modern random generator with almost no locking

hermunn:
This code introduces a very fast pseudo random generator with good statistical properties and almost no locking. It leverages the `__thread` word to store individual states per thread, and the platforms which we support all have linkers which can handle this mechanism well.

The period of the pseudo random generator is 2^128-1, and each new thread gets its own sequence of 2^64 quasi random numbers at the expense of one lock/unlock operation the first time the thread calls the random generator.

### DIFF
--- a/bin/varnishtest/tests/m00002.vtc
+++ b/bin/varnishtest/tests/m00002.vtc
@@ -9,10 +9,10 @@ varnish v1 -vcl+backend {
 	import std;
 
 	sub vcl_deliver {
-		set resp.http.rnd1 = std.random(0, 1);
-		set resp.http.rnd2 = std.random(0, 10);
-		set resp.http.rnd3 = std.random(8, 10);
-		set resp.http.rnd4 = std.random(99, 100);
+		set resp.http.rnd1 = std.random(0, 1, true);
+		set resp.http.rnd2 = std.random(0, 10, true);
+		set resp.http.rnd3 = std.random(8, 10, true);
+		set resp.http.rnd4 = std.random(99, 100, true);
 	}
 } -start
 
--- a/bin/varnishtest/tests/m00052.vtc
+++ b/bin/varnishtest/tests/m00052.vtc
@@ -1,0 +1,163 @@
+varnishtest "Test xorshiro128** random generator"
+
+varnish v1 -vcl {
+	import std;
+	import debug;
+
+	backend default none;
+
+	sub vcl_init {
+		debug.deterministic_random();
+	}
+
+	sub vcl_recv {
+		return (synth(200, "OK"));
+	}
+	sub vcl_synth {
+		set resp.http.r1 = 1000.0 * std.random52();
+		set resp.http.r2 = 1000.0 * std.random52();
+		set resp.http.r3 = 1000.0 * std.random52();
+		set resp.http.r4 = 1000.0 * std.random52();
+		set resp.http.r5 = 1000.0 * std.random52();
+		set resp.http.r6 = 1000.0 * std.random52();
+		return (deliver);
+	}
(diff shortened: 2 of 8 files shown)


nigoroll:
Questions:
* do we have anyone to judge on the claims of the authors or some references for independent validation?
* I wonder if investing quite some substantial per thread initialization cost is worth it. If we _really_ want to do this we probably should at least avoid the `open`/`close` in `VRND_RandomCrypto()` or maybe, even better, pre-fetch some entropy to use for thread initialization without the IO overhead

bsdphk:
I'm totally failing to see the point of this ?

We have two needs for randomness, cryptographic and testable.

For cryptographic randomness we will rely on what the system provides, because we're not in that game and do not want to be in that game.

For testable randomness, what we have is sufficient ?

hermunn:
> do we have anyone to judge on the claims of the authors or some references for independent validation?

Good point. I have not found anything like that, or peer reviewed articles for that matter.

> I wonder if investing quite some substantial per thread initialization cost is worth it. If we _really_ want to do this we probably should at least avoid the `open`/`close` in `VRND_RandomCrypto()` or maybe, even better, pre-fetch some entropy to use for thread initialization without the IO overhead

Doing the seeding earlier (when making the locks) is certainly OK with me. I would probably have used RDRAND in a private project, but that is specific to x64. A `open`,`close` of `/dev/urandom` on startup should be acceptable, in my humble opinion, even for use cases where the random generator is never called.

The main purpose of this PR is to get lock-less independent samples which are statistically well behaved. For a "secure" version, slapping a block cipher (with a random key, also from /dev/urandom) on top should be enough, but then there might be better options out there.

hermunn:
> I'm totally failing to see the point of this ?

In our fork, we have some applications where lock-less, statistically well behaved (but not secure), fast random generation is needed.

Unrelated to the above, we also did a very synthetic benchmark where we tried to get as many cache insertions as possible. Each object was made unique by adding "randomness" in `vcl_hash`, but with the built-in random generator, lock contention became an issue. 

The implementation in this PR solved this, and we got about 500k insertions per second on a massively parallel machine.

hermunn:
Bugwash questions:

```
15:26:41 <@phk> 1. Does changing PRNG (-testable) break VTCs ?  (if not, get rid of entire testable crap)
15:26:54 <@phk> 2. Is the proposed PRNG one that we want ?
15:27:02 <@phk> 3. Do we want to reduce locking ?
15:29:55 <@phk> 4. Do we need another PRNG/API ?!
```
I will offer my own answers to the questions, and we can discuss here.

After glancing at the code I'd say that I need a deeper look to understand question `1`. Independently of this, I don't think there are any use cases which needs a testable PRNG.

For `2` my humble opinion is: The is proposed PRNG is super fast, has good statistical properties, but I should ask someone who know even more math than I do, for input. What is good with the proposed PRNG is the existence of the jump function, and that no special assembly is used in the `next` function or in the scrambling part.

When it comes to `3` I would say yes, due to the fact that having all requests taking a single lock is generally a bad idea, and it does not scale when the number of CPU cores increase. In a VMOD which needs PRNG numbers, I would use the PR's code over the old VRND functions, but there might be even better alternatives.

For `4` the central word is *need*. With the built in VCL, there is no *need* for a random generator. The question is if we want to provide a locking or a non-locking PRNG in the standard VMOD.

If we can agree on Yes on question `3`, then I can continue working on this. If we resolving the lock contention issue is not considered "worth it", then we can just drop this PR.

hermunn:
I am closing this for now. I will do more research into this, and reopen or open a new one when the research is done.

PR #3213, opened 2020-02-10
Poxy vulnerability mitigation for builtin.vcl

nyov:
While `builtin.vcl` is certainly no place to try defending against numerous possible web exploits, this specific case is an easy security win at almost no cost, in today's botnet landscape.
And Varnish is just the the perfect place for it.

Perhaps you feel different. But the only reason, that I can see, where this could be unwanted, is if a backend server is used to track attempted exploits.

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -37,9 +37,6 @@ vcl 4.0;
 # Client side
 
 sub vcl_recv {
-    if (req.http.host) {
-        set req.http.host = req.http.host.lower();
-    }
     if (req.method == "PRI") {
         /* This will never happen in properly formed traffic (see: RFC7540) */
         return (synth(405));
@@ -54,6 +51,9 @@ sub vcl_recv {
       req.proto ~ "^(?i)HTTP/1.1") {
         /* In HTTP/1.1, Host is required. */
         return (synth(400));
+    }
+    if (req.http.host) {
+        set req.http.host = req.http.host.lower();
     }
     if (req.method != "GET" &&
       req.method != "HEAD" &&
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -44,6 +44,11 @@ sub vcl_recv {
         /* This will never happen in properly formed traffic (see: RFC7540) */
         return (synth(405));
     }
+    if (req.http.proxy) {
+        /* Poxy mitigation. Standards-compliant HTTP clients and servers will
+           never read or send this header. See https://httpoxy.org/#fix-now  */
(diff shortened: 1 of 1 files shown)


nigoroll:
> lgtm

but the two commits should be squashed

Dridi:
I think the two commits should remain separate. Also, I'm not sure whether the built-in VCL is the right location for this since it can easliy be bypassed. And since this is not a standard header, it also means that it could be used with semantics we can't anticipate besides an attack vector towards CGI backends.

Considering that we don't have CGI support in Varnish, I consider this mitigation should be part of the actual backend relying on CGI, especially of Varnish may not be its sole entry point. If someone publishes a vmod_cgi then I think it would make sense _there_ to make this a default behavior.

I'm not going to veto this change, but I think it crosses the tools vs policy line, and unless it makes its way in an RFC relevant to us I don't see a reason to do this by default.

bsdphk:
Agreed, we are close to the tools/policy line here, and that is why, *if* we want to do it, builtin_vcl is the right place, so it can be overridden by the user.

I'm not sure about the 400 response though.

Why not just unset the proxy header in vcl_backend_request{} ?

rezan:
Can we get more details on the vulnerability. Possibly in a 2020 context?

bsdphk:
Context: https://httpoxy.org/#fix-now

bsdphk:
Given that this is not our problem, this doesn't seem like it belongs in builtin.vcl.

After all:  How many workarounds do we want to accumulate i builtin.vcl for other peoples problems, and when can we ever get rid of them again ?

gquintard:
could this maybe be integrated into https://github.com/varnishcache/pkg-varnish-cache ?

nyov:
Thanks all for the consideration. I can agree with the decision.

@gquintard, I'm not sure how this would fit into https://github.com/varnishcache/pkg-varnish-cache - could you elaborate on that?

gquintard:
@nyov, offer it as `/etc/varnish/snippets/security/poxy.vcl`, allowing people to include it at the top of their vcl.
With globbing support coming, one can imagine telling user to

``` vcl
include glob "/etc/varnish/snippets/security/*.vcl";
```

to grab al lthe security snippets at once. To be discussed on the other repo though

Dridi:
Please note that our packages should never install VCLs in `/etc`, but in `/usr` instead (see `devicedetect.vcl`).

nyov:
@gquintard, okay, I have trouble finding any vcl snippets in that repo, or where it would go. Feel free to include it yourself, though, thanks.

gquintard:
so sad, you could have been a trailblazer :-)

Issue #3214, opened 2020-02-10
RFC: struct hdr_name instead of "\007string:"


Dridi:
As I was working on #3211 I noticed that our header constants aren't really constant and we need an `HTTP_Init()` step to fix their length. It also reminded me that despite understanding the optimization behind header name strings with their sizes encoded in the first octet, I'm really not happy with magic `+1` offsets whenever we need it (it also leaks in the `VCL_HEADER` type).

I'm asking for permission to go through the code base and introduce a new type instead, with the following construct:

```c
#include <stdint.h>
#include <stdio.h>

struct hdr_name {
	uint8_t		len;
	const char	str[];
};

#define HDR_NAME(name)	((const struct hdr_name*)			\
	    &(const struct { uint8_t _l; char _s[sizeof(name ":")]; })	\
	    { sizeof(name), name ":" })

/* convenience macro */
const struct hdr_name *const H_Host = HDR_NAME("Host");
const struct hdr_name *const H_Cache_Control = HDR_NAME("Cache-Control");

/* hard-coded layout */
const struct hdr_name *const H_Same_Layout = (struct hdr_name *)("\014Same-Layout:");

int
main(void)
{

	printf("%2u %s\n", H_Host->len, H_Host->str);
	printf("%2u %s\n", H_Cache_Control->len, H_Cache_Control->str);
	printf("%2u %s\n", H_Same_Layout->len, H_Same_Layout->str);
	return (0);
}
```

This only changes how we manipulate header names, not their layout, and it should result in more readable code:

```diff
--- i/bin/varnishd/cache/cache_http.c
+++ w/bin/varnishd/cache/cache_http.c
@@ -292,17 +292,14 @@ http_PutField(struct http *to, int field, const char *string)
 /*--------------------------------------------------------------------*/
 
 static int
-http_IsHdr(const txt *hh, const char *hdr)
+http_IsHdr(const txt *hh, const struct hdr_name *hn)
 {
-       unsigned l;
 
        Tcheck(*hh);
-       AN(hdr);
-       l = hdr[0];
-       assert(l == strlen(hdr + 1));
-       assert(hdr[l] == ':');
-       hdr++;
-       return (!strncasecmp(hdr, hh->b, l));
+       AN(hn);
+       assert(hn->len == strlen(hn->str));
+       assert(hn->str[hn->len - 1] == ':');
+       return (!strncasecmp(hn->str, hh->b, l));
 }
 
 /*--------------------------------------------------------------------*/
[...]

PR #3215, opened 2020-02-11
Introduce a struct gethdr_t type

dridi:
This replaces the (struct gethdr_s).what string, making it explicit in
the type system that the first octet of such strings contains a length
without changing the existing memory layout.

A convenience macro is here to create such a struct from a constant
string literal, which allows in turn to create true header constants
and drop the HTTP_Init() step.

The diff may be easier to follow with the --word-diff option.

---

This is a followup to #3214, to be reviewed after the 2020-03 release.
It breaks the VRT API but not its ABI.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -613,7 +613,8 @@ void http_PutResponse(struct http *to, const char *proto, uint16_t status,
 void http_FilterReq(struct http *to, const struct http *fm, unsigned how);
 void HTTP_Encode(const struct http *fm, uint8_t *, unsigned len, unsigned how);
 int HTTP_Decode(struct http *to, const uint8_t *fm);
-void http_ForceHeader(struct http *to, const char *hdr, const char *val);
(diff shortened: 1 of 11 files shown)


bsdphk:
While I fully agree that the current "first byte is magic" is suboptimal in every way but one, I really do not think this PR is an improvement.

Instead of relying on char being at least 8 bits, we now reply on variable size structs, on struct padding, on "const char" and "uint8_t" being able to coexist in the same word, on casting different structs to each other and ...

For the perceived benefits, this patch really does not carry its weight for me...

Dridi:
For `uint8_t` we can always use `unsigned char` instead. I happen to like the `[u]int<size>_t` types but if it feels like mixing apples and bananas for you I have no issue with changing that implementation detail.

bsdphk:
I have looked at this again with fresh eyes, and I still dont find it an improvement.

I still find our current "pascal-string" hack deeply problematic, but this is not an improvement in my eyes.

dridi:
Bugwash: to be revisited, check whether recent compilers complain about this construct.

I will remake the change on top of current master.

PR #3219, opened 2020-02-13
Add VRT_AllocStrandsWS() to allocate strands on a workspace

nigoroll:
So far, we have focused on transitioning vmod arguments to STRANDS.

To take the next step and also return STRANDS (which leverages the actual benefit - not copying unmodified strand elements), vmods need to allocate a struct strands on the workspace.

This commit adds a utility function for this common task.

Implementation note: A single `WS_Alloc()` with some pointer arithmetic would suffice, but using two results in cleaner code.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -181,6 +181,30 @@ VRT_GetHdr(VRT_CTX, VCL_HEADER hs)
 }
 
 /*--------------------------------------------------------------------
+ * Alloc Strands with space for n elements on workspace
+ *
+ * Error handling is deliberately left to the caller
+ */
+
+struct strands *
+VRT_AllocStrandsWS(struct ws *ws, int n)
+{
+	struct strands *s;
+	const char **p;
+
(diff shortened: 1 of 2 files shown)


Dridi:
No opinion.

> Implementation note: A single WS_Alloc() with some pointer arithmetic would suffice, but using two results in cleaner code.

I think we end up with contiguous allocations anyway, I'm pretty sure both allocations are pointer-aligned and that's the alignment workspace uses for all allocations. It shouldn't matter...

nigoroll:
> I think we end up with contiguous allocations anyway
correct

nigoroll:
bugwash generally in favor.
TODO: Survey use cases in tree

nigoroll:
Survey:
* vmod functions which could return some strands unmodified
  * `std.tolower` / `std.toupper`
  * `str.strstr`

* vmod functions which may append only (as in create a new `struct strands` with some strand(s) added)
  * `blob.encode` (for NUL-addition with encoding=IDENTITY)
  * `blob.transcode` same

vmod_debug disregarded

Issue #2990, opened 2019-04-30
Interactive varnishstat -f confusion


Dridi:
On multiple occasions I was told that people get confused when they limit the counters to a given list and yet only a subset shows up.

In general it is well understood that counters with the value zero don't show up, almost all of the time they don't realize that we have different kinds of counters and some may only show up only if we increase the verbosity with the `v` key.

There are a couple things I would like to consider:

- Start right away with verbosest verbosity when `-f` is used
- Exclude verbosity levels not covered by `-f` when rotating verbosity with `v`
- Consider an extra verbosity level to show all counters, even the zeroes

The second point may be to complicated to implement with dynamic VSCs if we use glob matching, and possibly not worth it.

nigoroll:
The second point may be too complicated to understand for someone with limited capacity like me ;)

+1 on the first

on the third point: we got `<d>`, should we maybe have a command like option as well?

bsdphk:
No opinion.

Dridi:
@nigoroll let's imagine that your `-f` selection leads to the following verbosity levels:

- normal: 3 counters
- diag: no counter
- debug: 1 counter

Cycling through with `<v>` or `<V>` could skip the diagnostic level, hence dynamic VSCs may render this too complicated and not worth it.

TIL about `<d>` so forget about the third point, I RTFM in general but forgot about that one (and others actually).

Dridi:
And f41055d326fdfabd75a1afd0653ea9357a13d25e is evidence that I rarely use `varnishstat` in interactive mode.

bsdphk:
Dridi tries to improve UX based on bugwash discussion.

nigoroll:
bugwash notes:
* add <h> for curses help
* set <v> by default to include all -f matches
* make <v> setting more prominent in interface, e.g.
* `press h for help` or toggle counters with the help docs?
* @Dridi wants to find someone to attempt a PR

Dridi:
Update: a while ago I pushed some ground work that would allow the key bindings reference to be shared by the event loop, man page and help screen. The help screen will follow at some point.

kn-cpt:
> set `<v>` by default to include all -f matches

This would fix `varnishstat -f '*overload*'` in that it actually shows the requested value without pressing `v`, just like the same invocation with an `-1` added shows it.

Above mentioned PR came out of this very confusion.
I think the user interface should show exactly what the user requested, hence to include all `-f` matches by default seems to be the right approach.

[...]

PR #3223, opened 2020-02-19
varnishstat.rst: Mention default log level in curses mode

kn-cpt:
The Keybindings section is the only place mentioning differnt log levels
and one might assume that `-1` or `-j` behave the same as the default
ncurses interface.

The latter however does not show all counters by default as the single
action flags do, hence `varnishstat -f '*overload*'` will not show
`MAIN.sc_overload` by default, with `-1` however it does.

Fix grammar while here.

### DIFF
--- a/doc/sphinx/reference/varnishstat.rst
+++ b/doc/sphinx/reference/varnishstat.rst
@@ -30,7 +30,7 @@ The following options are available:
 CURSES MODE
 ===========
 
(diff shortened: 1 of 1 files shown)


Dridi:
Thanks, see also #2990: I'd like to adjust the verbosity during startup to ensure `-f` matches are always shown by default. It will eventually reach the top of my plate :)

PR #3224, opened 2020-02-21
Fix formatting on 'ESI on invalid XML'

Napsty:
There's a tiny typo in the documentation users-guide/esi.rst which causes a wrong formatting in the section "ESI on invalid XML":

image


gquintard:
Merged, thanks!

PR #3226, opened 2020-02-21
TTL logging for bereq.uncacheable / Keep Age information on passes

mbgrydeland:
When doing a pass, we would remove the Age header from the backend, and
create a new one based on the time the fetch was initiated. This creates
problems when calculating the time to live in downstream caches (browser
cache or layered varnishes).

This patch set makes Varnish parse the Age header and update t_origin accordingly. That makes the synthesized Age header correct in delivery.

Also a patch is added that makes us consistently not log TTL log lines for private/pass objects.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -124,8 +124,9 @@ VRT_hit_for_pass(VRT_CTX, VCL_DURATION d)
 	oc->ttl = d;
 	oc->grace = 0.0;
 	oc->keep = 0.0;
-	VSLb(ctx->vsl, SLT_TTL, "HFP %.0f %.0f %.0f %.0f uncacheable",
-	    oc->ttl, oc->grace, oc->keep, oc->t_origin);
+	if (!(oc->flags & OC_F_PRIVATE))
+		VSLb(ctx->vsl, SLT_TTL, "HFP %.0f %.0f %.0f %.0f uncacheable",
+		    oc->ttl, oc->grace, oc->keep, oc->t_origin);
 }
 
 /*--------------------------------------------------------------------*/
--- a/bin/varnishd/cache/cache_vrt_var.c
+++ b/bin/varnishd/cache/cache_vrt_var.c
@@ -246,9 +246,10 @@ VRT_l_beresp_uncacheable(VRT_CTX, VCL_BOOL a)
 	}
 	oc = ctx->bo->fetch_objcore;
 
-	VSLb(ctx->vsl, SLT_TTL, "VCL %.0f %.0f %.0f %.0f %s",	\
-	    oc->ttl, oc->grace, oc->keep, oc->t_origin,		\
-	    ctx->bo->uncacheable ? "uncacheable" : "cacheable");\
+	if (!(oc->flags & OC_F_PRIVATE))
+		VSLb(ctx->vsl, SLT_TTL, "VCL %.0f %.0f %.0f %.0f %s",
+		    oc->ttl, oc->grace, oc->keep, oc->t_origin,
+		    ctx->bo->uncacheable ? "uncacheable" : "cacheable");
 }
 
 VCL_BOOL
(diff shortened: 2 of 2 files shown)


mbgrydeland:
b00002.vtc failes with this patch. From that test case:

```
# check that there are no TTL LogTags between the
# last header and VCL_return b deliver
logexpect l1 -v v1 -g request {
	  expect * 1002 Begin
	  expect * =	BerespHeader	{^Date:}
	  expect 0 =	VCL_call	{^BACKEND_RESPONSE}
	  expect 0 =	BerespHeader	{^x-ttl: 0.000}
	  expect 0 =	VCL_return	{^deliver}
} -start
```

I don't really understand the requirement in that test case and will argue that the test case is wrong.

mbgrydeland:
Further digging turned up this commit: d9b0919109b402d60d5978d8ab5f0fe370699785

This patch set seems to be somewhat of a reversal of that commit. This will need further discussion during the bugwash.

mbgrydeland:
@nigoroll I have updated and force pushed this PR with the changes we discussed during bugwash.

nigoroll:
@mbgrydeland I have merged the first three commits and took the liberty to rebrase your branch.
I will comment on the last commit

bin/varnishd/cache/cache_vrt.c
```diff
@@ -124,8 +124,9 @@ VRT_hit_for_pass(VRT_CTX, VCL_DURATION d)
 	oc->ttl = d;
 	oc->grace = 0.0;
 	oc->keep = 0.0;
-	VSLb(ctx->vsl, SLT_TTL, "HFP %.0f %.0f %.0f %.0f uncacheable",
-	    oc->ttl, oc->grace, oc->keep, oc->t_origin);
+	if (!(oc->flags & OC_F_PRIVATE))
```
nigoroll: so this is a `return(pass())` for a `bereq.uncacheable` object.

I wonder if it would rather be more stringent to log the error as in L119, but maybe with some reason added (called from the client side vs. called on an already uncacheable object).
And I wonder if that should rather be an `SLT_Notice`?

rezan:
Just a comment, I think we need to keep the commits intact in the PR. When we reference this for backporting or documentation, the commits are now lost.

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -246,9 +246,10 @@ VRT_l_beresp_uncacheable(VRT_CTX, VCL_BOOL a)
 	}
 	oc = ctx->bo->fetch_objcore;
 
-	VSLb(ctx->vsl, SLT_TTL, "VCL %.0f %.0f %.0f %.0f %s",	\
-	    oc->ttl, oc->grace, oc->keep, oc->t_origin,		\
-	    ctx->bo->uncacheable ? "uncacheable" : "cacheable");\
```
nigoroll: similar thing here, and I think we would already be doing the right thing, but IMHO we should rather just `return` in L243 (after the `SLT_VCL_Error`)

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -606,20 +607,20 @@ ttl_now(VRT_CTX)
...
+{									\
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);				\
+	a += (offset);							\
+	if (a < 0.0)							\
+		a = 0.0;						\
+	oc->fld = a;							\
+	if (!(oc->flags & OC_F_PRIVATE))				\
+		VSLb(ctx->vsl, SLT_TTL, "VCL %.0f %.0f %.0f %.0f %s",	\
+		    oc->ttl, oc->grace, oc->keep, oc->t_origin,		\
+		    ctx->bo->uncacheable ? "uncacheable" : "cacheable");\
```
nigoroll: Here we attempt to set ttl/grace/keep on a `bereq.uncacheable` object, and I would also prefer to output an `SLT_Error` or `SLT_Notice`.

On a related question, did we ever fix the semantics of grace & keep for HFM?

nigoroll:
@rezan the fix is referenced independently, but FTR, the following commits are merged:
* aaaf7b383339b18f368935ff3fd38a6575c15666
* b610e9a0b954fe963a86b95b44ee30089a9333d3
* cff50548b9633b676d71fc82c9103b368807e9a3

nigoroll:
btw @mbgrydeland thank you very much for fixing my regression.

Dridi:
> Just a comment, I think we need to keep the commits intact in the PR. When we reference this for backporting or documentation, the commits are now lost.

But the commits from the pull request are not authoritative since we rebase the pull request onto the target branch. You already end up with different hash since the committer field is changed to whomever proceeds with the merge (the author field is preserved).

rezan:
> But the commits from the pull request are not authoritative since we rebase the pull request onto the target branch. You already end up with different hash since the committer field is changed to whomever proceeds with the merge (the author field is preserved).

That fine, we aren't looking for the exact hashes which landed in the codebase, rather, we just want to see the code and commits which fix the problem the PR addresses.

If I were to come to this PR and didn't know it was rebased away, I would be completely lost when it came to understanding how the fix was implemented.

nigoroll:
@rezan has a point: It is obious which commits a merged PR references but for other merges we should just mention the commits. We did that in the past before we used rebase & ff merges.

Dridi:
If you want to have this discussion we can move it to a mailing list but specifically for this reason we stopped considering tickets as a source of back-ports. Issue fixes (commits containing "Fixes #issue¨) are sometimes preceded with prep work, and sometimes followed by "oops" or "polish" followup commits. Sometimes bug fixes aren't linked to a ticket at all. So when 6.0 back-ports started all commits in the log were considered instead of looking through the lens of the issue tracker.

Moving back to a ticket-based workflow doesn't work, unless all changes are attached to a ticket.

rezan:
My last comment on this, it just seems counter productive to create PRs for bugs and then have to hunt thru the commit log to find the commits. A lot of times the commit messages make no references to the bugs they fix since we now break up PRs into lots of small commit steps. People are free to examine each commit LOC to look for potential bug fixes, but in this case we have a very specific bug which goes back to before our LTS version and a patch which resolves it. Had I not noticed that the PR was rebased away, we would have easily wasted probably a few days trying to piece together how this bug was fixed. The best practice here is to leave PR commit sets as is and not alter them because when you do, the commits are pretty much lost forever.

If we want to continue with this practice, then I would ask that you add more context to the commit messages because in this case, we have 4 non linear commits in the history which are needed to fix this bug. Like I said, this would be near impossible to piece together had I not noticed this new PR rebasing practice.

nigoroll:
bugwash: @mbgrydeland will propose a different logging solution.

there was some discussion about how to log and handle ttl etc. for pass/private objects in general. Basically there are two perspectives:
* it does not matter so it can be (ab)used for anything, logging should reflect the vcl vs.
* logging and vcl behavior should reflect reality

Issue #2994, opened 2019-05-07
Probes don't work well with Envoy proxy, likely RFC violation by Varnish


MarcinFalkowski:
We encountered a problem running Varnish with Envoy proxy (https://www.envoyproxy.io) as a backend.
Varnish probes don't work well with Envoy because Envoy [delays connection closing](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto.html?highlight=delayed_close_timeout)  and the probe expects connection to be closed by the backend within the timeout to mark backend as healthy.

We discussed this already on Envoy github: https://github.com/envoyproxy/envoy/issues/6616 and current conclusion is this should be fixed on Varnish side. More below.

## Expected Behavior
* Varnish probe should succeed when it receives expected response. Connection closing shouldn't be necessary to pass the check.
* Varnish probe should close the connection by itself when it receives a response. This is required by rfc7230 section 6.6:
   > A client that sends a "close" connection option (...) MUST close the connection after reading the final response message corresponding to this request.

## Current Behavior
* Varnish probe, after receiving expected http response, doesn't mark the backend as healthy. It waits for connection closing by the backend. If the backend won't close the connection within the timeout, Varnish close the connection and mark backend as unhealthy. This is even stated in docs (https://varnish-cache.org/docs/6.2/reference/vcl.html#probes):
> Note that probes require the backend to complete sending the response and close the connection within the specified timeout

Here is a dump of Varnish <-> Envoy communication to illustrate current behaviour: https://github.com/MarcinFalkowski/envoy-varnish-testing-env/blob/master/captures/tcpdump-2019-04-22_23-00.pcap?raw=true  (192.168.0.3 is Varnish and 192.168.0.2 is Envoy). We can see that Varnish receives expected response after 1ms but still nobody closes the connection. Only after probe timeout (100ms) Varnish closes the connection and marks backend as unhealthy. You cannot see it but you can test it by yourself using this docker-compose environment: https://github.com/MarcinFalkowski/envoy-varnish-testing-env 

## Possible Solution
* Modify Varnish Probes logic to not require connection closing within the specified timeout to pass the health check.
* Close the connection immediately after receiving response from backend (as required by RFC7230)

## Other

Envoy delays connection closing to mitigate some possible race conditions with badly designed clients. More in the linked issue on Envoy github.

Some thoughts/questions:
[...]

PR #3228, opened 2020-02-24
RFC: Expansion of probe .request definitions

Dridi:
It is possible to share a probe between multiple backends today, but it
becomes impossible to share a probe with a .request field with a proper
Host header.

Either the backend is in cahoots with Varnish and will expect a specific
domain for health checks (for example Host: health) or we simply don't
send the header despite its mandatory status in HTTP/1.1.

This patch introduces incomplete and rudimentary %{variable} expansion
in the probe request in a way that could be outsourced in an include
table used both to build a probe request or generate documentation.

An initial set of two variables is implemented:

- %{host_header} for the reason mentioned above
- %{now} in order to generate a Date header

We could imagine more variables like %{version} as suggested by the test
case. The %{...} syntax was selected to avoid conflicts with varnishtest
macros and is only here to illustrate the mechanics. This is something I
would tend to check at vcl.load time, this is completely missing from
this RFC.

### DIFF
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -575,6 +575,55 @@ VBP_Status(struct vsb *vsb, const struct backend *be, int details, int json)
  */
 
 static void
+vbp_expand_host_header(struct vsb *vsb, const struct backend *be)
+{
+
+	VSB_printf(vsb, "Host: %s", be->hosthdr);
+}
+
+static void
+vbp_expand_now(struct vsb *vsb, const struct backend *be)
+{
+	char buf[VTIM_FORMAT_SIZE];
+
+	(void)be;
+	VTIM_format(VTIM_real(), buf);
+	VSB_cat(vsb, buf);
+}
+
+static void
+vbp_expand_req(struct vsb *vsb, const struct backend *be,
+    const char *s)
+{
+	const char *p, *q;
+
+	p = s;
+	while (1) {
+		q = strchr(p, '%');
+		if (q == NULL) {
+			VSB_cat(vsb, p);
+			break;
+		}
+		if (p < q) {
+			VSB_bcat(vsb, p, q - p);
+			p = q;
+		}
+#define VBP_MATCH_MACRO(m)				\
+		if (!strncasecmp(p, "%{" #m "}",	\
+		    sizeof("%{" #m "}") - 1)) {		\
+			vbp_expand_ ## m(vsb, be);	\
+			p += sizeof("%{" #m "}") - 1;	\
+			continue;			\
+		}
+		VBP_MATCH_MACRO(host_header)
+		VBP_MATCH_MACRO(now)
+#undef VBP_MATCH_MACRO
+		VSB_putc(vsb, *p);
+		p++;
+	}
+}
+
+static void
 vbp_build_req(struct vbp_target *vt, const struct vrt_backend_probe *vbp,
     const struct backend *be)
 {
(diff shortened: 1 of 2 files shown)


gquintard:
isn't it something that we should do via vcl subroutines?

This is an old subject, but this patch seems to bring the nginx config approach of obscure variables to vcl, and I think we can do something more idiomatic.

```
sub vcl_probe_fetch {
    set req.http.timestamp = now;
    if (req.backend.name ~ "^origin_img_[0-9]") {
        set req.http.host = "img.example.com";
    }
}
```

the esteemed reader will realize that I am not very subtle about opening the door to backend override, and possible response manipulation, but be assured that I know it :-)

Dridi:
Until probes stop having an incestuous relationship with backends, I think the probe definition is the right place to do that expansion.

If we want a VCL state machine dedicated to probes, this could be something done in the built-in VCL. But if you want a 6.0-able solution, I think (again) that variable expansion is the way to go with what we have today. I also don't see why it should be obscure once documented, and I specifically made it include-table-able (not implemented in the RFC) to make sure we don't run in an out-of-sync situation in vcl(7).

nigoroll:
I am not against it, but I am also surprised to see this suggestion right after we talked about the matter during bugwash and agreed that we wanted to make the solution to #2994 part of a generalized probe-in-vcl approach

nigoroll:
another note: the variable expansion would be an isolated syntax not found elsewhere in VCL
I predict that users would rightly ask themselves why they can use `%{host}` in a probe definition, but why not in `set req.http.myhdr = "my-%{req.http.host}"`
I am reluctant to introduce new language elements for a static definition when (at least in my mind) the way forward is to make probes (vmod) objects like anything else.

Dridi:
> I am not against it, but I am also surprised to see this suggestion right after we talked about the matter during bugwash and agreed that we wanted to make the solution to #2994 part of a generalized probe-in-vcl approach

I expected that kind of reaction. Let's say this is a way to both satisfy current probes, and how the initial `req` is built in a future probe state machine.

> another note: the variable expansion would be an isolated syntax not found elsewhere in VCL
I predict that users would rightly ask themselves why they can use `%{host}` in a probe definition, but why not in `set req.http.myhdr = "my-%{req.http.host}"`

Please don't focus on the expansion/interpolation syntax, like I said I came up with something on the spot and while we arguably have that with `vtc` macros (I know VTC is not VCL) but I didn't want to run into conflicting syntax so I bolted together something similar.

How the expansion is implemented internally and in VCL doesn't have to match, for example it could look like this:

```
.request = 1*(quoted-string / token) ";"
```

For example:

```vcl
probe default {
	.request =
	    "OPTIONS * HTTP/1.1"
	    date_header
	    host_header
	    "User-Agent: Varnish"
	    "Connection: close";
}
```

You lose the ability to have a fine-grained expansion, but with a mix of keywords and normal strings this should look less magic. In a state machine world I guess you could do complex string concatenations and for 6.0 LTS I would be perfectly fine with something more limited to back-port.

rezan:
I agree with @gquintard here, adding variables here goes against the nature of VCL. It should be possible to cookup a new probe parameter for 6.0 to do what you want if there is a strong need for this. Otherwise, I think VCL based probes is the path forward.

nigoroll:
@Dridi yes, I think the token approach would be more in line with VCL syntax.
So yeah, no veto, but no enthusiasm either

gquintard:
> Until probes stop having an incestuous relationship with backends, I think the probe definition is the right place to do that expansion.

Out of curiosity, is it that much an issue if the backend can't be changed in the `vcl_probe_*` subroutines?

I'll back out of that discussion, I don't like the expanded variable approach, but I want the feature it brings, so +0 from me.

bsdphk:
Firm -1 on this one.

Either we do the VIP we agreed to a few days ago, where this all becomes VCL'able, or we do what you propose here, we do not do both.

Dridi:
I'm closing this now, but I will come back with something similar depending on how the VIP looks like. My idea was that it could serve for today's probes, but also work with future VCL subroutine'd probes to define how the initial `req` looks like.

PR #3231, opened 2020-02-26
varnish.m4: Fix documentation mistake

xordspar0:
I think this got copied and pasted from the VARNISH_UTILITIES macro below. If I understand correctly, the macro is called VARNISH_COUNTERS and should be used as `VARNISH_COUNTERS([foo bar])`.

### DIFF
--- a/varnish.m4
+++ b/varnish.m4
@@ -437,7 +437,7 @@ clean-vsc-$1:
 #     @BUILD_VSC_FOO@
 #     @BUILD_VSC_BAR@
 #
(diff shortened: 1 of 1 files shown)


Dridi:
That's not a mistake, I always put typos and wrong copy-pasta on purpose to find when people start paying attention. Now I know that I can expect custom counters from third-party VMODs.

xordspar0:
😄 Good guess. We're actually working on adding counters to vmod_xkey.

PR #3232, opened 2020-02-26
varnish.m4: Correct the example VSC filenames

xordspar0:
Above, `VSC_$1.h: $1.vsc` seems to imply that the source VSC file should be called foo.vsc in this example. Also, if there's an example using VSC_foo.rst, then we should make sure to build that in the example.

### DIFF
--- a/varnish.m4
+++ b/varnish.m4
@@ -437,7 +437,7 @@ clean-vsc-$1:
 #     @BUILD_VSC_FOO@
 #     @BUILD_VSC_BAR@
 #
-# They take care of turning VSC_foo.vsc and VCS_bar.vsc into C code and
+# They take care of turning foo.vsc and bar.vsc into C code and
 # RST documentation.
 #
 # Just like the vcc_*_if.[ch] files, you need to manually add the generated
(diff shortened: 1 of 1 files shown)


Dridi:
This is getting increasingly embarrassing.

The problem I have with this bug is that I'm not sure whether we should fix the documentation or the implementation. In the Varnish source tree we name the files `VSC_*.vsc` and in hindsight it's probably more appropriate.

My fear if we fix the implementation is that someone's setup will break and for some reason they know where I live and they're really not happy with the change.

I will ponder this, but I'm leaning towards the docfix. Had I completed VSC support in VCDK I would probably have noticed myself, what a bummer. I'll sleep on it and probably merge tomorrow.

Thanks again for catching those issues!

varnish.m4
```diff
@@ -447,7 +447,8 @@ clean-vsc-$1:
 #             vcc_baz_if.c \
 #             vcc_baz_if.h \
 #             VSC_foo.c \
-#             VSC_foo.h
+#             VSC_foo.h \
+#             VSC_foo.rst
```
Dridi: `VSC_foo.rst` wouldn't be a source for the library, but for the manual instead.

Dridi:
I pushed a slightly different docfix, but at this point I'm expecting your next pull request pointing at the next silly mistake...

xordspar0:
Great, thank you. I updated my code to match your more accurate documentation.

PR #3239, opened 2020-03-02
Document more dynamic parameters bounds

Dridi:
This is fresh from the oven, hopefully docfixes we can get in before the release.

My original intent was to use the dynamic bounds reasons fields to document `vsl_reclen` and `vsl_buffer` dependencies. It turned into the beginning of an overhaul of `params.h` but while individual patches may look large, they aren't. Using `git show -U0 <hash>` might help for the review of large-looking patches. They are really just mechanical patches, courtesy of vim macros.

Along the way, I also found that `thread_pool_stack`'s dynamic nature was not documented.

I'm planning further progress in this area after the 6.4 release. Reviewing this area last week I found more opportunities for improvement.

### DIFF
--- a/bin/varnishd/common/common_param.h
+++ b/bin/varnishd/common/common_param.h
@@ -82,7 +82,8 @@ struct params {
 #define	ptyp_uint	unsigned
 #define	ptyp_vsl_buffer	unsigned
 #define	ptyp_vsl_reclen	unsigned
-#define PARAM(nm, ty, mi, ma, de, un, fl, st) ptyp_##ty nm;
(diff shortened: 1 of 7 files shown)


bsdphk:
I don't think we have ever had a docfix that was not appropriate right up to the very second we released.

Dridi:
> do you mean they are _impotent_ or _impoRtant_ ?

No, I mean those macros do nothing on account of being `#if 0`'d.

nigoroll:
bugwash OK from @bsdphk

Issue #2963, opened 2019-04-04
vcl_deliver restart does not abort backend fetch


HLeithner:
## Expected Behavior
Restart the connection and pipe the data.

## Current Behavior
Restart and piping seams to work, but it seams as the old connection is still running.

## Possible Solution
Closing the connection as soon as deliver returns restart?

## Steps to Reproduce (for bugs)
1. create a big file (ex. 5gb) on the backend server
2. use a varnish with limited memory 2gb
3. use the following vcl

```
sub vcl_recv {
  if(req.http.x-pipe && req.restarts > 0) {
    unset req.http.x-pipe;
    return (pipe);
  }
}
sub vcl_backend_response {
  // don't cache files larger than 20MB
  if (std.integer(beresp.http.Content-Length, 0) > 20242880) {
    set beresp.http.x-pipe = "1";
    set beresp.uncacheable = true;
    set beresp.do_gzip     = false;
    set beresp.do_gunzip   = false;
    set beresp.ttl         = 120s;
    return (deliver);
  }
}
sub vcl_deliver {
  if (resp.http.x-pipe && req.restarts == 0) {
    set req.http.x-pipe = "1";
  return (restart);
  }
}

```

4. varnish will run oom
5. directly pipe in vcl_recv without restart works

## Context
I'm trying to get around caching big files, in varnish < 4.0 I found code snipes for restarting and piping. But this doesn't work any longer in varnish >= 4.0. As alternative I tried 2 things.

Adding a x-pipe header in vcl_backend_response and restart in vcl_deliver and pipe the restarted connection in vcl_recv.

if you not disable gzip you will see that the process use 100% for compression even if the transfer is canceled.

## Your Environment

* Version used: 6.2
* Operating System and version: Debian Stretch
* Source of binary packages used: official packages
* haproxy with UDS

nigoroll:
I have not reproduced it, but this seems plausible to me: After (possibly aborted) delivery of the body on the client side, we `HSH_Abandon()` a private body in `cnt_deliver`:
https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_req_fsm.c#L407
So I guess we should do the same in `cnt_deliver` for `return(restart)` and `return(synth)`

PR #3240, opened 2020-03-02
POC/VIP26: limit Transient usage for private objects

ehocdet:
It's an old attempt to limit Transient usage for private objects (named "soft pipe" initially).
It looks like this attempt is more stable since #2963 has been fixed.
I guess my approach is a little too hackish but it can be a start for https://github.com/varnishcache/varnish-cache/wiki/VIP-26:-limit-private-prefetch.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -295,6 +295,8 @@ struct boc {
 	unsigned		refcount;
 	struct lock		mtx;
 	pthread_cond_t		cond;
+	struct lock		mtx_pipe;
+	pthread_cond_t		cond_pipe;
 	void			*stevedore_priv;
 	enum boc_state_e	state;
 	uint8_t			*vary;
--- a/bin/varnishd/cache/cache_esi_fetch.c
+++ b/bin/varnishd/cache/cache_esi_fetch.c
@@ -88,7 +88,7 @@ vfp_vep_callback(struct vfp_ctx *vc, void *priv, ssize_t l, enum vgz_flag flg)
 	VGZ_Ibuf(vef->vgz, vef->ibuf_o, l);
 	do {
 		dl = 0;
-		if (VFP_GetStorage(vc, &dl, &ptr) != VFP_OK) {
+		if (VFP_GetStorage(vc, &dl, &ptr, 0) != VFP_OK) {
 			vef->error = ENOMEM;
 			vef->tot += l;
 			return (vef->tot);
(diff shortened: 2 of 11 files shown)


nigoroll:
Your efforts are appreciated, thank you!

@ehocdet, do you intend this as a poc to give the issue some traction or do you actually want to work towards code fit for integration? I am asking before spending time on a comprehensive review...

ehocdet:
@nigoroll, i can work on this code, but the main functionality "limit private prefetch" greatly differ from VIP26: the stevedore gets notified by the requester as soon as it consumes a buffer. So, i think it will just a poc.

Dridi:
@ehocdet would you like to test #3480? It's another take on VIP26 without an additional lock and condvar.

ehocdet:
@Dridi  thanks you for the update.
#3480 have 2 issues : CondWait on slow clients is not good, lack of multi-buffering will slow down stream.
It's why I used additional lock and condvar on my poc.

mbgrydeland:
While working on #3480, I had a quick look at this. I believe there are cases where this patch set as is fails and causes deadlocks. Specifically, it assumes that there will always be a streaming client in order to update the fetch thread and not cause that to wait indefinitely.

The following test case from #3480, slightly adapted, demonstrates one mode in which this can happen:

```
varnishtest "Pass buffering deadlock test"

server s1 {
	rxreq
	txresp -status 404
} -start

server s2 {
	non_fatal
	rxreq
	txresp -bodylen 2000000
} -start

varnish v1 -vcl+backend {
	import vtc;
	sub vcl_recv {
		set req.backend_hint = s2;
		if (req.restarts == 1) {
			set req.backend_hint = s1;
		}
		return (pass);
	}
	sub vcl_deliver {
		if (req.restarts < 1) {
			vtc.sleep(0.5s);
			return (restart);
		}
	}
	sub vcl_backend_response {
	}
} -start

varnish v1 -cliok "param.set debug +syncvsl"

client c1 {
	txreq
	rxresp
	expect resp.bodylen == 0
	expect resp.status == 404
} -run

varnish v1 -expect VBE.vcl1.s1.conn == 0
varnish v1 -expect VBE.vcl1.s1.busy == 0
varnish v1 -expect VBE.vcl1.s2.conn == 0
varnish v1 -expect VBE.vcl1.s2.busy == 0

```

You also write that you have some reservations about the approach made in #3480. I think those are based on missunderstandings of the way this feature is implemented there. I added a comment there attempting to explain why. Please have a look and we can continue the discussion there on that ticket.

Would you be able to have the implementation in #3480 a go in your setups? Some real world testing of this would be much appreciated.

bsdphk:
See #3572

PR #3245, opened 2020-03-07
Improve vcl reference caching

nigoroll:
This PR consists of two commits, and I would kindly ask reviewers to also read the individual commit messages.

* The first commit enables vcl reference caching also for backend requests and ESI subrequests
* The second commit changes the `VCL_Recache()` function to also cache references to VCLs used for labels, which should increase chances of the cache being hit when labels are used.

### DIFF
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -90,6 +90,12 @@ VCL_Refresh(struct vcl **vcc)
 	VCL_Update(vcc, NULL);
 }
 
+static inline int
+vcl_cancache(const struct vcl *vcl)
+{
+	return (vcl == vcl_active || vcl->nlabels > 0);
+}
+
 void
 VCL_Recache(struct worker *wrk, struct vcl **vclp)
 {
@@ -98,12 +104,17 @@ VCL_Recache(struct worker *wrk, struct vcl **vclp)
 	AN(vclp);
 	CHECK_OBJ_NOTNULL(*vclp, VCL_MAGIC);
 
-	if (*vclp != vcl_active || wrk->vcl == vcl_active) {
+	if (! vcl_cancache(*vclp))
 		VCL_Rel(vclp);
-		return;
-	}
-	if (wrk->vcl != NULL)
+
+	if (wrk->vcl != NULL &&
+	    (*vclp != NULL || ! vcl_cancache(wrk->vcl)))
 		VCL_Rel(&wrk->vcl);
+
+	if (*vclp == NULL)
+		return;
+
+	AZ(wrk->vcl);
 	wrk->vcl = *vclp;
 	*vclp = NULL;
 }
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -97,7 +97,7 @@ VCL_Refresh(struct vcl **vclp, struct worker *wrk)
 }
 
 static inline int
-vcl_cancache(const struct vcl *vcl)
+vcl_cacheable(const struct vcl *vcl)
 {
 	return (vcl == vcl_active || vcl->nlabels > 0);
 }
@@ -142,18 +142,18 @@ VCL_Rel(struct vcl **vclp, struct worker *wrk)
 
 	/* lockless case first */
 	if (wrk != NULL && wrk->vcl == NULL &&
-	    vcl_cancache(vcl)) {
+	    vcl_cacheable(vcl)) {
 		wrk->vcl = vcl;
 		return;
 	}
 
 	Lck_Lock(&vcl_mtx);
 
-	if (wrk == NULL || ! vcl_cancache(vcl))
+	if (wrk == NULL || ! vcl_cacheable(vcl))
 		vcl_rel(vcl);
 
 	if (wrk != NULL && wrk->vcl != NULL &&
-	    (vcl != NULL || ! vcl_cancache(wrk->vcl)))
+	    (vcl != NULL || ! vcl_cacheable(wrk->vcl)))
 		vcl_rel(wrk->vcl);
 
 	Lck_Unlock(&vcl_mtx);
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -97,29 +97,6 @@ vcl_cancache(const struct vcl *vcl)
 }
 
 void
-VCL_Recache(struct worker *wrk, struct vcl **vclp)
-{
-
-	AN(wrk);
-	AN(vclp);
-	CHECK_OBJ_NOTNULL(*vclp, VCL_MAGIC);
-
-	if (! vcl_cancache(*vclp))
-		VCL_Rel(vclp);
-
-	if (wrk->vcl != NULL &&
-	    (*vclp != NULL || ! vcl_cancache(wrk->vcl)))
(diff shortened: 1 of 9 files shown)


bsdphk (dismissed):
I dont like spreading the knowledge about wrk->vcl all over the place.

A better idea would be to give VCL_Ref() wrk argument, which is allowed to be NULL.

nigoroll:
Re @bsdphk 
Thank you for the quick review.

This is a by-product of a rework of #2725, I just noticed that the existing vcl ref caching was only used in 1 of 3 places.
My understanding is that the caching in the `struct worker` was added because it is important and it was @mbgrydeland 's argument against #2725 that we should not require another mutex for each task. So it appeared consequential that further reducing `vcl_mtx` contention would help. As of now, I do not have own performance data on how contended it really is.

I have updated the patch to reflect your suggestion, thank you.

nigoroll:
As suggested by @bsdphk I have added a commit to merge `VCL_Recache()` and `VCL_Rel()`. Please see the third commit's message for reasoning.

nigoroll:
Also added a fourth commit with another bis of centralization of the wrk vcl ref caching just in case @bsdphk would have suggested that anyway. I am also fine with removing that commit again.

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -88,6 +88,12 @@ VCL_Refresh(struct vcl **vcc)
 	VCL_Update(vcc, NULL);
 }
 
+static inline int
+vcl_cancache(const struct vcl *vcl)
+{
+	return (vcl == vcl_active || vcl->nlabels > 0);
+}
+
```
Dridi: For lack of a better place, I will comment here about the general idea behind this change.

In a setup where the main VCL is only an empty shell that only dispatches requests to other VCLs via labels, caching anything but the active VCL is bound to be undone... on the client side. So it's not just opportunistic, it's even more of a gamble since recaching anything but the active VCL actively (pun intended) jeopardizes future client tasks, and the more the labels the lower the probability to benefit from recaching on the backend side.

Also, after 60s of inactivity an idle thread will release its VCL reference. Should we make the worker keep it instead when it's still the active VCL to trade vcl_mtx usage with the pool's mutex instead? I mean, 60s is probably less frequent than starting and completing tasks on a busy system.

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -138,22 +115,54 @@ VCL_Ref(struct vcl *vcl, struct worker *wrk)
...
+/*
+ * We do not garbage collect discarded VCL's here, that happens in VCL_Poll()
+ * which is called from the CLI thread.
+ */
+
+#define vcl_rel(vcl) do {			\
+	assert((vcl)->busy > 0);		\
+	(vcl)->busy--;				\
+	(vcl) = NULL;				\
+	} while (0)
```
Dridi, line 133: ws ocd

Also, why not a static inline function for readability?

Dridi (commented):
I'm really not convinced by the case for label caching, see main comment below.

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -107,11 +107,20 @@ VCL_Recache(struct worker *wrk, struct vcl **vclp)
...
-VCL_Ref(struct vcl *vcl)
+VCL_Ref(struct vcl *vcl, struct worker *wrk)
 {
 
 	CHECK_OBJ_NOTNULL(vcl, VCL_MAGIC);
+	CHECK_OBJ_ORNULL(wrk, WORKER_MAGIC);
 	assert(!vcl->temp->is_cold);
+
+	/* If the worker has the right vcl cached, we just take that
+	 * reference */
```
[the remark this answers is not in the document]
nigoroll: right :)

bin/varnishd/cache/cache_vrt_vcl.c
```diff
@@ -88,6 +88,12 @@ VCL_Refresh(struct vcl **vcc)
 	VCL_Update(vcc, NULL);
 }
 
+static inline int
+vcl_cancache(const struct vcl *vcl)
```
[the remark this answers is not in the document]
nigoroll: yeah, after all these years, I still find the word _cacheable_ hard to read, but we use it in many places already, so one more does not matter.

bin/varnishd/cache/cache_vrt_vcl.c
nigoroll: * It is used under `vcl_mtx`
* "autogen.des" builds with inlining disabled

bin/varnishd/cache/cache_vrt_vcl.c
nigoroll: I do not know enough about the circumstances which lead to the initial vcl caching iteration (maybe @bsdphk can tell us?), but I am well aware of the situation with labels and wanted to make incremental progress.
Regarding _jeopardizes future client tasks_ I am not quite sure what you mean. Do you think keeping a reference which is likely to be useless does any harm to future requests? I would think that, since we return the old reference and get the new one in one go, the difference would be marginal, but I agree that with labels, caching only benefits the backend side and the chances for it to be beneficial are inversely proportional to the number of labels.
Regarding the idle timeout, we could make it a parameter. Other than that I am also not quite sure where you are heading with that thought.
The other bit of context is that I have a long-term TODO List item: Backend Reference Counting. A previous iteration was rightfully criticized for introducing another central lock, so my intention is to propose re-using `vcl_mtx`.
So if we needed to improve vcl caching even further to optimize the many-labels-case, I would be interested to know in order to align the other plans.

bsdphk:
I'm timing out this ticket, because I think it has been overtaken by events.

We cache the active VCL in workers, to avoid `vcl_mtx` becoming a hotspot for all incoming requests.

Using`return(vcl)` means workers do not retire with a reference to the active VCL, and therefore each use of `return(vcl)` incurs the cost of two extra `vcl_mtx` operations.

That is not good.

Giving workers a two-slot VCL cache, can reduce that to 'at most one extra `vcl_mtx` operation', but that is a different patch than what is proposed here, which is therefore OBE.

As all number-theorists know, there are only three true quantities: Zero, One and Many, and thus the obvious next question becomes: "Why only two-slot ?"

As the size of the per worker cache approaches the number of different `return(vcl)` targets, the number of extra `vcl_mtx` operations go to zero.  That does seem like a valid optimization.

bsdphk:
I have spent the day looking at this, and the answer is that A) it's a lot of work and B) the savings are speculative at best.

Both when we start a new request and when `return(vcl)` we have to do vcl label resolution, and there's no escaping the `vcl_mtx` for that.

I did find one thing to optimize though: If a vcl has been parked on `req0` it should be recached, because it is very likely the active vcl.

nigoroll:
@bsdphk I agree with your analysis that proper VCL caching with labels is non trivial, yet this PR merely makes an opportunistic attempt to improve it.

Nevertheless, this PR proposes other improvements which are mostly relevant for the non-label case (which is still the most common one, at least in my world):

* The first commit enables vcl reference caching also for backend requests and ESI subrequests
* The third commit collapses `VCL_Recache()` into `VCL_Rel()`, cutting the number `vcl_mtx' operations in half for the former `VCL_Recache()` case

As always, I will accept your veto (if it is one), but in this case I would like to make sure that we do not throw away what I think is sensible optimization potential.

nigoroll:
I now see that I had rebased onto old code. Will look again

PR #3247, opened 2020-03-09
Allow *.body to take either a STRING or a BLOB

### LINKED ISSUE #3114: Inserting binary objects in vcl_backend_error

Dridi:
This patch series is a different take on #3114 that takes a different approach than #3123, as announced in https://github.com/varnishcache/varnish-cache/issues/3114#issuecomment-595805313.

It does mainly four things:

- Allow `set BODY [+]= STRINGS|BODY`
- Automatic cast from `STRINGS` to `BLOB`
- Migrate `std.fileread()` logic to a new VFC [1] subsystem
- Reuse VFC in a new `blob.fileread()` function

I followed the direction we agreed on and my only peeve is that VFC lives in `libvarnish` and it might not be the best venue. In order to avoid needless BLOB allocations I need VFC to hand over a BLOB directly which creates a "reverse" dependency from `libvarnish` to `vrt.h` that is not to my liking.

The other thing I'm not too happy about is the `vcc_PeekExpr` function. I had two approaches in store to deal with the alternate type:

- peeking and rewinding (the one I picked)
- evaluating the expression with both types

In hindsight, I think I picked the complicated option, so I will likely also give a try to my alternative.

Fixes #3114
Closes #3123

[1] Varnish File Cache, I figured VFIL was too crowded for that

### DIFF
--- a/bin/varnishd/cache/cache_vrt_var.c
+++ b/bin/varnishd/cache/cache_vrt_var.c
@@ -965,17 +965,29 @@ VRT_r_resp_do_esi(VRT_CTX)
 #define VRT_BODY_L(which)					\
 VCL_VOID							\
 VRT_l_##which##_body(VRT_CTX, enum lbody_e type,		\
-    const char *str, VCL_STRANDS s)				\
+    const char *str, VCL_BODY body)				\
 {								\
 	int n;							\
 	struct vsb *vsb;					\
+	VCL_STRANDS s;						\
+	VCL_BLOB b;						\
 								\
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);			\
+	AN(body);						\
 	CAST_OBJ_NOTNULL(vsb, ctx->specific, VSB_MAGIC);	\
-	assert(type == LBODY_SET || type == LBODY_ADD);		\
-	if (type == LBODY_SET)					\
+	if (type == LBODY_SET_STRING || type == LBODY_SET_BLOB)	\
 		VSB_clear(vsb);					\
+	if (type == LBODY_SET_BLOB || type == LBODY_ADD_BLOB) {	\
+		AZ(str);					\
+		b = body;					\
+		VSB_bcat(vsb, b->blob, b->len);			\
+		return;						\
+	}							\
 	if (str != NULL)					\
 		VSB_cat(vsb, str);				\
+	assert(type == LBODY_SET_STRING ||			\
+	    type == LBODY_ADD_STRING);				\
+	s = body;						\
 	for (n = 0; s != NULL && n < s->n; n++)			\
 		if (s->p[n] != NULL)				\
 			VSB_cat(vsb, s->p[n]);			\
--- a/lib/libvcc/vcc_compile.h
+++ b/lib/libvcc/vcc_compile.h
@@ -119,6 +119,8 @@ struct type {
 	const char		*tostring;
 	vcc_type_t		multype;
 	int			stringform;
+	int			bodyform;
+	int			noindent;
 };
 
 #define VCC_TYPE(UC, lc)	extern const struct type UC[1];
--- a/bin/varnishtest/tests/r03079.vtc
+++ b/bin/varnishtest/tests/r03079.vtc
@@ -26,7 +26,7 @@ client c1 {
 	expect resp.http.x-powered-by == varnishtest1002
 } -run
 
-# set BODY [+]= STRINGS;
+# set BODY [+]= STRINGS|BLOB;
 
 varnish v1 -vcl {
 	import blob;
@@ -50,11 +50,20 @@ varnish v1 -vcl {
 				set resp.body += "world";
 			}
 		}
+		if (req.url ~ "blob") {
+			set resp.body = blob.decode(HEX, encoded="1100");
+			if (req.url ~ "reset") {
+				set resp.body = blob.decode(HEX, encoded="221100");
+			}
+			if (req.url ~ "add") {
+				set resp.body += blob.decode(HEX, encoded="221100");
+			}
+		}
 		return (deliver);
 	}
 }
 
-client c1 {
+client c2 {
 	txreq -url "/synth"
 	rxresp
 	expect resp.body == hello
@@ -74,4 +83,16 @@ client c1 {
 	txreq -url "/string/add"
 	rxresp
 	expect resp.body == helloworld
+
+	txreq -url "/blob"
+	rxresp
+	expect resp.bodylen == 2
+
+	txreq -url "/blob/reset"
+	rxresp
+	expect resp.bodylen == 3
+
+	txreq -url "/blob/add"
+	rxresp
+	expect resp.bodylen == 5
 } -run
--- a/doc/sphinx/reference/vmod.rst
+++ b/doc/sphinx/reference/vmod.rst
@@ -302,6 +302,13 @@ BLOB
 	An opaque type to pass random bits of memory between VMOD
 	functions.
 
+BODY
+	C-type: ``const void *``
+
+	A type only used on the LHS of an assignment that can take
+	either a blob or an expression that can be converted to a
+	string.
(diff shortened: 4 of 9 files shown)


include/vfc.h
```diff
@@ -0,0 +1,41 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+struct vfc {
+	unsigned		magic;
+#define CACHED_FILE_MAGIC	0xa8e9d87a
+	char			*file_name;
+	struct vrt_blob		contents[1];
```
nigoroll: I see that embedding the blob saves one allocation, but in principle I would have preferred the `const` attribute from the `VRT_BLOB` typedef to make clear that it is immutable.

lib/libvcc/vcc_backend_util.c
```diff
@@ -106,17 +106,17 @@ vcc_IsField(struct vcc *tl, struct token **t, struct fld_spec *fs)
 			fs->found = t_field;
 			return;
 		}
-		VSB_cat(tl->sb, "Field ");
+		vcc_Complain(tl, "Field ");
 		vcc_ErrToken(tl, t_field);
-		VSB_cat(tl->sb, " redefined at:\n");
+		vcc_Complain(tl, " redefined at:\n");
```
nigoroll: I am not against it, but in general, I think we should handle refactoring (as introducing `vcc_Complain` and renaming `sb` -> `err_sb`) separate from such PRs.

lib/libvcc/vcc_compile.h
```diff
@@ -371,12 +374,13 @@ typedef void symwalk_f(struct vcc *tl, const struct symbol *s);
...
 /* vcc_token.c */
-void vcc_Coord(const struct vcc *tl, struct vsb *vsb,
-    const struct token *t);
+void vcc_Coord(const struct vcc *tl, struct vsb *vsb);
 void vcc_ErrToken(const struct vcc *tl, const struct token *t);
 void vcc_ErrWhere(struct vcc *, const struct token *);
 void vcc_ErrWhere2(struct vcc *, const struct token *, const struct token *);
 void vcc_Warn(struct vcc *);
+void vcc_Complain(const struct vcc *, const char *);
+void vcc_Complainf(const struct vcc *, const char *, ...);
```
nigoroll: decorate `v_printflike_` ?

lib/libvcc/vcc_compile.h
```diff
@@ -107,6 +107,7 @@ struct type {
 	const char		*tostring;
 	vcc_type_t		multype;
 	int			stringform;
+	int			stringcast;
```
nigoroll: can you explain why we need both `stringform` and `stringcast`?

lib/libvarnish/vfc.c
```diff
@@ -41,95 +42,105 @@
...
+	AZ(pthread_mutex_unlock(&vfc_mtx));
+	if (*vfcp != NULL)
+		vfc_destroy(vfcp);
 }
 
-VCL_STRING v_matchproto_(td_std_fileread)
-vmod_fileread(VRT_CTX, struct vmod_priv *priv,
-    VCL_STRING file_name)
+struct vfc *
+VFC_find(struct vfc *old, const char *file_name)
```
nigoroll: I am aware that the purpose of this is to share code between the `std` and `blob` vmods.
Yet I really do not like this to end up as an official API because it just caches the contents forever and provides no means no to. Also I provides no checks for updates etc. etc.
So _if_ we really make this move, I think we should at least mark the interface as deprecated and transient.

Before I go about enumerating all the deficits, I would rather want to point interested readers to [the UPLEX vmod_file](https://code.uplex.de/uplex-varnish/libvmod-file) where we attempt to address issues around caching, updates and file deletion.

bsdphk:
Please merge the first three commits (with due attention to @nigoroll's comments) to get them out of the way.

Dridi:
523bd18b89073dff7477419e90c96d797119c14b was made obsolete by 8a1948fa2b11a8f9d4ba2f81bf719f60ade3da36, looking at the two others.

Dridi:
We will also not need `blob.fileread()` from the last commit, meanwhile we added `std.blobread()`.

Dridi:
The ground has shifted a LOT since the last update of this pull request. Direct access to `struct vcc:sb` has expanded to more use cases and is a bit harder to contain with just `vcc_Complain()` and `vcc_Complainf()`. Those functions are needed to reduce the number of call sites where we need to inhibit error outputs.

An alternative could be to snapshot the VSB itself to let all errors write to a dead sink when we are peeking at an expression instead of what 0898c9c3c8ecabafd078e51742601b8df7b17604 does.

Dridi:
> An alternative could be to snapshot the VSB itself to let all errors write to a dead sink when we are peeking at an expression

I did just that yesterday, but with two somewhat long interruptions I wasn't able to submit it. I'm taking 5 minutes now to have this ready on Monday.

The patch series was reduced to 4 commits:

- 12f54a39d7
- 1d264ad6ed
- adaeb01701 (clean cherry pick from the old patch series)
- 86ff833418 (cherry pick from the old patch series with conflicts)

It is overall simpler but in the same spirit.

Dridi:
Force-pushed to reduce to an even less-intrusive single commit: f620709846cd0dd596b0b38890501bf41a2bd044.

bsdphk:
That is *much* neater!

Go from here.

Dridi:
Force-pushed to add docs in `vrt.h` and `vmod.rst`, with a bonus polish commit around strands.

Dridi:
Force-pushed to respect alphabetic order in `vmod.rst`, and rebased against current master.

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -965,17 +965,27 @@ VRT_r_resp_do_esi(VRT_CTX)
...
 		VSB_clear(vsb);					\
+	if (type == LBODY_SET_BLOB || type == LBODY_ADD_BLOB) {	\
+		AZ(str);					\
+		b = body;					\
+		VSB_bcat(vsb, b->blob, b->len);			\
+		return;						\
+	}							\
 	if (str != NULL)					\
 		VSB_cat(vsb, str);				\
+	s = body;						\
```
nigoroll, line 988: (minor) `assert(type == LBODY_SET_STRING || type == LBODY_ADD_STRING)` ?

nigoroll (approved):

PR #3263, opened 2020-03-26
Use CircleCI 'machine' executors with QEMU emulation to build packages for different architectures

martin-g:
This way CircleCI could be used to build, test and create multiarch packages.

Alpine Linux aarch64 packaging is currently commented out because it fails while installing the dependencies. Needs more debugging, but the issue must be in the base Docker image because the same Shell script works just fine on amd64.

Discussion at the mailing list: https://varnish-cache.org/lists/pipermail/varnish-dev/2020-March/009523.html


martin-g:
I'll add some more `echo`s in the scripts, for easier debugging when needed.
I'll also add a README file in `.circleci` folder explaining how all this works, and how to extend it for more architectures.

`distcheck` job was not setup in the current/previous version. Should I enable it after `dist` job ?

Also I was thinking whether it is a good idea to add jobs for testing the installation of the produced packages, i.e. to start a Docker instance for the respective architecture and run `dpkg -i varnish-*.deb`, for example. What do you think ?

Dridi:
This pull request does too much (for me) in one commit:

- extract some operations into reusable scripts
- reorganize existing jobs
- add aarch64 jobs

I would personally like to see commits of at least this granularity, more if I missed additional steps.

I find it too hard to review in one block, can you try to split the work into multiple commits?

martin-g:
> can you try to split the work into multiple commits

OK. I will try to do this!

martin-g:
The problem with the Alpine AARCH64 image is the following:

1. docker run -it --rm alpine:3 sh 
1. apk add sudo 
1. adduser xyz
1. echo "xyz ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers
1. su xyz
1. sudo ls -la

Here step 6 works fine - it prints the files and folders.

But if I change the Docker image at step 1) to `multiarch/alpine:aarch64-edge` then step fails with:
```
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
```

`suid` bit is there:
```
ls -la /usr/bin/sudo
-rwsr-xr-x    1 root     root        124504 Jan 30 23:12 /usr/bin/sudo
```
The file system on `/` is:
```
$ mount
overlay on / type overlay (rw,relatime,lowerdir=/var/lib/docker/overlay2/l/56YRL27MZBOOLIEBZPAXBBOUIP:/var/lib/docker/overlay2/l/UBFCWLE6VDFUICTAFQUYR5IYIZ:/var/lib/docker/overlay2/l/II33U7S4SBCCM5NSUWCRP4UEC7,upperdir=/var/lib/docker/overlay2/706a3e26cd377344d8cebd708cb9fd29ce45752c63624df77386748b9c9cac7d/diff,workdir=/var/lib/docker/overlay2/706a3e26cd377344d8cebd708cb9fd29ce45752c63624df77386748b9c9cac7d/work,xino=off)
```

Anyone have an idea what could be the reason ?

martin-g:
I've found the issue!
The QEMU registration needed one more argument: `--credential yes`.
I've also simplified the job by using references.

martin-g:
I've simplified the PR by generating the Dockerfile on the fly by passing the name of the base Docker image as a parameter.

I didn't yet split the first big commit into smaller ones! I will try to do it at the weekend.

The idea of the PR is:
1) extract the DEB/RPM/APK specific code into Shell script files
2) Get pkg-varnish-cache, as before
3) make dist, as before
4) a new job for the packaging for the different distros and architectures
5) collect the packages, as before

.circleci/make-rpm-packages.sh
```diff
@@ -0,0 +1,87 @@
...
+      dnf install -y 'dnf-command(config-manager)'
+      yum config-manager --set-enabled PowerTools
+      yum install -y diffutils python3-sphinx
+  else
+      yum install -y python-sphinx
+  fi
+fi
+
+yum install -y epel-release
+yum install -y automake jemalloc-devel git libedit-devel libtool libunwind-devel make pcre-devel python3 sudo rpm-build yum-utils
```
gquintard, line 27: this is wrong, the `yum-builddep` is responsible for installing the dependencies listed in the spec file.

(you do need `epel-release` and `PowerTools` repos to be enabled though)

.circleci/make-rpm-packages.sh
```diff
@@ -0,0 +1,87 @@
...
+
+
+export DIST_DIR=build
+
+# XXX: we should NOT have to do that here, they should be in the
+# spec as BuildRequires
+yum install -y make gcc
```
gquintard, line 34: let's open a PR on https://github.com/varnishcache/pkg-varnish-cache/ to fix it once and for all

.circleci/make-rpm-packages.sh
```diff
@@ -0,0 +1,87 @@
...
+
+adduser varnish
```
gquintard, line 36: unneeded?

.circleci/config.yml
```diff
@@ -39,13 +104,21 @@ jobs:
             fi
             ./autogen.des --quiet
             make dist -j 16
+            mkdir -p ~/workspace/tools
```
gquintard, line 107: why the switch to a subdir? since we know exactly what to save, why not grab them in place?

.circleci/config.yml
```diff
@@ -1,5 +1,69 @@
...
+        type: string
+      image:
+        description: the base Docker image for Dockerfile
+        type: string
+      workspace_extra:
+        description: extra paths for the workspace root
+        type: string
+        default: ""
+  package_job: &package_job
+    <<: *pkg_parameters
```
gquintard, line 35: I think there an extra level of indirection that we can cut here. `pkg_parameters` is never used elsewhere

.circleci/config.yml
```diff
@@ -62,14 +135,51 @@ jobs:
...
-          root: .
+          root: ~/workspace
           paths:
             - debian.tar.gz
             - redhat.tar.gz
             - alpine.tar.gz
+      - store_artifacts:
+          destination: ~/workspace
+          path: ~/workspace
+  x64-centos-7:
```
gquintard, line 151: wrong approach, instead, name only one job `package`, then, in the workflow, use the name attribute:
```
- package:
  name: aarch64-centos-7
  rest of the parameters
```

.circleci/config.yml
```diff
@@ -62,14 +135,51 @@ jobs:
             git clone https://github.com/varnishcache/pkg-varnish-cache.git .
             git checkout << pipeline.parameters.pkg-commit >>
             tar cvzf debian.tar.gz debian --dereference
-            tar cvzf redhat.tar.gz redhat --dereference 
+            tar cvzf redhat.tar.gz redhat --dereference
             tar cvzf alpine.tar.gz alpine --dereference
+            mkdir -p ~/workspace
```
gquintard, line 140: same, the use of an extra directory seems a bit gratuitous?

.circleci/config.yml
```diff
@@ -174,285 +287,162 @@ jobs:
...
-      - store_artifacts:
-          path: packages/<< parameters.dist >>/<< parameters.release >>
-          destination: ./
   collect_packages:
     docker:
       - image: centos:7
     steps:
       - attach_workspace:
-          at: ~/project
+          at: ~/workspace
```
gquintard, line 296: why change this too?

.circleci/make-deb-packages.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+set -eux
+
+export DEBIAN_FRONTEND=noninteractive
+export DEBCONF_NONINTERACTIVE_SEEN=true
+apt-get update
+apt-get install -y autoconf automake build-essential graphviz libncurses-dev libtool dpkg-dev ca-certificates debhelper devscripts equivs make gcc pkg-config libunwind-dev python3-docutils python3-sphinx ncurses-dev libpcre3-dev libedit-dev libjemalloc-dev apt-utils
```
gquintard, line 8: let `debian/control` specify these

.circleci/make-deb-packages.sh
```diff
@@ -0,0 +1,55 @@
...
+if test "x$PARAM_RELEASE" = "x"; then
+    echo "Env variable PARAM_RELEASE is not set! For example PARAM_RELEASE=8, for CentOS 8"
+    exit 1
+elif test "x$PARAM_DIST" = "x"; then
+    echo "Env variable PARAM_DIST is not set! For example PARAM_DIST=centos"
+    exit 1
+fi
+
+
+adduser --disabled-password --gecos "" varnish
```
gquintard, line 23: do we need it?

.circleci/make-rpm-packages.sh
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
+
+set -eux
+
+echo "PARAM_RELEASE: $PARAM_RELEASE"
+echo "PARAM_DIST: $PARAM_DIST"
+
+if test "x$PARAM_RELEASE" = "x"; then
```
gquintard, line 8: can we leave the 70s behind and just use `[ -z "$PARAM_RELEASE" ]` or is there a reason to continue using this pattern?

.circleci/make-rpm-packages.sh
```diff
@@ -0,0 +1,87 @@
...
+
+echo "Untar redhat..."
+tar xavf /workspace/redhat.tar.gz -C $DIST_DIR
+
+echo "Untar orig..."
+tar xavf /workspace/varnish-*.tar.gz -C $DIST_DIR --strip 1
+
+echo "Build Packages..."
+# use python3
+sed -i '1 i\%global __python %{__python3}' "$DIST_DIR"/redhat/varnish.spec
```
gquintard, line 52: I still hate that one, now that `centos:8` is out we may have a better case. Or even better, since `python3` is a requirement now, we may not even need it anymore.

I'll check

.circleci/make-rpm-packages.sh
```diff
@@ -0,0 +1,87 @@
...
+adduser varnish
+chown -R varnish:varnish /workspace
+
+cd /varnish-cache
+rm -rf $DIST_DIR
+mkdir $DIST_DIR
+
+
+echo "Untar redhat..."
+tar xavf /workspace/redhat.tar.gz -C $DIST_DIR
```
gquintard, line 45: I'd rather not have tight coupling and preexisting knowledge, can we just operate out of the current directory?

martin-g:
> this is a big PR, so I may have missed things. There are clearly a few things that need to change, but I think that a good half of my comments are really about collecting information about the usefulness of some changes

Thank you for the review, @gquintard !
I agree that this PR is rather big!
I've tried to refactor the code and preserve as much as possible of the old scripts.
Most of your comments are about things which I've just extracted into a Shell script so that they can be reused by the different build matrices, e.g. `make-deb-packages.sh` is reused for Ubuntu (xenial, bionic) & Debian (stretch, buster & bullseye), and also for both x86_64 and aarch64.

If you think that this PR is a good place to optimize the old code then I could do it here! Or I could open new PRs on top of this PR to keep them smaller. Just let me know which way you prefer!

I've made few attempts to split the first commit into smaller ones, as requested by @Dridi, but I was not able to do it properly. All attempts were either in unfinished state or again with rather big diff. 

I've added a README in `.circleci/` folder that explains how the new setup works. It should be helpful for the next person dealing with these matters. 

I'll go now over your comments one by one.

.circleci/make-rpm-packages.sh
martin-g: I'll move as many as possible to `redhat/varnish.spec` but `automake` needs to be installed here because of
   `VERSION=$("$DIST_DIR"/configure --version | awk 'NR == 1 {print $NF}')$WEEKLY` 
at line 54

.circleci/config.yml
martin-g: I've tried to keep the old file structure.
`varnish-x.y.z.tar.gz` and `.is_weekly` are stored in `~/workspace/`. 
But `mv tools/*.suppr ~/workspace/tools/` copies just the `.suppr` files, not all files under `tools/`.

.circleci/config.yml
martin-g: Done!

.circleci/config.yml
martin-g: This looks cool but then what would be the syntax for `requires` ?
```
- collect_packages:
          requires:
            - x64-ubuntu-xenial
            - x64-ubuntu-bionic
            - aarch64-ubuntu-xenial
            ...
```

.circleci/config.yml
martin-g: Maybe I don't quite understand how CircleCI's `persist_to_workspace` works.
Here is how I understood it:
```
            ...
            mkdir -p ~/workspace
            mv *.tar.gz ~/workspace
      - persist_to_workspace:
          root: ~/workspace
          paths:
            - debian.tar.gz
            - redhat.tar.gz
            - alpine.tar.gz
```

I create `~/workspace` as a folder that will contain the outcomes from all steps and all following steps may use any of the outcomes. This way is simpler to debug the steps - I just list the contents of `~/workspace` at the beginning of each step and it is easier to understand why some step fails.
Then I move the .tar.gz archives into it and tell CircleCI to export this folder. Maybe I can remove the `paths` parameter. I guess this way CircleCI will export the complete contents of the folder, not listing the files explicitly.

What do you suggest to simplify it ?

.circleci/config.yml
martin-g: I've decided to use `~/workspace` as the workspace folder for all steps. The outcome just accumulates there and any step can re-use any outcome from any previous step.

.circleci/make-deb-packages.sh
martin-g: it was there before. I just moved it from config.yml to the Shell script. I'll see whether it can be removed.

.circleci/make-rpm-packages.sh
martin-g: No reason! Changed it!

.circleci/make-rpm-packages.sh
martin-g: I guess we can. The other scripts do not use `-C $DIST_DIR`. I just preserved the old code.
I will change it!

.circleci/config.yml
gquintard: true, but you are already using a glob pattern in `persist_to_workspace` to only capture the `suppr` files
it's true that you need it for `store_artifacts` since you don't know the name of the files buI would argue that we possibly don't need the files as artifacts though, and we absolutely don't care about the `suppr` files or `.weekly` as artifacts.

.circleci/config.yml
gquintard: yes, that's the one, depends on the names that you specified for each job

.circleci/config.yml
gquintard: ok, I understand the approach now. My problem is that `persist_to_workspace` and `store_artifacts` and used pretty much the same way, which is a bit confusing.

I guess that renaming `workspace` to `saved` or something like that would have reduced confusion, but we are down to bikeshedding now, so it's fine

.circleci/make-rpm-packages.sh
gquintard: I don't think you do, `configure` should just be a `shell` script

.circleci/make-rpm-packages.sh
gquintard: I meant `/workspace/`. `DISTDIR` is a parameter that we could get rid of, but at least we can change it.

The fact that `/workspace/` is hardcoded means we can't run that script in parallel (outside of `circleci`). I think it makes more sense, if possible to go with

```
tar xavf redhat.tar.gz -C $DIST_DIR
```

martin-g: I've tried it with https://github.com/martin-g/varnish-cache/commit/1e1bb590ec64b4e5eaad57b9992cb860f0ab42b2 but it failed: https://app.circleci.com/pipelines/github/martin-g/varnish-cache/191/workflows/0b67272b-59ad-44df-9a4f-dd8d758e3b20/jobs/1039
It seems `rpmbuild` needs both a folder and a .tgz with the same name for `srcname`. 
It seems this is the reason redhat.tar.gz and varnish-cache.tar.gz to be extracted in `build/` folder.

.circleci/make-deb-packages.sh
martin-g: Removed it!

.circleci/config.yml
martin-g: Done with https://github.com/martin-g/varnish-cache/commit/4ffeb7dbb85c4cfede0e475a75da0bc55d9b20f9!

.circleci/make-rpm-packages.sh
martin-g: Removed it!

.circleci/config.yml
```diff
@@ -174,285 +250,178 @@ jobs:
...
+            - x64-debian-bullseye
+            - aarch64-debian-bullseye
+            - x64-debian-buster
+            - aarch64-debian-buster
+            - x64-debian-stretch
+            - aarch64-debian-stretch
+            - x64-centos-7
+            - x64-centos-8
+            - aarch64-centos-7
+#            - aarch64-centos-8
```
gquintard, line 422: `centos:8` doesn't work?

.circleci/make-rpm-packages.sh
gquintard: and that is fine, but I don't want to have to know that `redhat.tar.gz` is in `/workspace`. Same thing for the dist tarball.

```
tar xavf redhat.tar.gz -C $DIST_DIR
tar xavf varnish-*.tar.gz -C $DIST_DIR --strip 1
```

is perfectly fine, my beef with the current code is that it hardcodes the `/wokspace` when there's no need for it

.circleci/config.yml
martin-g: No. It should be an ARM64 image. `centos:8` is the x86_64 one.
I am experimenting now with `--persistent yes` flag of multiarch/qemu-user-static. With this flag I was able to run `docker run -it --rm arm64v8/centos:8 uname -m` here. But step `Activate QEMU` fails at CircleCI with:
```
docker run -it --rm --privileged multiarch/qemu-user-static --reset --credential yes --persistent yes
Unable to find image 'multiarch/qemu-user-static:latest' locally
latest: Pulling from multiarch/qemu-user-static

aa22dec6: Pulling fs layer 
7f1a5b79: Pulling fs layer 
5179ae11: Pulling fs layer 
9da9c97d: Pulling fs layer 
Digest: sha256:54ed5674b4d96a2200a75373993a48d8c5dc3433ed4cdeef95a0e947e88bb05f
Status: Downloaded newer image for multiarch/qemu-user-static:latest
^@^@Setting /usr/bin/qemu-alpha-static as binfmt interpreter for alpha
sh: write error: Invalid argument
Setting /usr/bin/qemu-arm-static as binfmt interpreter for arm
sh: write error: Invalid argument
```

https://app.circleci.com/pipelines/github/martin-g/varnish-cache/198/workflows/74a620b9-d05e-4ec7-a24b-10e579225bac/jobs/1192

.circleci/config.yml
martin-g: The use the artifacts in the intermediate steps as a debug tool - this is the only way I've found to be able to inspect what a given step produces as an outcome. I can remove it if it is confusing.
Only `collect_packages` step produces the final packages.tar.gz with all packages inside it as a meaningful artifact.

.circleci/make-rpm-packages.sh
martin-g: I will work on this now!

.circleci/make-rpm-packages.sh
martin-g: redhat/varnish.spec has this: `Requires:  python(abi) >= 3.4`
It seems python3 is a hard requirement.

.circleci/config.yml
martin-g: I've found the problem. It was the CircleCI `machine` executor uses Ubuntu 14.04 by default and kernel 4.4 which is too old for QEMU. I've changed it to Ubuntu 16.04 with 4.14 and now I use Docker images produced by `arm64v8` DockerHub organization, not from `multiarch` org.

.circleci/make-rpm-packages.sh
martin-g: Any idea what this error means:
```
+ /usr/lib/rpm/redhat/brp-strip-static-archive /usr/bin/strip
+ /usr/lib/rpm/brp-python-bytecompile /usr/bin/python 1
Compiling /root/rpmbuild/BUILDROOT/varnish-20200409.weekly-0.0.el7.x86_64/usr/share/varnish/vmodtool.py ...
  File "/usr/share/varnish/vmodtool.py", line 231
    print("ERROR: " + txt, file=sys.stderr)
                               ^
SyntaxError: invalid syntax

error: Bad exit status from /var/tmp/rpm-tmp.1T9NKA (%install)
```
https://app.circleci.com/pipelines/github/martin-g/varnish-cache/210/workflows/ba13bd1f-5cc6-42b7-8c1d-429632ad96ce/jobs/1354

martin-g: The syntax is OK with Python3, but fails with Python2.
It is because I commented out this line:
```
sed -i '1 i\%global __python %{__python3}' "$DIST_DIR"/redhat/varnish.spec
```
I'll revert this change and leave it for later

martin-g: https://github.com/varnishcache/pkg-varnish-cache/pull/138

The changes in Varnish Cache are at: https://github.com/martin-g/varnish-cache/pull/1/files
They can be merged to this PR only after https://github.com/varnishcache/pkg-varnish-cache/pull/138

.circleci/config.yml
```diff
@@ -174,285 +251,177 @@ jobs:
...
-            - packages/<< parameters.dist >>/<< parameters.release >>/*.rpm
-      - store_artifacts:
-          path: packages/<< parameters.dist >>/<< parameters.release >>
-          destination: ./
   collect_packages:
     docker:
       - image: centos:7
     steps:
       - attach_workspace:
-          at: ~/project
```
martin-g, line 395: @gquintard I am trying to get rid of `~/workspace` but I don't understand how the old `~/project` worked.
In the previous version lines 383-389 put `packages` into the workspace and the artifacts.
Then here, at line 395, the workspace is attached at `~/project` and later at line 399 the packages.tar.gz is created from `packages` folder. 
Why it is just `packages` but not `~/project/packages` ?
I don't see `~/project` being used later at all and I do not understand how `packages` magically appears in the current folder at line 399.

.circleci/config.yml
gquintard: since we can rerun a job and log into the container, I don't feel the artifacts are that useful. A borked file is often not enough and you need to see the broader environment, so I'd do away with intermediate file

.circleci/make-rpm-packages.sh
gquintard: I know :-) but i still had issues last time I tried to change the packaging, I just can't remember if it was pre or post `python3` migration

martin-g: See my PR in pkg-varnish-cache. I've moved this line directly in varnish.spec and the build is green.

.circleci/config.yml
martin-g: I'll remove the usage of store_artifacts in the intermediate steps!

.circleci/config.yml
martin-g: I got it! 
~/project is the PWD for the jobs.
I will replace ~/workspace with ~/project.

gquintard: in which case you don't even need to specify `~/project` since, unless specified, persisted objects paths are looked for in the workdir

martin-g: This didn't work. If I don't specify 
```
run:
   - attach_workspace
       at: ~/project
```
then the `packages` folder is not in CWD and there is nothing to tar.

martin-g:
I think I've addressed all issues. Please take a look at the new commits and the final diff.

martin-g:
Anything else I could improve/simplify here ?

gquintard:
Hi,

Sorry, haven't had time to look at it again, but it's on my plate.

.circleci/config.yml
```diff
@@ -87,7 +143,7 @@ jobs:
     working_directory: /workspace
     steps:
       - run:
-          name: Possibly activate centos:8 extra repos
+          name: Possibly << parameters.dist >>:<< parameters.release >> extra repos
```
gquintard, line 146:
```suggestion
          name: Possible << parameters.dist >>:<< parameters.release >> extra repos
```

.circleci/config.yml
```diff
@@ -174,285 +231,175 @@ jobs:
...
   version: 2
   build:
     jobs:
       - dist
       - tar_pkg_tools
-      - build_debs:
-          name: build_debian_stretch
-          dist: debian
-          release: stretch
+      - package:
```
gquintard, line 257: that whole list is going to be condensed with the newly added matrix support. You can have a look, or we can do that later on

.circleci/make-deb-packages.sh
gquintard: this still needs to go into `debian/control`

.circleci/make-rpm-packages.sh
gquintard: this hasn't been fixed yet. And most deps should be handled in the spec, not here.

gquintard:
there are a couple of points that still need to be addressed, but no new blockers

martin-g:
I've sent a PR at pkg-varnish-cache.
I cannot remove those before it is merged. See one of my comments above.
I'm on my mobile now and I can't give you a link to the comment.

gquintard:
ok, so focus need to go to that repo, i'll see what I can do

-- 
Guillaume Quintard

martin-g:
This is the PR: https://github.com/varnishcache/pkg-varnish-cache/pull/138

Once it is merged I can apply these changes here: https://github.com/martin-g/varnish-cache/compare/feature/create-aarch64-packages2...martin-g:feature/optimizations-from-pr-3263

martin-g:
By the way, I've written a blog article about this PR at https://medium.com/@martin.grigorov/building-linux-packages-for-different-cpu-architectures-with-docker-and-qemu-d29e4ebc9fa5?sk=782d0cb2c5bfde843d7742e5e9f1b684

.circleci/config.yml
martin-g: Done!

.circleci/config.yml
martin-g: I am not sure this will work in our case.
The new `matrix` support is a Cartesian product of all parameters. 
So it is going to look something like:
```
      - package:
          matrix:
            parameters:
              arch: ["aarch64", "x64"]
              dist: ["ubuntu", "debian", "centos", "alpine"]
              release: ["bionic", "xenial", "bullseye", "buster", "stretch", "7", "8", "3"]
              ext: ["deb", "rpm", "apk"]
              image: ["<< parameters.image_prefix >><< parameters.dist >>/<< parameters.release >>:<< parameters.version >>"]
              image_prefix: ["arm64v8/", ""]
              <<: *pkg_req
            exclude:
              - dist: ["ubuntu", "debian"]
                ext: ["rpm", "apk"]
              - dist: ["centos"]
                ext: ["deb", "apk"]
              - dist: ["alpine"]
                ext: ["rpm", "deb"]
              - arch: "aarch64"
                image_prefix: ""
              - arch: "x64"
                image_prefix: "arm64v8/"
```
And I didn't add exclusions for ubuntu+debian versions and vise versa.
IMO it is becoming more complicated with all the exclusions.
Also I needed to give up on the `-slim` versions of the Debian Docker images.

odidev:
Hi,  
I have tested the changes for arm64 debian platform it is working fine. It will be helpful if this PR is merged.

gquintard:
varnishcache/pkg-varnish-cache#138 has been merged, @martin-g , you can clean this one and we'll be good to go

martin-g:
@gquintard I've merged the latest changes!

Dridi:
The commit history isn't clean, you have merge commits (we avoid those) and it's currently conflicting with the target branch.

martin-g:
@Dridi What is your preferred way to get rid of the merge commits ?
I'm not sure it is feasible to do anything with `git rebase -i` at that point.
The simplest way for me would be to create a new branch from master and copy the contents of `.circleci/` folder from this PR. But then there will be one huge commit with all the changes.

Dridi:
I can try to reorganize the pull request myself if you allowed maintainers to make updates.

martin-g:
I didn't find a setting in my fork to allow all of you to collaborate in the fork.
So I've invited you and @gquintard . If anyone else needs access please let me know!

.circleci/config.yml
gquintard: arf, I'm now realizing that images are going to be an issue. You're right, let's keep it out of this PR

gquintard:
I'll let @Dridi handle that, but I'm personally in favor of just squashing all this in one single commit. The changes are now much clearer than they where when we started started (kudos to Martin) so I see no harm.

martin-g:
I've just found the setting `Allow edits by maintainers` for this particular PR and it was already checked, so all members of Varnish Cache team in GitHub should be able to push commits to my branch!

martin-g:
I've created a new Git branch and cherry-picked all non-merge commits to it.
If https://app.circleci.com/pipelines/github/martin-g/varnish-cache/241/workflows/57d14914-410a-424f-8fc4-03ff184b6178 passes then I will open a new PR and close this one.

martin-g:
Closing this PR in favour of https://github.com/varnishcache/varnish-cache/pull/3313

PR #3265, opened 2020-04-01
Replace vxid_lock with atomic fetch and add

fwsGonzo:
Removes the lock in favor of a single atomic operation (on x86). I looked at locks and they provide some additional features, so perhaps that is something you want to keep?


nigoroll:
I guess the two common questions to ask here are:
* do you have any data on the performance benefit?
* is this portable / how do get to a portable solution?

Dridi:
If we go for it, it should probably be guarded with `#ifdef`s.

fwsGonzo:
This header selects the either C11 stdatomic or implements using the older sync builtins (but supported all the way back to GCC 4.1):
https://github.com/freebsd/freebsd/blob/master/sys/sys/stdatomic.h#L316

The alternative of using inline asm is in my opinion a code smell, especially with all this existing support.

bsdphk:
A change like this needs to be backed by benchmark results which show that the disadvantages in terms of complexity and debugging is amply compensated by improved performance.

I close this ticket as I am pretty convinced that burden cannot be lifted in this case, but it can be reopened anytime with measurements proving me wrong.

Issue #3273, opened 2020-04-03
 Assert error in vbf_stp_condfetch: Condition((vbf_beresp2obj(bo)) == 0) not true


ema:
## Current Behavior
We have seen the following crash in production:

```
  Apr 02 12:13:42 cp3061 varnishd[29726]: Child (29751) died signal=6
  Apr 02 12:13:42 cp3061 varnishd[29726]: Child (29751) Panic at: Thu, 02 Apr 2020 12:13:42 GMT
                                        Assert error in vbf_stp_condfetch(),
        cache/cache_fetch.c line 797: Condition((vbf_beresp2obj(bo)) == 0) not true.
```
We are running a (modified) 5.1.3, but the issue affects master too.

The crash happens if: (a) transient storage is limited and full (b) main storage is also full and nuking fails (c) varnish issues a conditional backend fetch.

This VTC triggers the panic:

```
varnishtest "vbf_stp_condfetch crash"

server s1 {
    rxreq
    expect req.url == "/transient"
    txresp -bodylen 1048400

    rxreq
    expect req.url == "/malloc"
    txresp -hdr "Cache-Control: max-age=2" -hdr "Last-Modified: Fri, 03 Apr 2020 13:00:01 GMT" -bodylen 1048300

    rxreq
    expect req.http.If-Modified-Since == "Fri, 03 Apr 2020 13:00:01 GMT"
    expect req.url == "/malloc"
    txresp -status 304
} -start

varnish v1 \
 -arg "-s Transient=default,1m" \
 -arg "-s malloc,1m" \
 -arg "-p nuke_limit=0" \
 -syntax 4.0 \
 -vcl+backend {
    sub vcl_backend_response {
        if (bereq.url == "/transient") {
            set beresp.storage = storage.Transient;
            # Unset Date header to not change the object sizes
            unset beresp.http.Date;
        } 
    }
} -start

varnish v1 -cliok "param.set debug +syncvsl"

delay .1

client c1 {
    # Fill transient
    txreq -url "/transient"
    rxresp
    expect resp.status == 200
} -run

delay .1

varnish v1 -expect SM?.Transient.g_bytes > 1048000
varnish v1 -expect SM?.Transient.g_space == 8

client c1 { 
    # Fill malloc
    txreq -url "/malloc" -hdr "If-Modified-Since: Fri, 03 Apr 2020 12:00:01 GMT"
    rxresp
    expect resp.status == 200

    delay 3
} -run

[...]

PR #3274, opened 2020-04-03
vbf_stp_condfetch: handle storage error instead of crashing

ema:
Fixes: #3273


nigoroll:
merged as d3b6409ea3f52851aa63fde8120af5cfc430db8b - thank you

FTR:
* I think it is correct that we do not close the backend connection in this case. There is nothing wrong with it 
* any anyways, as @ema pointed out, the error handling is just the same as a few lines above

PR #3277, opened 2020-04-06
Tie workspace logs to their transactions

Dridi:
That is, when it's a request or backend transaction, otherwise it's
still logged with VXID 0. The goal is to allow a correlation between
VCL statements and their workspace footprint by sharing the same
transaction instead of being split between raw and vxid scopes. That
kind of debugging becomes possible when the workspace debug flag is
used and VCL_trace records are not masked.

Given the verbosity of both workspace and trace logs, we would need
new tools to process transactions for this purpose and until we can
link everything we have a chicken-egg problem.

This change introduces a new thread-local variable to keep track of a
worker's VSL when it is given one. Not all workspace operations from
a transaction are logged, the missing bits are early initialization
and teardown: in other words before and after VCL execution, not VCL
statements that could be debugged using such logs.

Grabbing the VSL from the thread-local req or bo isn't possible, the
VSL may not be ready to use. We could introduce a VSL_IsReady() or
some other means to find whether we can use it, but I felt that we
had a clear signal in the code regarding when we can start doing that
via the worker.

This introduces a dependency between workspace and workers, so the
latter need to be initialized earlier.

### DIFF
--- a/bin/varnishd/cache/cache_wrk.c
+++ b/bin/varnishd/cache/cache_wrk.c
@@ -738,6 +738,25 @@ static struct cli_proto debug_cmds[] = {
 	{ NULL }
 };
 
+void
+WRK_Log(enum VSL_tag_e tag, const char *fmt, ...)
+{
+	struct worker *wrk;
+	va_list ap;
+
+	AN(fmt);
+
+	wrk = THR_GetWorker();
+	CHECK_OBJ_ORNULL(wrk, WORKER_MAGIC);
+
+	va_start(ap, fmt);
+	if (wrk != NULL && wrk->vsl != NULL)
+		VSLbv(wrk->vsl, tag, fmt, ap);
+	else
+		VSLv(tag, 0, fmt, ap);
+	va_end(ap);
+}
+
 /*--------------------------------------------------------------------
  *
  */
--- a/bin/varnishd/cache/cache_main.c
+++ b/bin/varnishd/cache/cache_main.c
@@ -408,6 +408,8 @@ child_main(int sigmagic, size_t altstksz)
 
 	ObjInit();
 
+	WRK_Init();
+
 	VCL_Init();
 	VCL_VRT_Init();
 
@@ -431,8 +433,6 @@ child_main(int sigmagic, size_t altstksz)
 	STV_open();
 
 	VMOD_Init();
-
-	WRK_Init();
 
 	BAN_Compile();
 
--- a/bin/varnishd/cache/cache_varnishd.h
+++ b/bin/varnishd/cache/cache_varnishd.h
@@ -525,6 +525,7 @@ void VMOD_Panic(struct vsb *);
 /* cache_wrk.c */
 void WRK_Init(void);
 void WRK_AddStat(const struct worker *);
+void WRK_Log(enum VSL_tag_e, const char *, ...);
 
 /* cache_ws.c */
 void WS_Panic(struct vsb *, const struct ws *);
@@ -577,4 +578,10 @@ void SMP_Ready(void);
 	do {							\
 		if (DO_DEBUG(debug_bit))			\
 			VSL(SLT_Debug, (id), __VA_ARGS__);	\
+	} while (0)
+
+#define DSLb(debug_bit, ...)					\
+	do {							\
+		if (DO_DEBUG(debug_bit))			\
+			WRK_Log(SLT_Debug, __VA_ARGS__);	\
 	} while (0)
--- a/bin/varnishd/cache/cache_ws.c
+++ b/bin/varnishd/cache/cache_ws.c
@@ -44,7 +44,7 @@ WS_Assert(const struct ws *ws)
 {
 
 	CHECK_OBJ_NOTNULL(ws, WS_MAGIC);
-	DSL(DBG_WORKSPACE, 0, "WS(%p) = (%s, %p %zu %zu %zu)",
+	DSLb(DBG_WORKSPACE, "WS(%p) = (%s, %p %zu %zu %zu)",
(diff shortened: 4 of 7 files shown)


nigoroll:
I understand the usefulness of something like this for debugging workspace-related issues (or even largely unrelated issues, for which the workspace operations may provide indirect insight) and I am not against it.
Yet, I do have some comments/questions:
* For gaining insight on workspace consumption, I think the detailed debugging is really not the right tool. We should add the high watermark tracking, which could also be used by some vcl-centric trace facility
* While I see that you want to keep the impact low by using thread local storage (in this context, that is what TLS stands for), I am not sure if it is really the best option. What if, alternatively, we spent a `vsl` pointer for the `ws`? We could check `DBG_WORKSPACE` only when the workspace is initialized, which would make the other debug checks lokal to our `struct ws`.
* For glibc/nptl, the TLS lookup is a two-level array indirection, which should be quite efficient, but can we safely assume across platforms that TLS is efficient? In particular, could adding another key for basically all theads have a performance impact?

Dridi:
> For gaining insight on workspace consumption, I think the detailed debugging is really not the right tool. We should add the high watermark tracking, which could also be used by some vcl-centric trace facility

Adding a workspace field to the `VCL_trace` record could be an option, but only looking after a watermark would defeat the idea of learning the workspace footprint of a given VCL statement (for a given alignment of planets).

That would be overall much less verbose to combine VCL tracing and workspace accounting if the former included the latter.

> What if, alternatively, we spent a `vsl` pointer for the `ws`?

That was my first suggestion before producing any code and was considered too costly (since `struct ws` is itself allocated on the workspace it manages).

nigoroll:
@Dridi I was thinking more of something like `std.workspace()` which would log
* the amount of workspace currently used
* the high water mark so far
* delta since it was last called

But if you are already using the `VCL_trace` facility, I agree that adding something like this to it was also an option.

I am sorry to have missed that adding the `vsl` pointer was veto'ed. If that is the case, I am fine with your proposal. Otherwise let's consider if the cache locality argument has any merits to it

Dridi:
> I was thinking more of something like `std.workspace()` which would log [...]

I'm strongly against this for the purpose of this patch, that would remove the ability to trace existing VCL without modifying it.

I think your review convinced me that we should add two fields to `VCL_trace` and let `VPI_count` collect the "used" and "size" values of `ctx->ws`. Relying on freeform highly-verbose `Debug` records is probably a good recipe for self-induced pain down the road.

nigoroll:
@Dridi, we do agree. I wrote:
> But if you are already using the `VCL_trace` facility, I agree that adding something like this to it was also an option.

Integrating workspace usage information in the traces was the generic solution, which could also be used for automatic analysis.

Yet explicit trace points can be helpful for other scenarios.

bsdphk:
As far as the `cache_ws.c` part of this PR, I'm fine, and lets get that in first and separately.

However I think the thread-specific part is wrong.

For other debugging purposes we have `THR_[SG]et_{Name|BusyObj|Request}()`, which seems to indicate that maybe what we need is `THR_[SG]et_Worker()` ?

Dridi:
> As far as the `cache_ws.c` part of this PR, I'm fine, and lets get that in first and separately.

OK, so I assume that would be the logic centralization in `ws_Debug()`.

> However I think the thread-specific part is wrong.
> 
> For other debugging purposes we have `THR_[SG]et_{Name|BusyObj|Request}()`, which seems to indicate that maybe what we need is `THR_[SG]et_Worker()` ?

Something that `cache_wrk.c` could manage in a single location, in `WRK_Thread()` for example before and after calling `Pool_Work_Thread()`.

I have another work in progress around workspace, I would like to defer this change until after the other one is submitted. This one is simpler to rebase onto the next one :)

@bsdphk, any opinion about adding a field to `VCL_trace` to keep track of workspace?

Dridi:
Note to self: revisit with 3fc7847828963d0870b2eb4f4880512209304bfb in mind.

Dridi:
I finally took the chance of revisiting this change, and rebase it against current master as it was lagging 1533 commits behind.

@bsdphk's suggestion of `THR_[GS]etWorker()` functions was actually added in 3fc7847828963d0870b2eb4f4880512209304bfb for a different reason, and it resulted in less code to get to the same point.

The integration of those logs is less intrusive, it's now only a matter of switching from `DSL(DBG_WORKSPACE, 0, ...)` to `DSLb(DBG_WORKSPACE, ...)`, a macro also introduced by this change.

Trivia: none of the functions moved to `cache_ws_common.c` have debug logs.

Dridi:
Rebased against master.

nigoroll (approved):
OK, but we should add a comment to `WRK_Log()` that it really is for debugging only.
For regular logs, we should avoid the cost of the additional `THR_GetWorker()` de-tour.

bin/varnishd/cache/cache_wrk.c
```diff
@@ -738,6 +738,28 @@ static struct cli_proto debug_cmds[] = {
...
+{
+	struct worker *wrk;
+	va_list ap;
+
+	AN(fmt);
+
+	wrk = THR_GetWorker();
+	CHECK_OBJ_ORNULL(wrk, WORKER_MAGIC);
+
+	if (wrk == NULL)
```
nigoroll, line 752: discussed during bugwash, this should lead to logging with vxid 0 instead of dropping vsls

Dridi: Done.

Dridi:
> OK, but we should add a comment to `WRK_Log()` that it really is for debugging only.
> For regular logs, we should avoid the cost of the additional `THR_GetWorker()` de-tour.

I was going to do this, but I had originally already put it under the "Debugging aids" comment.

Issue #3169, opened 2019-12-19
Handle changing Content-Encoding for inconsistent backends?


Dridi:
Test case by @daghf:

```
varnishtest "Ensure we don't override Content-Encoding on a cond fetch"

server s1 {
    rxreq
    txresp -hdr {ETag: "foo"} -body {****}

    rxreq
    expect req.http.If-None-Match == {"foo"}
    txresp -status 304 -hdr {ETag: W/"foo"} -hdr "Content-Encoding: gzip"

    rxreq
    expect req.url == "/1"
    txresp -hdr {Etag: "bar"} -gzipbody {asdf}

    rxreq
    expect req.url == "/1"
    expect req.http.If-None-Match == {"bar"}
    txresp -status 304

} -start

varnish v1 -vcl+backend {
    sub vcl_backend_response {
        set beresp.ttl = 1s;
        set beresp.grace = 0s;
        set beresp.keep = 1d;
    }
} -start

client c1 {
    txreq -hdr "Accept-Encoding: gzip"
    rxresp
    expect resp.status == 200
    expect resp.http.ETag == {"foo"}
    expect resp.http.Content-Length == "4"
    expect resp.http.Content-Encoding == <undef>

    delay 1.5

    txreq -hdr "Accept-Encoding: gzip"
    rxresp
    expect resp.status == 200
    expect resp.http.ETag == {W/"foo"}
    expect resp.http.Content-Length == "4"
    expect resp.http.Content-Encoding == <undef>

    # Also check that we're still OK in the general case
    txreq -url "/1" -hdr "Accept-Encoding: gzip"
    rxresp
    expect resp.status == 200
    expect resp.http.ETag == {"bar"}
    expect resp.http.Content-Encoding == "gzip"

    delay 1.5

    txreq -url "/1" -hdr "Accept-Encoding: gzip"
    rxresp
    expect resp.status == 200
    expect resp.http.ETag == {"bar"}
    expect resp.http.Content-Encoding == "gzip"
} -run
```

nigoroll:
I am not sure if the test case is valid. Wouldn't a sound backend have to send `Vary: Accept-Encoding` if it decided to send a different `Content-Encoding` for a 304?

Dridi:
I believe @daghf worked around the problem for the support case that brought this up, and left me a note saying "warrants more research". I should probably have mentioned that in the first place, I opened this issue so Someone:tm: could do that research.

nigoroll:
[...]

PR #3281, opened 2020-04-07
Ensure we don't override Content-Encoding in a 304 conditional fetch backport

andrewwiik:
Commits from VCP:
e68a10b6603edb95b3c90ed679bc3c816fa5ef77

We diverging from the ``master`` idea that an new vcl sub be introduced: https://github.com/varnishcache/varnish-cache/issues/3169


Dridi:
Please submit this patch for the master branch and then back-port it to 6.0, back-ports must originate from trunk.

The patch is OK and will only be superseded when `vcl_backend_refresh` is introduced and it is still at the design phase.

Dridi:
Actually don't, I will push this patch directly to master and add a comment here when you have a commit to back-port.

Dridi:
d921b1f263ba7403427cc9f8c4e82ec05469b90a

PR #3282, opened 2020-04-07
Body Free Race Backport

andrewwiik:
After some consideration we decided to backport this, although it changes a private API because well the API was private and undocumented so vmod writers should not have been using it.

commits:
f88b47951415fc212c8b4bb929b5a0db2c0835c0
d4b6228e1e32cda3014eeff0a5cb60cd7a0b5474
4e6e4eedcf078795d3197792d17b1294d79ff616

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -390,7 +390,8 @@ struct busyobj {
 	 * All fields from retries and down are zeroed when the busyobj
 	 * is recycled.
 	 */
-	int			retries;
+	unsigned		retries;
+	enum req_body_state_e	initial_req_body_status;
 	struct req		*req;
 	struct sess		*sp;
 	struct worker		*wrk;
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -403,6 +403,7 @@ struct busyobj {
 	struct http		*bereq0;
 	struct http		*bereq;
 	struct http		*beresp;
+	struct objcore		*bereq_body;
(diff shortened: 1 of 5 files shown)


rezan:
Going to have a discussion on this soon and will have a final decision. This does not break anything, rather, it silently removes a racy path, so I think this should be ok.

rezan:
Load tested, everything looks good.

bin/varnishd/cache/cache.h
```diff
@@ -390,7 +390,8 @@ struct busyobj {
 	 * All fields from retries and down are zeroed when the busyobj
 	 * is recycled.
 	 */
-	int			retries;
+	unsigned		retries;
+	enum req_body_state_e	initial_req_body_status;
```
Dridi, line 394: This is an ABI breakage, unclear whether it belongs to the VRT or strict ABI, but most likely strict and OK: #3355.

bin/varnishd/cache/cache.h
```diff
@@ -403,6 +403,7 @@ struct busyobj {
 	struct http		*bereq0;
 	struct http		*bereq;
 	struct http		*beresp;
+	struct objcore		*bereq_body;
```
Dridi, line 406: Same ABI problem since the field is not appended at the end of the struct.

Both could have been avoided, but probably fine with our stability promise.

Dridi (commented):
Post-merge review for future reference in case anyone is interested.

rezan:
Good catch, that slipped thru. Fixed 18bf1d491bf857f05b703c29f690d1be33a9bddc.

PR #3284, opened 2020-04-07
ESI workspace assert - backport

andrewwiik:
Commits used:
0f3af407f65ac4bc6d446c61fb3e3b50747194cd
d1bd80e83a88a0caaaf7e5021472e23d244a8ac6

(needed to change the gzip[ flag back to 6.0)

### DIFF
--- a/bin/varnishd/cache/cache_esi_fetch.c
+++ b/bin/varnishd/cache/cache_esi_fetch.c
@@ -170,6 +170,7 @@ vfp_esi_gzip_init(struct vfp_ctx *vc, struct vfp_entry *vfe)
 		return (VFP_ERROR);
 	}
 	vef->vgz = VGZ_NewGzip(vc->wrk->vsl, "G F E");
+	AN(vef->vgz);
 
 	vef->ibuf_sz = cache_param->gzip_buffer;
 	vef->ibuf = calloc(1L, vef->ibuf_sz);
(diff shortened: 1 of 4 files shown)


Dridi:
I would personally back-port 54af42d4c328089625a59eb063e75d5c72125852 since this is outside of VRT, and outside of our stability commitment. This technically doesn't even break the ABI, so in addition to grabbing this commit I would add one more commit that does this:

```c
#define OF_CHGGZIP OF_CHGCE
#define OA_CHGGZIP OA_CHGCE
```

This preserves source compatibility.

Afterwards the two desired commits apply without a conflict, and on my system at least the result looks stable.

rezan:
Is this related to the 304 gzip backport? This backport just addresses an assert.

> This preserves source compatibility.

Right now the scope for backporting is anything security related, bugs, and sometimes features. I think we are too far gone in terms of source compability to add this to our backporting scope.

andrewwiik:
@rezan I agree with Dridi here

rezan:
Sure, but I'm just concerned we might be optimizing for the wrong thing here. Maintaining clean cherry picks and bringing in unrelated commit dependencies means we are bringing in a lot of changes to the LTS. We have been bitten by this in the past which is why its best to keep the scope of backports as narrow as possible. 6.0 is extremely stable these days, so I would prefer to keep the changes to a minimum.

I'm also concerned that this practice will get harder and more complicated the more these versions diverge. Its easier to just be practical here and accept that cherry picks will have to be tweaked a bit to get them backported into the LTS.

If you feel strongly about this, I won't stop you.

andrewwiik:
Closed in favor of https://github.com/varnishcache/varnish-cache/pull/3324

PR #3285, opened 2020-04-08
RFC/POC high watermarks for workspaces

nigoroll:
Unless `DBG_WORKSPACE` is used or additional tracing added to VCL (see for example `vtc.workspace_free()`), we usually have no idea how much workspace we are actually using.

We know when it's not enough because we fail requests and increase the `ws_*_overflow` stats counters, but in particular for production deployments, we cannot (easily) answer the question how much workspace headroom is left, if it is getting any close to "too small" or if we are wasting memory for good no reason.

So we should really add a facility to improve visibility.

A trivial approach would be to just log `ws->f - ws->s` at the end of each transaction, but that would not account for the cases where we use workspace as "scratch" space for some intermediate result.

To add such reporting, we need to remember a high water mark for the maximum amount of workspace we used for some transaction.

This patch adds the `(struct ws)` `h`(igh water mark) member and demos reporting the high water mark at the end of each client request:

```
**** v1   vsl|       1009 Timestamp       c Resp: 1586355523.554209 0.489777 0.250743
**** v1   vsl|       1009 Debug           c wshwm client 34376
**** v1   vsl|       1009 Debug           c wshwm session 96
**** v1   vsl|       1009 Debug           c wshwm thread 1128
**** v1   vsl|       1009 ReqAcct         c 0 0 0 0 111 111
**** v1   vsl|       1009 End             c
```

Please note that the reporting in this patch is by no means meant to be final, we should probably have an SLT for that purpose and also need to add reporting for backend requests.

To track the high water marks for temporary allocations, we need to consider two basic patterns:

* `WS_Snapshot()` / `WS_Reset()`

  Workspace is used as normal and then reset back to a snapshot.

  To track the high watermark, we check in `WS_Reset()`if the `ws->f` pointer is a new high and, if yes, record it.

* `WS_Reserve()*` / `WS_Release()*`

  A Reservation is made and later returned fully or in part.

  For this pattern, we need to change the interface and ask callers to cooperate: They need to tell us with a `WS_Release*()` what their highest usage was.

  For this purpose, we add `WS_ReleaseH()` and `WS_ReleaseHP()` which take another argument for the caller to pass their high water mark.

This patch is related to #3277, which concerns tracing workspace usage in detail. The two are complementary in that this patch intents to provide the big picture and guidance on where to look closer, while #3277 aims to provide better means to corellate allocations with VCL code.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -144,6 +144,7 @@ struct ws {
 	char			*f;		/* (F)ree/front pointer */
 	char			*r;		/* (R)eserved length */
 	char			*e;		/* (E)nd of buffer */
+	const char		*h;		/* (H)igh watermark */
 };
 
 /*--------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


nigoroll:
on reporting: should be maybe have status counters with some buckets? e.g. for <50% | 50%.. 90% | >90% of workspace used?

nigoroll:
bugwash: Because we might never be able to ask all callers to report a high watermark, we need an "at least" flag when we had reservations without a high watermark reported

PR #3287, opened 2020-04-09
Add error handling to the VFP/VDP API & streamline definition-time privs

nigoroll:
There is no way for a vmod to determine if a filter name is already taken, so we add error handling for the cases that

* A filter is already registered when adding or not registered when removing
* A filter name is already taken

----

edit: As it concerns the same code, I added a second commit to streamline the defintiion-time priv of struct vfp/vdp

### DIFF
--- a/bin/varnishd/cache/cache_filter.h
+++ b/bin/varnishd/cache/cache_filter.h
@@ -34,6 +34,13 @@ struct vfp_entry;
 struct vfp_ctx;
 struct vdp_ctx;
 
+enum vrt_filter_err {
+	FLT_OK = 0,
+	FLT_DUPLICATE = -1,
+	FLT_CONFLICT = -2,	// name conflict
+	FLT_MISSING = -3
+};
+
 /* Fetch processors --------------------------------------------------*/
 
 enum vfp_status {
@@ -91,8 +98,8 @@ struct vfp_ctx {
 enum vfp_status VFP_Suck(struct vfp_ctx *, void *p, ssize_t *lp);
 enum vfp_status VFP_Error(struct vfp_ctx *, const char *fmt, ...)
     v_printflike_(2, 3);
-void VRT_AddVFP(VRT_CTX, const struct vfp *);
-void VRT_RemoveVFP(VRT_CTX, const struct vfp *);
+enum vrt_filter_err VRT_AddVFP(VRT_CTX, const struct vfp *);
+enum vrt_filter_err VRT_RemoveVFP(VRT_CTX, const struct vfp *);
 
 /* Deliver processors ------------------------------------------------*/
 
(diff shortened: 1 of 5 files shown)


bin/varnishd/cache/cache_filter.h
```diff
@@ -33,6 +33,12 @@ struct req;
 struct vfp_entry;
 struct vfp_ctx;
 
+enum vrt_filter_err {
+	FLT_OK = 0,
+	FLT_LOOKUP = -1,	// already found / not found
+	FLT_NAME_TAKEN = -2,
+};
```
Dridi, line 40: I'm not fond of the names that I don't find very descriptive, I would rather avoid using one entry for two different cases.

For example:

- ok
- duplicate (already added)
- conflict (other filter with same name present)
- missing (cannot remove)

bin/varnishd/cache/cache_filter.h
```diff
@@ -58,7 +58,7 @@ struct vfp {
 	vfp_init_f		*init;
 	vfp_pull_f		*pull;
 	vfp_fini_f		*fini;
-	const void		*priv1;
+	const void		*vfp_priv;
```
Dridi, line 61: Should we go further and rename `vfp_entry`'s `priv1` to `body_priv` and `priv2` to something also more descriptive?

nigoroll: yes, but that might bring even more sensible changes, so I would like to keep it out of here.

bsdphk:
This is unnecessary complexity.

VMOD filters should be prefixed with the vmod name, like everything else in the vmod.

Dridi:
Should we then give VMODs and varnishd a magic cookie to prevent collisions during lookup?

nigoroll:
Re @bsdphk 

> This is unnecessary complexity.

I cannot see any complexity in the suggested changes.

> VMOD filters should be prefixed with the vmod name, like everything else in the vmod.

We allow for arbitrary, non-conflicting VMOD object names in VCL and I do not see why would need to treat filters differently. In fact, we (UPLEX) have, for the filters we implemented, named dynamic filters (filters created by VCL) identical to VMOD objects.

Varnish-Cache would always register its vital filters first, so vmods could not interfere. On the other hand, I do not see why it should not be possible to have multiple implementations of, say, a `deflate` filter. Swapping a vmod would then allow remaining VCL code to remain identical. @Dridi has made good suggestions on how to make it clear who registered a vmod first when a conflict arises.

The requirement to prefix would throw more name space questions on the table, just to mention those which immediately come to my mind: a binding vmod registry, clarify naming with `import ... as`, reserving a separator. 

And I sincerely hope that we will not end up in java hell and require vmods to be named `de.uplex.brotli`.

Re @Dridi 

> Should we then give VMODs and varnishd a magic cookie to prevent collisions during lookup?

I think the collisions could be prevented easily, but I like your suggestion from IRC:

> (12:50:17) dridi: so, should we add a "cookie" in VRT_CTX that tells us who is calling vrt functions? (varnishd or vmod_foo)
> (12:50:33) dridi: this way we can have graceful conflict error reporting
> (12:51:04) dridi: could not register filter x from vmod_foo, already registered by vmod_bar
> (12:51:13) dridi: ..., already registered by varnishd

nigoroll:
pow wow: we might want to restrict filter creation to vcl_init time

bsdphk:
Superseded by 19f77efe69aceb78 and ff8f8bff5d4aaaf.

Issue #2790, opened 2018-10-04
vsl binary log incompatibility


nigoroll:
6.1 log utilities (varnishlog, varnishncsa etc) fail to read log files written with previous versions.

for example, `varnishncsa -r ${old_log}` will usually produce no output.

PR #3288, opened 2020-04-09
Backend connection reuse logging

rezan:
When a backend connection gets reused, it gets logged as `BackendOpen`. Its more correct to log it as `BackendReuse`. This is important information to know when looking thru backend logs as `BackendOpen` is a bit misleading.


nigoroll:
So until now, `BackendReuse` is issued when a connection is made available for reuse, that is, this tag is used where otherwise `BackendClose` would be issued.
We should not use the tag for both, so if we wanted to change the meaning, we should at least not use `BackendReuse` for the other purpose.
Alternatively, I wonder if it made more sense to add another field to `BackendOpen` indicating if the connection was reused.

rezan:
Ya, I would agree with that. I was considering changing `BackendReuse` to `BackendRecycle` since that better describes what we are doing. That also aligns with the counters:

https://github.com/varnishcache/varnish-cache/blob/d33f4a86098ac93c141c97ebf4f710cab417052f/bin/varnishd/cache/cache_backend.c#L241-L244

https://github.com/varnishcache/varnish-cache/blob/d33f4a86098ac93c141c97ebf4f710cab417052f/bin/varnishd/cache/cache_tcp_pool.c#L504-L505

nigoroll:
@rezan that might even make more sense

rezan:
ok, second take.

`BackendReuse` is renamed `BackendRecycle`

`BackendReuse` is then re-introduced as an alternate tag to `BackendOpen`

This leaves us with the following backend connection workflow:

```
BackendOpen [ > BackendRecycle <=> BackendReuse ] > BackendClose
```

With a connection being able to be recycled and reused any number of times. So now we can clearly see in the logs when a connection gets recycled and eventually reused.

include/tbl/vsl_tags.h
```diff
@@ -117,9 +117,22 @@ SLTM(BackendOpen, 0, "Backend connection opened",
...
+	"\t|  |  |  |  |  +- Local port\n"
+	"\t|  |  |  |  +---- Local address\n"
+	"\t|  |  |  +------- Remote port\n"
+	"\t|  |  +---------- Remote address\n"
+	"\t|  +------------- Backend display name\n"
+	"\t+---------------- Connection file descriptor\n"
+	"\n"
+)
+
+SLTM(BackendRecycle, 0, "Backend connection recycled",
```
nigoroll, line 134: new tags need to be added at the end of the file to preserve binary compatibility of vsl files

Dridi:
I'm all for renaming `BackendReuse` to `BackendRecycle` but that may break existing VSL queries. Is there a precedent of such a rename since Varnish 4.0? Can we have something similar to the `vcl_dir`/`vcl_path` transition where for a while one was an alias for the other?

Adding a new SLT is a waste in my opinion, so I would rather change `BackendOpen` like this:

```c
SLTM(BackendOpen, 0, "Backend connection opened",
	"Logged when a new backend connection is opened.\n\n"
	"The format is::\n\n"
	"\t%d %s %s %s %s %s %s\n"
	"\t|  |  |  |  |  |  |\n"
	"\t|  |  |  |  |  |  +- "connect" or "reuse"\n"
	"\t|  |  |  |  |  +---- Local port\n"
	"\t|  |  |  |  +------- Local address\n"
	"\t|  |  |  +---------- Remote port\n"
	"\t|  |  +------------- Remote address\n"
	"\t|  +---------------- Backend display name\n"
	"\t+---------------- Connection file descriptor\n"
	"\n"
)
```

And if at this point you feel that `BackendOpen` is no longer the right name for the tag, I believe that if we can rename `BackendReuse` we can rename both.

nigoroll:
@Dridi I suggested the same along the way:
> Alternatively, I wonder if it made more sense to add another field to `BackendOpen` indicating if the connection was reused.

The reasons why I thought the (breaking) rename was warranted are:
* The existing name is actually misleading as to which event it refers to
* We are on the path to a major release
* In my world, `BackendReuse` is not used that frequently in automated queries

Dridi:
> I suggested the same along the way:

I kinda jumped right to the conclusion seeing you almost approved the patch.

> The reasons why I thought the (breaking) rename was warranted are:

I just want to make sure we discuss this before we break. Especially since I suspect @rezan would ultimately want to back-port this to 6.0, having a discussion about compatibility that doesn't introduce a new SLT should be on the proverbial table.

In particular, if we add a new SLT and wish to back-port it then we also need to add unused SLTs to 6.0 to maintain binary compatibility (and we have a handy gadget to document them as such). And if we rename a field without a contingency, then this is unfit for a stable release.

nigoroll:
@Dridi 
* By no means did I intend to merge this without prior concensus
* In my mind, the patch in its current state cannot be backported
* The VSL binary compatibility is orthogonal to that. Adding an SLT is backwards, but not forwards compatible.

Dridi:
> The VSL binary compatibility is orthogonal to that. Adding an SLT is backwards, but not forwards compatible.

We do currently have compatibility both ways! See how `VCL_use` was ported to 6.0.

Renaming a tag is not orthogonal.

nigoroll:
Re @Dridi it seems I have not expressed myself clearly:

By only ever appending to `vsl_tags.h`, we achieve backward compatibility: Newer binaries can still parse old VSL files. But we do not achieve forward compatibility: VSL files written with newer binaries cannot, if they contain the new tags, be parsed with old binaries.

With respect to 6.0 commit 46921d19dd24f2e39e384e1361b0e599a0d73278 : Yes, _append only_ implies that we consistently need to add all the tags such that the binary id is equivalent across code which is supposed to be compatible.

I am sure that you understand all of this, so maybe we have differing definitions of _compatibility_, but I would think when done right, we do get backwards compatibility. And we^WI have done it wrong at one point, so we also know when we fail. See 5ca0b5d3e749be81eec39988d7db7a0fdf13e73b #2790 for the bad example.

With respect to renaming a tag, the VSL could still be read, but the old/new name would be output. I think this can be considered compatible.

With _orthogonal_ I meant to express that the rename is unrelated to the fact that we should not the SLTs.

Dridi:
#2790 was indeed fixed in point releases, a regression we took care of.

> By only ever appending to `vsl_tags.h`, we achieve backward compatibility: Newer binaries can still parse old VSL files.

Agreed. And by back-porting newer tags to supported branches, we enable forward compatibility in newer binaries from those branches.

> With respect to renaming a tag, the VSL could still be read, but the old/new name would be output. I think this can be considered compatible.

I agree on paper, except that passing the same string before and after renaming the tag to the appropriate `VSL_*()` or `VUT_*()` functions may change the outcome in a non-additive way.

So either we fix the documentation, or if we rename the tag we should add an alias mechanism like parameters may have. And overall the simplest way forward is to just docfix `Backend{Open,Reuse}` and add a field to `BackendOpen`.

nigoroll:
@Dridi ACK.
Let's take this to bugwash for a joint conclusion.

rezan:
This will not be backported to 6.0, it breaks too much. The thinking here is that this is a major release, so lets break things and get it right.

Dridi:
> The thinking here is that this is a major release, so lets break things and get it right.

I thought we were suppose to preserve binary compatibility until "binary" VSL would land (the one with binary fields instead of space- or dquote-delimited text fields). Currently TheRightThing(tm) is not to break.

Also, considering that the newly introduced `BackendReuse` would be the exact same as `BackendOpen` in terms of contents, getting it right could also be the addition of a field to distinguish the two cases.

rezan:
Sure, lets discuss at the next bugwash and figure out what we want to do here. I don't feel strongly about anything here, so im ok with whatever. I just want some kind of indication in the VSL that a fresh connection was made vs a re-used connection.

rezan:
Bugwash consensus is that we go back to just 2 tags `BackendOpen` and `BackendClose`, depricate `BackendReuse`, and then add a new fields to signal if the connection is reused/recycled.

PR #3289, opened 2020-04-14
Centralize control over Range headers in the range VDP

### LINKED ISSUE #3251: We set Accept-Range on pass, is that OK ?
Shouldn't we leave it to the backend to set A-R on pass ?

nigoroll:
This is an alternative approach to the ideas recorded in #3251: By centralizing control over the range-request related headers, range handling can be disabled by removing the `range` vdp from the response filters.

By default, we now always push the `range` VDP as long as `http_range_support` is on. If that is the case, behaviour is just the same as before, but the `Accept-Ranges` response header is now set in the range VDP's init callback.

The difference is that, if the range vdp is removed from the filters, no `Accept-Ranges` header will be sent.

Note that the `range` vdp will not be active during body delivery if it only sets `Accept-Ranges` (by returning 1 from the init callback).

Fixes #3251

### DIFF
--- a/bin/varnishd/cache/cache_range.c
+++ b/bin/varnishd/cache/cache_range.c
@@ -246,7 +246,14 @@ vrg_range_init(struct req *req, void **priv)
 	const char *r;
 	const char *err;
 
-	assert(http_GetHdr(req->http, H_Range, &r));
+	if (http_GetStatus(req->resp) != 200)
+		return (1);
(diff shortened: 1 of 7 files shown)


nigoroll:
veto'ed by @bsdphk during bugwash. He will come up with a better suggestion in #3251

bsdphk:
This is a Firm no.

See #3251 for full comments.

PR #3290, opened 2020-04-14
Worker scheduling back-ports for 6.0

Dridi:
This is a back-port of #2942 and #2796 with their dependencies to avoid conflicts.

Conflicts still exist, mostly because I didn't port #3099 in the process.


rezan:
ok to close? This was completed in https://github.com/varnishcache/varnish-cache/pull/3323.

PR #3291, opened 2020-04-14
varnishtest: make process -expect-text on the same scale as vX

benatto:
Currently on process pX -expect-text varnishtest uses at most
300000 us pauses. This make it offscale when compared on what
varnish vX -expect does by trying 50 times with 100us pause
in a total of 5s.

Changing term_expect_text() wait 3s before failing avoid tests
to cause false positive under stress (i.e.: r02990) but putting
process pX -expect-text and varnish vX -expect at the same time
scale.

Signed-off-by: Marco Benatto <mbenatto@redhat.com>


nigoroll:
merged as 6f7af3a5a6c99a943c7f016a18a7b7d6c9693d22

PR #3294, opened 2020-04-15
allow u00015.vtc to run without python

gquintard:


bin/varnishtest/tests/u00015.vtc
```diff
@@ -1,5 +1,7 @@
 varnishtest "varnishstat -j"
 
+feature ignore_unknown_macro
```
Dridi, line 3: I would rather mark the test as skipped instead of faking its results:

```
feature cmd "${PYTHON} -c 'import json'"
```

vtc.am
```diff
@@ -8,7 +8,7 @@
 # The current assumption is that all the tests for the Makefile are VTCs
 # and may be revisited if needed.
 
-VTC_LOG_COMPILER = $(top_builddir)/bin/varnishtest/varnishtest -v -i
+VTC_LOG_COMPILER = $(top_builddir)/bin/varnishtest/varnishtest -v -i -D PYTHON=$(PYTHON)
```
Dridi, line 11: Should `varnishtest -i` discover and define `${PYTHON}`?

bin/varnishtest/tests/u00015.vtc
gquintard: ok, so split the test in two, `varnishstat -j` on one side, and `varnishstat -j | ${PYTHON} ...` on the other?

vtc.am
gquintard: if this goes in, we might as well. Python is stable, widespread and useful enough that I feel this is warranted.
But I'm a bit worried of reimplementing the discovery in a way that doesn't match `autotools`'

Dridi: You are correct, and I definitely don't want to diverge. We can always ask nicely:

```
grep ^PYTHON= config.log
```

bin/varnishtest/tests/u00015.vtc
Dridi: No, with my suggested `feature` statement the test is skipped if you don't have the `json` module!

Dridi: Alternatively, can we maybe use `jq` for this?

Dridi: I think this is overall cleaner:

```vcl
varnishtest "varnishstat -j"

feature cmd "command -v jq"

server s1 -repeat 5 {
        rxreq
        txresp
} -start

varnish v1 -vcl+backend {} -start

client c1 -repeat 5 {
        txreq
        rxresp
} -run

varnish v1 -expect MAIN.client_req == 5

shell -expect 1 {
        varnishstat -n ${v1_name} -j |
        jq .version
}

shell -expect 5 {
        varnishstat -n ${v1_name} -j |
        jq '.counters."MAIN.client_req".value'
}

shell -expect {"Good client requests received"} {
        varnishstat -n ${v1_name} -j |
        jq '.counters."MAIN.client_req".description'
}
```

gquintard: > No, with my suggested feature statement the test is skipped if you don't have the json module!

if you have `python3` then you should have the `json` module, so it's really just a matter of testing for python. But I feel my intent was misunderstood: I still want to make sure that `varnishstat -j` works even if I can't test the output, hence the proposed second, always-on, test.

> Alternatively, can we maybe use jq for this?

sure, `jq` would be nicer in this specific case, and I'm probably overthinking it, but I think it'd be nice to not rely on too many external tools and python is more widely useful.

But I'm totally ok with just slapping a `feature` statement on the whole file and calling it a day

gquintard: I'm fine with your proposed code, I'll amend the PR

gquintard:
@Dridi re-pushed with your changes

fgsch:
What is the benefit of doing `command -v jq` vs just ensuring jq runs using e.g. `jq -V`?

Dridi:
> What is the benefit of doing `command -v jq` vs just ensuring jq runs using e.g. `jq -V`?

One checks that jq can run from the PATH, the other one checks that jq runs from the PATH. The only benefit for `jq -V` is a shorter command, either way no opinion.

bsdphk:
Bugwash:

Just use `feature cmd "python3 -c 'import json'"` for now.

We may or may not need `jq` later, but not for this.

gquintard:
force pushed, and switched to `-match` for stricter checks

PR #3298, opened 2020-04-27
Try to send VDP_END with the last bytes

nigoroll:
Follow-up to #3125 :

That PR added `VDP_END`, but always issued another `VDP_bytes()` call with no data at the end of the object.

We now make it the responsibility of the object iterator to send `VDP_END` and opportunistically try to issue it with the last bytes in order to avoid the additional `VDP_bytes()` call with no data.

As noted by Martin in https://github.com/varnishcache/varnish-cache/pull/3125#issuecomment-600620905 the advantage is that "in H/2 there is the need to send a frame with a last bit set to mark the end of the stream" and in H/1 we could generate the end chunk right after the last bit of data.

This change does not implement these optimizations yet, but prepares the grounds.

In addition to the suggested changes, could reviewers please also let me know if you would support renaming `OBJ_ITER_FINAL` to `OBJ_ITER_END` to make it more consistent with `VDP_END`? I would think this would also make sense because we already use _final_ to denote `OC_F_PRIVATE | OC_F_HFM | OC_F_HFP`.

### DIFF
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -315,6 +315,9 @@ ved_vdp_esi_bytes(struct vdp_ctx *vdx, enum vdp_action act, void **priv,
 	struct ecx *ecx;
 	int retval = 0;
 
+	if (act == VDP_END)
+		act = VDP_FLUSH;
+
 	AN(priv);
 	CHECK_OBJ_NOTNULL(vdx, VDP_CTX_MAGIC);
 	CAST_OBJ_NOTNULL(ecx, *priv, ECX_MAGIC);
@@ -432,7 +435,6 @@ ved_vdp_esi_bytes(struct vdp_ctx *vdx, enum vdp_action act, void **priv,
 			return (retval);
 		case 3:
 		case 4:
-			assert(act != VDP_END);
 			/*
 			 * There is no guarantee that the 'l' bytes are all
 			 * in the same storage segment, so loop over storage
--- a/bin/varnishd/cache/cache_deliver_proc.c
+++ b/bin/varnishd/cache/cache_deliver_proc.c
@@ -217,8 +217,16 @@ VDP_Close(struct vdp_ctx *vdc)
 static int v_matchproto_(objiterate_f)
 vdp_objiterator(void *priv, unsigned flush, const void *ptr, ssize_t len)
 {
+	enum vdp_action act;
 
-	return (VDP_bytes(priv, flush ? VDP_FLUSH : VDP_NULL, ptr, len));
+	if (flush == 0)
+		act = VDP_NULL;
+	else if ((flush & OBJ_ITER_END) != 0)
+		act = VDP_END;
+	else
+		act = VDP_FLUSH;
+
+	return (VDP_bytes(priv, act, ptr, len));
 }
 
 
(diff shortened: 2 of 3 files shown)


bin/varnishd/cache/cache_esi_deliver.c
```diff
@@ -311,6 +311,9 @@ ved_vdp_esi_bytes(struct req *req, enum vdp_action act, void **priv,
 	struct ecx *ecx;
 	int retval = 0;
 
+	if (act == VDP_END)
```
mbgrydeland, line 314: Looking at how I did this, I have the following snippet (I had it named VDP_LAST, but that is the same meaning as VDP_END):

```
	if (act == VDP_LAST) {
		/* It's tricky to know when to signal VDP_LAST down the
		 * VDP stack. Run with VDP_FLUSH, and signal an empty
		 * VDP_LAST afterwards. */
		retval = ved_vdp_esi_bytes(req, VDP_FLUSH, priv, ptr, len);
		if (retval == VDP_OK)
			retval = VDP_bytes(req, VDP_LAST, NULL, 0);
		return (retval);
	}

```

Point being that the VDPs of the suburequests should be able to see a VDP_END event if they so require. But of course this should not be propagated to the parent request during an ESI subrequest handling, and that will then be filtered away again in ved_bytes() (I see this filter is already present). So I think the snippet given here needs to be added.

mbgrydeland (commented):
Plus one on renaming ITER_FINAL to ITER_END.

bsdphk:
Having thought about this for some time, I wonder if we are attacking the problem from the wrong end ?

I trust we agree that for a cached object, the stevedore should have no trouble correctly sending VDP_END with the final data.

In streaming deliveries, (Ignore VFP transformations, they dont change the picture), we have two cases:  Backend sends C-L or not.

If we have a C-L from the backend, we know for certain which byte is the final one, the moment we read it from the backend socket, and that information should be able to travel with that byte, all the way up through the VFP and down the VDP.

If we do not have a C-L from the backend and use H1/Chunked, we will only know that we are done, when we read the zero-chunkhead, and at that time we already passed the previous chunk up VFP and down VDP.

Hypothetical H2 backends come in the same two variants, depending on how the backend sends C-L and if not, if the FINAL bit comes in a separate frame.

It seems to me then, that the right place to start, is the VFP_Extend() call in cache/cache_fetch.c:

                AZ(vfc->failed);
                vfps = VFP_Suck(vfc, ptr, &l);
                if (l > 0 && vfps != VFP_ERROR) {
                        VFP_Extend(vfc, l);
                        if (est >= l)
                                est -= l;
                        else
                                est = 0;
                }

If vfps is VFP_FINAL, we should pass info that to VFP_Extend(), so it can tell ObjExtend(), so it can tell the streaming subscribers, that no more will be coming.

nigoroll:
@bsdphk I agree, and this is complementary. This patch turns the `OBJ_ITER_FINAL` flag into a `VDP_END` be pushed down delivery.
Your comment is concerned about how to get the  `OBJ_ITER_FINAL` for the last bit fetched and we should look after that, too.

bin/varnishd/cache/cache_esi_deliver.c
nigoroll: Right.
This only concerns state 3 (literal bytes), because the other `VDP_Bytes` calls already have explicit actions, but I agree that my patch did not get that case right if the literal bytes came last.
Also, by your requirement that the filters on all ESI levels need to get to see 'VDP_END`, I got case 2 wrong in de6288cbe4ac102ebd3ca66cddb64bc12069ee49.
Yet I would want to avoid the zero-length `VDP_END` by your own argument. So I have pushed a different patch.

nigoroll: I went back and forth a bit, but now addressed the relevant issues both in master and in this PR.

bsdphk:
OK, with me.

PR #3303, opened 2020-04-29
Stop VCL processing if vcl_recv failed

Dridi:
VCL failure should abort execution, but `vcl_recv` would possibly modify
`req` and even continue execution in `vcl_hash`.

Better diff with the --ignore-all-space option.

---

I noticed this reviewing a test log, working on something unrelated. In my case `vcl_recv` failed because of a workspace overflow.

There may be other scenarios (`vcl_hit`?) where that happens, I didn't have time to dig.

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -910,6 +910,7 @@ cnt_recv(struct worker *wrk, struct req *req)
 
 	/* We wash the A-E header here for the sake of VRY */
 	if (cache_param->http_gzip_support &&
+	     (recv_handling != VCL_RET_FAIL) &&
 	     (recv_handling != VCL_RET_PIPE) &&
 	     (recv_handling != VCL_RET_PASS)) {
 		if (RFC2616_Req_Gzip(req->http)) {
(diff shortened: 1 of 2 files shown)


nigoroll:
I am not strictly against it, but I thought we wanted to deliberately handle workspace overflows such that vcl execution _could_ continue in order to avoid more error handling all over the place?

Dridi:
> I am not strictly against it, but I thought we wanted to deliberately handle workspace overflows such that vcl execution _could_ continue in order to avoid more error handling all over the place?

I noticed it because a test case of mine overflowed, but the test case I wrote for this change does an explicit `return (fail)`.

> Maybe you could elaborate a bit about why the current behaviour was problematic for you and, in particular, why you consider it a bug?

I'm very surprised that VCL failure in a subroutine proceeds to the next subroutine.

nigoroll:
> I noticed it because a test case of mine overflowed

Yes, but is there any actual harm being caused?

I am not sure if it makes sense to optimize the fail path.

If we wanted to do this, the natural place, IMHO, would be `CNT_Request()`. But you are right that, as `cnt_recv()` calls `VCL_hash_method` directly, we would need a special case handling there.

Dridi:
> Yes, but is there any actual harm being caused?

We can't anticipate what people will do with their VCL, especially when VMODs are involved.

And for the record this hitting me square in the POLA :)

nigoroll:
FTR: I changed my judgement on this: If we returned to the FSM loop with `VCL_RET_FAIL`, we would stop processing and Dridi is right that we should do the same for `vcl_hash`, which just happens to be called from `cnt_recv` to ensure that we always have a hash.

Dridi:
Bugwash summary:

- do an early return
- see whether we could `memset` req or at least the digest in `Req_New()`
  - if we touched this, we [c]ould move the memset from the mempool code into all the callers

Dridi:
I committed the early return, but didn't close this ticket because the question of the digest remains.

Shouldn't it be enough to zero it at the end of `cnt_recv_prep()`?

bsdphk:
@Dridi  and I tried to remember what "the question of the digest" was, and as best we an tell it might be an performance optimization relating to zeroing happening in mempool or by caller.

If that is relevant, it needs code, benchmarks and a new PR.

PR #3304, opened 2020-04-30
Import vmod_header as std.hdr_* functions

gquintard:
follow-up of https://github.com/varnishcache/varnish-cache/pull/3055.

Functions are prefixed with `hdr_` but I've no real opinion on that.

### DIFF
--- a/lib/libvmod_std/Makefile.am
+++ b/lib/libvmod_std/Makefile.am
@@ -4,7 +4,10 @@ libvmod_std_la_SOURCES = \
 	vmod_std.c \
 	vmod_std_conversions.c \
 	vmod_std_fileread.c \
+	vmod_std_hdr.c \
 	vmod_std_querysort.c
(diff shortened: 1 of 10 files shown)


nigoroll:
questions:
* do we really want this as part of `std`? Which are the reasons this has been preferred over `header`?
* I would like to see https://github.com/varnish/varnish-modules/pull/148 added, but that can be done after the merge (and I would be interested if this is considered useful)

bsdphk:
Tentatively bugwashed:

header.append(...) looks more sanitary than std.hdr_append(...)

Is .dyn() missing or should that be type-specific, for instance: bereq.dyn("foobar")

More bugwashing needed to make sure we get this right before goldplating.

bsdphk:
Bugwash consensus:  This ticket is not the way forward, most likely type-specific methods are.

PR #3305, opened 2020-04-30
Document how to make counters in a VMOD

xordspar0:
The only official documentation about defining counters from a VMOD currently is the [Varnish 6.0 release notes](https://varnish-cache.org/docs/trunk/whats-new/changes-6.0.html#under-the-hood-mostly-for-developers). This patch fixes that by adding a counter section to the [VMOD reference guide](https://varnish-cache.org/docs/trunk/reference/vmod.html).

I'm using the diff from the [xkey patch that added counters](https://github.com/varnish/varnish-modules/commit/21d0c84d876315ce2377a1f1168a6aa794df5f8f) as a guide.

### DIFF
--- a/doc/sphinx/reference/vmod.rst
+++ b/doc/sphinx/reference/vmod.rst
@@ -700,3 +700,8 @@ format
 
 After these parameters, a counter can have a longer description, though
 this description has to be all on one line in the .vsc file.
+
+You should call ``VSC_*_New()`` when your VMOD is loaded and
+``VSC_*_Destroy()`` when it is unloaded. See the generated
+``VSC_*.h`` file for the full details about the structure that contains
+your counters.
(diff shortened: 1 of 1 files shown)


xordspar0:
My VSC code sample is getting highlighted as Python code. Is there a way to prevent that?

gquintard:
a quick search yields `..higlight::` and `..code-block::` (https://www.sphinx-doc.org/en/1.5/markup/code.html) but i have no idea how compatible we are with those

xordspar0:
Thanks. `..highlight::` affects everything until the next `..highlight::` directive. I used `..code-block::` and it did the trick.

gquintard:
I like it BUT, isn't it more of and educational addition that would fit https://github.com/varnishcache/libvmod-example better?

xordspar0:
I'm not sure. I think there is a place for reference documentation about counters, but it's true that this is a mix of reference docs and how-to so far.

nigoroll:
merged because I think this documentation is fit as-is. The fact that we can still add example code should not keep us from making progress.

nigoroll:
Oh and thank you @xordspar0

PR #3308, opened 2020-05-06
Handle unformatable VCL_TIME to string conversion failures

mbgrydeland:
For VCL_TIME values that would convert to a year element that can not fit
in an int, gmtime_r would fail, and VTIM_format() would use random stack
values when picking weekday and month strings.

This patch changes VTIM_format to return "" when gmtime_r reports
failures. This way the API is not changed. Callers can test for empty
string to catch the failure if needed.

VRT_TIME_string is patched to catch the VTIM_format error, and return NULL
on failure.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -770,8 +770,11 @@ VRT_TIME_string(VRT_CTX, VCL_TIME t)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	p = WS_Alloc(ctx->ws, VTIM_FORMAT_SIZE);
-	if (p != NULL)
+	if (p != NULL) {
 		VTIM_format(t, p);
+		if (*p == '\0')
+			p = NULL;
+	}
 	return (p);
 }
 
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -767,13 +767,17 @@ VCL_STRING v_matchproto_()
 VRT_TIME_string(VRT_CTX, VCL_TIME t)
 {
 	char *p;
+	uintptr_t snapshot;
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	snapshot = WS_Snapshot(ctx->ws);
 	p = WS_Alloc(ctx->ws, VTIM_FORMAT_SIZE);
 	if (p != NULL) {
 		VTIM_format(t, p);
-		if (*p == '\0')
+		if (*p == '\0') {
 			p = NULL;
+			WS_Reset(ctx->ws, snapshot);
+		}
 	}
 	return (p);
 }
--- a/lib/libvarnish/vtim.c
+++ b/lib/libvarnish/vtim.c
@@ -167,11 +167,16 @@ VTIM_format(vtim_real t, char *p)
 	struct tm tm;
 	time_t tt;
 
+	AN(p);
 	tt = (time_t) t;
-	(void)gmtime_r(&tt, &tm);
-	AN(snprintf(p, VTIM_FORMAT_SIZE, "%s, %02d %s %4d %02d:%02d:%02d GMT",
-	    weekday_name[tm.tm_wday], tm.tm_mday, month_name[tm.tm_mon],
-	    tm.tm_year + 1900, tm.tm_hour, tm.tm_min, tm.tm_sec));
+	if (gmtime_r(&tt, &tm) != NULL)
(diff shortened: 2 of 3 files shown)


bin/varnishd/cache/cache_vrt.c
```diff
@@ -721,8 +721,11 @@ VRT_TIME_string(VRT_CTX, VCL_TIME t)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	p = WS_Alloc(ctx->ws, VTIM_FORMAT_SIZE);
-	if (p != NULL)
+	if (p != NULL) {
 		VTIM_format(t, p);
+		if (*p == '\0')
+			p = NULL;
+	}
```
Dridi, line 728: We are wasting `VTIM_FORMAT_SIZE` bytes of workspace on the failure condition. Either we `WS_Snapshot()` then `WS_Reset()` on error, or `WS_ReserveSize()` then `WS_Release()` accordingly.

Dridi (dismissed):

lib/libvarnish/vtim.c
```diff
@@ -167,11 +167,16 @@ VTIM_format(vtim_real t, char *p)
...
-	    weekday_name[tm.tm_wday], tm.tm_mday, month_name[tm.tm_mon],
-	    tm.tm_year + 1900, tm.tm_hour, tm.tm_min, tm.tm_sec));
+	if (gmtime_r(&tt, &tm) != NULL)
+		AN(snprintf(p, VTIM_FORMAT_SIZE,
+			"%s, %02d %s %4d %02d:%02d:%02d GMT",
+			weekday_name[tm.tm_wday],
+			tm.tm_mday, month_name[tm.tm_mon],
+			tm.tm_year + 1900, tm.tm_hour, tm.tm_min, tm.tm_sec));
+	else
+		*p = '\0';
```
nigoroll, line 179: assert(errno == EOVERFLOW)?

bin/varnishd/cache/cache_vrt.c
mbgrydeland: This is a very exceptional case, so it wouldn't be a big deal, but added for completeness.

lib/libvarnish/vtim.c
mbgrydeland: Not sure about that. Do we really care why it failed?

nigoroll: the idea would be that we understand why

bsdphk:
Shouldn't this trigger a VCL failure ?

nigoroll:
> Shouldn't this trigger a VCL failure ?

In this particular case - maybe. @mbgrydeland can you maybe explain under which circumstances the issue surfaces? Is this more a configuration error or some value determined at runtime?

I recently came across a loosely related issue: With the `blob` vmod it is not possible to try to decode an invalid base64 (still need to ticket this), because a VCL failure is triggered. In other words, it is not possible to gracefully handle invalid user input.

So I think that we might want to think twice about VCL failures.

mbgrydeland:
This is a configuration problem. Was discovered through @fwsGonzo's excellent fuzzing, but using some strange VCL in order to feed fuzzing data into VCL that used our conversion functions to produce VCL_DURATIONs, which failed when attempting to convert them back to a timestamp string.

For the error behaviour, I tried to make it bahve the same as the other automatic conversions work, ie VRT_IP_string, VRT_backend_string etc. They all seem to return NULL rather than cause VRT_fail(). Specifically, its not that it necessarily is a bad time, its just unformatable.

ingvarha:
This test fails on armv7hl aka 32bit arm

https://gist.github.com/ingvarha/d84de0f557bb84f5ab685724fcdcbb38

rezan:
Advanced apologies, this backport to 6.0 should not have been merged.

rezan:
Looks like this vtc produced `Tue, 19 Jan 2038 03:14:07 GMT`, which is equivalent to 32bit INT_MAX: `2147483647`. Without access to a 32bit arm system, hard to say which function failed to detect the overflow/NaN. Not even sure how we can get this test to fail on both 32bit and 64bit.

rezan:
Discussed with @nigoroll on IRC, this VTC probably needs a `feature 64bit` heading.

nigoroll:
OK'ed by Dridi via irc

nigoroll (approved):

nigoroll:
pow-wow: return `""` from `VTIM_format()` but `VRT_fail()` from `VRT_TIME_string()`. Reason: vmods can call `VTIM_format()`, `VRT_TIME_string()` is used directly from VCL (via the `.tostring()` folding), so it should fail.

PR #3310, opened 2020-05-08
fix out-of-tree builds

### LINKED ISSUE #3309: building html/man broken for out-of-tree build
I tried to get back to clean out-of-tree builds and noticed that they got broken since I last had a go.

To illustrate:
```
cd $SRCDIR
make maintainer-clean || true
./autogen.sh

cd $TMPDIR
$SRCDIR/configure
make
make html
```

* `make` fails while building man pages (`$SRCDIR` substituted)

```
make[2]: *** No rule to make target '$SRCDIR/doc/sphinx/include/cli.rst', needed by 'varnish-cli.7'.  Stop.
```

* `make html`:

```
sphinx-build -W -q -N -b html -d ./build/doctrees -D latex_elements.papersize=a4  $SRCDIR/doc/sphinx ./build/html
Error: Config directory doesn't contain a conf.py file.
Makefile:579: recipe for target 'html' failed
make[2]: *** [html] Error 1
```

The root cause is that we build some of the documentation source files while we provide others as manually written sources and the tools we use do not (as far as I know) offer an _include path_.

I had originally addressed this issue by copying the source files to the build dir, such that the doc tools can run only on the build dir. Some of the more relevant commits are 5822d0ce45b99e03ee2e611c5ef0d74bfc6a585d and 531ecb40733d3df41007784d7039f0fe9adb0150

This method did work with the disadvantage of not picking up changes to the hand written source files immediately.

subsequent commits, for example 883fddfea4e185ff417762fb64207b81d4cea971, broke this method, so now out-of-tree builds do not work any more.

Questions:
* Does anyone but me care about out-of-tree builds?
* If yes, are there any good ideas how to solve this?

I would suggest do either go back to the method from 5822d0ce45b99e03ee2e611c5ef0d74bfc6a585d (copy rst files from `$(srcdir)` to `$(builddir)`) or use the same idea but with symlinks, which should also make incremental rebuilds work for source file changes.

Thoughts?

nigoroll:
This PR fixes #3309 with the symlink approach:

For out-of-tree builds (aka [VPATH builds aka parallel builds](https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html) in automake parlor), we symlink all documentation source files to the builddir in order to find generated documentation and documentation source files within the same tree.

More background is given in #3309

### DIFF
--- a/doc/sphinx/Makefile.am
+++ b/doc/sphinx/Makefile.am
@@ -5,9 +5,9 @@
 SPHINXOPTS    =
 SPHINXBUILD   = $(SPHINX) -W -q -N
 PAPER         = a4
-BUILDDIR      = $(builddir)/build
+BUILDDIR      = build
 
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees -D latex_elements.papersize=$(PAPER) $(SPHINXOPTS) $(srcdir)
+ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees -D latex_elements.papersize=$(PAPER) $(SPHINXOPTS) $(builddir)
 
 .PHONY: help clean html dirhtml pickle json htmlhelp qthelp latex changes linkcheck doctest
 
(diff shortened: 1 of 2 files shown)


bsdphk:
This is fine with me, I just wish there were a neater way to solve this.

nigoroll:
> This is fine with me, I just wish there were a neater way to solve this.

I could not agree more

nigoroll:
possible alternative option: [`AC_CONFIG_LINKS`](http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Configuration-Links.html#Configuration-Links)

PR #3311, opened 2020-05-11
Make .resolve() a type specific method of BACKEND (#2680 redux)

bsdphk:
This is an alternative implementation of #2680

VTC by @nigoroll from #2680

### DIFF
--- a/bin/varnishd/cache/cache_director.c
+++ b/bin/varnishd/cache/cache_director.c
@@ -80,27 +80,32 @@ VDI_Ahealth(const struct director *d)
 
 /* Resolve director --------------------------------------------------*/
 
+VCL_BACKEND
+VRT_VDI_Resolve(VRT_CTX, VCL_BACKEND d)
+{
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+
+	while (d != NULL) {
+		CHECK_OBJ_NOTNULL(d, DIRECTOR_MAGIC);
+		AN(d->vdir);
+		if (d->vdir->methods->resolve == NULL)
+			break;
+		d = d->vdir->methods->resolve(ctx, d);
+		CHECK_OBJ_ORNULL(d, DIRECTOR_MAGIC);
+	}
+        return (d);
+}
+
 static VCL_BACKEND
 VDI_Resolve(VRT_CTX)
 {
 	const struct director *d;
-	const struct director *d2;
 	struct busyobj *bo;
 
 	bo = ctx->bo;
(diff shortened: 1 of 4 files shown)


bin/varnishd/cache/cache_director.c
```diff
@@ -80,27 +80,32 @@ VDI_Ahealth(const struct director *d)
 
 /* Resolve director --------------------------------------------------*/
 
+VCL_BACKEND
+VRT_VDI_Resolve(VRT_CTX, VCL_BACKEND d)
```
nigoroll, line 84: So except for slightly different assertions, the `VRT_VDI_Resolve()` from this PR is identical to `VRT_DirectorResolve()` which I proposed in #2680 so, sure, I am all for it (and would think that we should still polish the assertions).

bin/varnishd/cache/cache_director.c
```diff
@@ -80,27 +80,32 @@ VDI_Ahealth(const struct director *d)
...
-	    d->vdir->methods->resolve != NULL; d = d2) {
-		CHECK_OBJ_NOTNULL(d, DIRECTOR_MAGIC);
-		AN(d->vdir);
-		d2 = d->vdir->methods->resolve(ctx, d);
-		if (d2 == NULL)
-			VSLb(bo->vsl, SLT_FetchError,
-			    "Director %s returned no backend", d->vcl_name);
-	}
-
+	d = VRT_VDI_Resolve(ctx, bo->director_req);
```
nigoroll, line 108: compared to #2680, the suggested additional detail level of error reporting has been omitted, but we can split that from the other changes.

lib/libvcc/vcc_expr.c
```diff
@@ -839,6 +839,7 @@ static const struct vcc_methods {
 
 	{ STRINGS, STRING, "upper", "VRT_UpperLowerStrands(ctx, \vT, 1)", 1 },
 	{ STRINGS, STRING, "lower", "VRT_UpperLowerStrands(ctx, \vT, 0)", 1 },
+	{ BACKEND, BACKEND, "resolve", "VRT_VDI_Resolve(ctx, \v1)", 1 },
```
nigoroll, line 842: Sure, this is the better place for the resolve function.
But mind you, when I originally proposed #2680, we did not have type methods.

nigoroll:
I am going to merge because @Dridi had already approved of #2680 and I know he is a big fan of type methods...

Dridi:
> I am going to merge because @Dridi had already approved of #2680 and I know he is a big fan of type methods...

I liked `VRT_DirectorResolve()` better than `VRT_VDI_Resolve()`, I'll bring it up in #3315.

Issue #3299, opened 2020-04-28
Connection to admin-healthy backend with sick probe fails


nigoroll:
found by @uplexPaul: 
Forcing a backend with a failing probe to be healthy using `varnishadm backend.set_health ... healthy` does not work as expected, trying to connect to it still fails with `FetchError      b backend ...: unhealthy`
Logging issue for completeness, fix is on the way.

nigoroll:
FTR I checked if this is a regression from b16a9d9a8d000a0c9fcac2245ec0e6e0bd5cbecb, but it is not, the vtc to be committed already fails for the commit before, 6ab26126194fc81017c1847925b7ea4fc2b10969

rezan:
Just want to comment that this does **not** affect 6.0.

nigoroll:
you are right @rezan , sorry for the wrong label.
Actually from a quick look it seems this was originally introduced with 401d1dd92aec97e9a41733549442737100375266

PR #3315, opened 2020-05-13
Diffentiate "no backend" from "none resolved" in VDI_Resolve()

nigoroll:
recover a minor improvement from #2680:

This commit is to emit slightly different error messages to differentiate between the cases "no backend set in vcl" and "none returned from the director".

Another error logging change from #2680 was to check for administratively unhealthy, but we should better leave that check to the backend (see #3299)

### DIFF
--- a/bin/varnishd/cache/cache_director.c
+++ b/bin/varnishd/cache/cache_director.c
@@ -103,24 +103,26 @@ VRT_VDI_Resolve(VRT_CTX, VCL_BACKEND d)
 static VCL_BACKEND
 VDI_Resolve(VRT_CTX)
 {
-	const struct director *d;
+	VCL_BACKEND d;
 	struct busyobj *bo;
 
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	bo = ctx->bo;
 	CHECK_OBJ_NOTNULL(bo, BUSYOBJ_MAGIC);
-	CHECK_OBJ_ORNULL(bo->director_req, DIRECTOR_MAGIC);
-	d = VRT_VDI_Resolve(ctx, bo->director_req);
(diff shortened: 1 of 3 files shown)


bin/varnishd/cache/cache_director.c
```diff
@@ -103,24 +103,26 @@ VRT_VDI_Resolve(VRT_CTX, VCL_BACKEND d)
...
-		if (d->vdir->admin_health->health == 0)
-			d = NULL;
+	if (bo->director_req == NULL) {
+		VSLb(bo->vsl, SLT_FetchError, "No backend");
+		return (NULL);
 	}
 
-	return (d);
+	CHECK_OBJ(bo->director_req, DIRECTOR_MAGIC);
+	d = VRT_VDI_Resolve(ctx, bo->director_req);
```
Dridi, line 119: `VRT_VDI_Resolve()` is not part of any release yet, should we enforce that callers provide a non-null director?

Dridi: Since we are following #2680 up, I liked `VRT_DirectorResolve()` better than `VRT_VDI_Resolve()`.

nigoroll: > `VRT_VDI_Resolve()` is not part of any release yet, should we enforce that callers provide a non-null director?

`NULL` is now a perfectly valid backend (called `none` in VCL) and I do not see why we should add an assertion. Do you have any specific scenario in mind?

Dridi: > `NULL` is now a perfectly valid backend (called `none` in VCL) and I do not see why we should add an assertion.

I forgot.

> Do you have any specific scenario in mind?

No, it was more a case of whether it was considered. I forgot that legit VCL can yield a null backend.

PR #3316, opened 2020-05-14
implements memory barrier for arm

devnexen:


nigoroll:
Shouldn't all three macros be used?

Dridi:
My skeptical self needs more than a one-liner diff and a one-liner commit log :)

> Shouldn't all three macros be used?

Something we could do after the `#ifdef` dance is:

```c
#ifndef VWMB
#  define VWMB()	VMB()
#endif
#ifndef VRMB
#  define VRMB()	VMB()
#endif
```

This way we don't have to worry about how fine-grained our barriers are for each architecture or system.

nigoroll:
@Dridi that is what I had in mind

bsdphk:
@mbgrydeland 

Martin, do you know if we can ditch vmb.h and use <stdatomic.h> instead ?

mbgrydeland:
I assume you ask me wrt to the shmlog-stuff using the V*MB()-macros. Those should be expressable as release-consume fence operations. Would probably be a good time to have a critical view on some of that usage. I'm not at all certain that we are consistent.

I am a bit concerned with the availability of stdatomic.h though. That is C11, and if I did the research  correct it appeared in GCC 4.9. Centos 7 is a platform that is still very much relevant for us, and GCC in Centos7 it is at GCC 4.8.

I'm sure that these issues can be solved through some ifdef'ery, but we should be prepared to accept that we may not find stdatomic.h universally available on all platforms.

nigoroll:
>  we may not find stdatomic.h universally available on all platforms.

Yes. It is not available on our smartos vtest box yet, but I have not updated that for long.

bsdphk:
See PR in #3363

bsdphk:
#3363 moved us onto <stdatomic.h> for "new" platforms, including ARM.

Issue #3319, opened 2020-05-15
Buffer overflow (read) in VRY_Validate()


Dridi:
This is harmless by default and should only become actionable under some unlikely circumstances to trigger a panic for example:

```
***  v1   debug|Assert error in VRY_Validate(), cache/cache_vary.c line 379:
***  v1   debug|  Condition(strlen((const char*)vary+3) == vary[2]) not true.
***  v1   debug|version = varnish-trunk revision 551d2bd39820dc87ac61c2ffb032251c2bdbe3b8, vrt api = 11.0
***  v1   debug|ident = Linux,5.6.8-300.fc32.x86_64,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,epoll
***  v1   debug|now = 456675.113616 (mono), 1589543476.922098 (real)
***  v1   debug|Backtrace:
***  v1   debug|  0x456180: varnishd() [0x456180]
***  v1   debug|  0x455f2a: varnishd() [0x455f2a]
***  v1   debug|  0x519855: varnishd(VAS_Fail+0x55) [0x519855]
***  v1   debug|  0x46f52c: varnishd() [0x46f52c]
***  v1   debug|  0x46f3d7: varnishd(VRY_Finish+0x27) [0x46f3d7]
***  v1   debug|  0x460593: varnishd() [0x460593]
***  v1   debug|  0x45decc: varnishd(CNT_Request+0x6fc) [0x45decc]
***  v1   debug|  0x4a803f: varnishd() [0x4a803f]
***  v1   debug|  0x4a739d: varnishd() [0x4a739d]
***  v1   debug|  0x49489b: varnishd() [0x49489b]
***  v1   debug|pthread.self = 0x7f9a34940700
***  v1   debug|pthread.name = (cache-worker)
***  v1   debug|pthread.attr = {
***  v1   debug|  guard = 4096,
***  v1   debug|  stack_bottom = 0x7f9a34933000,
***  v1   debug|  stack_top = 0x7f9a34941000,
***  v1   debug|  stack_size = 57344,
***  v1   debug|}
***  v1   debug|thr.req = 0x7f9a2980ac20 {
***  v1   debug|  vxid = 1001, transport = HTTP/1 {
***  v1   debug|    state = HTTP1::Proc
***  v1   debug|  }
***  v1   debug|  step = R_STP_LOOKUP,
***  v1   debug|  req_body = none,
***  v1   debug|  restarts = 0, esi_level = 0,
[...]

Issue #3318, opened 2020-05-15
Assert error in vca_mk_tcp(): Condition((SES_Reserve_local_addr(sp, &sa)) != 0) not true.


Dridi:
This can easily be observed with this change:

```diff
--- i/bin/varnishtest/tests/o00005.vtc
+++ w/bin/varnishtest/tests/o00005.vtc
@@ -214,7 +214,7 @@ logexpect l1 -wait
 # vtc.alloc(-x), yet there is no vcl code being run before we parse
 # proxy headers
 
-varnish v1 -cliok "param.set workspace_session 402"
+varnish v1 -cliok "param.set workspace_session 0.25k"
 
 delay 1
```

It's a kind of "you asked for it" situation, but maybe we could prevent it from happening in the first place.

PR #3320, opened 2020-05-15
Workspace sanitizer

Dridi:
### Preamble

It all started with our resident fuzzmeister complaining that asan (address sanitizer) wasn't able to tap into workspace allocations, and resulted with the following changes:

- ban direct field access inside `struct ws`
- revise the workspace API
- add a workspace sanitizer (wssan)

### API changes

Avoiding direct field access was in some cases trivial, for example `ws->f` could be replaced with `WS_Front(ws)` and in other instances we needed new APIs to support "exotic" use cases.

Summary of API changes:

- removed `WS_Front()`, it never makes sense, except during a reservation
- added `WS_Reservation()`, similar to `WS_Front()` but returns `NULL` outside of a reservation
- added `WS_Reserved()`, needed by the HTTP/1 pipelining code
- changed `WS_ReserveSize()` to reserve exactly what was requested or nothing
- changed `WS_Copy()`, it returns `const void *`
  - only `std.querysort()` makes in-place modifications inside the copy
- changed `WS_Printf()`, it returns `const char *`
- added `WS_ReservationOffset()` and `WS_AtOffset()`
  - needed by the session attributes

We also have an inconsistent use of `unsigned` and `size_t` in the workspace API and since we are making breaking changes it might be the occasion to standardize on one type.

### Workspace sanitizer

Enabled by a debug flag, wssan's main goal is to help detect buffer overflows inside a workspace. In order to do that it adds red zones before and after each allocation and puts canaries inside the red zones that are checked whenever `WS_Assert()` is used.

Since allocations are aligned, the "lost" bytes are also treated as a red zone with their dedicated canaries.

This one is not exactly a wssan feature but the end of workspace single-byte canary was extended until the end of the available space.

It was found after the facts that wssan can also detect violations of the snapshot/reset contract.

### Problems uncovered

We haven't used wssan for fuzzing yet, but found a few things:

- `std.ip()` does not respect the snapshot/reset contract (fixed)
- `blob.{de,en}code()` uses `WS_Front()` before making a reservation (fixed)
- useless snapshot in `blob.encode()` (removed)
- division by zero in `WS_ReserveLumps()` (fixed)
- #3318
- #3319 (fixed)


Dridi:
Following today's bugwash discussion something occurred to me and resulted in 8d794d9.

Bugwash summary:

- we aren't reconsidering the workspace facility (that was discussed)
- we probably want wssan but not enabled by default (what this PR does)

I mentioned that we can enable wssan for the whole VTC test suite:

```
make check AM_VTC_LOG_FLAGS=-pdebug=+wssan
```

Some tests will fail because of workspace exhaustion, caused by wssan's default overhead of 32B per allocation. The important thing in a CI context is not to find panics in the test log.

Dridi:
The failing `distcheck_alpine` circleci job was fixed in trunk and has nothing to do with this branch: 48845c95bcf30769a6ab8d7eb83850791efd396f.

Dridi:
Rebased against current master, which leaves us with only 13 commits dedicated to implementing the wssan feature. Everything else was already merged via #3342.

Dridi:
powwow summary:

- we agreed to remove `WS_Inside()` from the API
- in exchange a `WS_Allocated()` function will be added
- `WS_Assert_Allocated()` can go away too for obvious reasons
- instead of a workspace sanitizer, we can have a workspace emulator
  - combined with asan, the emulator would be even more effective
  - it shall be a compile time option instead of a debug flag

I will close this pull request with the wsemu one once it is ready.

Dridi:
https://github.com/varnishcache/varnish-cache/pull/3644 introduces a workspace emulator.

PR #3327, opened 2020-05-16
Avoid referencing explicit executable paths

zaninime:
Not all systems have an executable available at `/bin/*`, such as on NixOS. These executables are present in other places specified on the `PATH`. From the man page of `execl(3)`:

> If the specified filename includes a slash character, then PATH is ignored, and the file at the specified pathname is executed.

This PR removes all the path segments from the `path` argument but the filename.

### DIFF
--- a/bin/varnishd/mgt/mgt_vcc.c
+++ b/bin/varnishd/mgt/mgt_vcc.c
@@ -170,7 +170,7 @@ run_cc(void *priv)
 	AZ(VSB_finish(sb));
 
 	(void)umask(027);
-	(void)execl("/bin/sh", "/bin/sh", "-c", VSB_data(sb), (char*)0);
+	(void)execlp("sh", "sh", "-c", VSB_data(sb), (char*)0);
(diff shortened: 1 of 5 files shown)


bsdphk:
On the NixOS homepage it says:

"NixOS is a Linux distribution with a unique approach to package and configuration management." 

Not having a compatible shell available under the filename "/bin/sh" is certainly a "unique approach", which I am not particular inclined to cater for in Varnish.

My guess is your patch fails CI checks because execl(3) calls needs changed to execlp(3) to actually take the PATH into account ?

zaninime:
You're right, I'm wondering why it worked on my system with `execl` alone. I will change it.

While NixOS definitely has a unique approach, relying on a standard mechanism for discovery of executables (relying on PATH) doesn't seem to have any drawback in this context, IMHO.

Also, the original motivation for this change, was the missing `/bin/rm`, and not `/bin/sh`, which is also available on NixOS.

Would it make sense to treat the two execs differently, relying on `execlp` for `rm` and `execl` for `/bin/sh`?

PR #3332, opened 2020-05-19
varnish.m4: Support generated VCC files

Dridi:
With a separate autoconf macro to preserve existing behavior. While at it, I got rid of some intermediate processing in the generated configure script that should result in slightly smaller scripts for third party VMODs.

---

Use case prompted by @nigoroll, the existing macro in `varnish.m4` doesn't play well with dynamically generated VCC files at build time.

Not sure when this should be merged, the change speculatively targets Varnish 7 for the Sep 2020 release and would need to be amended if we end up with a 6.5 release. And that's typically the kind of thing we'd forget when it's release time.

This needs to be tested extensively, so far so good on my end.

### DIFF
--- a/varnish.m4
+++ b/varnish.m4
@@ -31,7 +31,7 @@
 # OF THE POSSIBILITY OF SUCH DAMAGE.
 
 # varnish.m4 - Macros to build against Varnish.         -*- Autoconf -*-
-# serial 10 (varnish-6.0.0)
+# serial 11 (varnish-7.0.0)
 #
 # This collection of macros helps create VMODs or tools interacting with
 # Varnish Cache using the GNU build system (autotools). In order to work
@@ -186,8 +186,8 @@ AC_DEFUN([_VARNISH_VMOD_CONFIG], [
 	AC_SUBST([VMOD_TEST_PATH], [$VARNISH_TEST_PATH])
 ])
 
-# _VARNISH_VMOD(NAME)
-# -------------------
+# _VARNISH_VMOD(NAME, MODE)
+# -------------------------
 AC_DEFUN([_VARNISH_VMOD], [
 
 	AC_REQUIRE([_VARNISH_VMOD_CONFIG])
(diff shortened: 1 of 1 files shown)


varnish.m4
```diff
@@ -201,7 +201,15 @@ AC_DEFUN([_VARNISH_VMOD], [
 	dnl Define the VCL directory for automake
 	AC_SUBST([vmod_$1_vcldir], [\${vcldir}/$1])
 
-	VMOD_RULES="
+	m4_case([$2],
+		[static],
+		[m4_pushdef([VCC_SRC], [\$(srcdir)/vmod_$1.vcc])],
+		[generated],
+		[m4_pushdef([VCC_SRC],
+			[\$(builddir)/vmod_$1.vcc \$(srcdir)/vmod_$1.vcc])],
```
nigoroll, line 209: this would check for both locations while, when the semantics are `generated`, we would only need to look in `$(builddir)`.
This way, the change would, in my mind, be fully backwards compatible with the previous vmodtool and we would not need to tie merging this change to the release, or do I miss something?

Dridi: We had this discussion a couple weeks ago but I'll summarize it here: this wouldn't work in a project that produce autotools dist archives that are meant to support multiple Varnish versions. So a new macro was added and documented to only work with Varnish 7, since vmodtool would not understand the VPATH trick with older releases.

nigoroll: You are right about compatibility with existing releases, and we avoid any issues in this area with a new macro.
What I am asking is this: For `generated`, the vcc can be expected in `$(builddir)` always. So I think we do not need the second vmodtool argument.

Dridi: Not if you ship it in the dist archive, downstream to `make dist` it belongs in `$(srcdir)`.

nigoroll: ah, I see. thank you!

Dridi: > We had this discussion a couple weeks ago

;-)

nigoroll: I apologize for my bad memory, my bad spelling and the fact that I should on public mailing lists when someone mentions water transport systems in conjunction with mollusk exoskeletons and a particular shape of hair.

PR #3343, opened 2020-06-03
[RFC] mgt/linux_jail: Adding linux jail support

benatto:
Adding support for an experimental specific linux jailing.
The basic functionality is the same as UNIX jailing, using
the same parameters and same semantic for cgroups and UIDs.

The biggest initial difference is the use of SECCOMP to filter allowed
system calls. The filters are split in three groups:

	1) Manager filter (mgt_filter parameter);
	2) VCC compiler filter (vcc_filter parameter);
	3) Worker filter (wrk_filter parameter);

Each of these parameters is a plain text file containing the allowed
syscall numbers, one per line.

The default behavior of Linux jailing is killing the process which
is trying to execute a disallowed syscall with SIGSYS signal sent by
the kernel. In order to allow one to profile and create its own filter
there's an audit mode enabled by the parameter 'audit'.

The audit mode allow the execution of any syscall but it logs the call
to Linux default audit.log file. This can futher by queried by using
ausearch command.

This patch also includes standard filters for each process group as
example, they are located at /usr/share/doc/varnish. Those filters
are based on the default VCL shipped as example.

To start varnishd using the Linux jailing mechanism the following
command may be used:

  -j linux_experimental,mgt_filter=/usr/share/doc/varnish/wrk_filter,
  vcc_filter=/usr/share/doc/varnish/vcc_filter,
  wrk_filter=/usr/share/doc/varnish/wrk_filter

Signed-off-by: Marco Benatto <mbenatto@redhat.com>


Dridi:
> idk how stable syscall numbers are these days, should we have some additional checks if the tables are used on a system compatible to that where they were generated?

I'd rather have syscall names in the files, and let the linux jail convert them to numbers during startup.

bin/varnishd/mgt/mgt_main.c
```diff
@@ -158,6 +158,9 @@ usage(void)
 	printf(FMT, "-j jail[,options]", "Jail specification");
 #ifdef HAVE_SETPPRIV
 	printf(FMT, "", "  -j solaris");
+#endif
+#ifdef WITH_LINUX_JAIL
+	printf(FMT, "", "  -j linux_experimental");
 #endif
```
Dridi, line 164: For consistency I'd rather see something like `HAVE_SECCOMP` or something similar instead of `WITH_LINUX_JAIL`.

That also means changes in `configure.ac` which I promised I'd help make.

Dridi: With the current `AC_CHECK_LIB` in `configure.ac` we should already have a macro called `HAVE_LIBSECCOMP` out of the box!

configure.ac
```diff
@@ -365,9 +365,30 @@ else
...
 
+
 AM_CONDITIONAL([WITH_UNWIND],
 	[test "$with_unwind" = yes])
 
+AC_ARG_WITH([linux_jail],
+			[AS_HELP_STRING([--with-linux-jail],
+			[Turn on experimental linux jailing support])],
+			[],
+			[with_linux_jail=check])
```
Dridi, line 376: I'd rather stick to `--with-PACKAGE` or `--enable-FEATURE`.

My preference goes to `--with-seccomp` and auto-detection by default. I will help with that.

benatto:
Thanks for the initial review,

I'm replying both comments inline:

> I only had a quick first look, I guess this will need extensive testing and, if possible, real world exposure.

Yes. I tested it a few times but surely not enough for real world.

> some questions:
> 
>     * should we have documentation on how to use audit mode and generate the syscall lists?

I can do that, I didn't included on the patch but I have a shell script that parses ausearch output
and generate the list.

> 
>     * could the syscall lists have comments which syscall names the numbers map to to facilitate manual review?
> 
>     * idk how stable syscall numbers are these days, should we have some additional checks if the tables are used on a system compatible to that where they were generated?

seccomp_rule_add() should handle this gracefully and return a ERRNO in case it doesn't find the
proper syscall number for a given arch. I opted to continue in case of rule insertion failed and just throw a warning, but if needed I can make the process stop. I'd rather warning about lower protection than eventually crash varnish but it's up to you guys.

> 
>     * should we really check in (as in `git commit`) the syscall tables or should there be a way to build them?

The syscall table can be recovered from kernel headers in the system. We can think about commit it as syscalls are pretty much very stable and doesn't change frequently but I don't think it's necessary on a first glance. On the other hand the numbers and available calls changes within arch so for this we'd need on for each supported arch.
> 
>     * I would guess this could be quite fragile for vmods. Do we need to worry if people decide to load a new vmod and suddenly see syscalls denied?

It may happen depending on what the vmod does. I need to dig a bit more on how vmods are written and handled, but I can imagine in some cases it may required a not allowed syscall. When this happens one needs to start it on audit mode to catch what syscalls are being denied and included it
on the respective filter file.

> 
>     * It would be helpful to have at least a broad idea on the performance impact this may or may not have
> I'd rather have syscall names in the files, and let the linux jail convert them to numbers during startup.

This is a point I'm not happy as well. libseccomp has a way to resolve syscall names to its numbers but it doesn't solve all of them (and most of those we need). The biggest issue is the relation between name and numbers are solve during compilation time as they are macros, making it impossible to quickly solve on runtime.

We can:
1) Have some time of parser on the startup that parses the syscall table and solve it;
2) Start accepting comments on filter files;
3) I think seccomp may let dump the BPF program from it, on this case it'd be possible to have a utilitary to generate a BPF program and we load this data instead of parsing a text file.

I think we need a more robust filter file syntax eventually as well, make this better is my next step.

configure.ac
Dridi: But in that case you can also enable the jail if either dependencies are satisfied, making it possible for someone to build varnish with linux kernel support but not seccomp.

benatto: Got it, it makes sense! I was thinking more about naming than functionality itself.

bsdphk:
bugwash:

Some concern about the length if -j arguments.  (Possible solution: allow multiple -j arguments, first must always select provider)

Now that rewinding argc/argv works most places, we could remove restriction that -j arg must be first.

benatto:
@bsdphk @Dridi should I go for multiple arguments or for a better configuration file with more structure information, which would allow concentrate the filters all under the same file?

Dridi:
I think one thing that could happen first is to lift the `varnishd -j` restriction: currently it has to be the very first option.

Instead we could first `getopt()` in a loop that only looks at the (single) `-j` option, rewind `optind` and handle the rest of the options as usual. Such a change doesn't have to come with this pull request.

Martin had some legitimate reserves about having multiple `-j` options, since they would ultimately refer to the same (single) jail.

benatto:
@Dridi I think it's fair to lift the restriction but I have a feeling it's better to have one single cfg file to clean the options than multiple -j parameters being sent. However I'm up to you guys.

If I understand correctly, if we go down the multiple -j options road I'd need to wait the restriction and the rewind to be finish right?

Dridi:
> @Dridi I think it's fair to lift the restriction but I have a feeling it's better to have one single cfg file to clean the options than multiple -j parameters being sent. However I'm up to you guys.

Initially when I suggested multiple `-j` options it was to do something like this:

```
varnishd -j unix,<args> -j experimental_linux,<args>
```

That suggestion didn't convince, and there are other ways to reuse unix jail features in the linux jail.

Then phk suggested this:

```
varnishd -j unix,<args> -j <more-args>
```

To which martin had reserves, which I share.

> If I understand correctly, if we go down the multiple -j options road I'd need to wait the restriction and the rewind to be finish right?

I was describing how lifting the restriction would look in case you are interested.

nigoroll:
This has been overtaken by #4141 
If there is continued interest in the seccomp additions, I think the previous comments should be addressed and the patch reworked on the now existing linux jail.
Other than that I apologize for this project's tendency to sometimes not follow up to PRs in a ways which we ideally should.

PR #3344, opened 2020-06-05
 Generalize type methods (take 3)

Dridi:
The second iteration (#3158) of this change has been running in production successfully for months now (with I assume a variety of VMODs and a fair amount of VCL) but @bsdphk requested a third attempt.

The `_type` top-level symbol used to avoid conflicts was not as categorically rejected as the `::` hack from the first iteration and instead we decided to try with multiple symbol tables. This is what this patch series does.

This pull request is mostly composed of the exact same commits as #3158 (which I'm keeping open as requested by @nigoroll) and the main difference is the replacement of d5875b833cfeda80567d1c3a43be0c4efafc3c2a by e5a0cdc30 (with a few conflicts to resolve afterwards).

Again, like #3158 and #3147 this third installment does not change any VCL behavior, and only touches libvcc internals to accommodate future works around type methods.

### DIFF
--- a/lib/libvcc/vcc_compile.h
+++ b/lib/libvcc/vcc_compile.h
@@ -76,6 +76,7 @@ struct expr;
 struct vcc;
 struct vjsn_val;
 struct symbol;
+struct vmod_obj;
 
 struct source {
 	VTAILQ_ENTRY(source)	list;
@@ -276,6 +277,7 @@ struct vcc {
 
 	VTAILQ_HEAD(, symbol)	sym_objects;
 	VTAILQ_HEAD(, symbol)	sym_vmods;
+	VTAILQ_HEAD(, vmod_obj)	vmod_objects;
 
 	unsigned		unique;
 	unsigned		vmod_count;
--- a/lib/libvcc/vcc_compile.h
+++ b/lib/libvcc/vcc_compile.h
@@ -402,6 +402,8 @@ extern const struct symmode SYMTAB_PARTIAL[1];
 struct symbol *VCC_SymbolGet(struct vcc *, vcc_ns_t, vcc_kind_t,
     const struct symmode *, const struct symxref *);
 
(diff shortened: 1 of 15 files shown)


lib/libvcc/vcc_compile.h
```diff
@@ -100,6 +100,15 @@ struct token {
 
 typedef const struct type	*vcc_type_t;
 
+/*
+ * A type attribute is information already existing, requiring no processing
+ * or resource usage.
+ *
+ * A type method call and may do (significant processing, change things, eat
```
bsdphk: line 107 reads badly, but I think I can guess what it is meant to say.

Dridi: Bad refactoring of:

https://github.com/varnishcache/varnish-cache/blob/be1191ea9d5b0230ebb6c187ec40a51e23844d83/lib/libvcc/vcc_expr.c#L823-L824

Will fix.

Dridi:
Rebased against master, bad comment fixed.

Dridi:
I had another idea to split the monolithic symbol table: the main one and one per type.

This would remove the need to build a fake token:

https://github.com/varnishcache/varnish-cache/blob/a037118de188da286fa075359646743eb7cbd0f8/lib/libvcc/vcc_symb.c#L374-L382

See a037118de188da286fa075359646743eb7cbd0f8:

> Technically the lookup is `TYPE.METHOD` so it's not as simple as calling `VCC_SymbolGet()` with the correct namespace.

If my hypothesis is correct, it would become as simple as that.

Dridi:
Rebased against master to resolve the conflict from fca8c59aa07e9eb6b761dd2a5258855630c950e9 (trivial).

Unless CI checks fail I'm going to merge this. To meet our ambitions we will need a take 4 that will undo some of the changes from this patch series, but I have no visibility and will instead work on top of this pull request once I'm ready to tackle this. It's not a straight route but that will take us closer to our next destination in terms of VCL design.

PR #3345, opened 2020-06-05
VRE and VRT functions to quote regular expressions

Dridi:
Using PCRE's \Q and \E escape sequences, accounting for the possibility
of a \E sequence in the input string.

Supersedes #3206

### DIFF
--- a/bin/varnishtest/tests/v00065.vtc
+++ b/bin/varnishtest/tests/v00065.vtc
@@ -1,0 +1,33 @@
+varnishtest "VRT_re_quote() coverage"
+
+varnish v1 -vcl {
+	import debug;
+
+	backend be none;
+
+	sub vcl_recv {
+		return (synth(200));
+	}
+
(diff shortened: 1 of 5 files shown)


bsdphk:
Shouldn't you also account for a \Q in the input string, or do they not nest ?

Dridi:
Once you reach a first `\Q` everything is considered raw, including other `\Q` sequences, up to the first `\E`.

bsdphk:
Why do we need both `VRE_quote()` and `VRT_re_quote()` ?

(Vmods can include "vre.h" as far as I know ?)

Dridi:
I can live without `VRT_re_quote()`, the rationale was to have a strict VRE function and a more lenient VRT interface. If we only keep `VRE_quote()` it should probably tolerate null strings.

Dridi:
Approved by @bsdphk out of band.

Issue #3341, opened 2020-06-02
Empty bodies and IMS/INM/304


rezan:
I believe an optimization was made to not attempt IMS/INM requests when we have no body. This prevents a backend from doing a 304 response in such cases. Some backends, like S3, would consider this non-optimal. Its also not clearly documented anywhere and can trip people up, especially when writing VTCs.

Should we revert this optimization and go back to attempting IMS/INM on all bodies?

Original commit: https://github.com/varnishcache/varnish-cache/commit/9371b61d76ae1e653cd2168b827a0e104a0d968d

nigoroll:
In my mind, not asking for a conditional response should, under all circumstances, be more efficient for a backend than making a useless conditional request.
In other words, if any S3 implementation required clients to make conditional requests on empty objects for better performance, I would both be interested in a reference and claim that the S3 implementation was inefficient to begin with.

Dridi:
> In my mind, not asking for a conditional response should, under all circumstances, be more efficient for a backend than making a useless conditional request.

I disagree, it might be cheaper to reply to a last-modified or etag request if they are for example indexed somehow on the backend side.

The 304 status is supposed to be a fast path so any sane server implementation making use of conditionals should (in my mind ;) be more efficient at treating the conditionals.

> Should we revert this optimization and go back to attempting IMS/INM on all bodies?

A middle ground could be a feature flag, but that's more complicated.

I don't personally have an opinion, should we check what the RFCs have to say and put this under the #3246 compliance umbrella?

rezan:
> In my mind, not asking for a conditional response should, under all circumstances, be more efficient for a backend than making a useless conditional request.

It might be better if you could explain why serving a 200 is faster than a 304?

nigoroll:
> I disagree, it might be cheaper to reply to a last-modified or etag request if they are for example indexed somehow on the backend side.

Not having to run the validation should be faster than running one no matter if indexed or not.

But anyway, I did not mean to blow this up. Again, a reference to _actual_ implementation data and the reasons might help.

rezan:
> Not having to run the validation should be faster than running one no matter if indexed or not.

So you are talking about using `strcmp()` on 2 headers, right? So `strcmp()` has more overhead than delivering an object payload? Not buying this. Definitely not true for Varnish.

> Again, a reference to actual implementation data and the reasons might help.

[...]

PR #3347, opened 2020-06-10
Conditional requests when there is no body

rezan:
This addresses https://github.com/varnishcache/varnish-cache/issues/3341. It reverts the change previously made to disable conditional requests when there is no body. I also add a regression VTC to cover this case.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -202,7 +202,7 @@ vbf_beresp2obj(struct busyobj *bo)
  */
 
 static enum fetch_step
-vbf_stp_mkbereq(struct worker *wrk, struct busyobj *bo)
+vbf_stp_mkbereq(const struct worker *wrk, struct busyobj *bo)
 {
 	const char *q;
 	struct objcore *oc;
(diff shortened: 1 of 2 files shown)


rezan:
What would be the case for keeping this?

bsdphk:
That it has worked fine for a long time for, as far as we know, N-1 users.

nigoroll:
re @rezan cases for keeping it:
* see https://github.com/varnishcache/varnish-cache/issues/3341#issuecomment-640615695
* when we already know the object is empty, a 304 has no benefit in terms of no body being sent

rezan:
So regarding the comment that corrupted empty response can come in, that's not possible. You need either a content length or a chunked response. Both of these conditions make an empty response invalid for modern HTTP.

And right, the goal here was to allow the backend to make the decision for a 304. I understand that Varnish sees no benefit for 304 vs 200 as both will insert a fresh object into cache.

Feel free to close this.

bsdphk:
@rezan I'm going to pull out my veto, because having you and slink haggle about this is counter productive.

This can go in on if controlled by a default-off feature flag.

rezan:
No problem, I will close this.

nigoroll:
@rezan if the empty body (in the form as a correct http response) is an error in itself, the current logic helps.

PR #3350, opened 2020-06-12
Backport send_timeout

rezan:
This adds more coverage for send_timeout on a response.

Reference https://github.com/varnishcache/varnish-cache/pull/3201


rezan:
Testing came out clean.

Issue #3246, opened 2020-03-09
7.0 Compliance improvements


bsdphk:
We have agreed to improve RFC compliance in the next "big" release (7.0/vcl x.y)

This ticket is meant to be the coordination point for this, in particular for the necessary documentation updates.

Each of the subtasks will have their own tickets, listed here:

#3016 (status 502)

#3121 (status 412) 

#3205 (status 413/414)

Dridi:
#3211 (private headers)

Dridi:
Registering the idea here: we should try to improve RFC coverage in the code as we make progress in this area.

By that I mean comments like:

https://github.com/varnishcache/varnish-cache/blob/e896529982027a4325dc3bf19f97c8ebbff4ef53/bin/varnishd/cache/cache_req_fsm.c#L414-L420

Dridi:
I'm wondering whether [VIP24](https://github.com/varnishcache/varnish-cache/wiki/VIP-24%3A-Hitpass-turning-into-hitmiss-after-ttl) also falls under the compliance umbrella.

ThijsFeryn:
I had a look at the [Edge Architecture Specification](https://www.w3.org/TR/edge-arch/), and noticed something that is not properly implemented in our `vcl_backend_response` logic.

> The Surrogate-Control response header contains several directives that influence entity cacheability; specifically, "no-store", "no-store-remote", and "max-age" (see "Surrogate-Control Header" for more information). Collectively, these directives and their behaviors are described by the capability token
>
>Surrogate/1.0
>
>This token should be included in all requests sent by compliant surrogates (see "Surrogate-Capabilities Header").
>
>When any of these directives are present, they override any HTTP cacheability information present in the response.

### Surrogate-Control: no-store

In https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/builtin.vcl#L154, we unconditionally allow `Surrogate-Control: no-store` to have precedence over `Cache-Control`.

According to the spec `Surrogate-Capability: abc="Surrogate/1.0"` should be set first.

If we want to improve compliance, we could change the Surrogate check as follows:
```
bereq.http.Surrogate-Capability  ~ "(?i)[^=]+=\"Surrogate\/1\.0\"" && beresp.http.Surrogate-control ~ "(?i)no-store"
```

### Surrogate-Control: max-age

[...]

PR #3364, opened 2020-07-16
Implementing the Cache-Control "must-revalidate" directive

ThijsFeryn:
This PR implements `must-revalidate` based on [the spec](https://tools.ietf.org/html/rfc7234#section-5.2.2.1), stating that when `must-revalidate` is part of the `Cache-Control` header, no stale data should be served from cache.

In terms of Varnish, this means *grace* is set to zero.

## Example header

This is what such a header could look like:

```
Cache-Control: public, max-age=3600, must-revalidate
```

## Combining `must-revalidate` and `stale-while-revalidate`

By putting this *if-clause* behind the `stale-while-revalidate` check, any occurrence of `must-revalidate` will supercede `stale-while-revalidate`.

If for some weird reason, the backend would send back both directives, *grace* would be set to zero, as illustrated in the example below:

```
Cache-Control: public, max-age=3600, must-revalidate, stale-while-revalidate=30
```

## This is probably a breaking change

Because this PR changes the behavior of how *grace* is set, this could be considered a breaking change.

If you would accept this patch, I'm not sure at what point this would end up in a future release of Varnish.


bin/varnishtest/tests/b00073.vtc
```diff
@@ -0,0 +1,31 @@
...
+	rxresp
+	expect resp.http.grace == 10.000
+
+	txreq -url /2
+	rxresp
+	expect resp.http.grace == 0.000
+
+	txreq -url /2
+	rxresp
+	expect resp.http.grace == 0.000
```
Dridi, line 30: Isn't this a cache hit?

Dridi:
I added this pull request to #3246 ( 7.0 Compliance improvements)

bin/varnishd/cache/cache_rfc2616.c
```diff
@@ -220,6 +220,16 @@ RFC2616_Ttl(struct busyobj *bo, vtim_real now, vtim_real *t_origin,
 		*grace = rfc2616_time(p);
 	}
 
+	/*
+	 * The "must-revalidate" directive prohibits the cache
+	 * from serving stale data. 
+	 * This sets the grace to zero.
```
Dridi, line 226: We should add an RFC comment for coverage too.

bin/varnishtest/tests/b00073.vtc
ThijsFeryn: Yes, that was a typo, I fixed this.

bin/varnishd/cache/cache_rfc2616.c
ThijsFeryn: Just added this and referred to `RFC7234, 5.2.2.1`.

Dridi: We have a special comment format for RFC coverage, see for example:

https://varnish-cache.org/gcov/varnish-cache/bin/varnishd/cache/cache_range.c.html#191

Comments like `// rfc7233,l,455,456` are then referenced:

https://varnish-cache.org/rfc/rfc7233.html#452

bin/varnishd/cache/cache_rfc2616.c
```diff
@@ -220,6 +220,17 @@ RFC2616_Ttl(struct busyobj *bo, vtim_real now, vtim_real *t_origin,
...
 
+	/*
+	 * The "must-revalidate" directive prohibits 
+	 * the cache from serving stale data,
+	 * according to RFC7234, 5.2.2.1.
+	 * This sets the grace to zero.
+	 */
+	if (*ttl >= 0 && http_GetHdrField(hp, H_Cache_Control,
+	    "must-revalidate", &p)) {
+		*grace = 0;
```
Dridi, line 231: Maybe change `0` to `0.`.

bin/varnishd/cache/cache_rfc2616.c
ThijsFeryn: @Dridi There's no corresponding page for RFC 7234 on https://varnish-cache.org/rfc/. Is that something you can fix?

bin/varnishd/cache/cache_rfc2616.c
ThijsFeryn: Done!

bin/varnishd/cache/cache_rfc2616.c
Dridi:
```
curl -s https://tools.ietf.org/rfc/rfc7234.txt | less -N
```

That should help you find the relevant lines covered by your commit. I don't know how @bsdphk set this up but I assume rfc7243 will automagically show up once we start reporting coverage in the master branch.

bin/varnishd/cache/cache_rfc2616.c
```diff
@@ -220,6 +220,17 @@ RFC2616_Ttl(struct busyobj *bo, vtim_real now, vtim_real *t_origin,
...
 	}
 
+	/*
+	 * The "must-revalidate" directive prohibits 
+	 * the cache from serving stale data,
+	 * according to RFC7234, 5.2.2.1.
+	 * This sets the grace to zero.
+	 */
+	if (*ttl >= 0 && http_GetHdrField(hp, H_Cache_Control,
+	    "must-revalidate", &p)) { // rfc7234,l,1303,1305
```
ThijsFeryn, line 230: @Dridi Here's where I refer to the (currently) non-existent page for RFC7234. Lines 1303-1305 talk about the staleness aspect. Hope the page magically appears once this comment is processed.

bin/varnishd/cache/cache_rfc2616.c
```diff
@@ -220,6 +220,17 @@ RFC2616_Ttl(struct busyobj *bo, vtim_real now, vtim_real *t_origin,
 		*grace = rfc2616_time(p);
 	}
 
+	/*
+	 * The "must-revalidate" directive prohibits 
+	 * the cache from serving stale data,
+	 * according to RFC7234, 5.2.2.1.
+	 * This sets the grace to zero.
+	 */
```
Dridi, line 228: Not sure we need this comment now that we have an RFC reference, and the code is rather self-explanatory.

ThijsFeryn: @Dridi I removed that part.

Dridi:
One thing that should be noted is the risk of serialization introduced by this change, negating the non-zero `default_grace`:

```http
Cache-Control: public, must-revalidate
```

Such a change should get a fair share of red tape in the release notes and relevant documentation.

For this use case we would need a non-zero `keep`, and in the spirit of better compliance we could change the default timings:

- `default_ttl=0`
- `default_grace=0`
- `default_keep=2m`

This way we could cache resources with no TTL based on the revalidation status (200 or 304). This would however be a major VCL breaking change since a zero TTL is interpreted today as uncacheable. To avoid serialization it would be necessary to wake up anyone in the waiting list regardless of the outcome (200 or 304) to be served with a fresh response.

Even a negative TTL shouldn't be interpreted as uncacheable (and more precisely unstorable) if we allow grace and keep.

That doesn't change my review, I wanted to mention the risks and where we could go if we dial this compliance knob up to eleven. I had mentioned some of that to @ThijsFeryn prior to his submission but didn't have time to elaborate here last week.

rezan:
> the risk of serialization introduced by this change

Request serialization can only happen when you set a zero TTL. Having a zero grace value does not affect this, so its safe.

Dridi:
> Request serialization can only happen when you set a zero TTL. Having a zero grace value does not affect this, so its safe.

This is prevented with the current `default_ttl` and `default_grace`, but if you touch those defaults (as a Varnish user) and override the initial `beresp.grace` (directly from the backend) via the `must-revalidate` directive you introduce this risk. That's why I'm advocating for significant red tape in the docs because it can be difficult to diagnose.

This is what meant by "risk", we are pulling a rug under existing VCL.

And I guess I forgot some thoughts:

It can be easy to turn a cacheable response into a hitmiss object because the TTL is negative, especially in a tiered setup where Varnish's backend might be another Varnish instance serving a graced object. Making an object storable based on ttl+grace+keep would make this a non-issue.

The `Cache-Control: public, must-revalidate` serialization risk also goes away if ttl+grace+keep is enough to store a response since we have a usable cache insertion regardless of the 200 vs 304 outcome, which should be enough to wake clients on the waiting list.

Still, it breaks the assumption that a zero or negative TTL can be used to tell the built-in VCL and the core code that a response is uncacheable:

```vcl
varnishtest "cache negative ttl"

server s1 {
        rxreq
        txresp -hdr {ETag: "123abc"} -body helloworld

        rxreq
        expect req.http.If-None-Match == {"123abc"}
        txresp -status 304 -hdr {ETag: "123abc"}
} -start

varnish v1 -vcl+backend {
        sub vcl_backend_response {
                set beresp.ttl = -1s;
                set beresp.grace = 0s;
                set beresp.keep = 2m;
                return (deliver);
        }
} -start

client c1 -repeat 2 {
        txreq
        rxresp
        expect resp.body == helloworld
} -run
```

Such a scenario would pass with this change in core code:

```diff
--- i/bin/varnishd/cache/cache_hash.c
+++ w/bin/varnishd/cache/cache_hash.c
@@ -435,7 +435,7 @@ HSH_Lookup(struct req *req, struct objcore **ocp, struct objcore **bocp)
                        continue;
                }
 
-               if (oc->ttl <= 0.)
+               if (oc->ttl + oc->grace + oc->keep <= 0.)
                        continue;
 
                if (BAN_CheckObject(wrk, oc, req)) {
```

And we'd need something similar in the built-in VCL.

That's it for my compliance detour, I hope I clarified my previous comment. I'm also fully aware of my lack of scrutiny, I haven't researched all the problems that this suggestion could introduce, for now I'm putting it on the proverbial table.

PR #3367, opened 2020-07-21
Haiku build fix

devnexen:
Mainly WCOREDUMP is undefine but WIFCORED has similar semantic.

### DIFF
--- a/bin/varnishd/mgt/mgt_child.c
+++ b/bin/varnishd/mgt/mgt_child.c
@@ -34,7 +34,6 @@
 #include "config.h"
 
 #include <sys/types.h>
-#include <sys/wait.h>
 
 #include <poll.h>
 #include <signal.h>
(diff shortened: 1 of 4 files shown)


bin/varnishd/mgt/mgt_child.c
```diff
@@ -59,6 +59,13 @@
 #include "common/heritage.h"
 #include "common/vsmw.h"
 
+#ifndef WCOREDUMP
+#  ifdef WIFCORED
+#    define WCOREDUMP(s) WIFCORED(s)
+#  else
+#    define WCOREDUMP(s) (-1)
+#endif
+
```
Dridi, line 68: Aren't you missing an `#endif` for the nested `#ifdef`?

Dridi:
I'm wondering whether we should move this to `vsig.h` or something else instead of making the same ifdefery everywhere.

devnexen:
I though of it and would have preferred but did not dare to do disruptive changes.

nigoroll:
backporters see also 2d4da769f5f59374a6baf9f9f6efdf327d0dee8a

PR #3371, opened 2020-07-27
Complete the VSL memory fence pair

mbgrydeland:
The log readers would not complete the acquire/release pair of memory
fence operations when waiting for new data to appear in the shared memory
log. With this, a VRMB() is issued after having waited for the
VSL_ENDMARKER disappears, and before using any data after it (including
the deciphered length data).

Ref: #3363

### DIFF
--- a/lib/libvarnishapi/vsl_cursor.c
+++ b/lib/libvarnishapi/vsl_cursor.c
@@ -138,15 +138,6 @@ vslc_vsm_next(const struct VSL_cursor *cursor)
 		t = *(volatile const uint32_t *)c->next.ptr;
 		AN(t);
 
-		if (t == VSL_WRAPMARKER) {
-			/* Wrap around not possible at front */
-			assert(c->next.ptr != c->head->log);
-			c->next.ptr = c->head->log;
-			while (c->next.priv % VSL_SEGMENTS)
-				c->next.priv++;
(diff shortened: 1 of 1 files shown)


bsdphk:
So as I understood my SHM lecture from people with more clue than me, no currently relevant archs need this, but I think it is a good idea to be correct on general principles.

lib/libvarnishapi/vsl_cursor.c
```diff
@@ -156,6 +147,19 @@ vslc_vsm_next(const struct VSL_cursor *cursor)
 			return (vsl_end);
 		}
 
+		/* New data observed. Ensure load ordering with the log
+		 * writer. */
+		VRMB();
+
+		if (t == VSL_WRAPMARKER) {
+			/* Wrap around not possible at front */
```
nigoroll, line 155: I would think the `VRMB()` had to go here (after the `if (t == VSL_WRAPMARKER`):
In `vsl_wrap()`, we prepare the new head and then set the wrap marker. So on the reader side, we should have a read barrier after reading the tag, but before reading the new head.

nigoroll: Ah never mind, my comment was without substance. `t` is being read further up in line 138

PR #3373, opened 2020-07-28
Binheap: miniobj-ify, add destructor, install for vmods

nigoroll:
* (cleanup) switch to the common `miniobj.h` macros
* Add a desctructor
* Install `binary_heap.h` for use in vmods

### DIFF
--- a/lib/libvarnish/binary_heap.c
+++ b/lib/libvarnish/binary_heap.c
@@ -538,14 +538,6 @@ vrnd_lock(void)
 {
 }
 
-#if defined(__SANITIZER) || __has_feature(address_sanitizer)
-int __lsan_is_turned_off(void);
-int __lsan_is_turned_off(void)
-{
-	return (1);
-}
-
-#endif
 int
 main(void)
 {
@@ -649,8 +641,10 @@ main(void)
 		}
 		fprintf(stderr, "%d updates OK\n", M);
 	}
-	while ((fp = binheap_root(bh)) != NULL)
+	while ((fp = binheap_root(bh)) != NULL) {
 		binheap_delete(bh, fp->idx);
+		FREE_OBJ(fp);
+	}
 	binheap_destroy(&bh);
 	AZ(bh);
 	return (0);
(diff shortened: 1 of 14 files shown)


lib/libvarnish/binary_heap.c
```diff
@@ -244,6 +244,21 @@ binheap_new(void *priv, binheap_cmp_t *cmp_f, binheap_update_t *update_f)
...
 }
 
+void
+binheap_destroy(struct binheap **bhp)
+{
+	struct binheap *bh;
+	unsigned u;
+
+	TAKE_OBJ_NOTNULL(bh, bhp, BINHEAP_MAGIC);
+	AZ(binheap_root(bh));
```
Dridi, line 254: Why should it be forbidden to release a non-empty binheap?

nigoroll: my thinking was that the binheap might be the only reference and a delete is likely to require more than just the binheap delete

Dridi: Good point!

Dridi:
Suggested additional change following the logic of my previous comment, free entries as we delete them in the test case since we go through the trouble of clearing the binheap anyway. We would no longer need the lsan workaround:

```diff
--- i/lib/libvarnish/binary_heap.c
+++ w/lib/libvarnish/binary_heap.c
@@ -538,14 +538,6 @@ vrnd_lock(void)
 {
 }
 
-#if defined(__SANITIZER) || __has_feature(address_sanitizer)
-int __lsan_is_turned_off(void);
-int __lsan_is_turned_off(void)
-{
-       return (1);
-}
-
-#endif
 int
 main(void)
 {
@@ -649,8 +641,10 @@ main(void)
                }
                fprintf(stderr, "%d updates OK\n", M);
        }
-       while ((fp = binheap_root(bh)) != NULL)
+       while ((fp = binheap_root(bh)) != NULL) {
                binheap_delete(bh, fp->idx);
+               FREE_OBJ(fp);
+       }
        binheap_destroy(&bh);
        AZ(bh);
        return (0);
```

nigoroll:
good point, @Dridi , thank you!

bsdphk:
Should we rename it VBH for consistency ?

Dridi:
Makes sense!

nigoroll:
discussed on IRC:
I will add another commit with the rename and merge

PR #3376, opened 2020-08-04
VENC vs. vmod_blob

bsdphk:
This PR is here so we can discuss if vmod_blob should be using systemwide base64 encoding functions or not.

The first two commits are part of the 'structured-fields' work in VCC, they lay the ground, but can otherwise be ignored.

The third commit adds functions so VENC can support vmod_blob

The fourth commit makes vmod_blob use VENC for base64 decoding, and it raises two questions:

1) Should we do this, or are there good reasons to let vmod_blob be self-contained (bugwards and backwards portability etc.)

2) I think there is a bug in the base64 decoding in vmod_blob today, which is not in the VENC code, and if we decide not to venc-ify vmod_blob, that should probably (unless bugwards compat ?) be fixed in vmod_blob.

The bug is exemplified by the base64 sequence `VB==`, which leaves "dribble-bits" after the payload bytes:

    V      B      =      =
    010101 000001 ------ ------
    01010100 0001---- --------

The addition to m00037 explictly tests for this, and if we do not venc-ify vmod_blob, I think this should be committed with a fix.

The test commented out in m00042 relies dribble-bits being ignored instead of rejected.  I cannot tell if the test can be removed or if it should be modified to fulfill other test-requirements.

Another important detail to note is that since the VENC functions uses VSB, and VSB always NUL-terminates, even for binary contents (it doesn't know it is binary) the buffer must be one byte larger for these decode functions, and the way I've done that now is only "proof-of-concept" quality.

### DIFF
--- a/bin/varnishd/cache/cache_expire.c
+++ b/bin/varnishd/cache/cache_expire.c
@@ -208,6 +208,7 @@ EXP_Insert(struct worker *wrk, struct objcore *oc)
 	if (remove_race) {
 		ObjSendEvent(wrk, oc, OEV_EXPIRE);
 		tmpoc = oc;
+		assert(oc->refcnt >= 2); /* Silence coverity */
 		(void)HSH_DerefObjCore(wrk, &tmpoc, 0);
 		AZ(tmpoc);
 		assert(oc->refcnt >= 1); /* Silence coverity */
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -738,7 +738,7 @@ MCF_ParamConf(enum mcf_which_e which, const char * const param,
 	va_end(ap);
 	AZ(VSB_finish(vsb));
 	mcf_dyn_vsb(which, pp, vsb);
-	VSB_delete(vsb);
+	VSB_destroy(&vsb);
 }
 
 /*--------------------------------------------------------------------*/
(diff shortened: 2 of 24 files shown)


nigoroll:
quick notes after irc chat with @bsdphk 

on  "dribble bits":

* [RFC 3548](https://tools.ietf.org/html/rfc3548#section-3) clearly states _When fewer than 24 input bits are available in an input group, zero bits are added (on the right) to form an integral number of 6-bit groups._ but I have not found any statement on whether or not input with "dribble bits" should be refused by decoders
* decoding of "dribble bits" does not make a difference if padding is used, but for no padding, we get an ambiguity: should only a "full byte" be output or also a "partial byte"? I have not conducted extensive research on this.
* for the padding case, we have checked if "dribble bits" are accepted:
  *  :x: freebsd `b64decode`
  * :heavy_check_mark: linux `base64`
  * :heavy_check_mark: SmartOS `base64`
  * :heavy_check_mark: python `base64` module
* For `vmod_blob`, I would prefer to remain _postel_-ish and optionally add a switch to enable strict checking
* This touches the general issue that `vmod_blob` should not `VRT_fail()` on invalid input in order to allow graceful handling of untrusted inputs.

On whether to use venc, I would like to understand the performance difference.

nigoroll:
OBE for the most part:
cd41b81c93eda7dc24535b99c744906d6b718809

I would still want to understand if using VENC in vmod_blob would imply any performance hit.

nigoroll:
See #3380 #3381 :
the `venc`/`VSB_putc` hit is in the order of x 2.7 on my system.
As one base64 call is microseconds, this could still easily be debated for absolute insignificance, but I would still prefer to either not use VSB_putc in venc or keep the current blob vmo_code.

bsdphk:
I've decided we leave vmod_blob alone for now

PR #3377, opened 2020-08-05
cli: Teach vcl.discard to operate on multiple VCLs

Dridi:
To put it simply, let's take a simple CLI script:

    vcl.discard vcl1
    vcl.discard vcl2
    [...]
    vcl.discard vclX

We can now achieve the same with a single command:

    vcl.discard vcl1 vcl2 ... vclX

But there is slighty more to it, because vcl.discard operates on both
VCLs and VCL labels, and dependencies can exist between both. So in
addition to operate on multiple VCLs it also does so in the correct
order.

### DIFF
--- a/include/tbl/cli_cmds.h
+++ b/include/tbl/cli_cmds.h
@@ -125,6 +125,20 @@ CLI_CMD(VCL_LIST,
 	0, 0
 )
 
+CLI_CMD(VCL_DEPS,
+	"vcl.deps",
+	"vcl.deps [-j]",
+	"List all loaded configuration and their dependencies.",
+	"  Unless ``-j`` is specified for JSON output, the"
+	" output format is up to two columns of dynamic width"
+	" separated by white space with the fields:\n\n"
+	"  * VCL: a VCL program\n\n"
+	"  * Dependency: another VCL program it depends on\n\n"
+	"Only direct dependencies are listed, and VCLs with"
+	" multiple dependencies are listed multiple times.",
+	0, 0
+)
+
 CLI_CMD(VCL_SHOW,
 	"vcl.show",
 	"vcl.show [-v] <configname>",
--- a/include/tbl/cli_cmds.h
+++ b/include/tbl/cli_cmds.h
@@ -99,10 +99,12 @@ CLI_CMD(VCL_STATE,
 
 CLI_CMD(VCL_DISCARD,
 	"vcl.discard",
-	"vcl.discard <configname|label>",
-	"Unload the named configuration (when possible).",
-	"",
-	1, 1
+	"vcl.discard <configname|label>...", /* XXX: allow globs */
+	"Unload the named configurations (if possible).",
+	" If more than one named configuration is specified the command"
+	" is equivalent to individual commands in the right order with"
+	" respect to configurations dependencies.",
+	1, -1
 )
 
 CLI_CMD(VCL_LIST,
--- a/bin/varnishd/mgt/mgt_vcl.c
+++ b/bin/varnishd/mgt/mgt_vcl.c
@@ -678,28 +678,15 @@ mgt_vcl_discard(struct cli *cli, struct vclprog *vp)
 	mgt_vcl_del(vp);
 }
 
-static struct vclprog *
-mgt_vcl_can_discard(struct cli *cli, const char * const *av)
+static void
+mgt_vcl_discard_depfail(struct cli *cli, struct vclprog *vp)
 {
-	struct vclprog *vp;
 	struct vcldep *vd;
 	int n;
 
 	AN(cli);
-	AN(av);
-	AN(*av);
-
-	vp = mcf_find_vcl(cli, *av);
-	if (vp == NULL)
-		return (NULL);
-	if (vp == active_vcl) {
-		VCLI_SetResult(cli, CLIS_CANT);
-		VCLI_Out(cli, "Cannot discard active VCL program %s\n",
-		    vp->name);
-		return (NULL);
-	}
-	if (VTAILQ_EMPTY(&vp->dto))
-		return (vp);
+	AN(vp);
+	assert(!VTAILQ_EMPTY(&vp->dto));
 
 	VCLI_SetResult(cli, CLIS_CANT);
 	AN(vp->warm);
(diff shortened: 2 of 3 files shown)


include/tbl/cli_cmds.h
```diff
@@ -99,10 +99,14 @@ CLI_CMD(VCL_STATE,
 
 CLI_CMD(VCL_DISCARD,
 	"vcl.discard",
-	"vcl.discard <configname|label>",
-	"Unload the named configuration (when possible).",
-	"",
-	1, 1
+	"vcl.discard <configname|label>...",
+	"Unload the named configurations (when possible).",
```
nigoroll, line 103: unrelated to the patch: could "when possible" be read as "maybe delayed"? Would "if possible" be better?

include/tbl/cli_cmds.h
```diff
@@ -123,6 +127,20 @@ CLI_CMD(VCL_LIST,
 	0, 0
 )
 
+CLI_CMD(VCL_DEPS,
```
nigoroll, line 130: I am not sure how useful this is. For the tabular format, doesn't `vcl.list` already give the relevant information?
For the JSON format, should we actually have a new command or rather extend `vcl.list -j` in a backwards compat manner?

include/tbl/cli_cmds.h
Dridi: It doesn't show up in the line range you selected but "when possible" was already present, and it is sometimes "not possible" to discard a VCL (it is active or a dependency of others).

include/tbl/cli_cmds.h
Dridi: > I am not sure how useful this is.

I know, I initially wrote `vcl.deps` first until it lead me to a dead end for my use case. I kept it in hindsight to make it convenient to visualize dependencies. I didn't want to generate a dot graph directly so I also kept the standard `tsort(1)` format.

> For the tabular format, doesn't `vcl.list` already give the relevant information?

No, it tells you that your VCL is labelled, and that the label is used in a `return(vcl)` transition.

> For the JSON format, should we actually have a new command or rather extend `vcl.list -j` in a backwards compat manner?

I find it harder to leverage a JSON output in this case rather than the plain text counterpart. But I have a strong shell bias.

nigoroll: > No, it tells you that your VCL is labelled, and that the label is used in a return(vcl) transition.

There must be something obvious which I am missing. Do you have a side-by-side example with `vcl.list` and `vcl.deps` illustrating your case?

include/tbl/cli_cmds.h
nigoroll: yes, I know, which is why I wrote _unrelated to the patch_

Dridi: Yes, I know, but I can't read...

Dridi: ("if possible" +1)

include/tbl/cli_cmds.h
Dridi: I made this change:

```diff
--- bin/varnishtest/tests/c00077.vtc
+++ bin/varnishtest/tests/c00077.vtc
@@ -64,4 +64,6 @@ varnish v1 -clierr 106 "vcl.label vclA vcl3"
 
 varnish v1 -cliok "vcl.symtab"
 
+varnish v1 -cliok "vcl.list"
+varnish v1 -cliok "vcl.deps"
 varnish v1 -cliok "vcl.discard vcl1 vcl2 vcl3 vcl4 vcl5 vcl6 vcl7 vclA vclB"
```

This is the output:

```
**   top  === varnish v1 -cliok "vcl.list"
**** v1   CLI TX|vcl.list
***  v1   CLI RX  200
**** v1   CLI RX|available   auto     warm         0    vcl1    <-    (1 label)
**** v1   CLI RX|available   label    warm         0    vclA    ->    vcl1 (5 return(vcl)'s)
**** v1   CLI RX|available   auto     warm         0    vcl2    <-    (1 label)
**** v1   CLI RX|available   label    warm         0    vclB    ->    vcl2 (1 return(vcl))
**** v1   CLI RX|available   auto     warm         0    vcl3
**** v1   CLI RX|available   auto     warm         0    vcl4
**** v1   CLI RX|available   auto     warm         0    vcl5
**** v1   CLI RX|available   auto     warm         0    vcl6
**** v1   CLI RX|available   auto     warm         0    vcl7
**** v1   CLI RX|active      auto     warm         0    vcl8
**   v1   CLI 200 <vcl.list>
**   top  === varnish v1 -cliok "vcl.deps"
**** v1   CLI TX|vcl.deps
***  v1   CLI RX  200
**** v1   CLI RX|vcl1
**** v1   CLI RX|vclA   vcl1
**** v1   CLI RX|vcl2   vclA
**** v1   CLI RX|vclB   vcl2
**** v1   CLI RX|vcl3   vclB
**** v1   CLI RX|vcl4   vclA
**** v1   CLI RX|vcl5   vclA
**** v1   CLI RX|vcl6   vclA
**** v1   CLI RX|vcl7   vclA
**** v1   CLI RX|vcl8
**   v1   CLI 200 <vcl.deps>
**   top  === varnish v1 -cliok "vcl.discard vcl1 vcl2 vcl3 vcl4 vcl5 vcl6...
**** v1   CLI TX|vcl.discard vcl1 vcl2 vcl3 vcl4 vcl5 vcl6 vcl7 vclA vclB
***  v1   CLI RX  200
**** v1   CLI RX|VCL vcl3 discarded
**** v1   CLI RX|VCL vclB discarded
**** v1   CLI RX|VCL vcl2 discarded
**** v1   CLI RX|VCL vcl4 discarded
**** v1   CLI RX|VCL vcl5 discarded
**** v1   CLI RX|VCL vcl6 discarded
**** v1   CLI RX|VCL vcl7 discarded
**** v1   CLI RX|VCL vclA discarded
**** v1   CLI RX|VCL vcl1 discarded
**   v1   CLI 200 <vcl.discard vcl1 vcl2 vcl3 vcl4 vcl5 vcl6 vcl7 vclA vclB>
```

nigoroll: ah ok. So I agree that the vcls used for `vclA` returns could be named. But It would appear helpful in `vcl.list`

Dridi: But you can hardly do that in the tabular output, it would only work for the json output. That's why I kept `vcl.deps` in my patch series.

nigoroll: ok.
but vcl.list should still show the fact, e.g. `<- (x return(NAME)'s)`

nigoroll:
bugwash: we would prefer if a discard was _atomic_ in that it either succeeded fully or failed without any change.

Dridi:
I have an idea on how to make this atomic without much more effort, but I have one question:

If we go atomic should we tolerate duplicate VCL names?

```
vcl.discard foo bar foo
```

I feel that once we go atomic it becomes acceptable.

nigoroll:
@Dridi I see no problem with duplicate names in the discard command

Dridi:
Same feature, but this time all-or-nothing.

nigoroll (approved):

bsdphk:
OK with me.

PR #3380, opened 2020-08-07
WIP FYI B64 vmod blob perf

nigoroll:
to supplement #3376, continuing there
```
**** v1   vsl|       1001 RespHeader      c t256: 3.021
**** v1   vsl|       1001 RespHeader      c t512: 5.200
**** v1   vsl|       1001 RespHeader      c t1024: 10.050
```


bsdphk:
I've decided we leave vmod_blob alone for now

PR #3386, opened 2020-08-14
Pipe workspace fix

### LINKED ISSUE #3329: Reason is lost when returning synth from pipe

### LINKED ISSUE #3385: ctx in vcl_pipe mixes req and bo priv pointers
So I was hunting a very odd bug when backporting some of the recent shard patches around `vcl_pipe`. This bug seems to effect anything using a dynamic priv pointer in `vcl_pipe`, so I believe its different than (#3329 and #3330).

Basically, when we construct the `ctx` for for `vcl_pipe`, we favor `bo`:

https://github.com/varnishcache/varnish-cache/blob/83ab25a259374194bda0ae222a5484cb1f2091c4/bin/varnishd/cache/cache_vrt_vcl.c#L430-L443

And `ctx->ws` is a `bo` workspace.

Here, when we pull a dynamic priv, we favor `req`:

https://github.com/varnishcache/varnish-cache/blob/83ab25a259374194bda0ae222a5484cb1f2091c4/bin/varnishd/cache/cache_vrt_priv.c#L147-L162

And this line mixes `req` and `bo`:

https://github.com/varnishcache/varnish-cache/blob/83ab25a259374194bda0ae222a5484cb1f2091c4/bin/varnishd/cache/cache_vrt_priv.c#L150-L151

This means that we are going to always put bo workspaced privs on the req priv list.

This has slipped under the radar for a few reasons. The main one being that we dont have enough checks in place when freeing the priv list. We also zero out the workspace when done, which basically masks the problem when we iterate the priv list, as that skips the free and terminates the loop.

Anyway, a few different ways to fix this. I would like to discuss this before making the PR to fix this, so please do not jump out ahead with a fix, thank you :)

rezan:
This uses `req->ws` for `ctx->ws` when piping. Previously the `bo` workspace was used, which lead to privs being registered in the wrong context and client data being lost after the `bo` completes.

This is a 6.0 backport patch.

Fixes #3329
Fixes #3385

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -768,6 +768,7 @@ cnt_pipe(struct worker *wrk, struct req *req)
 	THR_SetBusyobj(bo);
 	bo->sp = req->sp;
 	SES_Ref(bo->sp);
+	VCL_TaskEnter(bo->vcl, bo->privs);
 
 	HTTP_Setup(bo->bereq, bo->ws, bo->vsl, SLT_BereqMethod);
 	http_FilterReq(bo->bereq, req->http, 0);	// XXX: 0 ?
@@ -802,6 +803,7 @@ cnt_pipe(struct worker *wrk, struct req *req)
 	}
 	http_Teardown(bo->bereq);
 	SES_Rel(bo->sp);
+	VCL_TaskLeave(bo->vcl, bo->privs);
 	VBO_ReleaseBusyObj(wrk, &bo);
 	THR_SetBusyobj(NULL);
 	return (nxt);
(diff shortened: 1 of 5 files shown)


rezan:
Tested, looks good.

rezan:
Adding https://github.com/varnishcache/varnish-cache/pull/3408/commits/810e06998a4c47082f9248a59a7a4f4feb0c80fb

PR #3387, opened 2020-08-18
Get rid of the malloc'ed VSB for synth bodies

bsdphk:
This is a PR for discussion purposes relative to making *.body able to ingest BLOBs (ie: #3123 bis)

I'm not entirely happy with the gains of this patch.

There are also a couple of error conditions which seems to fall through the cracks.


bsdphk:
The r1990 failure is known and expected.

I should add that the end-goal here, is that one can set up a VFP filter-list so that a synth response can use ESI or gzip or ...

That is vastly complicted by the fact that one can start composing a body, then scrap that and start over again.

PR #3395, opened 2020-08-25
vsc: New 'F' argument to force field inclusion

### LINKED ISSUE #3394: varnishstat with both -f and curses is weird
This applies to all versions since at least `6.0`. It might just be me being dense, and it might only require a doc fix, but trying to filter out lines from the `curses` output just didn't work.

For example, I wanted to snip the backend counters and went with:

```
varnihstat -1 -f '^VBE.*'
```

And I was greeted by a screen with only four counters, the ones allowlisted here:

```
// bin/varnishstatvarnishstat.c +337
	if (curses) {
		if (has_f) {
			AN(VSC_Arg(vsc, 'f', "MGT.uptime"));
			AN(VSC_Arg(vsc, 'f', "MAIN.uptime"));
			AN(VSC_Arg(vsc, 'f', "MAIN.cache_hit"));
			AN(VSC_Arg(vsc, 'f', "MAIN.cache_miss"));
		}
		do_curses(vd, vsc);
	}
```

so they just filtered everything else when i only wanted to block the `VBE` ones.

In my case, commenting out the four `VSC_Arg` lines produced the right output, but I assume they were placed here for a reason, like always including them by default in case of a positive glob.

Dridi:
We need some fields to always be present in some cases with varnishstat
and as a result it disturbs end users settings. The new 'F' VSC argument
adds include patterns taking precedence over 'f' settings and is meant
for internal VUT usage.

Since varnishstat already needs to force some fields in curses mode, it's
the first consumer of the new argument.

Fixes #3394

---

The commits from this branch can be cleanly cherry-picked in the 6.0 branch as of the time of writing. This doesn't break the ABI and is therefore safe to back-port in the LTS branch. The test case fails in 6.0, but I don't understand why, yet:

```
*    top   1.7 RESETTING after bin/varnishtest/tests/r03394.vtc
**** p1    1.7 Sent signal 15
**** p1    1.7 stdout|\33[24;1H\33[?12l\33[?25h\33[?1049l\33[23;0;0t\r\33[?1l\33>
**** p1    1.7 stdout read -1
**** p1    1.7 stderr read 0
**   p1    1.7 WAIT4 pid=1223089 status=0x0100 (user 0.025696 sys 0.032119)
*    p1    1.7 Expected exit: 0x0 signal: 15 core: 0
---- p1    1.7 Bad exit status: 0x0100 exit 0x1 signal 0 core 0
---- p1    2.7 Failed to send signal 9 (No such process)
*    top   2.7 failure during reset
#    top  TEST bin/varnishtest/tests/r03394.vtc FAILED (2.696) exit=2
```

The error occurs after the test script succeeded, during the tear down phase.

This change is incomplete though, it lacks a bit of error handling where the `INCOMPL()` macro is used. Following the documentation I think we have a copy-pasta mistake:

```c
int VSC_Arg(struct vsc *, char arg, const char *opt);
        /*
         * Handle standard stat-presenter arguments
         *      'f' - filter
         *
         * Return:
         *      -1 error, VSM_Error() returns diagnostic string
         *       0 not handled
         *       1 Handled.
         */
```

There is no practical way to forward a diagnostic that will be output by `VSM_Error()` so I suspect it was meant to be `VSC_Error()` except that this function does not exist, probably because there was no diagnostic needed until now.

Adding the new function is kept outside of the scope of this PR because when I looked closer I realized that VSM and VSL code basically share the same diagnostic/error reporting code, and I don't think adding another copy in VSC code is the best way forward. Also, by keeping `VSC_Error()` separate we save ourselves ABI headaches in the 6.0 LTS branch where we can have this feature, and it is technically fine if a knob that is meant to only be used internally (not tainted by arbitrary input) to panic when misused.

### DIFF
--- a/bin/varnishstat/varnishstat.c
+++ b/bin/varnishstat/varnishstat.c
@@ -165,8 +165,7 @@ do_once_cb_first(void *priv, const struct VSC_point * const pt)
 		return (0);
 	op = priv;
 	AZ(strcmp(pt->ctype, "uint64_t"));
-	if (strcmp(pt->name, "MAIN.uptime"))
-		return (0);
+	AZ(strcmp(pt->name, "MAIN.uptime"));
 	val = *(const volatile uint64_t*)pt->ptr;
 	op->up = (double)val;
 	return (1);
--- a/bin/varnishstat/varnishstat.c
+++ b/bin/varnishstat/varnishstat.c
@@ -336,10 +336,10 @@ main(int argc, char * const *argv)
 
 	if (curses) {
 		if (has_f) {
-			AN(VSC_Arg(vsc, 'f', "MGT.uptime"));
-			AN(VSC_Arg(vsc, 'f', "MAIN.uptime"));
-			AN(VSC_Arg(vsc, 'f', "MAIN.cache_hit"));
-			AN(VSC_Arg(vsc, 'f', "MAIN.cache_miss"));
+			AN(VSC_Arg(vsc, 'F', "MGT.uptime"));
+			AN(VSC_Arg(vsc, 'F', "MAIN.uptime"));
+			AN(VSC_Arg(vsc, 'F', "MAIN.cache_hit"));
+			AN(VSC_Arg(vsc, 'F', "MAIN.cache_miss"));
 		}
 		do_curses(vd, vsc);
 	}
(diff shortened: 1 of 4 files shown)


Dridi:
I just solved the 6.0 test failure, now it can really be cherry-picked safely on the stable branch.

Dridi:
@bsdphk would like to replace the `-f` option with explicit and in-order `-I` and `-X` options, I will attempt a competing pull request with this approach.

nigoroll:
#3396 looks like the way forward

Dridi:
Bugwash consensus: see https://github.com/varnishcache/varnish-cache/pull/3396#issuecomment-683869976.

Dridi:
Done in  #3399

PR #3396, opened 2020-08-25
varnishstat: New -I and -X options

Dridi:
See https://github.com/varnishcache/varnish-cache/pull/3395#issuecomment-680058098

Like #3395, it can be cherry-picked cleanly on the 6.0 stable branch.

Unlike #3395, this isn't a strictly additive change because the ordering of filters matter now. This is a small breakage I'm willing to live with in both trunk and 6.0 LTS.

I left the removal of `varnishstat -f` outside of the scope of this patch series, we can remove it next dot-zero release.


Dridi:
Unsurprisingly this change got a bit of push back because it's breaking so let me clarify why I can live with this case.

First, @bsdphk's feedback against #3395:
- `varnishstat -f` has complicated semantics
- some VSL utilities support `-I` and `-X` options
- multiple filtering globs should be order-sensitive

To which I agree, #3395 was an attempt at fixing #3394 without breaking anything. Where I stopped agreeing with the feedback was the lack of solution for required fields to not interfere with regular filtering. This PR ended up with a very similar solution to #3395 by adding a new VSC argument meant for internal use in `varnishstat`.

Unfortunately the push back I got for this PR stopped short at the "breaking" mention in the description without bothering to look at the finer prints in the patches themselves so first let's look at what doesn't break:

- new `-I` and `-X` options (additive)
- `-f glob` is equivalent to `-I glob` (emulated)
- `-f ^glob` is equivalent to `-X glob` (emulated)

The three options can be used together, and the breakage is in the handling of multiple options: they are evaluated in order and the first to match defines the outcome. Like today, if we have at least one inclusion glob, fields that don't match anything are filtered out.

The only case where this becomes a breaking change is for the case where an inclusion glob may overlap with an exclusion glob.

Today the documentation states that inclusion globs take precedence, and VSC filtering does the exact opposite. The filtering code has been in its current state since the 5.2.0 release for the major VSM change to individual maps, and nobody complained that overlapping patterns don't behave as documented.

For these reasons I'm OK with this small breakage.

nigoroll (approved):

Dridi:
Bugwash consensus: merge this patch series in trunk and back-port #3395 in 6.0, but use the same `'R'` VSC argument (instead of `'F'`).

PR #3399, opened 2020-08-31
vsc: Add an 'R' argument for required fields (back-port)

### LINKED ISSUE #3394: varnishstat with both -f and curses is weird

Dridi:
This is a 6.0 variant of #3395 that addressed #3394 for the master branch without breaking any prior behavior, but using the same VSC argument as #3396 instead.

Fixes #3394

### DIFF
--- a/bin/varnishstat/varnishstat_options.h
+++ b/bin/varnishstat/varnishstat_options.h
@@ -36,14 +36,6 @@
 	    "Instead of presenting a continuously updated display,"	\
 	    " print the statistics to stdout."				\
 	)
-#define STAT_OPT_f							\
-	VOPT("f:", "[-f <glob>]", "Field inclusion glob",		\
-	    "Field inclusion glob."					\
-	    " Use backslash to escape characters. If the argument"	\
-	    " starts with '^' it is used as an exclusive glob."		\
-	    " Multiple -f arguments may be given. Inclusive globs"	\
-	    " are accumulative and are run before exclusive ones."	\
-	)
 #define STAT_OPT_j							\
 	VOPT("j", "[-j]", "Print statistics to stdout as JSON",		\
 	    "Print statistics to stdout as JSON."			\
@@ -59,7 +51,7 @@
 	)
 
 STAT_OPT_1
-STAT_OPT_f
+VSC_OPT_f
 VUT_OPT_h
 STAT_OPT_j
 STAT_OPT_l
--- a/bin/varnishstat/varnishstat.c
+++ b/bin/varnishstat/varnishstat.c
@@ -319,10 +319,10 @@ main(int argc, char * const *argv)
 
 	if (curses) {
 		if (has_f) {
-			AN(VSC_Arg(vsc, 'f', "MGT.uptime"));
-			AN(VSC_Arg(vsc, 'f', "MAIN.uptime"));
-			AN(VSC_Arg(vsc, 'f', "MAIN.cache_hit"));
-			AN(VSC_Arg(vsc, 'f', "MAIN.cache_miss"));
+			AN(VSC_Arg(vsc, 'R', "MGT.uptime"));
+			AN(VSC_Arg(vsc, 'R', "MAIN.uptime"));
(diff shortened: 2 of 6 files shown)


bin/varnishstat/varnishstat_options.h
```diff
@@ -36,14 +36,6 @@
 	    "Instead of presenting a continuously updated display,"	\
 	    " print the statistics to stdout."				\
 	)
-#define STAT_OPT_f							\
```
rezan, line 39: Does this have to be redefined `VSC_OPT_f`?

Dridi: It is, but in the correct file.

rezan: Our number one backport rule is don't break anything unless there is a security or safety issue. So breaking this parameter for a new feature is a no go. Your first PR did not have this change, so can we go back to that?

Dridi: I don't understand, `varnishstat_options.h` is a private header so this change increases the visibility of an option that should have been exposed from day one. Removing a `STAT_OPT_?` definition is not breaking change.

#3395 very much had this change too.

bin/varnishstat/varnishstat_options.h
```diff
@@ -59,7 +51,7 @@
 	)
 
 STAT_OPT_1
-STAT_OPT_f
+VSC_OPT_f
```
Dridi, line 54: What you see as a breaking change is visible here has having no effect on `varnishstat`. Instead of a private `STAT_OPT` option this is now a public `VSC_OPT`.

bin/varnishstat/varnishstat_options.h
rezan: I get that, but we also have a patch which does not do this header alteration (#3395 does not alter any headers). So this is definitely pushing the breakage boundary and the new feature boundary, simultaneously. If anything, leave the old `STAT_OPT_f` definition in place and simply add a new definition for `VSC_OPT_f`. I could even ask for 2 separate PRs, 1 for the fix and 1 for the new feature, but in the spirit of getting this thru, I will let this slide :)

Dridi: My bad, this change is not part of #3395, I could have sworn it was in both #3395 and #3396 since both PRs came with a bit of polish in the area.

HOWEVER that still doesn't change the fact that `varnishstat_options.h` is a PRIVATE header that has NO END USER except `varnishstat` because we DON'T INSTALL IT so this is NOT A BREAKING CHANGE. There is nothing to let slide, except how picky you are being on a non-issue.

I hope the ALL-CAPS will help you CONNECT THE DOTS.

rezan: ok, this is an unfortunate response. Anyway, going to put this on hold for a bit and discuss with the backport team. We already finalized 6.0.7, so if this moves forward, we have plenty of time to get it on the next release.

Dridi:
Proceed as you wish, I will leave this to your capable hands.

PR #3400, opened 2020-09-03
vbe: Don't recycle Connection:close sessions

Dridi:
When a user adds a Connection:close header to the bereq, we end up
recycling the session and relying solely on the backend to close
it. This is not ideal when we deal with a misbehaving backend.

Conversely, if a backend replies with a Connection:close header,
or if a user manually adds it to the beresp, we should also not be
recycling sessions.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -235,7 +235,9 @@ vbe_dir_finish(VRT_CTX, VCL_BACKEND d)
 	if (PFD_State(pfd) != PFD_STATE_USED)
 		assert(bo->htc->doclose == SC_TX_PIPE ||
 		    bo->htc->doclose == SC_RX_TIMEOUT);
-	if (bo->htc->doclose != SC_NULL || bp->proxy_header != 0) {
+	if (bo->htc->doclose != SC_NULL || bp->proxy_header != 0 ||
+	    http_GetHdrField(bo->bereq, H_Connection, "close", NULL) ||
+	    http_GetHdrField(bo->beresp, H_Connection, "close", NULL)) {
 		VSLb(bo->vsl, SLT_BackendClose, "%d %s close", *PFD_Fd(pfd),
 		    VRT_BACKEND_string(bp->director));
 		VTP_Close(&pfd);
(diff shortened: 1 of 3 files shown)


bin/varnishd/cache/cache_backend.c
```diff
@@ -235,7 +235,9 @@ vbe_dir_finish(VRT_CTX, VCL_BACKEND d)
 	if (PFD_State(pfd) != PFD_STATE_USED)
 		assert(bo->htc->doclose == SC_TX_PIPE ||
 		    bo->htc->doclose == SC_RX_TIMEOUT);
-	if (bo->htc->doclose != SC_NULL || bp->proxy_header != 0) {
+	if (bo->htc->doclose != SC_NULL || bp->proxy_header != 0 ||
+	    http_GetHdrField(bo->bereq, H_Connection, "close", NULL) ||
+	    http_GetHdrField(bo->beresp, H_Connection, "close", NULL)) {
```
nigoroll, line 240: good catch, but shouldn't we better make sure to set `bo->htc->doclose` appropriately before?

Dridi: done!

bin/varnishd/cache/cache_fetch.c
```diff
@@ -450,6 +450,10 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 	if (http_IsStatus(bo->beresp, 304) && vbf_304_logic(bo) < 0)
 		return (F_STP_ERROR);
 
+	if (bo->htc->doclose == SC_NULL &&
+	    http_GetHdrField(bo->bereq, H_Connection, "close", NULL))
+		bo->htc->doclose = SC_REQ_CLOSE;
+
 	VCL_backend_response_method(bo->vcl, wrk, NULL, bo, NULL);
```
nigoroll, line 458: ... and shouldn't the check for the beresp Connection header happen here?

Dridi: No preference, I'll go with the consensus. At the very least the case where the backend sends a `Connection: close` and we overwrite it is covered.

Dridi: On second thought I find it cleaner to have the distinction between `REQ_CLOSE` and `RESP_CLOSE`, and to avoid touching `beresp` headers `vbe_dir_finish()`.

Done.

Dridi:
Bugwash decision: split this change in two, the `bereq` part should not be added to this September release.

Dridi:
The merge-able half is available at https://github.com/varnishcache/varnish-cache/pull/3405. I'll submit something for `bereq` once #3405 is merged.

PR #3401, opened 2020-09-04
Document correctly vcl_hit built-in VCL

hermunn:
It has been a while since the built-in vcl_hit changed, and this
updates the user guide.

Since 6.0.1 we have had the corresponding change in TTL, grace, keep
change, and this is also mentioned.


hermunn:
I am a bit unsure if the history lesson is desired, but I did the research. Please comment soon, so that we can merge this before the release.

gquintard (approved):

nigoroll:
good catch, it was about time we fix the documentation in this regard, thank you.
Also I think the history lesson is helpful.

Issue #3402, opened 2020-09-07
Poll-read loop for backend probe HTTP responses reading never updates its poll timeout


delthas:
## Current Behavior

In [the code for reading backend probe HTTP reponses](https://github.com/varnishcache/varnish-cache/blob/fe8602b42d98f56379b53eed6ab55ff53984ce10/bin/varnishd/cache/cache_backend_probe.c#L351):
```c
while(1) {
  // ...
  tmo = (int)round((t_end - t_now) * 1e3);
  // ...
  i = poll(pfd, 1, tmo);
  // ...
  i = read(s, vt->resp_buf + rlen, sizeof vt->resp_buf - rlen);
  rlen += i;
  // ...
}
```

`tmo = (int)round((t_end - t_now) * 1e3);` is currently a noop because `t_now` is never updated.

Suppose there is a 2s timeout for backend probes, and connecting takes 0.5s. This leaves us with 1.5s for reading the response, so `tmo = 1500` (*milliseconds*).

Then Varnish runs  `poll(1500); read; poll(1500); read; ...` until a `poll` times out. Which means that if the backend sends 1 byte every 1s, the polls will never timeout. The timeout is effectively an incorrect between-bytes timeout instead of an entire per-probe timeout.

## Expected Behavior / Possible solution

The timeout should be a per-probe timeout. Since `t_end` effectively acts as a deadline for the whole probe, the poll timeout must be recomputed before every poll by running `t_now = VTIM_real()`, *then* updating `tmo`, so that `poll` only has the remaining time until that `t_end` deadline.

ie:
```c
while(1) {
  // ...
  t_now = VTIM_real(); // new!
  tmo = (int)round((t_end - t_now) * 1e3);
  // ...
  i = poll(pfd, 1, tmo);
  // ...
  i = read(s, vt->resp_buf + rlen, sizeof vt->resp_buf - rlen);
  rlen += i;
```

## Context

I found this when working with the Varnish code, not by finding it in production and tracing it back to the code.

PR #3403, opened 2020-09-07
vbe: Update the poll timeout before polling-reading in backend probes

delthas:
Varnish uses a global per-probe timeout for backend probes. When reading
the backend response, Varnish tries to poll and read in a loop, until a
poll timeouts, the streams EOFs or there is an error.

The poll is supposed to timeout when the per-probe timeout ends. This is
currently setup so that `t_end` is the deadline for the probe, set when
the function starts, then the poll waits until `t_end`.

Previously, the poll timeout was never updated, and was always set to
`t_end - t_now` without updating `t_now`, which means that it was
effectively a between-bytes timeout instead of a proper per-probe
timeout.

This fixes this issue by updating `t_now` before updating the
`t_end - t_now` timeout so that the timeout passed to poll effectively
corresponds to a deadline of `t_end`.

See the issue fixed by this commit for more details.

Fixes: #3402

### DIFF
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -348,6 +348,7 @@ vbp_poke(struct vbp_target *vt)
 	while (1) {
 		pfd->events = POLLIN;
 		pfd->revents = 0;
+		t_now = VTIM_real();
 		tmo = (int)round((t_end - t_now) * 1e3);
 		if (tmo <= 0) {
 			bprintf(vt->resp_buf,


Dridi (approved):
I couldn't find a regression leading to that, as far as I know this timestamp update has always been missing.

It would be nice to add a test case.

Dridi:
Bugwash consensus: @dridi to write a test case and merge.

Dridi:
ba98c0935a437a543622799a9188135fce4ca995...9d8cfbee2f391d74c2020c2c3f5f93932e27257c

PR #3407, opened 2020-09-09
add a trivial VRT function to check if a string is a valid header value

nigoroll:
Ref: https://httpwg.org/specs/rfc7230.html#header.fields

>  field-value    = \*( field-content / obs-fold )
>  field-content  = field-vchar [ 1\*( SP / HTAB ) field-vchar ]
>  field-vchar    = VCHAR / obs-text

- we do not accept the obsolete line folding (obs-fold)
- VCHAR is 0x20 to 0x7e
- obs-text is 0x80 to 0xff

So we end up with `!VCT_CTL || HTAB (LF)`

This change would be the basis for new use cases, like

- a type method `.valid_header`
- a feature flag `+validate_headers` which would call `VRT_ValidHdr()` for all `VRT_SetHdr()` calls

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -590,6 +590,35 @@ VRT_UpperLowerStrands(VRT_CTX, VCL_STRANDS s, int up)
 	return (r);
 }
 
+// RFC7232, 3.2 without obsolete line folding:
+// ASCII VCHAR + TAB + obs-text (0x80-ff)
+static inline VCL_BOOL
+validhdr(const char *p)
+{
+	AN(p);
+	for(;*p != '\0'; p++)
+		if (vct_isctl(*p) && !vct_issp(*p))
+			return (0);
+	return (1);
+}
+
+/*--------------------------------------------------------------------*/
+VCL_BOOL
+VRT_ValidHdr(VRT_CTX, VCL_STRANDS s)
+{
+	int i;
+
+	(void) ctx;
+
+	for (i = 0; i < s->n; i++) {
+		if (s->p[i] == NULL || s->p[i][0] == '\0')
+			continue;
+		if (! validhdr(s->p[i]))
+			return (0);
+	}
+
+	return (1);
+}
 /*--------------------------------------------------------------------*/
 
 VCL_VOID
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -597,7 +597,7 @@ validhdr(const char *p)
 {
 	AN(p);
 	for(;*p != '\0'; p++)
-		if (vct_isctl(*p) && !vct_issp(*p))
+		if (! vct_ishdrval(*p))
 			return (0);
 	return (1);
 }
--- a/include/vct.h
+++ b/include/vct.h
@@ -84,6 +84,7 @@ vct_is(int x, uint16_t y)
 #define vct_isxmlnamestart(x) vct_is(x, VCT_XMLNAMESTART)
 #define vct_isxmlname(x) vct_is(x, VCT_XMLNAMESTART | VCT_XMLNAME)
 #define vct_istchar(x) vct_is(x, VCT_ALPHA | VCT_DIGIT | VCT_TCHAR)
+#define vct_ishdrval(x) (((x) >= 0x20 && (x) != 0x7f) || (x) == 0x09)
 
 static inline int
 vct_iscrlf(const char* p, const char* end)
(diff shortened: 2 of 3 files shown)


bsdphk:
In these days and times, shouldn't it be a VRT_Failure to try to set _any_ header to violate basic syntax ?

nigoroll:
I would not want to take away the option for users to know what they are doing, but yes, the proposed `+validate_headers` flag should default to `true` for 7.0

bin/varnishd/cache/cache_vrt.c
```diff
@@ -590,6 +590,35 @@ VRT_UpperLowerStrands(VRT_CTX, VCL_STRANDS s, int up)
...
 }
 
+// RFC7232, 3.2 without obsolete line folding:
+// ASCII VCHAR + TAB + obs-text (0x80-ff)
+static inline VCL_BOOL
+validhdr(const char *p)
+{
+	AN(p);
+	for(;*p != '\0'; p++)
+		if (vct_isctl(*p) && !vct_issp(*p))
```
Dridi, line 600:
```c
/* vct.h ? */
#define vct_ishdrval(x) (!vct_isctl(x) || vct_issp(x))
```

nigoroll: actually, if we touch it, we might as well have 
```c
#define vct_ishdrval(x) (((x) >= 0x20 && (x) <= 0x7e) || (x) >= 0x80)
```
which should be quite a bit more efficient (saving the table lookup).
And,  at least in my mind, it would actually be easier to understand and match to the RFC

Dridi: Or, if I get your simplification of the grammar right:

```c
#define VCT_HEADERVALUE (~VCT_CTL)
#define vct_ishdrval(x) vct_is(x, VCT_HEADERVALUE)
```

nigoroll: oh sorry, I forgot to consider HTAB in my last comment, this is why I have `!vct_issp(*p)` in the PR commit.
So the simplification is
```c
#define vct_ishdrval(x) (((x) >= 0x20 && (x) <= 0x7e) || (x) >= 0x80 || (x) == 0x09)
```
`~VCT_CTL` would not work as it would match for anything but 'VCT_CTL`, e.g. `VCL_CRLF`, which we would not want.

nigoroll:
```c
#define vct_ishdrval(x) (((x) >= 0x20 && (x) != 0x7f) || (x) == 0x09)
```

can we compress it further? ;)

Dridi: Good point, but we can maybe find the right bit set:

```c
#define VCT_HEADERVALUE (~VCT_CTL & ~VCL_CRLF [& ~...])
```

nigoroll: that would be something to miss when more labels are added to `VCT_CTL` later

nigoroll:
pow wow decision:
- no type method
- a feature flag `+validate_headers` which would call `VRT_ValidHdr()` for all `VRT_SetHdr()` calls, default to `on`

nigoroll:
todo: benchmark vct vs. specific `#define vct_ishdrval(x) (((x) >= 0x20 && (x) != 0x7f) || (x) == 0x09)`

nigoroll:
see #3413 the direct comparison is some 10-25% faster for me with with default developer cflags and, interestingly, with `-O6` the VCT variant is 6% / 10% faster for 256byte and 1024byte strings, respectively.

nigoroll:
> with -O6 the VCT variant is 6% / 10% faster for 256byte and 1024byte strings

see https://github.com/varnishcache/varnish-cache/pull/3413#issuecomment-694135983 : i am skeptical now how much of a value the `-O6` results have.

nigoroll:
todo: implement check in SetHdr, controlled by feature flag, default to on
use
```c
#define vct_ishdrval(x) (((uint8_t)(x) >= 0x20 && (uint8_t)(x) != 0x7f) || (uint8_t)(x) == 0x09)
```

andrewwiik:
Wouldn’t this hurt the fact that sometimes we need to store special strings as headers because VCL doesn’t have variables? Like what if I am making a string to pass to a VMOD that is not header compliant but is compliant for the vmod?.

nigoroll:
@andrewwiik I'd recommend to not use headers any more, see [here](https://code.uplex.de/uplex-varnish/varnish-objvar).
But if you wanted to (ab)use headers for possibly invalid values, you could still toggle the feature.

nigoroll:
I'll implement the feature flag in a separate commit

nigoroll:
apologies for the overlooked SQUASHME

Dridi:
To answer @andrewwiik's concern, yes, it's been established so far that people use headers as stand-in variables and this is the only way to pass state between tasks (but only to some extent).

At the same time almost all combinations of 8bit words are valid for a header value, it's only ASCII control characters that are problematic and I doubt that would wreck much havoc in real life.

PR #3408, opened 2020-09-11
Pipe workspace and safety fixes

### LINKED ISSUE #3329: Reason is lost when returning synth from pipe

### LINKED ISSUE #3385: ctx in vcl_pipe mixes req and bo priv pointers

rezan:
This fixes a handful of workspace and safety issues around `vcl_pipe` and privs. I will basically describe each commit and what it addresses:

fd0b113 - Task privs were never properly setup or free'ed in `vcl_pipe`. This leads to silently leaking privs. The setup just nulls a pointer, which was the case without this call.
ea511b0 - `vcl_pipe` has a special director resolve function which mixes `req` and `bo` workspaces.
4caad2c1858afb399f75230916183d31f57fc908 - Since `struct vrt_priv` is never magic checked, zero'ed, or removed, we can either double free or leak. This covered up a lot of unsafeness floating around. When I did a single step of zero'ing out the magic, I started getting a lot of VTC failures, so I went with full safety here. Given we are dealing with user memory, I think we need to be opting for safety over performance.
d239163 - Use `req->ws` as the `ctx` workspace. This is the safe choice since the req workspace outlives everything and we can jump back into client states.

Fixes #3329
Fixes #3385

### DIFF
--- a/bin/varnishd/cache/cache_director.c
+++ b/bin/varnishd/cache/cache_director.c
@@ -208,6 +208,7 @@ VDI_Http1Pipe(struct req *req, struct busyobj *bo)
 	INIT_OBJ(ctx, VRT_CTX_MAGIC);
 	VCL_Req2Ctx(ctx, req);
 	VCL_Bo2Ctx(ctx, bo);
+	ctx->ws = req->ws;
 
 	d = VDI_Resolve(ctx);
 	if (d == NULL || d->vdir->methods->http1pipe == NULL) {
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -719,7 +719,7 @@ cnt_pipe(struct worker *wrk, struct req *req)
 	SES_Ref(bo->sp);
 	VCL_TaskEnter(bo->privs);
 
-	HTTP_Setup(bo->bereq, bo->ws, bo->vsl, SLT_BereqMethod);
+	HTTP_Setup(bo->bereq, req->ws, bo->vsl, SLT_BereqMethod);
 	http_FilterReq(bo->bereq, req->http, 0);	// XXX: 0 ?
 	http_PrintfHeader(bo->bereq, "X-Varnish: %u", VXID(req->vsl->wid));
 	http_ForceHeader(bo->bereq, H_Connection, "close");
@@ -730,7 +730,10 @@ cnt_pipe(struct worker *wrk, struct req *req)
 	}
 
 	bo->wrk = wrk;
-	VCL_pipe_method(req->vcl, wrk, req, bo, NULL);
+	if (WS_Overflowed(req->ws))
+		wrk->handling = VCL_RET_FAIL;
+	else
+		VCL_pipe_method(req->vcl, wrk, req, bo, NULL);
 
 	switch (wrk->handling) {
 	case VCL_RET_SYNTH:
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -717,6 +717,7 @@ cnt_pipe(struct worker *wrk, struct req *req)
 	THR_SetBusyobj(bo);
 	bo->sp = req->sp;
 	SES_Ref(bo->sp);
+	VCL_TaskEnter(bo->privs);
 
 	HTTP_Setup(bo->bereq, bo->ws, bo->vsl, SLT_BereqMethod);
 	http_FilterReq(bo->bereq, req->http, 0);	// XXX: 0 ?
@@ -760,6 +761,7 @@ cnt_pipe(struct worker *wrk, struct req *req)
 	}
 	http_Teardown(bo->bereq);
 	SES_Rel(bo->sp);
+	VCL_TaskLeave(bo->privs);
 	VBO_ReleaseBusyObj(wrk, &bo);
 	THR_SetBusyobj(NULL);
 	return (nxt);
--- a/bin/varnishd/cache/cache_vrt_priv.c
+++ b/bin/varnishd/cache/cache_vrt_priv.c
@@ -220,12 +220,11 @@ VCL_TaskLeave(struct vrt_privs *privs)
 {
 	struct vrt_priv *vp, *vp1;
 
-	/*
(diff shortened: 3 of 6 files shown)


rezan:
I think there is one more bug floating around. As mentioned above, when I remove the `VRBT_REMOVE()`, I get panics. So it smells like uninitialized memory or something more subtle. Will spend another hour or 2 on this today.

nigoroll (changes requested):
we need to resolve https://github.com/varnishcache/varnish-cache/pull/3330#issuecomment-654346165

rezan:
No bug, I was zero'ing out the entire RB tree node while iterating it, which essentially corrupts the RB tree. So this was a bug in my thinking :) I did change the order of a magic check, it was after a dereference.

rezan:
Added 1 more commit. `bereq.http.*` was using the `bo` workspace and this too can leak out to the client. Also added a workspace overflow check to make sure the previous pipe headers made it into the `bereq`. This was spotted by @mbgrydeland.

nigoroll:
> we need to resolve [#3330 (comment)](https://github.com/varnishcache/varnish-cache/pull/3330#issuecomment-654346165)

I still think that this is necessary

Dridi:
I have currently settled on the opinion that `vcl_pipe` is a client subroutine, and that we should do everything in a client context. I first expressed that idea in https://github.com/varnishcache/varnish-cache/pull/3330#issuecomment-633677367.

The shard director should not care whether its current task is client or backend, and should work within the scope of its task. The latest change @rezan pushed was found by @mbgrydeland after we discussed what the semantics of `vcl_pipe` should be and I strongly believe that to make everything easier to reason about and less complicated it should be strictly client-side.

There was one worry expressed, namely the workspace consumption of the `busyobj` infrastructure. I believe that in the absence of a `resp` to set up, workspace consumption sort-of balances itself out. And of course if users run out of workspace because they used to consume a lot of `workspace_backend` for piped sessions, then obviously they can always increase `workspace_client`. Anecdotal, I have yet to see a pipe setup heavy on workspace usage.

nigoroll:
@Dridi yes, my area of concern still is the `busyobj` infrastructure and `PRIV_TASK`s in particular. As the shard director uses a `PRIV_TASK` to remember load balancing decisions, in the absence of `PRIV_TASK` cloning from the client to the backend side, it also needs additional attention.
I am tasking myself with another review.

nigoroll:
Oh, and another major concern is `std.rollback()`.

Dridi:
I haven't reviewed this closely, but I have a few cents to toss.

This patch series fails to address this:

https://github.com/varnishcache/varnish-cache/blob/bc46b11d22088073e75554a07cc72912f834c69f/bin/varnishd/cache/cache_vrt_priv.c#L137-L142

I stated before that I believe that `vcl_pipe` should be treated as a client task, and in its current form `VRT_priv_task()` does the right thing. For this reason I don't see the point of having `bo` privs in fd0b1135b11332f6b9fd691d093779a4fe1d957c since they should currently be (and in my opinion remain) unreachable. Looking at the test cases they seem to be reachable though, so I'm curious but can't engage in a proper review yet.

Second, we need to make sure we don't run out of workspace in core code. I suggest we add a workspace-free function `http_ForceStaticHeader()` that takes static strings like `"\xxConnection: close"` that are known to outlive the task. As long as we use `workspace_backend` this is mostly safe unless you tweak parameters in a way that could overflow right off the bat (waving in @nigoroll's direction) but such a setup would be doomed to never succeed a fetch anyway. However when we switch to `workspace_client` the risk of running out of workspace at that step becomes a more legitimate question (initially raised by @mbgrydeland).

I didn't check whether this pull request comprehensively moves all workspace operations on the client side, or for that matter I also didn't verify that we don't even bother grabbing a slab from `pool_vbo`.

Dridi:
We might take `vcl_pipe` in a different direction for the March release, and on the question of whether it should be strictly client or backend side, pipe mode might actually end up with two threads (client \*and\* backend) working each in one direction. It is not yet defined how pipe mode would look in VCL.

The 6.0 stable branch has a bug fixes that didn't reach consensus on how to address the pipe quirks, but it has the merit of removing known panics.

nigoroll:
In addition to [Dridis update](https://github.com/varnishcache/varnish-cache/pull/3408#issuecomment-723794532) we also discussed the option to remove `return(synth)` from `vcl_pipe {}` in master.
For bugwash: can we nail this down?

nigoroll:
#3470 is basically a review of this PR with some changes

nigoroll:
continued in #3470

PR #3409, opened 2020-09-14
vsb: Harden destroy and fini operations

Dridi:
They either take both a dynamic struct and buffer, or both are static.

We can also use macros from miniobj to tidy the functions and try to
ensure that memset() is not optimized away.

---

I'm wondering whether it makes sense to keep some of the macros
from sbuf and not go straight for miniobj, vas and stdlib facilities. We
already diverged a lot with sbuf, both in terms of downstream changes
and lagging behind upstream.

### DIFF
--- a/lib/libvarnish/vsb.c
+++ b/lib/libvarnish/vsb.c
@@ -529,6 +529,7 @@ VSB_fini(struct vsb *s)
 
 	assert_VSB_integrity(s);
 	assert(!VSB_ISDYNAMIC(s));
+	assert(!VSB_ISDYNSTRUCT(s));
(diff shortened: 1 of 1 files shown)


nigoroll (approved):
OK, but I wonder if we really need to rush it into the release a day before.

Dridi:
I realized that the new API is all-or-nothing while reviewing the release notes, and I think we should ship this change with the appropriate checks.

Dridi:
Removed the miniobj part of the change, we still want to track upstream sbuf as closely as we can.

nigoroll (approved):

PR #3414, opened 2020-09-18
client.ip equals remote.ip, not local.ip

ThijsFeryn:
If *PROXY protocol* is not used, the value of `client.ip` will be the same as `remote.ip`, not `local.ip` as mentioned in the docs.


nigoroll:
thank you!

PR #3416, opened 2020-09-21
Strip Keep-Alive header from any backend response when the client is …

feld:
…using HTTP/2.0

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -109,6 +109,12 @@ sub vcl_miss {
 }
 
 sub vcl_deliver {
+    # Some misbehaving backends force Keep-Alive header which breaks
+    # clients like Safari and Curl.
+    if (req.proto ~ "HTTP/1") {
+        unset resp.http.keep-alive;
+    }
+
     return (deliver);
 }
 


feld:
This happens reliably with Hitch+Varnish serving the Sonarr and Radarr (Mono-based) apps. They shared a common codebase which may be the source of the issue, but I haven't been able to identify if it's the app that's misbehaving or Mono's webserver stack.

Interestingly anyone using Chrome or Firefox can still access the website as they ignore the header.

Dridi:
I think this is a more general problem of filtering out connection/transport headers headers. It should probably be done by core code itself just in case such headers are set in `vcl_deliver` and the built-in is skipped in addition to the filtering happening during client/backend context switches.

feld:
@Dridi actually I think this approach is wrong anyway. After some further testing I observe some performance issues and I think what I'm seeing is that Keep Alive header gets stripped from the response between the backend and Varnish, so Varnish can't reuse connections. I haven't had a chance to verify it, but it seems to make sense based on my understanding of the Varnish architecture.

I believe the only correct approach (in VCL) is to do this in `vcl_backend_response` but we don't have access to the information we need there to make this decision. I'll just have to apply this to the specific backends I know are problematic.

Getting a proper fix into core of Varnish would be great though. If you manage to make a patch I'd be glad to test.

Dridi:
You can set your own `connection: keep-alive` in `vcl_backend_response` if the backend forgot to mention it and Varnish will filter it out from the response during the backend-client switch.

If you want to discuss this in greater details feel free to hit the varnish-misc mailing list. I will reopen this ticket for the sake of discussing whether we should actively filter headers that operate on the connection rather than the application level. Keep-Alive is one such connection header.

feld:
I believe I made the correct fix in https://github.com/varnishcache/varnish-cache/pull/3417

Dridi:
Yes, #3417 looks much better and I believe you have a VCL workaround for now,

PR #3417, opened 2020-09-22
Keep-Alive, RFC 2616 Section 13.5.1

feld:
Treating Keep-Alive like the other Hop-to-Hop headers fixes clients like Safari and Curl which strictly implement HTTP/2.0 and refuse to proceed if they receive a Keep-Alive header on a response to an HTTP/2.0 request.

--

The following HTTP/1.1 headers are hop-by-hop headers:

    - Connection
    - Keep-Alive
    - Proxy-Authenticate
    - Proxy-Authorization
    - TE
    - Trailers
    - Transfer-Encoding
    - Upgrade

### DIFF
--- a/include/tbl/http_headers.h
+++ b/include/tbl/http_headers.h
@@ -52,7 +52,6 @@
 #define S HTTPH_A_PASS
 #define H(s,e,f) HTTPH(s, e, f)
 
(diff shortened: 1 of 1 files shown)


bsdphk:
I miss a VTC ?

bsdphk:
Bugwash conclusion:  Probably more needed than this (correct) patch.  VTCs good idea so we agree what we are fixing.

PR #3420, opened 2020-09-24
Change CCI behaviour for the release branch

mbgrydeland:
This changes the .circleci/config.yaml for the release branch. This accomplishes two things:

The nightly "build packages" behaviour is now for all commits to the release branch, so no need to wait for the nightly cron trigger to get a package building type of build initiated by circleci.

It uses the 6.5 branch of pkg-varnish-cache, which was forked off of the master pkg-varnish-cache for this release. This makes it so that any future changes to pkg-varnish-cache won't influence the building of the release branch. This will be very useful if there ever is a need to make another point release on the release branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ parameters:
     default: "HEAD"
   pkg-commit:
     type: string
-    default: "master"
+    default: "6.5"
 jobs:
   dist:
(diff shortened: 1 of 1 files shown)


mbgrydeland:
Hmm, this PR wants to merge into master instead of the 6.5 release branch. Will close and create a new PR.

PR #3424, opened 2020-09-29
http2: Don't send resp connection-specific headers

Dridi:
Some browsers are strict about this and simply drop responses containing such headers. Since this is not filtering a context switch between a client and a backend transaction (or cache hit) a new filtering flag is added to the HTTP headers table for connection-specific headers. This new flag cannot be compounded as `HTTPH_R_FETCH|HTTPH_A_INS` because the TE header is an exception and left alone, even though trailers aren't supported.

Better diff with the `--ignore-all-space` option.

We could go further and consider any client request containing one as malformed as mandated by RFC 7540.

Closes #3416
Closes #3417

### DIFF
--- a/bin/varnishd/http2/cache_http2_deliver.c
+++ b/bin/varnishd/http2/cache_http2_deliver.c
@@ -241,6 +241,9 @@ h2_build_headers(struct vsb *resp, struct req *req)
 		r = strchr(hp->hd[u].b, ':');
 		AN(r);
 
+		if (http_IsFiltered(hp, u, HTTPH_C_SPECIFIC))
+			continue; //rfc7540,l,2999,3006
+
 		hps = hp_idx[tolower(*hp->hd[u].b)];
 		sz = 1 + r - hp->hd[u].b;
 		assert(sz > 0);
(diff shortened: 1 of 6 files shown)


bin/varnishd/http2/cache_http2_deliver.c
```diff
@@ -241,6 +241,9 @@ h2_build_headers(struct vsb *resp, struct req *req)
 		r = strchr(hp->hd[u].b, ':');
 		AN(r);
 
+		if (http_IsFiltered(hp, u, HTTPH_C_SPECIFIC))
+			continue; //rfc7540,l,2999,3006
```
mbgrydeland, line 245: Wouldn't it be better to reference sections instead of line numbers? When looking up RFCs online they usually don't show line numbers.

nigoroll (approved):

bin/varnishd/http2/cache_http2_deliver.c
nigoroll: true, but we had that format defined before

Dridi: I don't disagree but this is how we get coverage:

https://varnish-cache.org/rfc/rfc7540.html

And this doesn't even cover the entire section.

Dridi:
Added a compliance note for the req side in #3246.

PR #3426, opened 2020-09-30
vmod: Objects can have methods named like constructors

Dridi:
I came across a VMOD that has a method named like the constructor to
conceptually reinitialize the object's state. That's an implementation
detail of the VMOD in question and arguably undefined behavior in
Varnish.

In order to not accidentally break this assumption, add some test
coverage and document this possibility.


Dridi:
I'd be surprised that we'd go against maintaining status quo, but I didn't push directly just in case.

nigoroll (approved):

PR #3429, opened 2020-10-07
vslc: New mmap cursor type for regular files

Dridi:
This cursor is the simplest of all as of now, and also the fastest way
to skim through a -r input when it's a regular file. This is essentially
a file cursor with much less memory churn.

On a 13GB log dump captured during a performance benchmark varnishlog
takes around 6m15s to print all the transactions to /dev/null but the
mmap cursor consistently reduces this time to around 1m01s.

This is particularly useful to audit logs at rest, be it error logs
after an incident or transactions captured during a benchmark-type
scenario.

### DIFF
--- a/lib/libvarnishapi/vsl_cursor.c
+++ b/lib/libvarnishapi/vsl_cursor.c
@@ -456,10 +456,14 @@ static enum vsl_check v_matchproto_(vslc_check_f)
 vslc_mmap_check(const struct VSL_cursor *cursor, const struct VSLC_ptr *ptr)
 {
 	struct vslc_mmap *c;
+	const char *t;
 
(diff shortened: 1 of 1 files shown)


Dridi:
FYI, that's an old idea from 2018:

https://varnish-cache.org/lists/pipermail/varnish-dev/2018-May/009251.html

I already covered points 1 and 2 in #3001 and this is bonus point 4. I have wasted too many cycles waiting for VSL queries lately and decided to finally give it a stab.

Dridi:
The speedup for regular files seems to be around 9x for raw grouping and around 6x for everything else.

JonathanHuot:
That will be very useful for us! post-analysis will be faster...

nigoroll (approved):

mbgrydeland (approved):

PR #3434, opened 2020-10-08
VRB_Free() refcount checking

### LINKED ISSUE #3433: VRB_Free() refcount panic
```
***  v1   debug|Error: Child (16017) Panic at: Thu, 08 Oct 2020 18:58:55 GMT
***  v1   debug|Assert error in VRB_Free(), cache/cache_req_body.c line 272:
***  v1   debug|  Condition(r == 0 || r == 1) not true.
***  v1   debug|version = varnish-trunk revision fb36b7b1e29edb347a4296f1f143ba8197cdfd97, vrt api = 13.0
***  v1   debug|ident = Linux,4.15.0-118-generic,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,epoll
***  v1   debug|now = 14619.502026 (mono), 1602183535.606084 (real)
***  v1   debug|Backtrace:
***  v1   debug|  0x457560: varnishd() [0x457560]
***  v1   debug|  0x457312: varnishd() [0x457312]
***  v1   debug|  0x51dcd5: varnishd(VAS_Fail+0x55) [0x51dcd5]
***  v1   debug|  0x45e7e4: varnishd(VRB_Free+0x114) [0x45e7e4]
***  v1   debug|  0x45fd9a: varnishd(CNT_Request+0xa7a) [0x45fd9a]
***  v1   debug|  0x4a9172: varnishd() [0x4a9172]
***  v1   debug|  0x4a83a2: varnishd() [0x4a83a2]
***  v1   debug|  0x496fe6: varnishd() [0x496fe6]
***  v1   debug|  0x496802: varnishd() [0x496802]
***  v1   debug|  0x49652b: varnishd() [0x49652b]
```

rezan:
We can't really assume any refcount value in VRB_Free(). In theory, it can be any number from 0 to 1 + number of restarts.

Fixes #3433

### DIFF
--- a/bin/varnishd/cache/cache_req_body.c
+++ b/bin/varnishd/cache/cache_req_body.c
@@ -259,17 +259,12 @@ VRB_Ignore(struct req *req)
 void
 VRB_Free(struct req *req)
 {
-	int r;
-
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 
 	if (req->body_oc == NULL)
 		return;
 
-	r = HSH_DerefObjCore(req->wrk, &req->body_oc, 0);
-
-	// a busyobj may have gained a reference
-	assert (r == 0 || r == 1);
+	(void) HSH_DerefObjCore(req->wrk, &req->body_oc, 0);
 }
 
 /*----------------------------------------------------------------------
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -1028,9 +1028,6 @@ vbf_fetch_thread(struct worker *wrk, void *priv)
 	VCL_TaskLeave(bo->privs);
 	http_Teardown(bo->bereq);
 	http_Teardown(bo->beresp);
-	// XXX after 6.4 release:
-	// bereq_body should have 0 or 1 references remaining,
-	// see VRB_Free() for the other end
 	if (bo->bereq_body != NULL)
 		(void) HSH_DerefObjCore(bo->wrk, &bo->bereq_body, 0);
 
--- a/bin/varnishtest/tests/r03433.vtc
+++ b/bin/varnishtest/tests/r03433.vtc
@@ -1,0 +1,41 @@
+varnishtest "req.body and restarts"
+
+server s1 -repeat 4 {
+	rxreq
+	txresp -bodylen 1000
+} -start
+
+varnish v1 -vcl+backend {
+	import std;
+	import vtc;
+
+	sub vcl_recv {
+		std.cache_req_body(1MB);
+		return (hash);
+	}
+
+	sub vcl_backend_response {
+		set beresp.ttl = 0.1s;
+	}
+
+	sub vcl_deliver {
+		if (!req.restarts) {
+			set req.url = "/2";
+			return (restart);
+		}
+	}
+} -start
+
+client c1 {
+	txreq -req PUT -url /1 -bodylen 250000
+	rxresp
+	expect resp.status == 200
+
+	delay 0.2
+
+	txreq -req PUT -url /1 -bodylen 250000
+	rxresp
+	expect resp.status == 200
+} -run
+
+delay 0.2


nigoroll:
The VTC is helpful without question, but I think we should be able to fix this properly. At this point I think the bug is to hold a reference across a restart, will look.

nigoroll:
OK, I see it now: By changing the url, we can fire up up to `req.restarts` backend fetches, good point!
But still I think we should just reflect this insight in the assertion.

bin/varnishd/cache/cache_req_body.c
```diff
@@ -259,17 +259,12 @@ VRB_Ignore(struct req *req)
...
-
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 
 	if (req->body_oc == NULL)
 		return;
 
-	r = HSH_DerefObjCore(req->wrk, &req->body_oc, 0);
-
-	// a busyobj may have gained a reference
-	assert (r == 0 || r == 1);
```
nigoroll, line 272: alternative suggestion:
```diff
diff --git a/bin/varnishd/cache/cache_req_body.c b/bin/varnishd/cache/cache_req_body.c
index a3a771163..4cb01e815 100644
--- a/bin/varnishd/cache/cache_req_body.c
+++ b/bin/varnishd/cache/cache_req_body.c
@@ -268,8 +268,8 @@ VRB_Free(struct req *req)
 
        r = HSH_DerefObjCore(req->wrk, &req->body_oc, 0);
 
-       // a busyobj may have gained a reference
-       assert (r == 0 || r == 1);
+       // each busyobj may have gained a reference
+       assert (r <= req->restarts + 1);
 }
 
 /*----------------------------------------------------------------------

```
?

nigoroll (changes requested):

bin/varnishd/cache/cache_req_body.c
Dridi: Shouldn't we make `HSH_DerefObjCore()` return `unsigned` in addition? It doesn't seem to ever return a negative value.

nigoroll: yes

rezan: I feels a bit odd to assert that we are less than the number of restarts. Is there a particular refcount scenario we are guarding against? Most usages of `HSH_DerefObjCore()` do not bother trying to do this. Im ok leaving this as is.

nigoroll: This is to guard against reference leaks.

rezan: We can still leak and pass thru this assert. We could in theory pull extra valid references and then incorrectly trigger this assert (for example writing the req.body somewhere else). So that's why I feel this doesn't really accomplish much.

Dridi: That's basically to maintain the role of the previous assert that forgot to take restarts into account.

rezan: Right, this is what I'm saying, the previous assert doesn't accomplish anything. It doesn't protect from actual leaks and only serves as a hidden tripwire.

nigoroll: Both asserts are equivalent for the common `req->restarts == 0` case .
In that case, if we were to accidentally get a second reference, we would trigger.
So the assert *does* accomplish something and it *does* protect from actual leaks.
Also, even for the case which you have spotted, it made us think more and understand the code better.

rezan: > In that case, if we were to accidentally get a second reference, we would trigger.

No, because that reference could be properly returned or taken after `VRB_Free()`. Or it could be taken, lost, and then it falls under that extra `+ 1` in the calculation. Again, `refcnt` cannot be safely used to tell us anything in terms of the number of references taken or potential leaks.

Im ok putting a comment in place of the assert describing all the paths a reference could be taken. What do you think?

nigoroll: this is ridiculous. Unfortunately. Again.

PR #3435, opened 2020-10-09
directors: Bring the client director back to life

Dridi:
Or almost: allow the hash director to not take a key and fall back to client.identity instead.

Refs #3276

---

If there's one thing I miss about the Varnish 3 days, it's the client director. It had an abstract ``client.identity`` placeholder with a sensible default value.


Dridi:
Bugwash decision: expose `client.identity` on the backend side.

Otherwise, regarding the future of the hash director itself, #3276.

Dridi:
bc46b11d22088073e75554a07cc72912f834c69f

PR #3438, opened 2020-10-14
Clear vut->sighup even if sighup_f is not defined

### LINKED ISSUE #3437: 6.0.6 LTS: high VUT fstat overhead if sighup_f is not defined
This is very similar to #3436, but affecting the LTS branch. The [do-while mitigation](https://github.com/varnishcache/varnish-cache/blob/b10f51e5ce09004cbf3b859c4dcd90fa8bf61d62/lib/libvarnishapi/vut.c#L423) for #2788 looks like this:

```
do
    i = VSLQ_Dispatch(vut->vslq, vut_dispatch, vut);
while (i == vsl_more && !vut->sighup && !vut->sigusr1);
```
The sighup callback is called [at the beginning of the while loop](https://github.com/varnishcache/varnish-cache/blob/b10f51e5ce09004cbf3b859c4dcd90fa8bf61d62/lib/libvarnishapi/vut.c#L367), and `vut->sighup` is set back to 0:

```
if (vut->sighup && vut->sighup_f) {
    /* sighup callback */
    vut->sighup = 0;
    i = vut->sighup_f(vut);
```

If the client using VUT does not define `vut->sighup_f`, the branch is never taken, `vut->sighup` is never set back to 0, and the same behavior described in #3436 can be observed whenever a HUP signal is received.

ema:
Fixes #3437

### DIFF
--- a/lib/libvarnishapi/vut.c
+++ b/lib/libvarnishapi/vut.c
@@ -364,12 +364,15 @@ VUT_Main(struct VUT *vut)
 	AN(vut->vslq);
 
 	while (!vut->sigint) {
-		if (vut->sighup && vut->sighup_f) {
-			/* sighup callback */
+		if (vut->sighup) {
 			vut->sighup = 0;
-			i = vut->sighup_f(vut);
-			if (i)
-				break;
+
+			if (vut->sighup_f) {
+				/* sighup callback */
+				i = vut->sighup_f(vut);
+				if (i)
+					break;
+			}
 		}
 
 		if (vut->sigusr1) {


rezan:
I think this needs to go into master? We don't merge anything into 6.0 directly.

nigoroll:
This patch LGTM for 6.0, master needs a different patch, see #3439 
Or, more changes would need to be backported to 6.0, https://github.com/varnishcache/varnish-cache/commit/55995ab0f96c458b1b585afe49c3af737e363205 in particular

Dridi:
55995ab0f96c458b1b585afe49c3af737e363205 might be unfit for 6.0.

That patch was even debatable in trunk without a major bump of libvarnishapi. `struct VUT` is not opaque so this is technically breaking the ABI down the line, but we also handle everything in `VUT_Init()` so "sane" VUT clients shouldn't be impacted.

Overall I think back-porting both 55995ab0f96c458b1b585afe49c3af737e363205 and #3439 should be fine, but I'm raising the red flag and ask for extra vigilance before merging anything in 6.0-plus. I need to do some research to see whether I had an opinion on the signal change at the time.

ema:
> [55995ab](https://github.com/varnishcache/varnish-cache/commit/55995ab0f96c458b1b585afe49c3af737e363205) might be unfit for 6.0.

Agreed, and that's why I proposed this patch for 6.0.x instead.

I tried clearing `vut->sighup` at runtime with SystemTap on hosts affected by #3437 :

```
sudo stap -g -e 'probe process("/usr/lib/x86_64-linux-gnu/libvarnishapi.so.2.0.0").statement("VUT_Main@./lib/libvarnishapi/vut.c:425") { if (pid() == 29240) { $vut->sighup = 0 ; exit() } }'
```
And that does immediately solve the issue, so I'm fairly sure that this patch does not break anything unrelated. At least not in obvious ways. :-)

Dridi:
I'm wondering whether a `bin/varnishdebug` could help write VUT test cases similar to what we do with vmod_debug and vmod_vtc. At least for cases like this where the problem is not visible by just observing varnishlog.

Dridi (approved):

PR #3441, opened 2020-10-16
synth: Don't close the connection upon VCL failure

Dridi:
Before that, try to send a minimal 500 response and see whether we can
avoid losing the session, especially for h2 connections.

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -295,7 +295,9 @@ cnt_synth(struct worker *wrk, struct req *req)
 
 	if (wrk->handling == VCL_RET_FAIL) {
 		VSB_destroy(&synth_body);
-		req->doclose = SC_VCL_FAILURE;
+		(void)VRB_Ignore(req);
+		if (req->transport->minimal_response(req, 500))
+			req->doclose = SC_VCL_FAILURE;
 		VSLb_ts_req(req, "Resp", W_TIM_real(wrk));
 		http_Teardown(req->resp);
 		return (REQ_FSM_DONE);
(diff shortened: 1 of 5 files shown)


Dridi:
I didn't fully realize that we would close the connection on purpose for a failure in any client subroutine.

I think we need to revisit this because failing becomes show stopper for h2 sessions. In its current state the pull request only solves the problem for `vcl_synth` and is therefore incomplete.

Dridi:
CI failure is related to #3442.

nigoroll:
can you maybe motivate this change a bit more? In my mind, a vcl failure should be rather exceptional and thus justifies a full connection loss, which has the advantage of reducing risks from other flakyness, of which the vcl failure may just be symptom.
To rephrase, would it not make more sense to avoid vcl failures instead of handling them more efficiently?

Dridi:
Below is a (not so) quick summary of my motivation and a few insights I shared during bugwash yesterday.

It was initially brought to my attention that failing in `vcl_synth` would close the connection without even sending a response. So I initially planned to send a minimal response depending on the client transport. Adding coverage I realized that h2 connections were subject to sudden death too: I find inconceivable to sacrifice potentially dozens of ongoing streams just because one failed. In particular, if the client connection doesn't actually belong to one actual client because there are other hops in front of Varnish, one bad apple shouldn't take the whole tree down.

As I understand that failing should mean failing hard, I also firmly believe that we shouldn't need to leak the failure outside of the task itself. In that regard h2 makes a distinction between stream errors and protocol errors, and I believe that a client task error should be treated similarly to a stream error.

Even h1 clients may do aggressive pipe-lining of requests, or may be subject to running out of ephemeral ports if they establish connections too frequently (multi-tier Varnish cluster for example) and for this reason I would also prefer we preserve the connection when there is nothing wrong at the transport/protocol level.

Finally, I would like to remind ourselves that we used to panic on workspace exhaustion and that failure resilience has dramatically improved. My main motivation is to keep moving towards more resilience.

Issue #3374, opened 2020-07-29
autocrap-issue: doc/sphinx/include is mode 755


bsdphk:
If I `git clone` trunk and run `autogen.des` the `doc/sphinx/include` directory gets created with mode 755 which means I cannot write files into it:

    critter phk> ls -la doc/sphinx/include/ | head
    total 2
    drwxr-xr-x   2 phk  phk   2 Jul 29 21:55 .
    drwxrwxr-x  11 phk  phk  19 Jul 29 21:55 ..

Therefore `make distcheck` fails.

bsdphk:
Finally found out how to reproduce this:

    mkdir /somewhere/empty
    cd /somewhere/empty
    git clone \
        https://github.com/varnishcache/varnish-cache.git \
        varnish-cache

    sed '
    /^submit/a\
            return
    ' varnish-cache/tools/vtest.sh > vtest.sh

    diff -u varnish-cache/tools/vtest.sh vtest.sh || true

    env MAXRUNS=1 sh -x vtest.sh

Dridi:
Unfortunately the reproducer didn't fail on my machine.

Dridi:
It should be fixed by #3443, please reopen if it still happens.

PR #3443, opened 2020-10-21
sphinx: Accomodate VPATH builds

Dridi:
It is not possible to give sphinx-build multiple source directories, and
instead of trying to link missing files in $(builddir) to their $(srcdir)
counterpart we can merge them both directly in a new directory. But we
can't also merge in a way that would overwrite files:

    cp -Rf $(srcdir) $(builddir) <mergedir>

During a distcheck, on systems that support it, the source directory is
write-protected, and to some extent so are build directories. So instead
we need to proceed the other way around from $(builddir) to $(srcdir)
where files are copied in the merge directory only if there isn't already
one.

To spice things up a notch, we also need sources from a different tree,
and we shouldn't recursively copy $(builddir)/build in the merge
directory.

As it is not practical to encode all those rules into a single make
target it is moved to a dedicated script.

Refs #3374


bsdphk:
This seems to solve #3374 for me. (n=1 :-)

Dridi:
@bsdphk can you please try again? @nigoroll reminded me of a corner case, so I undid changes in the branch and went for a straight fix.

bsdphk:
I tested this with "make -j 8 distcheck" on FreeBSD and it passed with no problems.

nigoroll (approved):
:+1:

Dridi:
0e12fb1d8...4232439a7

PR #3445, opened 2020-10-22
vsl: New Timestamp records for backend tasks

Dridi:
Unlike client transactions where it's usually easy to attribute latency, we lack some serious insights on the backend side. I had a hard time finding names and eventually gave up on the idea not to collide with client timestamps, but client and backend tasks already share Timestamp:Start so maybe it's fine to actually repurpose some?

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -168,6 +168,7 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, struct backend *bp,
 		return (NULL);
 	}
 
+	VSLb_ts_busyobj(bo, "Connected", W_TIM_real(wrk));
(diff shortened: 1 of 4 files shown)


bsdphk:
Not thrilled about the "Process" name, it does not say "only headers processed" to me.

nigoroll (approved):

Dridi:
> Not thrilled about the "Process" name,

I'm not thrilled either by non-constructive comments, especially when I said in the description that I had a hard time finding proper names myself. So yes, we agree that it may not be the best. The "connected" timestamp for example works for our current HTTP/1 backends, but wouldn't be adequate for a stream-based transport.

> it does not say "only headers processed" to me.

And yet that's what it currently does on the client side.

nigoroll:
no objections from bugwash

PR #3448, opened 2020-10-26
Add """ ... """ as a long string format in VCL

andrewwiik:
This PR adds """ ... """ as a long string format to VCL. The purpose of this is to allow JSON to be more easily written in VCL as the current string format(s) in VCL heavily conflict with JSON. While I wanted to use backticks (`) @bsdphk said we should go with triple double quotes to keep in line with borrowing things from python.

### DIFF
--- a/bin/varnishtest/tests/v00019.vtc
+++ b/bin/varnishtest/tests/v00019.vtc
@@ -18,6 +18,11 @@ varnish v1 -errvcl {Unterminated long-string, starting at} {
 	{" }
 }
 
+varnish v1 -errvcl {Unterminated long-string, starting at} {
+	backend b { .host = "127.0.0.1"; }
+	""" ""
+}
+
 varnish v1 -errvcl {Unterminated string at} {
 	backend b { .host = "127.0.0.1"; }
 	"
--- a/bin/varnishtest/tests/v00066.vtc
+++ b/bin/varnishtest/tests/v00066.vtc
@@ -1,0 +1,22 @@
+varnishtest "long string coverage"
+
+varnish v1 -vcl {
+	backend default none;
+
+	sub vcl_recv {
+		return (synth(200));
+	}
+
+	sub vcl_synth {
+                set resp.body = """{"key":"value"}""";
+                return (deliver);
+	}
+} -start
+
+client c1 {
+	txreq
+	rxresp
+	expect resp.status == 200
+        expect resp.bodylen == 15
+	expect resp.body == {{"key":"value"}}
+} -run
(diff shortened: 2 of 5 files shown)


bin/varnishtest/tests/v00019.vtc
```diff
@@ -18,6 +18,11 @@ varnish v1 -errvcl {Unterminated long-string, starting at} {
 	{" }
 }
 
+varnish v1 -errvcl {Unterminated long-string, starting at} {
+	backend b { .host = "127.0.0.1"; }
+	""" ""
```
gquintard, line 23: isn't the problem here that even if the string was fine, it'd be in the wrong place. In other words: are you testing the right thing?

bin/varnishtest/tests/v00066.vtc
```diff
@@ -0,0 +1,22 @@
...
+
+varnish v1 -vcl {
+	backend default none;
+
+	sub vcl_recv {
+		return (synth(200));
+	}
+
+	sub vcl_synth {
+                set resp.body = """{"key":"value"}""";
```
gquintard, line 11: this needs more `\n` and `{"`

bin/varnishtest/tests/v00019.vtc
andrewwiik: According to the previous test no as we are testing throwing this specific error.

bin/varnishtest/tests/v00066.vtc
andrewwiik: This is already tested in the built-in VCL for {", if you think a different string for the test would be better please let me know what that string would be.

gquintard:
```
std.log("""
{" "} "  
""");
```

would work for me, but `varnishtest` may not cooperate.

bin/varnishtest/tests/v00019.vtc
gquintard: fair enough, I missed that you are checking the error string too

bin/varnishtest/tests/v00066.vtc
Dridi: If you have a `{...}` balance issue in `varnishtest` multi-line strings you can often work around it with VCL:

```
varnish v1 -vcl+backend {
    import std;
    sub vcl_recv {
        std.log("{");
        # adding '}' for balance
    }
}
```

Add the right brace at the right location in a comment.

Dridi (approved):

Dridi:
I haven't decided whether it would be worth it to fold `{"..."}` and `"""..."""` parsing together since it's the exact same code.

andrewwiik:
@Dridi I thought about that too but there are a lot of places where you need to either add multiples of 2 or 3 and I thought doing a lot of ternary operators would be annoying there.

nigoroll:
bugwash: add docs, then merge

andrewwiik:
@Dridi merge me pls :)

Dridi:
7a12b1c3ff...3ec74a3aa2

PR #3451, opened 2020-10-28
vsl: Give up when it's too late to buffer records

Dridi:
When a VUT is slow enough, it might very well be overrun while it is
scanning logs. For our built-in VUTs like varnishncsa or varnishlog
this can happen if writing the output can block waiting for IO ops
or when the output is piped to a slow consumer.

---

The assertion from 8a4e28f17c2ee39237a2c56fa0694f2f0b73e3f8 triggered in production.


nigoroll (approved):
asking @mbgrydeland for the final word

mbgrydeland (approved):

PR #3454, opened 2020-11-05
Support copying PRIV_TASKs from the client to the backend side

nigoroll:
Until now, more or less the only option to copy data from the VCL client side to the backend side is through headers, as `bereq.http` is initialized as a copy of `req.http` (additions to this are the recently changed `client.identity` and `req.backend_hint`).

We now add the option for vmods to register a callback per `PRIV_TASK` which can then be used to copy vmod owned data from the client to the backend side.

Callback registrations are kept on a list hanging off `struct req` on demand. When a backend request is created, the registered functions are called in order of registration.

The base memory cost to this is one pointer added to `struct req`. If no registrations are used, this is the only impact on memory requirements.
The runtime overhead when no callbacks are registered is close to a noop: one function call to `vbf_privs_copy()`, which only checks a pointer (besides the common `CHECK_OBJ_*` argument checks).

Alternatives considered:

A straight forward alternative would be a per `PRIV_TASK` copy function pointer paralleling the free pointer. This option would charge every `PRIV_TASK` with an additional pointer. Under the assumption that `PRIV_TASK` copy operations will be required for select data only, this option was dismissed. It should be noted though that this option could be more memory efficient if a relevant proportion of `PRIV_TASKs` requested copying (one pointer vs. approx 4 pointers for this solution).

Another option was to add an event signalled to vmods for backend request creation (basically `STP_MKBEREQ`). This would allow for the same functionality as proposed here, but vmods would be responsible to track all `PRIV_TASK`s requiring copying themsevles. This tracking would be similar to what is suggested herein, but without the reuse.


rezan:
Interesting, but I think the better way to approach this would be to create a data serializer with the serialization format being HTTP safe. This would better encompass all the different kinds of architectures Varnish is run in. We did this for our VHA product with great success [[1](https://docs.varnish-software.com/varnish-high-availability/architecture/)]. Serialization would allow for the data to flow across process, container, and server boundaries. It also means the consumer of this data does not even need to be Varnish. There is definitely a growing trend of people using Varnish in highly distributed architectures where the client and backend are not on the same server and VCL is executed in a distributed manner [[2](https://developer.fastly.com/learning/concepts/clustering/)][[3](https://info.varnish-software.com/blog/creating-self-routing-varnish-cluster)]. This is a *very* powerful feature of VCL, whether intended for or not.

In the specific case where you want to serialize on the same server across the workspace boundary, you would just place the 2 serializers inline (serialize => deserialize) and your memory state is properly transferred, probably thru a header. Not even sure if anything needs to be implemented for this in Varnish core, your vmods can just expose a serialization API and your all set.

[1] https://docs.varnish-software.com/varnish-high-availability/architecture/
[2] https://developer.fastly.com/learning/concepts/clustering/
[3] https://info.varnish-software.com/blog/creating-self-routing-varnish-cluster

nigoroll:
This has been discussed during bugwash. An alternative implementation by replacing `(struct vmod_priv).free` with a pointer to a method struct was generally favored.
I will follow up.

Notes:
* race when copy function creates another priv
* this implementation will need to iterate all privs

nigoroll:
As [@rezan 's comment](https://github.com/varnishcache/varnish-cache/pull/3454#issuecomment-723150111) came up in a discussion:
I am with you that serializing should be possible ([I am also using the same concept, just more explicitly in VCL](https://code.uplex.de/uplex-varnish/libvmod-cluster)), and, in my mind, it could happen in the copy function. While I agree that this could work without it in traditional vmod functions, the copy call back could optimize for the case that the delivery and fetch functions (to use fastly speak) reside on the same node and result in a cleaner interface for the case that they are not.

rezan:
Sorry if this was previously discussed, but what is the usecase for this feature?

nigoroll:
we have used bugwash to clarify :) but ftr:
* right now, the shard director can not be used fully on the client side because it needs to create state which is the needed on the backend side
* in general, we already have `req.backend_hint` copied to `bereq.backend` and `client.identity` copied to the backend side also. It would appear inconsequential to bless certain objects but not vmod state in general.

I also mentioned that I am by no means interested in breaking distributed setups, all my important projects depend on those working smoothly. Details would need to be worked out, but I think extending the priv methods with a (de)serialization interface could greatly improve such setups.

PR #3456, opened 2020-11-10
Count ESI subrequests

rezan:
This has been on my TODO for many years, finally got around to making the PR. The problem is that when you make an ESI subrequest, its not considered a `client_req`. However, it gets accounted for in all the other counters like `cache_hit`, `cache_miss`, etc. This leads to situations where you can see way more hits and misses than requests, which is confusing. Adding this counter will bring these counts back into equilibrium.

### DIFF
--- a/bin/varnishd/VSC_main.vsc
+++ b/bin/varnishd/VSC_main.vsc
@@ -97,6 +97,12 @@
 
 	The count of parseable client requests seen.
 
+.. varnish_vsc:: esi_req
+	:group: wrk
+	:oneliner:	ESI subrequests
+
+	Number of ESI subrequests made.
+
 .. varnish_vsc:: cache_hit
 	:group: wrk
 	:oneliner:	Cache hits
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -139,6 +139,7 @@ ved_include(struct req *preq, const char *src, const char *host,
 
 	VSLb_ts_req(req, "Start", W_TIM_real(wrk));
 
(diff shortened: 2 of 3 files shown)


bin/varnishd/VSC_main.vsc
```diff
@@ -875,6 +875,12 @@
 	:oneliner:	HCB Inserts
 
 
+.. varnish_vsc:: esi_req
+	:group: wrk
+	:oneliner:	ESI subrequests
+
+	Number of ESI subrequests made.
+
```
nigoroll, line 883: would this more naturally placed below `client_req`?

nigoroll (approved):
makes a lot of sense to me, I wonder why I never missed that counter

bsdphk:
OK with me.

bin/varnishd/VSC_main.vsc
Dridi: Agreed.

Dridi (changes requested):
I would like to see one additional commit to fix the hit ratio computation in `varnishstat`.

bin/varnishd/VSC_main.vsc
rezan: Moved

rezan:
> I would like to see one additional commit to fix the hit ratio computation in `varnishstat`.

Looks like the hit rate is calculated by just looking at the hits and misses, so this was never broken for ESI:

https://github.com/varnishcache/varnish-cache/blob/4d858a111c105d9fb6be9a2b8fa18a69d04ac7de/bin/varnishstat/varnishstat_curses.c#L311

PR #3457, opened 2020-11-10
Introduce req.hash_always_pass

gquintard:
The current way to bypass the cache is to either `return(pass)` in
`vcl_recv` or to trigger an HfP/HfM on the backend side.

The former forces you to interrupt the request processing, such as
header normalization, or to set a variable/header to remember your
choice so that you can return `pass` once processing is done.

The latter is nicer but has to deal with request coalescing and it
doesn't really make sense to wait that long before make this decision.

`req.hash_always_pass` mirrors `req.hash_always_miss` but really just
transforms a `return(hash)` into a `return(pass)`. This means that it
has priority over `req.hash_always_miss` but also operate slightly
earlier, at the end of vcl_recv` rather than at the end of `vcl_hash`

`req.uncacheable` might be a better fit, not sure.

### DIFF
--- a/bin/varnishd/cache/cache_req.c
+++ b/bin/varnishd/cache/cache_req.c
@@ -246,6 +246,7 @@ Req_Cleanup(struct sess *sp, struct worker *wrk, struct req *req)
 	req->req_body_status = NULL;
 
 	req->hash_always_miss = 0;
+	req->hash_always_pass = 0;
 	req->hash_ignore_busy = 0;
 	req->esi_level = 0;
 	req->is_hit = 0;
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -855,6 +855,7 @@ cnt_recv_prep(struct req *req, const char *ci)
 		req->d_grace = -1;
 		req->disable_esi = 0;
 		req->hash_always_miss = 0;
+		req->hash_always_pass = 0;
 		req->hash_ignore_busy = 0;
 		req->client_identity = NULL;
 		req->storage = NULL;
@@ -933,6 +934,10 @@ cnt_recv(struct worker *wrk, struct req *req)
 	if (req->req_body_status == BS_ERROR) {
 		req->doclose = SC_RX_BODY;
 		return (REQ_FSM_DONE);
+	}
+
+	if (wrk->handling == VCL_RET_HASH && req->hash_always_pass) {
+		wrk->handling = VCL_RET_PASS;
 	}
 
 	recv_handling = wrk->handling;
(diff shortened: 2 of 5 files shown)


nigoroll:
I am undecided on this for now. On the one hand, it is a trivial change with very low impact, on the other hand, the settable req flags have always bothered me because they conflict with the concept of having code control varnish, not settings. On the latter, I do see a strong case for the existing `hash_always_miss` and `hash_ignore_busy`, which steer `HSH_Lookup()` in a place where we really do not want to call into VCL.
That said, as you noted yourself, it seems there exists no such strong argument for adding a replacement for `return(pass)`. Your motivation for it appears vague to me, can you maybe illustrate the case better with a VCL example?

bsdphk:
I would put the "request processing" in a separate function can call it both from the normal and the pass branches of vcl_recv{}

gquintard:
The main use-case I have is where the VCL is split between generic and specific sections, with a simple top VCL example could look like this:

``` vcl
include "framework_pre.vcl";
include "site.vcl";
include "framework_post.vcl";
```

if `site.vcl` returns early, `framework_post.vcl` is ignored. And as Poul-Henning mentioned, it's possible to ask the writers to explicitly call the post-treatment subroutine, but:
- the goal of the framework is to simplify the writing of `site.vcl`
- `site.vcl` will often be written by less-experiences VCL jockeys
- it's easy to forget to call the post-processing subroutine, specially if there are multiple `return`s
- it's easier to have a "don't `return`" rule

It's possibly not a strong enough case, and I'm absolutely fine with just dropping the PR. I would however try to make the case that the VCL is a bit confusing at the moment, notably with `return(pass)` actually sending to `vcl_hash` and that asymmetry with `req.hash_always_miss` and the lack of `return(miss)`. I believe it doesn't help new users and that PR could maybe do something about this..

nigoroll:
thank you for the illustration of the context, @gquintard 

This reminds me of a request @martin-uplex has bothered me with numerous times: to add "pre return" hooks, e.g. `sub vcl_recv_return_hash` to be called after a `return(hash)` from `vcl_recv {}`. For your case, that would be a place where you could turn a `return(hash)` into a `return(pass)`.
In contrast to this PR, this would at least be a generic concept, not limited to the specific case of turning fetches into passes.
My argument to dismiss this request has always been that VCL authors simply need to adhere to a convention to call a sub instead of returning to achieve the same - for the example above, one would just need to replace `return(hash)` with `call framework_recv_return_hash` or whatever a suitable name.
That said, being able to _hook_ `return()` could result in cleaner code.

So _if_ we actually wanted to address the use case you describe, I would rather want to go this route than special-case the turn fetch-into-pass scenario.

gquintard:
> So if we actually wanted to address the use case you describe, I would rather want to go this route than special-case the turn fetch-into-pass scenario.

I can't argue with that :-)

> That said, being able to hook return() could result in cleaner code.

Totally agree. I have a pretty fun prototype using inline-c that does that, but I'd rather have a solution have a solution that doesn't include "fun", "prototype" or "inline-c"

I'll close this

PR #3459, opened 2020-11-12
vsm: Notify failures to lock shared memory segments

dridi:
This gives a chance to operators to notice failures to prevent VSM to
swap out and an opportunity to increase limits accordingly.

---

With one caveat: `VTC_MODE`.


nigoroll:
unintentionally overtaken by a8b3a8b812acc4eae5f30222bd5f80e7ad7bf6e4

PR #3464, opened 2020-11-18
Methods for vmod_priv

nigoroll:
This is the refactoring we agreed on to enable an alternative implementation of #3454. This PR does not yet introduce the copy callback needed to add the functionality suggested in #3454.

We replace the `.free` pointer of `struct vmod_priv` with a pointer to a methods struct with callbacks. As of this commit, it only contains the former `.free` callback renamed to `.fini`. The purpose of the refactoring is to allow addition of more callbacks later.

The new `struct vmod_priv_methods` also contains a `.type` member pointing to a string to contain an arbitrary description of the type of data any priv holds which uses these methods.

Implementation:

relevant changes are in `cache_vrt_priv.c` and `vrt.h`, other changes are to the documentation and bundled vmods.

The implementation is a simple refactoring for indirection of the call to the `.fini` callback via the methods structure.

### DIFF
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -55,6 +55,8 @@
  * 13.0 (2021-03-15)
  *	Calling convention for VDP implementation changed
  *	Added VRT_ValidHdr()
+ *	struct vmod_priv_methods added
+ *	struct vmod_priv free member replaced with methods
  * 12.0 (2020-09-15)
  *	Added VRT_DirectorResolve()
  *	Added VCL_STRING VRT_BLOB_string(VRT_CTX, VCL_BLOB)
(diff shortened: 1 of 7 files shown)


nigoroll:
bugwash:
* add ctx to fini method if it can be added safely (re-check if we should zero parts of the ctx)
* add a magic to `struct vmod_priv`
* keep the `length` attribute

Dridi:
One question not addressed:

> 16:04 \< slink\> one reason for the free_f signature might have been identity to that of libc::free()

That's merely a convenience, and we can provide that back with a `VRT_libc_free()` symbol for example to spare `malloc()` users to reinvent the same `free_f` over and and over again if we break the signature. We shift the convenience.

include/vrt.h
```diff
@@ -583,11 +585,20 @@ void VRT_Format_Proxy(struct vsb *, VCL_INT, VCL_IP, VCL_IP, VCL_STRING);
...
 typedef int vmod_event_f(VRT_CTX, struct vmod_priv *, enum vcl_event_e);
 
-typedef void vmod_priv_free_f(void *);
+/* vmod_priv related */
+typedef void vmod_priv_fini_f(void *);
+
+struct vmod_priv_methods {
+	unsigned			magic;
+#define VMOD_PRIV_METHODS_MAGIC	0xcea5ff99
+	const char			*type;
```
bsdphk, line 594: Bikeshedding:  Would "owner" be a better name than type ?

nigoroll:
@Dridi and @bsdphk agreed that we should merge this. I have the other changes
* add ctx to fini method if it can be added safely (re-check if we should zero parts of the ctx)
* add a magic to struct `vmod_priv`
near ready and will take them to a new PR.

include/vrt.h
nigoroll: resolved on irc

Dridi (approved):
This does not change current behavior and makes it easier to extend the priv facility in the future.

nigoroll:
FTR on my TODOs before merge:

> add ctx to fini method if it can be added safely (re-check if we should zero parts of the ctx)

this is #3472 

> add a magic to struct vmod_priv

I do not think this makes much because `struct vmod_priv` is `embedded in `struct vrt_priv` . We should rather have https://github.com/varnishcache/varnish-cache/pull/3470/commits/f6b72b0e48fe33ab51768e0986ba7060ae035d14 as part of #3470

PR #3467, opened 2020-11-20
clarify first byte timeout FetchError message

nigoroll:
For a first byte timeout on a new connection, we emit

	FetchError: first byte timeout

For a first byte timeout being reached on a recycled connection, we
should keep the error message similar.

Ref eecd409d13f145765de3aeee4984179e0ae008b5


bsdphk:
I thought we made them different precisely so you could tell if it was a fresh connection ?

Dridi:
It is different, it says "first byte timeout (reused connection)".

FWIW with #3346 it's become much easier to figure whether the connection is fresh so I don't think the distinction is still relevant here.

nigoroll:
bugwash was ok / don't care

PR #3468, opened 2020-11-23
vsl: Performance and stability improvements

Dridi:
This patch series is an attempt at improving the reliability of log utilities:

- reduce opportunities for overruns
- report VSL diagnostics to users
- avoid indefinite error conditions

To fight overruns during traffic peaks, I tried to limit the collection of transactions to only those relevant as early as possible. As a result I pushed a lot of `varnishncsa`'s processing in libvarnishapi, teaching the `-E` option to other log processing tools. In exchange, they stopped processing ESI requests as part of `-c` to be consistent with `varnishncsa`'s behavior. And with this shift of feature towards libvarnishapi, `varnishncsa` was greatly simplified in return.

As I was trying to kill VSL bottlenecks I ran into a VSM segment corruption and had the unpleasant surprise that it was not recoverable and could go on indefinitely without informing the user, for which I also propose a solution.

I'm still looking for more avenues to improve VSL performance, but I consider this patch series ready to be reviewed. Details can be found in individual commit messages.


Dridi:
This is a direct continuation of #3451, because that assertion helped detect frequent overruns (and the fact that they were so frequent they increased the risk of triggering it).

nigoroll:
first review: lgtm. I would like to have a second look at the condidate function though

nigoroll:
FTR: @bsdphk noted on irc that he would abstain from review of this PR

lib/libvarnishapi/vsl_dispatch.c
```diff
@@ -1274,15 +1274,54 @@ vslq_shmref_check(struct VSLQ *vslq)
...
+	struct VSL_data *vsl;
+	enum VSL_tag_e tag;
+	unsigned p_vxid;
+	int i;
+
+	CHECK_OBJ_NOTNULL(vslq, VSLQ_MAGIC);
+	AN(ptr);
+
+	assert(vslq->grouping != VSL_g_raw);
+	if (vslq->grouping != VSL_g_vxid)
```
nigoroll, line 1291: I think the commit message could point out that the change only affects vxid grouping.

lib/libvarnishapi/vsl_dispatch.c
```diff
@@ -1274,15 +1274,54 @@ vslq_shmref_check(struct VSLQ *vslq)
...
+	assert(tag == SLT_Begin);
+	i = vtx_parse_link(VSL_CDATA(ptr), &type, &p_vxid, &reason);
+
+	if (i != 3 || type == VSL_t_unknown)
+		return (0);
+	if (vsl->c_opt && !vsl->b_opt && !VSL_CLIENT(ptr))
+		return (0);
+	if (vsl->b_opt && !vsl->c_opt && !VSL_BACKEND(ptr))
+		return (0);
+	if (type == VSL_t_sess)
```
nigoroll, line 1309: I do not understand why we can do this. Why can we just drop the session transactions here?

nigoroll:
on the review, I would like to add that I do not feel 100% confident and would prefer to leave the final word with martin.
Yet my concern is mostly about `vslq_candidate()` and I think most other changes could be added outside a PR

lib/libvarnishapi/vsl_dispatch.c
Dridi: Because we know that we have at least one of `c_opt` or `b_opt` set, and sessions show up for neither.

nigoroll: side note: as the sessions might live for a long time, is the issue maybe more about the fact that many "client" queries are not interested in the session info (but rather only requests)? Could it make more sense to give sessions their own option on the same level as b, c and

nigoroll: thank you.

lib/libvarnishapi/vut.c
```diff
@@ -382,17 +383,44 @@ VUT_Fini(struct VUT **vutp)
...
 	int i = -1;
-	int hascursor = -1;
+	int hascursor = -1, trycursor = 0;
+	vtim_mono failcursor = NAN;
 
 	CHECK_OBJ_NOTNULL(vut, VUT_MAGIC);
 	AN(vut->vslq);
 
 	while (!VSIG_int && !VSIG_term) {
+		if (trycursor == 10) {
```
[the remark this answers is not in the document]
Dridi: We already have the `-t` option that does something similar for the VSM, the `trycursor` limit is only for the VSL once we are already "connected" to a varnish server. Or maybe I'm missing something else?

lib/libvarnishapi/vsl_dispatch.c
Dridi: It does, but maybe poorly?

> We also need to collect irrelevant transactions for the sake of
grouping, because even when we do not wish to output them we have
to maintain the hierarchy.

nigoroll: I think that is a particularly complicated way to say "only affects vxid grouping". but never mind, it's not so important

lib/libvarnishapi/vut.c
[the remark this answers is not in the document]
nigoroll: I have not tried this myself, but could this be relevant for crashes, when the VSM stays intact?

nigoroll (approved):
approved but would like @mbgrydeland to have the final wor

lib/libvarnishapi/vsl_dispatch.c
```diff
@@ -1274,15 +1274,54 @@ vslq_shmref_check(struct VSLQ *vslq)
...
+	CHECK_OBJ_NOTNULL(vslq, VSLQ_MAGIC);
+	AN(ptr);
+
+	assert(vslq->grouping != VSL_g_raw);
+	if (vslq->grouping != VSL_g_vxid)
+		return (1);
+
+	vsl = vslq->vsl;
+	CHECK_OBJ_NOTNULL(vsl, VSL_MAGIC);
+	if (!vsl->c_opt && !vsl->b_opt)
```
mbgrydeland, line 1296: Isn't there an argument that if (grouping != VSL_g_sess && type == VSL_t_sess), then we should not keep it? Regardless of the state of c_opt and b_opt. It looks to me like we may have been doing a lot of unnecessary session type tracking as was before, and also with this patch.

lib/libvarnishapi/vsl_dispatch.c
```diff
@@ -1299,22 +1338,26 @@ vslq_next(struct VSLQ *vslq)
...
+		ptr = c->rec.ptr;
 	}
 	assert(len > 0);
 	if (vxid == 0)
 		/* Skip non-transactional records */
 		return (r);
 
 	vtx = vtx_lookup(vslq, vxid);
-	if (vtx == NULL && tag == SLT_Begin) {
+	keep = tag != SLT_Begin || vslq_candidate(vslq, ptr);
```
mbgrydeland, line 1356: Should the test of tag shouldn't be moved into vslq_candidate as well, to keep all the logic in one place?

lib/libvarnishapi/vut.c
```diff
@@ -382,12 +383,34 @@ VUT_Fini(struct VUT **vutp)
...
+	}
+}
+
 int
 VUT_Main(struct VUT *vut)
 {
 	struct VSL_cursor *c;
 	int i = -1;
 	int hascursor = -1;
+	vtim_mono failcursor = NAN;
```
mbgrydeland, line 413: Rename to t_failcursor? Makes it clearer that it is a timestamp.

lib/libvarnishapi/vut.c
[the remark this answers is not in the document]
mbgrydeland: I am uncertain of this feature. With this patch, we'd allow a maximum of 1 second (10 * 0.1s) before the log section reappears. I worry that for our very large persistent caches, then that may not be enough. Also I believe the OS may take more than a sec to clean up the old child process when handling a very large memory footprint process like the Varnish child, especially so if there was swap usage involved. Perhaps some logging to stderr would be better, but otherwise continue to retry indefinetely?

lib/libvarnishapi/vsl.c
```diff
@@ -346,6 +346,8 @@ VSL_PrintTransactions(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 			case VSL_t_req:
 				if (!vsl->c_opt)
 					continue;
+				if (t->reason == VSL_r_esi && !vsl->E_opt)
```
mbgrydeland, line 349: Won't this change `varnishlog -c` to not any longer display ESI by default?

mbgrydeland (commented):
Some source line comments added that we should discuss.

lib/libvarnishapi/vsl.c
Dridi: Yes, it is mentioned explicitly as a breaking change in case Someone:tm: would like to bring some patches to a stable branch: 65503181e6b479538698bcc7ef8e7323e13f80ec.

My reasoning is that ESI requests are not client requests, but requests coming from Varnish itself, so it's ok to leave them out of `-c` and grant them a new `-E`.

lib/libvarnishapi/vsl_dispatch.c
Dridi: Before, we definitely tracked sessions for nothing in this scenario. However this patch should prevent session tracking when we don't need them. At least according to my testing.

I focused on vxid grouping because that's where `varnishncsa` operates by default, but it might still be the case for request grouping that sessions are collected for no good reason.

lib/libvarnishapi/vut.c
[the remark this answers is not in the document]
Dridi: My concern today is that a the purely textual diagnostic doesn't tell you whether the error is recoverable or not. For example, X attempts at acquiring a corrupted VSL should be enough to leave because you'll likely loop forever. If the child is down, then you would eventually recover. We probably need a way to know the diagnostic severity to exit the loop.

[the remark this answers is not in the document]
nigoroll: To add facts to this discussion:
```
$ varnishadm param.set auto_restart false
$ varnishlog &
[2] 13430
$ kill 13188
Log abandoned (vsl)
Log abandoned (vsm)
Failed to acquire log: No VSL chunk found (child not started ?)
Log abandoned (vsm)
Log abandoned (vsm)
Log abandoned (vsm)
Log abandoned (vsm)
Log abandoned (vsm)
Log abandoned (vsm)
Log abandoned (vsm)
Log abandoned (vsm)
Log abandoned (vsm)
Failed to acquire log too many times

[2]+  Exit 1                  varnishlog
```
The varnishlog exit happened "immediately", after 1 second.

So my verdict has changed to a clear _no_. We should not have this short, fixed timeout. We might want to rate limit diagnostic output (I like exponential backoff, waving to @slimhazard ), and we might want to offer a parameter for the maximum time to wait, but this definitely is not a good default (and in particular not fixed) behavior.

nigoroll (changes requested):
see comment on timeout showstopper

lib/libvarnishapi/vsl.c
nigoroll: I agree with this

lib/libvarnishapi/vut.c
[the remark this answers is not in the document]
nigoroll: also note the documentation of `-t`:
> -t <seconds|off> ... Defaults to 5 seconds.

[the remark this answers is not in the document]
nigoroll: bugwash: remove from PR, come back to it later ifneedbe

lib/libvarnishapi/vsl_dispatch.c
mbgrydeland: I tested this, and confirmed that we do still with this patch set, keep track of session transactions for request grouping. As these will never be reported (filtered out in vslq_callback()), that work is wasted.

I wonder if this function should look something like:

```
static unsigned
vslq_candidate(struct VSLQ *vslq, const uint32_t *ptr)
{
	enum VSL_transaction_e type;
	enum VSL_reason_e reason;
	struct VSL_data *vsl;
	enum VSL_tag_e tag;
	unsigned p_vxid;
	int i;

	CHECK_OBJ_NOTNULL(vslq, VSLQ_MAGIC);
	AN(ptr);

	assert(vslq->grouping != VSL_g_raw);
	if (vslq->grouping == VSL_g_session)
		/* All are needed */
		return (1);

	vsl = vslq->vsl;
	CHECK_OBJ_NOTNULL(vsl, VSL_MAGIC);
	if (vslq->grouping == VSL_g_vxid) {
		if (!vsl->c_opt && !vsl->b_opt)
			/* Implies also !vsl->E_opt */
			return (1);
		if (vsl->c_opt && !VSL_CLIENT(ptr))
			return (0);
		if (vsl->b_opt && !VSL_BACKEND(ptr))
			return (0);
		/* Need to parse the Begin tag - fallthrough to below */
	}

	tag = VSL_TAG(ptr);
	assert(tag == SLT_Begin);
	i = vtx_parse_link(VSL_CDATA(ptr), &type, &p_vxid, &reason);
	if (i != 3 || type == VSL_t_unknown)
		return (0);

	if (type == VSL_t_sess)
		return (0);

	if (vslq->grouping == VSL_g_vxid && reason == VSL_r_esi && !vsl->E_opt)
		return (0);

	return (1);
}
```

Dridi: We had a discussion about this at some point, I thought you could make `-q`uery sessions but not display them with `-g request` and @mbgrydeland told me irrelevant transactions are removed before testing queries.

So yes, I agree with the suggestion.

Dridi: I pushed 55494ae1a1f5184afc2fdc831241e2b912d07a4b on @mbgrydeland's behalf with the following change:

```diff
        if (vslq->grouping == VSL_g_vxid) {
                if (!vsl->c_opt && !vsl->b_opt)
                        return (1); /* Implies also !vsl->E_opt */
-               if (vsl->c_opt && !VSL_CLIENT(ptr))
+               if (!vsl->b_opt && !VSL_CLIENT(ptr))
                        return (0);
-               if (vsl->b_opt && !VSL_BACKEND(ptr))
+               if (!vsl->c_opt && !VSL_BACKEND(ptr))
                        return (0);
                /* Need to parse the Begin tag - fallthrough to below */
        }
```

Otherwise candidates are discarded too early. This is essentially what my code would do, shaped slightly differently.

Dridi:
Rebased against master, pushed one step further with @mbgrydeland's suggestion to avoid collecting sessions also for request grouping, without the VSL acquire timeout.

nigoroll (approved):

mbgrydeland (approved):

PR #3469, opened 2020-11-25
Added missing timestamp for SessOpen VSL tag

ThijsFeryn:
There is a mismatch between what the `SessOpen` tag returns, and what is documented.

- The documentation has 6 fields
- The real output has 7 fields

Here's an example:

```
-   SessOpen       127.0.0.1 51726 a0 127.0.0.1 80 1606299467.161264 19
```
The missing field is a timestamp. This PR adds the timestamp to `include/tbl/vsl_tags.h`, hoping this change will result in an automatic documentation update.


nigoroll:
thank you and, yes, the documentation on the homepage should auto-update.

PR #3470, opened 2020-11-27
a compromise on workspace / priv handling of vcl_pipe

nigoroll:
This is a follow-up  to #3408 because I accidentally pushed 5cbb3586649839b83438bd99f557cf30c9c78a66 while working on #3464. This also exposed #3385. (the reason for my accidental push was that, unfortunately, I looked at an old `make check` before pushing - my apologies).
At any rate, this has made me re-iterate over #3329, #3385 and #3408 :
So far my position has been [that we need to make vcl_pipe {} live either on the client, or the backend side](https://github.com/varnishcache/varnish-cache/pull/3330#issuecomment-654346165) but not both, which, while it formally is client side, is currently indirectly implied by the availability of symbols like `bereq.http` and `bereq.backend` and the fact that `ctx->ws` is the backend workspace (which is also the cause for #3385 )
While I stand by my point that the backend side would be the right place for clarity, I partly surrender my position in order to solve the currently open bugs for the existing `vcl_pipe {}`, in the hope that we will replace it with something better and cleaner soon.
This basically boils down to making the decision I asked for as _client side_.
I want to emphasize though that I _still fear we might be overlooking something important_ even with this PR: Along this path, we need to always consider `vcl_pipe {}` specifically. As an example, if, hypothetically, in `VRT_r_client_identity()`, `ctx->bo->client_identity` was considered first, we could easily leak backend workspace to the client side again.

But I also do not want to see us starve locked into a corner.

So, why this PR?

I went through the patches from #3408, which basically move all of `vcl_pipe {}` processing to the client side, and made some changes:
* https://github.com/varnishcache/varnish-cache/pull/3408/commits/fd0b1135b11332f6b9fd691d093779a4fe1d957c I did not pick, because it points in the wrong direction. We hand out task privs via `VRT_priv_task()` and as long as that always prefers the `req` privs, it is cleaner if privs do not even exist on the busyobj
* https://github.com/varnishcache/varnish-cache/pull/3408/commits/4caad2c1858afb399f75230916183d31f57fc908 is not a good idea. We do not want to remove objects from an rbtree which we are about to destroy.
* https://github.com/varnishcache/varnish-cache/pull/3408/commits/76e09408a5cda6416e980f102daa615257d8926a and https://github.com/varnishcache/varnish-cache/pull/3408/commits/48911e9efe50a719d383221f885537ad04e0b2bd make `ctx.ws` the client workspace. I have added a commit achieving the same, but in a different way: by simply calling `Req2Ctx()` last such that its changes win.
* https://github.com/varnishcache/varnish-cache/pull/3408/commits/810e06998a4c47082f9248a59a7a4f4feb0c80fb is important to ensure the bereq's `http->ws` points to the client workspace.
* I added a patch to fail `std.rollback(req)`, which would break things big time when the bereq lives on the client workspace. See the last commit for details, we could also solve this in VCC if we accepted potentially breaking vmod functions taking a `req` argument.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -795,6 +795,10 @@ VRT_Rollback(VRT_CTX, VCL_HTTP hp)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	CHECK_OBJ_NOTNULL(hp, HTTP_MAGIC);
+	if (ctx->method & VCL_MET_PIPE) {
+		VRT_fail(ctx, "Cannot rollback in vcl_pipe {}");
+		return;
+	}
 	if (hp == ctx->http_req) {
 		CHECK_OBJ_NOTNULL(ctx->req, REQ_MAGIC);
 		Req_Rollback(ctx->req);
--- a/bin/varnishtest/tests/m00017.vtc
+++ b/bin/varnishtest/tests/m00017.vtc
@@ -3,6 +3,8 @@ varnishtest "Test std.rollback"
 # bug regressions:
 # - #3009
 # - #3083
+# - #3329
+# - #3385
 
 server s1 {
 	rxreq
@@ -335,3 +337,40 @@ client c6 {
 } -run
 
 server s1 -wait
+
+# this could work, but would need additional coding to save
+# the right snapshot of the bereq on the req ws
+varnish v1 -errvcl {Not available in subroutine 'vcl_pipe'} {
+	import std;
+
+	backend proforma None;
+
+	sub vcl_pipe {
+		std.rollback(bereq);
+	}
+}
+
+# vcl_pipe { std.rollback(req); } cannot work unless it also implied
+# rolling back the bereq first.
+# We would want to remove req from vcl_pipe, but that could break
+# vmods, so we fail specifically at runtime
+
+varnish v1 -vcl {
+	import std;
+
+	backend proforma None;
+
+	sub vcl_recv {
+		return (pipe);
+	}
+	sub vcl_pipe {
+		std.rollback(req);
+	}
+}
+
+client c7 {
+	txreq -url /
+	rxresp
+	expect resp.status == 503
+	expect resp.reason == "VCL failed"
+} -run
--- a/bin/varnishd/cache/cache_director.c
+++ b/bin/varnishd/cache/cache_director.c
@@ -206,9 +206,8 @@ VDI_Http1Pipe(struct req *req, struct busyobj *bo)
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 	CHECK_OBJ_NOTNULL(bo, BUSYOBJ_MAGIC);
 	INIT_OBJ(ctx, VRT_CTX_MAGIC);
-	VCL_Req2Ctx(ctx, req);
 	VCL_Bo2Ctx(ctx, bo);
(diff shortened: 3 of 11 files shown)


nigoroll:
from pow wow, just to clarify:
https://github.com/varnishcache/varnish-cache/pull/3470/commits/0b1ebd4e83397f99622a2203f4827503c1384e99 moves pipe vsl to the client side. This could be relevant wrt backwards compatibility

nigoroll:
@bsdphk OKed during pow-wow. TODO: document the "if you want to rollback, do it in vcl_recv{}"
(this deserves more than one line of changelog->release notes anyway)

bin/varnishd/cache/cache_vrt.c
```diff
@@ -795,7 +795,10 @@ VRT_Rollback(VRT_CTX, VCL_HTTP hp)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	CHECK_OBJ_NOTNULL(hp, HTTP_MAGIC);
-	if (hp == ctx->http_req) {
+	if (ctx->method & VCL_MET_PIPE) {
+		VRT_fail(ctx, "Cannot rollback in vcl_pipe {}");
+		return;
+	} else if (hp == ctx->http_req) {
```
Dridi, line 801: nit-pick: I don't see the point of if-else-chaining with the other conditions since the pipe one returns.

bin/varnishtest/tests/m00017.vtc
```diff
@@ -335,3 +337,40 @@ client c6 {
 } -run
 
 server s1 -wait
+
+# this could work, but would need additional coding to save
+# the right snapshot of the bereq on the req ws
+varnish v1 -errvcl {Not available in subroutine 'vcl_pipe'} {
```
Dridi, line 343: I was surprised by this one and surveyed the `bereq` symbols not available in `vcl_pipe`. I think that `bereq.xid` could be made available since there technically is a transaction tied to `bereq`. I would rather see this land in this patch series since this is squaring the pipe semantics.

bin/varnishtest/tests/m00017.vtc
```diff
@@ -335,3 +337,40 @@ client c6 {
...
+	import std;
+
+	backend proforma None;
+
+	sub vcl_pipe {
+		std.rollback(req);
+	}
+	sub vcl_recv {
+		return (pipe);
+	}
```
Dridi, line 368: nit-pick: the subs order disturbs me.

Dridi (changes requested):
Overall LGTM, I only have nit-picks in the diff.

One additional nit-pick in the commit log: can you please credit @rezan as you did previously in the vrt_privs commit (currently 30d71bb2b510f136092fcfb0676cdca35db8b22a) since he had it first even though you came to the same change independently. This mention disappeared with this pull request.

bin/varnishd/cache/cache_vrt.c
nigoroll: it was for the structure, but anyway

bin/varnishtest/tests/m00017.vtc
nigoroll: done

Dridi (approved):
LGTM but I would like @rezan and/or @martin to have a look too.

nigoroll:
@rezan @mbgrydeland my apologies, I had not seen Dridis "but" when I merged.
Would you still have a look?

Issue #3471, opened 2020-11-30
VIP30, plumbing a better vcl_pipe


bsdphk:
See: https://github.com/varnishcache/varnish-cache/wiki/VIP30%3A-Plumbing%3A-vcl_raw%28%29-and-vcl_pipe%28%29

This ticket for discussion.

nigoroll:
I do fully agree that we should make improvements in this area and I think the general approach of adding ``vcl_raw {}`` and a new "pipe-fsm" is the right one.

``vcl_raw {}`` not only lays the foundation for the new pipe handling, but also allows moving C-code handling to VCL, as with the X-F-F case in #3453

Low overhead pipe handling in a single (or a low number of) thread(s) will substantially improve a relevant practical use case.

So, in short, I am very much :+1: overall.

## Suggested clarifications

### separate fsms == separate tasks

My understanding is that the tunnel fsm would be separate from both the client and the backend fsm. This implies that it would also represent a separate _task_ wrt ``PRIV_TASK``, as ``vcl_raw {}`` would.

So we should make clear that the scope of PRIV_TASKs created within ``vcl_raw {}`` and ``vcl_tunnel_* {}`` will be limited that (set of) sub(s).

### separate fsms == separate subset of vcl subs

the VIP specifies ``return(synth)`` from ``vcl_tunnel{}`` and ``vcl_raw{}``. These would require separate incarnation of the synth sub, possibly called ``vcl_tunnel_synth {}`` and ``vcl_raw_synth {}``.

### ``vcl_raw {}``, no rollbacks/restarts and ``req0``

We should clarify that ``vcl_raw {}`` does neither support rollbacks nor restarts.

As whatever modification done in ``vcl_raw {}`` will be the state that the client fsm would roll back to, we should use ``req0`` (or a another name different from ``req``) in ``vcl_raw {}``.

### ``return(pipe)`` from ``vcl_tunnel_* {}``

We should use a different name to disambiguate from the existing ``return(pipe)``. For example, ``return(stream)`` or ``return(connect)``.

## Suggested changes

### Keep ``vcl_pipe {}`` for transition

[...]

PR #3472, opened 2020-11-30
add a VRT_CTX argument to the vmod_priv fini function

nigoroll:
Continuation of #3464:

we take the chance of the breaking change in the previous commit to also add a `VRT_CTX` argument to `vmod_priv_fini_f`. This allows a `vmod_priv` fini function, for example, access to the request or simpler and better logging (because the task's vsl buffer can be used where otherwise only unbuffered vsl was possible).

Implementation:

The meat of this commit really only is the change to the `vmod_priv_fini_f` typedef and `VRT_priv_fini()`.

All other changes are either to bundled vmods or the straight forward infrastructure to make available a ctx to `VRT_priv_fini()`, which, for client and backend context, is called via `VCL_TaskLeave()`.

Consequently, it made sense to also change the signature `Req_Rollback()` and `Bereq_Rollback()`.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -134,8 +134,12 @@ vbf_cleanup(struct busyobj *bo)
 }
 
 void
-Bereq_Rollback(struct busyobj *bo)
+Bereq_Rollback(VRT_CTX)
 {
+	struct busyobj *bo;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	bo = ctx->bo;
 	CHECK_OBJ_NOTNULL(bo, BUSYOBJ_MAGIC);
 
 	if (bo->htc != NULL &&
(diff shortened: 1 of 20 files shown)


Dridi:
Following https://github.com/varnishcache/varnish-cache/issues/3471#issuecomment-738865201 I think the `sub vcl_end` use case would remove the need to get a full context in the `priv::fini()` method.

nigoroll:
@Dridi I can't follow. Can you please elaborate?

nigoroll:
> Following [#3471 (comment)](https://github.com/varnishcache/varnish-cache/issues/3471#issuecomment-738865201) I think the `sub vcl_end` use case would remove the need to get a full context in the `priv::fini()` method.

nigoroll:
sorry

nigoroll:
We should decide on this before release because we already broke the `PRIV_*` interface for the next release and if we want this change, we should avoid a second breaking change in another release.

Also in reply to @Dridi whoose comment I did not quote understand, the advantages I see here are
* have a vsl with the right xid available at fini time, which makes logging from the fini callback significantly more helpful (see for example the changes to v00041.vtc)
  * we have other places where we already save the vsl in the priv to be able to have it available at fini time, which is kinda at the edge of API, and we would make these cases official
* be able to officially call `VRT_fail()` from the fini (see case with your favorite quote _thou shalt not rollet back_ ;) )

nigoroll:
FTR: as phk might not be able to join bugwash today, he gave his OK beforehand.

nigoroll:
bugwash OKed the interface change, @Dridi still to review the code

Dridi (approved):

PR #3475, opened 2020-12-02
Turn the "x" parameter from ExpKill LRU into "vxid"

ThijsFeryn:
The `x` parameter is not very clear. It apparently refers to the _VXID_, so I believe we should name it accordingly.

### DIFF
--- a/bin/varnishd/storage/storage_lru.c
+++ b/bin/varnishd/storage/storage_lru.c
@@ -204,7 +204,7 @@ LRU_NukeOne(struct worker *wrk, struct lru *lru)
 	/* XXX: We could grab and return one storage segment to our caller */
 	ObjSlim(wrk, oc);
 
-	VSLb(wrk->vsl, SLT_ExpKill, "LRU x=%u", ObjGetXID(wrk, oc));
+	VSLb(wrk->vsl, SLT_ExpKill, "LRU vxid=%u", ObjGetXID(wrk, oc));
 	(void)HSH_DerefObjCore(wrk, &oc, 0);	// Ref from HSH_Snipe
 	return (1);
 }
--- a/include/tbl/vsl_tags.h
+++ b/include/tbl/vsl_tags.h
@@ -399,7 +399,7 @@ SLTM(ExpKill, 0, "Object expiry event",
 	"\tE=%f         Old expiry time (unix epoch)\n"
 	"\tf=0x%x       Objcore flags\n"
 	"\tr=%d         Objcore refcount\n"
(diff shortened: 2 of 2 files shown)


Dridi:
I don't think we need to change that, since the `x` field is already documented:

https://github.com/varnishcache/varnish-cache/blob/61a5d17fbae296a80aab70a543a461311cf27849/include/tbl/vsl_tags.h#L395-L402

If anything, I'd rather mask `ExpKill` by default because in my experience it's noise to users most of the time and some of its output is really for debugging, or requires at least some degree of internals knowledge.

bsdphk:
Maybe reconsider what needs to be logged if we're going to change this in the first place ?

nigoroll:
my vote:
- replace nuking with a more sensible, stevedore-specific housekeeping job, which ensures that there is always some free space available and on-demand freeing actually works
- mask ExpKill
- still merge this PR because it makes sense

nigoroll (approved):

Dridi:
Last week's bugwash consensus:

- retire `ExpKill`
- use `Debug` instead
- revisit fields names

Dridi:
I'm having second thoughts on the consensus.

While I still agree that `ExpKill` is mostly noise we could avoid today, I wouldn't retire it since we wouldn't gain an SLT slot anyway, and yet we would lose the documentation of its meaning in the `vsl(7)` manual for the rare times we need to understand what's happening in the expiry thread.

We also use `ExpKill` in test cases, and for a similar reason it's actually more expressive than `Debug` logs.

If we were to mask `ExpKill` by default it would in effect remove the noise. Demoting it this way would be the occasion to slightly tweak its format to make it slightly more actionable:

- before `ExpKill event [field=val]...`
- after `ExpKill event: [field=val]...`

In other words, make the first field a prefix usable in queries.

If we mask it by default, we need to enable it explicitly in 3 test cases, another data point showing how seldom we use it.

On the question of renaming the fields, I'm still leaning on the consistent single-letter names are fine and documented anyway so I wouldn't change that.

That makes this a much less intrusive change:

- add to default mask
- enable in 3 VTCs

nigoroll:
I am OK with Dridi. I never had strong feelings about the field rename, his argument makes sense.

Dridi:
Bugwash consensus: @bsdphk has no opinion, I'll proceed as suggested.

One more reason to do so:

> 15:34 \< martin\> i don't mind keeping them, but think we really ought to silence them by default as suggested. the lock contention they generate on the logs likely are significant

PR #3476, opened 2020-12-02
refactoring to prepare for working with vcl sets more

nigoroll:
@Dridi would this work for you?

This is to facilitate things like
* `vcl.discard *glob*`
* `vcl.discard -a` # all available vcls
* `vcl.discard -k <n>` # keep <n> vcls

### DIFF
--- a/bin/varnishd/mgt/mgt_vcl.c
+++ b/bin/varnishd/mgt/mgt_vcl.c
@@ -45,6 +45,7 @@
 #include "vcli_serve.h"
 #include "vct.h"
 #include "vev.h"
+#include "vav.h"
 #include "vtim.h"
 
 struct vclstate {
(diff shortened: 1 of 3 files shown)


Dridi:
I'm not convinced that we should rely on a new `VAV_Count()` for this since globs could yield more than one VCL. It looks like something that would be undone for the same goal this refactoring does it.

bsdphk:
It seems somewhat counter-intuitive to me to first have VAV_Parse offer to return the argc, but decline, and then later ask VAV what it would have been if we hadn't refused it :-)

As Dridi, I'm not sure this is the proper direction.

We may want a general facility to operate on "set of strings", but this is not it.

Dridi:
I'm wondering whether a binheap would be better, with a comparison function that would look at the number of dependencies and the name. We'd need a VBH_Find() to avoid duplicates prior to insertion. Then we could pop VCLs in topological order to remove them.

We'd probably also need a way to iterate over the entries to find missing deps at some point.

Many blanks to fill even with a binheap...

nigoroll:
Thank you for your quick feedback

Re @Dridi 

> I'm not convinced that we should rely on a new `VAV_Count()` for this since globs could yield more than one VCL

Right. This function could become redundant again, but I have not written the code for the next steps.

Regarding the globs, I thought it might be safest to just allocate the set with as many members as we have vcls, but, again, have not looked closer yet.

Re @bsdphk 

> It seems somewhat counter-intuitive to me to first have VAV_Parse offer to return the argc, but decline, and then later ask VAV what it would have been if we hadn't refused it :-)

Can you please point out the piece of code you are referring to? It is not clear to me.

> We may want a general facility to operate on "set of strings", but this is not it.

I thought about "set of strings", but here the more appropriate data structure is set of vcls.

> I'm wondering whether a binheap would be better

I doubt we should optimize for a large number of VCLs loaded. Yes, an array does not scale, but I see no use in building up a complicated data structure a reuse factor of maybe 2 or 3.

PR #3480, opened 2020-12-07
feature: Transit buffer

fwsGonzo:
See: https://github.com/varnishcache/varnish-cache/wiki/VIP-26%3A-limit-private-prefetch

Ended up using beresp.pass_prefetch as the only difference.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -286,7 +286,9 @@ struct boc {
 	void			*stevedore_priv;
 	enum boc_state_e	state;
 	uint8_t			*vary;
-	uint64_t		len_so_far;
+	uint64_t		fetched_so_far;
+	uint64_t		delivered_so_far;
+	uint64_t		pass_prefetch;
 };
 
 /* Object core structure ---------------------------------------------
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -536,6 +536,10 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 		bo->uncacheable = 1;
 		wrk->handling = VCL_RET_DELIVER;
 	}
+	if (!bo->uncacheable || !bo->do_stream) {
+		/* Pass prefetch should be disabled when not streaming. */
+		oc->boc->pass_prefetch = 0;
+	}
 	if (bo->do_pass || bo->uncacheable)
 		oc->flags |= OC_F_HFM;
 
@@ -584,6 +588,12 @@ vbf_stp_fetchbody(struct worker *wrk, struct busyobj *bo)
 		}
 		AZ(vfc->failed);
 		l = est;
+		/* Limit fetches by pass readahead size */
+		oc = bo->fetch_objcore;
+		if (oc->boc->pass_prefetch > 0 &&
+			l > oc->boc->pass_prefetch) {
+			l = oc->boc->pass_prefetch;
+		}
 		assert(l >= 0);
 		if (VFP_GetStorage(vfc, &l, &ptr) != VFP_OK) {
 			bo->htc->doclose = SC_RX_BODY;
@@ -656,7 +666,7 @@ vbf_stp_fetch(struct worker *wrk, struct busyobj *bo)
 	if (oc->flags & OC_F_PRIVATE)
 		AN(bo->uncacheable);
 
-	oc->boc->len_so_far = 0;
+	oc->boc->fetched_so_far = 0;
 
 	if (VFP_Open(bo->vfc)) {
 		(void)VFP_Error(bo->vfc, "Fetch pipeline failed to open");
@@ -721,7 +731,7 @@ vbf_stp_fetchend(struct worker *wrk, struct busyobj *bo)
 	   give predictable backend reuse behavior for varnishtest */
 	vbf_cleanup(bo);
 
-	AZ(ObjSetU64(wrk, oc, OA_LEN, oc->boc->len_so_far));
+	AZ(ObjSetU64(wrk, oc, OA_LEN, oc->boc->fetched_so_far));
 
 	if (bo->do_stream)
 		assert(oc->boc->state == BOS_STREAM);
@@ -878,7 +888,7 @@ vbf_stp_error(struct worker *wrk, struct busyobj *bo)
 	VSLb_ts_busyobj(bo, "Error", now);
 
 	if (oc->stobj->stevedore != NULL) {
-		oc->boc->len_so_far = 0;
+		oc->boc->fetched_so_far = 0;
 		ObjFreeObj(bo->wrk, oc);
 	}
 
@@ -949,6 +959,9 @@ vbf_stp_error(struct worker *wrk, struct busyobj *bo)
 		VSB_destroy(&synth_body);
 		return (F_STP_FAIL);
 	}
+
+	/* Synthetics don't have any backend fetch to pace */
+	oc->boc->pass_prefetch = 0;
 
 	ll = VSB_len(synth_body);
 	o = 0;
--- a/bin/varnishd/cache/cache_obj.c
+++ b/bin/varnishd/cache/cache_obj.c
@@ -88,6 +88,7 @@
 #include "cache_obj.h"
 #include "vend.h"
 #include "storage/storage.h"
+#include "vtim.h"
 
 static const struct obj_methods *
 obj_getmethods(const struct objcore *oc)
@@ -109,6 +110,7 @@ obj_newboc(void)
 	Lck_New(&boc->mtx, lck_busyobj);
 	AZ(pthread_cond_init(&boc->cond, NULL));
 	boc->refcount = 1;
+	boc->pass_prefetch = cache_param->pass_prefetch;
 	return (boc);
 }
 
@@ -220,6 +222,7 @@ void
 ObjExtend(struct worker *wrk, struct objcore *oc, ssize_t l, int final)
 {
 	const struct obj_methods *om = obj_getmethods(oc);
+	AN(om->objextend);
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(oc->boc, BOC_MAGIC);
@@ -227,10 +230,25 @@ ObjExtend(struct worker *wrk, struct objcore *oc, ssize_t l, int final)
 	assert(l >= 0);
 
 	Lck_Lock(&oc->boc->mtx);
-	AN(om->objextend);
 	if (l > 0) {
+		if (oc->boc->pass_prefetch > 0) {
+			assert(oc->flags & (OC_F_PRIVATE | OC_F_HFM | OC_F_HFP));
+			/* delivered_so_far will be re-read due to memory clobber
+			   from the CondWait opaque function call */
+			while (!(oc->flags & OC_F_ABANDON) &&
+				oc->boc->fetched_so_far > oc->boc->delivered_so_far + oc->boc->pass_prefetch) {
+				/* The client is too far behind, wait for it to catch up and
+				   signal the condvar on the clients part, avoiding deadlocks. */
+				(void)Lck_CondWait(&oc->boc->cond, &oc->boc->mtx, VTIM_real() + 0.1);
+				/* Fallback: Check if we are alone waiting on this object */
+				if (oc->refcnt == 1)
+					break;
+			}
+		}
+
+		/* Fill more bytes */
 		om->objextend(wrk, oc, l);
-		oc->boc->len_so_far += l;
+		oc->boc->fetched_so_far += l;
 		AZ(pthread_cond_broadcast(&oc->boc->cond));
 	}
 	Lck_Unlock(&oc->boc->mtx);
@@ -253,13 +271,19 @@ ObjWaitExtend(const struct worker *wrk, const struct objcore *oc, uint64_t l)
 	CHECK_OBJ_NOTNULL(oc->boc, BOC_MAGIC);
 	Lck_Lock(&oc->boc->mtx);
 	while (1) {
-		rv = oc->boc->len_so_far;
+		rv = oc->boc->fetched_so_far;
 		assert(l <= rv || oc->boc->state == BOS_FAILED);
+		if (oc->boc->pass_prefetch > 0) {
+			assert(oc->flags & (OC_F_PRIVATE | OC_F_HFM | OC_F_HFP));
+			/* Signal the new client position */
+			oc->boc->delivered_so_far = l;
+			AZ(pthread_cond_signal(&oc->boc->cond));
+		}
 		if (rv > l || oc->boc->state >= BOS_FINISHED)
 			break;
 		(void)Lck_CondWait(&oc->boc->cond, &oc->boc->mtx, 0);
 	}
-	rv = oc->boc->len_so_far;
+	rv = oc->boc->fetched_so_far;
 	Lck_Unlock(&oc->boc->mtx);
 	return (rv);
 }
--- a/bin/varnishd/cache/cache_vrt_var.c
+++ b/bin/varnishd/cache/cache_vrt_var.c
@@ -297,6 +297,36 @@ VRT_r_beresp_uncacheable(VRT_CTX)
 
 /*--------------------------------------------------------------------*/
 
+VCL_BYTES
+VRT_r_beresp_pass_prefetch(VRT_CTX)
+{
+    struct objcore *oc;
+
+    CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+    CHECK_OBJ_NOTNULL(ctx->bo, BUSYOBJ_MAGIC);
+
+    oc = ctx->bo->fetch_objcore;
+    CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
+
+    return oc->boc->pass_prefetch;
+}
+
+VCL_VOID
+VRT_l_beresp_pass_prefetch(VRT_CTX, VCL_BYTES value)
+{
+    struct objcore *oc;
+
+    CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+    CHECK_OBJ_NOTNULL(ctx->bo, BUSYOBJ_MAGIC);
+
+    oc = ctx->bo->fetch_objcore;
+    CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
+
+    oc->boc->pass_prefetch = value;
+}
+
+/*--------------------------------------------------------------------*/
+
 VCL_STRING
 VRT_r_client_identity(VRT_CTX)
 {
--- a/doc/sphinx/reference/vcl_var.rst
+++ b/doc/sphinx/reference/vcl_var.rst
@@ -1024,6 +1024,18 @@ beresp.filters
 	After beresp.filters is set, using any of the beforementioned
 	``beresp.do_*`` switches is a VCL error.
 
+beresp.pass_prefetch
+
+	Type: BYTES
+
+	Readable from: vcl_backend_response
+
+	Writable from: vcl_backend_response
+
+	Set the maximum number of bytes the client can be ahead of the
+	backend during a streaming pass.
+
+
 obj
 ~~~
 
--- a/include/tbl/params.h
+++ b/include/tbl/params.h
@@ -465,6 +465,19 @@ PARAM_SIMPLE(
 )
 
 PARAM_SIMPLE(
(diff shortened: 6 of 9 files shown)


bin/varnishd/cache/cache.h
```diff
@@ -287,6 +287,8 @@ struct boc {
 	enum boc_state_e	state;
 	uint8_t			*vary;
 	uint64_t		len_so_far;
+	uint64_t		client_data_so_far;
```
Dridi, line 290: I would rename `len_so_far` to `fetched_so_far` and `client_data_so_far` to `delivered_so_far` for clarity.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -532,6 +532,13 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 		bo->uncacheable = 1;
 		wrk->handling = VCL_RET_DELIVER;
 	}
+	if (bo->do_pass && !bo->do_stream) {
+		/* Pass readahead should be disabled when not streaming. */
+		if (bo->fetch_objcore->boc->pass_prefetch > 0) {
+			bo->fetch_objcore->boc->pass_prefetch = 0;
+			VSLb(bo->vsl, SLT_Error, "Cannot enable pass readahead when streaming is disabled");
```
Dridi, line 539: I would rather make this case a silent no-op.

Dridi (changes requested):
Besides unforgivable whitespace violations, I have a couple minor suggestions.

bin/varnishd/cache/cache_obj.c
```diff
@@ -220,15 +222,31 @@ void
...
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(oc->boc, BOC_MAGIC);
 	(void)final;
 	assert(l >= 0);
 
 	Lck_Lock(&oc->boc->mtx);
-	AN(om->objextend);
 	if (l > 0) {
+		/* Pass readahead applies when only sending to one client */
```
nigoroll, line 234: how could more than one client read an `OC_F_PRIVATE` object?

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -297,6 +297,24 @@ VRT_r_beresp_uncacheable(VRT_CTX)
...
+
+    return cache_param->pass_prefetch;
+}
+
+VCL_VOID
+VRT_l_beresp_pass_prefetch(VRT_CTX, VCL_BYTES value)
+{
+    CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+
+    cache_param->pass_prefetch = value;
```
nigoroll, line 313: no. we do not want to update a global parameter from vcl.

bin/varnishtest/tests/proxyclose.vtc
```diff
@@ -0,0 +1,31 @@
+varnishtest "Pass buffering with early close"
```
nigoroll, line 1: please keep the vtc nomenclature.

bin/varnishtest/vtc_main.c
```diff
@@ -510,6 +510,21 @@ i_mode(void)
 	vsb = VSB_new_auto();
 	AN(vsb);
 
+	/* CMake build system puts all binaries in the root of the
```
nigoroll, line 513: this change is unrelated to the PR

doc/sphinx/reference/vcl_var.rst
```diff
@@ -1024,6 +1024,18 @@ beresp.filters
...
 
+beresp.pass_prefetch
+
+	Type: BYTES
+
+	Readable from: vcl_backend_response
+
+	Writable from: vcl_backend_response
+
+	Contains the numbers of bytes the client is ahead of the backend
```
nigoroll, line 1035: the description does not match the implementation

include/tbl/params.h
```diff
@@ -464,6 +464,19 @@ PARAM_SIMPLE(
...
+	/* name */	pass_prefetch,
+	/* type */	bytes,
+	/* min */	"4k",
+	/* max */	NULL,
+	/* def */	"1M",
+	/* units */	"bytes",
+	/* descr */
+	"The default prefetch amount used during a single private transaction.\n"
+	"Enabling this will prevent running out of memory when "
+	"there are big transfers going on in the background."
```
nigoroll, line 477: this could be confused with bgfetches, I suggest making clear that this applies to streaming passes.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -580,6 +586,12 @@ vbf_stp_fetchbody(struct worker *wrk, struct busyobj *bo)
 		}
 		AZ(vfc->failed);
 		l = est;
+		/* Limit fetches by pass readahead size */
+		oc = bo->fetch_objcore;
+		if (oc->boc->pass_prefetch > 0 &&
+			l > 2 * oc->boc->pass_prefetch) {
```
nigoroll, line 592: why `2 *` ?

nigoroll (changes requested):

bin/varnishtest/tests/proxyclose.vtc
nigoroll: FTR, this applies to all the VTCs
That said, I think we could have more descriptive vtc names, but if so, only after a deliberate decision has been made.

nigoroll:
besides my technical review comments, I am very much in favor :+1:

bin/varnishtest/tests/proxyclose.vtc
Dridi: Probably tests in the `v*.vtc` category.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -532,6 +532,12 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 		bo->uncacheable = 1;
 		wrk->handling = VCL_RET_DELIVER;
 	}
+	if (bo->do_pass && !bo->do_stream) {
+		/* Pass readahead should be disabled when not streaming. */
+		if (bo->fetch_objcore->boc->pass_prefetch > 0) {
+			bo->fetch_objcore->boc->pass_prefetch = 0;
```
nigoroll, line 538: nit-picking: the `if()` serves no purpose

bin/varnishd/cache/cache_obj.c
```diff
@@ -253,13 +271,18 @@ ObjWaitExtend(const struct worker *wrk, const struct objcore *oc, uint64_t l)
 	CHECK_OBJ_NOTNULL(oc->boc, BOC_MAGIC);
 	Lck_Lock(&oc->boc->mtx);
 	while (1) {
-		rv = oc->boc->len_so_far;
+		rv = oc->boc->fetched_so_far;
 		assert(l <= rv || oc->boc->state == BOS_FAILED);
+		if (oc->flags & OC_F_PRIVATE) {
+			/* Update with the new client position */
+			oc->boc->delivered_so_far = l;
+			AZ(pthread_cond_broadcast(&oc->boc->cond));
```
nigoroll, line 279: can there be more than one waiter? Otherwise, should this be `pthread_cond_signal` for clarity?

bin/varnishd/cache/cache_obj.c
```diff
@@ -220,17 +222,33 @@ void
...
 	if (l > 0) {
+		/* Pass readahead applies when only sending to one client */
+		if (oc->flags & OC_F_PRIVATE) {
+			/* delivered_so_far will be re-read due to memory clobber
+			   from the CondWait opaque function call */
+			while (!(oc->flags & OC_F_ABANDON) &&
+				oc->boc->pass_prefetch > 0 &&
+				oc->boc->fetched_so_far > oc->boc->delivered_so_far + oc->boc->pass_prefetch) {
+				/* Wait 1ms for client to catch up */
+				(void)Lck_CondWait(&oc->boc->cond, &oc->boc->mtx, VTIM_real() + 1e-3);
```
nigoroll, line 242: Can you please comment on the reasons for the timeout?
In my mind, a perfect solution would ensure that we signal the CV under all possible circumstances and not use any timeout.
We might still decide on the compromise of a timeout to avoid mishaps, but I fear relevant overhead for 1000 CondWaits a seconds.

nigoroll (changes requested):
I took the time for a second review and have more questions.
That said, I still agree with the overall concept.

bin/varnishd/cache/cache_fetch.c
Dridi: Bugwash:

> 15:04 \< dridi> the only reason i could see would be for a '/2' instead
15:04 \< dridi> up to two segments of half pass_prefetch
15:04 \< slink> I don't see a case for / 2 either
15:04 \< dridi> meaning that client can be delivered a segment while fetch is happening on the next segment
15:05 \< slink> I see the case for / 2 too
15:05 \< dridi> without more than pass_prefetch in flight

nigoroll:
bugwash:
* CV timeout was discussed. It could act as a fallback, but was questioned if necessary. Some argued that doing without one could become very hairy
* 1ms timeout was generally considered too short
* needs more work

bin/varnishd/cache/cache_fetch.c
nigoroll: /me forgot a ;) at the second statement. Dridi convinced me

bin/varnishd/cache/cache_obj.c
fwsGonzo: I made the timeout 30ms now

bin/varnishd/cache/cache_obj.c
fwsGonzo: I guess that's fair! Fixed.

bin/varnishd/cache/cache_fetch.c
fwsGonzo: I need to see some kind of agreement between everyone here, especially martin who worked this math out with me last time.

bin/varnishd/cache/cache_fetch.c
fwsGonzo: Fixed.

bin/varnishd/cache/cache_obj.c
fwsGonzo: The comment seems correct to me

bin/varnishd/cache/cache_fetch.c
nigoroll: @mbgrydeland Dridi and me think this should be `/ 2`?

bin/varnishd/cache/cache_obj.c
nigoroll: I thought it was causing more confusion than do any good, but nevermind

bin/varnishd/cache/cache_obj.c
```diff
@@ -220,17 +222,33 @@ void
...
 	if (l > 0) {
+		/* Pass prefetch applies when only sending to one client */
+		if (oc->flags & OC_F_PRIVATE) {
+			/* delivered_so_far will be re-read due to memory clobber
+			   from the CondWait opaque function call */
+			while (!(oc->flags & OC_F_ABANDON) &&
+				oc->boc->pass_prefetch > 0 &&
+				oc->boc->fetched_so_far > oc->boc->delivered_so_far + oc->boc->pass_prefetch) {
+				/* Wait 30ms for client to catch up */
+				(void)Lck_CondWait(&oc->boc->cond, &oc->boc->mtx, VTIM_real() + .030);
```
nigoroll, line 242: I can live with it for the time being, but we might want to revisit later. I still think the timeout should not be necessary

nigoroll (approved):

bin/varnishd/cache/cache_obj.c
Dridi: Maybe something on the scale of `idle_send_timeout` to still have a safety net but much less lock operations?

bin/varnishd/cache/cache_fetch.c
daghf: I think this should just be `l`.

In its current form the fetch thread will wait for the client to consume the full segment before continuing.

If we at a later point improve on this with a double-buffered approach that allows fetch and delivery to happen concurrently, then `l/2` would make more sense.

fwsGonzo: Alright, I think that clears up this point

bin/varnishd/cache/cache_obj.c
fwsGonzo: I've changed this to 2.5 seconds, just to avoid kernel round-trips on slow clients.

bin/varnishd/cache/cache_obj.c
```diff
@@ -220,17 +222,33 @@ void
...
-	AN(om->objextend);
 	if (l > 0) {
+		/* Pass prefetch applies when only sending to one client */
+		if (oc->flags & OC_F_PRIVATE) {
+			/* delivered_so_far will be re-read due to memory clobber
+			   from the CondWait opaque function call */
+			while (!(oc->flags & OC_F_ABANDON) &&
+				oc->boc->pass_prefetch > 0 &&
+				oc->boc->fetched_so_far > oc->boc->delivered_so_far + oc->boc->pass_prefetch) {
+				/* Wait 2.500s for client to catch up */
```
mbgrydeland, line 241: I think this comment isn't clear. Also the timeout is on the high side I think, maybe 0.1s would be better? (This controls for how long a stuck fetch thread could be sitting waiting for a gone client).

Suggested new comment wording:

The client is too far behind, wait for it to catch up and signal the condvar. Lck_CondWait() with timeout is used to catch any failure to signal the condvar on the client's part (e.g. a restart or late hangup detection), avoiding deadlocks.

bin/varnishd/cache/cache_obj.c
```diff
@@ -220,17 +222,33 @@ void
...
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(oc->boc, BOC_MAGIC);
 	(void)final;
 	assert(l >= 0);
 
 	Lck_Lock(&oc->boc->mtx);
-	AN(om->objextend);
 	if (l > 0) {
+		/* Pass prefetch applies when only sending to one client */
+		if (oc->flags & OC_F_PRIVATE) {
```
mbgrydeland, line 235: Having looked at #3240, I think this should be extended to cover (oc->flags & (OC_F_PRIVATE | OC_F_HFM | OC_F_HFP)) like is done there.

bin/varnishd/cache/cache_obj.c
```diff
@@ -253,13 +271,18 @@ ObjWaitExtend(const struct worker *wrk, const struct objcore *oc, uint64_t l)
 	CHECK_OBJ_NOTNULL(oc->boc, BOC_MAGIC);
 	Lck_Lock(&oc->boc->mtx);
 	while (1) {
-		rv = oc->boc->len_so_far;
+		rv = oc->boc->fetched_so_far;
 		assert(l <= rv || oc->boc->state == BOS_FAILED);
+		if (oc->flags & OC_F_PRIVATE) {
```
mbgrydeland, line 276: Same as HFM/HFP comment above.

mbgrydeland (commented):
I think also this should be added to HSH_Cancel, to wake a sleeping fetch thread earlier without having to rely on the Lck_CondWait timeout for the cases where HSH_Cancel is called.

```
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -794,6 +794,13 @@ HSH_Cancel(struct worker *wrk, struct objcore *oc, struct boc *boc)
 
        if (boc != NULL) {
                hsh_abandon(oc);
+               if (oc->flags & (OC_F_PRIVATE | OC_F_HFM | OC_F_HFP)) {
+                       /* Signal the boc condvar that we have been
+                        * abandoned, waking up the fetch thread early in
+                        * case it was sleeping waiting for the client to
+                        * make progress. */
+                       AZ(pthread_cond_broadcast(&boc->cond));
+               }
                ObjWaitState(oc, BOS_FINISHED);
        }

```

mbgrydeland:
From comment by @ehocdet on another issue:

> #3480 have 2 issues : CondWait on slow clients is not good, lack of multi-buffering will slow down stream.
> It's why I used additional lock and condvar on my poc.

I don't think these are valid concerns. The way this patch works, the fetch and delivery dance will not change for the case where the client is faster than the backend. It is only when the client is slower than the backend that the new wait point is introduced, forcing the fetch thread to pause for a signal before continuing. The timeout used on that CondWait is there only as a backstop against clients that spuriously disappeared without proper signalling. It was implemented this way as getting the signalling just right, and never be able to go wrong seems very complex, and not made any easier by the condvar being shared for both types of waiting. There is multi-buffering also the way this patch set works, just like there was previously, defined by the fetch limit number of bytes. So the fetch and delivery will be allowed to run simultaneously as long as they are not more than fetch limit bytes apart.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -532,6 +532,10 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 		bo->uncacheable = 1;
 		wrk->handling = VCL_RET_DELIVER;
 	}
+	if (bo->do_pass && !bo->do_stream) {
+		/* Pass prefetch should be disabled when not streaming. */
+		bo->fetch_objcore->boc->pass_prefetch = 0;
+	}
```
Dridi, line 538: I think the condition should actually be `if (!bo->uncacheable || !bo->do_stream)` since `bo->do_pass` is off-topic on account of belonging to `bereq` and we need both to let `pass_prefetch` kick in.

bin/varnishd/cache/cache_obj.c
Dridi: The condition should rather be:

```c
if (l > 0) {
        if (oc->boc->pass_prefetch > 0) {
                assert(oc->flags & ...);
                ...
        }
        ...
}
```

Since we wipe `pass_prefetch` when it is irrelevant, it should be the condition to enter the cond loop. That would also simplify the `while` loop condition that is currently redundant. We also don't need a comment at this point, we know why we enter the nested branch.

Not commenting about the set of flags that might enable pass prefetch, I didn't look at that part.

bin/varnishd/cache/cache_obj.c
Dridi: Same comment as above, test `oc->boc->pass_prefetch > 0` to signal changes in `oc->boc->delivered_so_far`.

Dridi:
@fwsGonzo followed my and @mbgrydeland's suggestions to:

- rely on `beresp` fields to disable pass prefetch (instead of `bereq`)
- a `pass_prefetch > 0` as the condition to enable this during streaming
- add assertions on flags that used to be the condition to enable

It's not obvious because the commit was amended instead of adding a followup review patch, but manually diffing the last force push shows the above changes.

This exploded in CI and after a quick look I found that synthetic objects don't have any of the expect flags, and being synthetic anyway there is no actual backend fetch to  pace so the following diff fixes the current failures:

```diff
--- i/bin/varnishd/cache/cache_fetch.c
+++ w/bin/varnishd/cache/cache_fetch.c
@@ -534,7 +534,7 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
        }
        if (!bo->uncacheable || !bo->do_stream) {
                /* Pass prefetch should be disabled when not streaming. */
-               bo->fetch_objcore->boc->pass_prefetch = 0;
+               oc->boc->pass_prefetch = 0;
        }
        if (bo->do_pass || bo->uncacheable)
                oc->flags |= OC_F_HFM;
@@ -956,6 +956,7 @@ vbf_stp_error(struct worker *wrk, struct busyobj *bo)
                return (F_STP_FAIL);
        }
 
+       oc->boc->pass_prefetch = 0;
        ll = VSB_len(synth_body);
        o = 0;
        while (ll > 0) {
```

Also, once this change is in, please rebase against master.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -580,6 +584,12 @@ vbf_stp_fetchbody(struct worker *wrk, struct busyobj *bo)
 		}
 		AZ(vfc->failed);
 		l = est;
+		/* Limit fetches by pass readahead size */
+		oc = bo->fetch_objcore;
+		if (oc->boc->pass_prefetch > 0 &&
+			l > oc->boc->pass_prefetch) {
+				l = oc->boc->pass_prefetch;
+		}
```
Dridi, line 592: whitespace offense :police_car:

Dridi:
Pow wow consensus: rename `pass_prefetch` to `transit_buffer`.

nigoroll:
FTR: Continued in #3572

Issue #3218, opened 2020-02-11
optimize string functions for STRANDS


nigoroll:
Ref 0c3ec169282427e629f211dc17c4afd1df2bb9da

Here we went from
```
*b++ = (char)toupper(*p++);
```
to
```
VSB_putc(vsb, (char)toupper(*p));
```

now that we use VSBs more (which I appreciate), should be inline `VSB_putc()` and `VSB_put_byte()` to reduce overhead?

bsdphk:
Only if we can measure a difference in some microbenchmark ?

nigoroll:
So, I've run some benchmarks...

## TL;DR

* the WS_VSB does have an impact
* with inlining we can almost compensate for it

## summary

* the code before the switch to the new `WS_VSB` (see 0c3ec169282427e629f211dc17c4afd1df2bb9da) was fastest in all cases
* for gcc with `-O2` (which I assume a typical production build will use)
  * VSB_putc/WS_VSB is ~3x slower than the original code
  * the same with inlining is only ~10% slower than the original code (= ~2.8x faster than no inlining)
* for a developer build (without compiler optimizations and inlining)
  * VSB_putc/WS_VSB is ~4.5x slower than the original code
  * and inlining does not help because it is disabled
* for `-O6` which I tried out of curiosity
  * VSB_putc/WS_VSB also is ~3x slower than the original code (with both being faster than `-O2`)
  * inlining only gives a ~40% improvement and results in slower code than with `-O2`

## methodology

* to benchmark std.tolower(), I have used `vtc.benchmark()` from #3163 / https://github.com/varnishcache/varnish-cache/pull/3163/commits/902235a7be4244cbf0dac7524fa5a51b6eab7848 which uses the proposed `VRT_call`
* the vtc https://github.com/nigoroll/varnish-cache/commit/20da8cb6eb60707043c121423fa42660da2e080e takes a header of 3780 bytes length and repeatedly calls `std.tolower` on it. (the length is completely arbitrary except that I did not want to use too short a string for relevance)
* `if (std.tolower(..) == "")` is called 10 times in a vcl sub. The comparison is used to avoid the result to be discarded by any compiler optimization.
* This sub with the 10 `std.tolower` is timed using `vtc.benchmark()` with 10000 iterations which has been found to produce quite stable results
* three compiler settings and three different variants of the code have been tried
* the vtc has been run multiple times manually to check if results compare broadly (which they do). Typical results have been picked for the result tables.

[...]

PR #3482, opened 2020-12-08
vsb: Speed VSB_cat() up using VSB_bcat()

Dridi:
We don't need to assert the integrity of the VSB after every single
byte, it can become very expensive depending on the workload.


nigoroll (approved):
generally in favor, see also https://github.com/varnishcache/varnish-cache/issues/3218#issuecomment-587144176 for a rough idea of the performance impact

bsdphk:
Ok with me.

PR #3483, opened 2020-12-10
Various VSC consistency improvements

Dridi:
See the commit log for details:

* push stats in more relevant cases (by @mbgrydeland)
* adjust negative gauges to zero
  * add a 'r' control everywhere relevant
* push stats from the HTTP/1 session loop

### DIFF
--- a/bin/varnishd/cache/cache_wrk.c
+++ b/bin/varnishd/cache/cache_wrk.c
@@ -398,10 +398,32 @@ Pool_Work_Thread(struct pool *pp, struct worker *wrk)
 				i = Lck_CondWait(&wrk->cond, &pp->mtx, tmo);
 				if (i == ETIMEDOUT)
 					VCL_Rel(&wrk->vcl);
-			} while (wrk->task->func == NULL);
-			tpx = *wrk->task;
-			tp = &tpx;
-			wrk->stats->summs++;
+				if (wrk->task->func != NULL) {
+					/* We have been handed a new task */
+					tpx = *wrk->task;
+					tp = &tpx;
+					wrk->stats->summs++;
+				} else if (pp->b_stat != NULL &&
+				    pp->a_stat->summs) {
+					/* Woken up to release the VCL,
+					 * and noticing that there are
+					 * pool stats not pushed to the
+					 * global stats and no active
+					 * thread currently doing
+					 * it. Remove ourself from the
+					 * idle queue and take on the
+					 * task. */
+					assert(pp->nidle > 0);
+					VTAILQ_REMOVE(&pp->idle_queue,
+					    wrk->task, list);
+					pp->nidle--;
+					tps.func = pool_stat_summ;
+					tps.priv = pp->a_stat;
+					pp->a_stat = pp->b_stat;
+					pp->b_stat = NULL;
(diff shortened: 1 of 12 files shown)


bin/varnishd/cache/cache_wrk.c
```diff
@@ -398,10 +398,32 @@ Pool_Work_Thread(struct pool *pp, struct worker *wrk)
...
-			} while (wrk->task->func == NULL);
-			tpx = *wrk->task;
-			tp = &tpx;
-			wrk->stats->summs++;
+				if (wrk->task->func != NULL) {
+					/* We have been handed a new task */
+					tpx = *wrk->task;
+					tp = &tpx;
+					wrk->stats->summs++;
+				} else if (pp->b_stat != NULL &&
```
nigoroll, line 442: I see the point and I am not against it, but it would be nice if we could maybe refactor a bit more to collapse the two `pool_stat_summ` cases. Wouldn't something similar to
```
if (i == ETIMEDOUT) {
    VCL_Rel(&wrk->vcl);
    Lck_Unlock(&pp->mtx);
    continue;
}
```
at L399 work as well?

lib/libvarnishapi/libvarnishapi.map
```diff
@@ -195,3 +195,11 @@ LIBVARNISHAPI_2.5 {	/* 2020-09-15 release */
     local:
 	*;
 };
+
+LIBVARNISHAPI_2.6 {	/* 2020-03-15 release */
```
nigoroll, line 199: add a reminder to bump the API version for release?

bin/varnishd/cache/cache_wrk.c
```diff
@@ -160,28 +160,39 @@ WRK_Thread(struct pool *qp, size_t stacksize, unsigned thread_workspace)
...
-	dst->summs++;
-	VSC_main_Summ_wrk_wrk(dst, src);
-	memset(src, 0, sizeof *src);
+	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
+	pp = wrk->pool;
+	CHECK_OBJ_NOTNULL(pp, POOL_MAGIC);
+	if (locked)
+		Lck_AssertHeld(&pp->mtx);
+
+	if ((tp == NULL && wrk->stats->summs > 0) ||
```
nigoroll, line 174: I think we should move this test `CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);` and return unless true to avoid useless assertion tests on `pp`

Dridi: I didn't want to do anything before making all assertions, the current order was chosen on purpose.

lib/libvarnishapi/libvarnishapi.map
Dridi: Let's have that chat offline.

bin/varnishd/cache/cache_wrk.c
Dridi: From bugwash:

> \< martin> that whole loop is getting a bit unwieldy

I agree but haven't tried anything yet to improve that aspect of the change,

PR #3484, opened 2020-12-14
PoC: Built-in VCL split

Dridi:
This is a minimal not too intrusive implementation of a built-in VCL split. It's incomplete in the sense that only one branch was split out of `vcl_recv` and the relevant error message does not reflect this change, plus other details.

This should be enough to get a general idea.

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -1,35 +1,34 @@
-/*-
- * Copyright (c) 2006 Verdens Gang AS
- * Copyright (c) 2006-2015 Varnish Software AS
- * All rights reserved.
- *
- * Author: Poul-Henning Kamp <phk@phk.freebsd.dk>
- *
- * SPDX-License-Identifier: BSD-2-Clause
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in the
- *    documentation and/or other materials provided with the distribution.
- *
- * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
- * ARE DISCLAIMED.  IN NO EVENT SHALL AUTHOR OR CONTRIBUTORS BE LIABLE
- * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
- * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
- * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
- * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
- * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
- * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
- * SUCH DAMAGE.
- *
- * This is the builtin VCL code
- */
+#-
+# Copyright (c) 2006 Verdens Gang AS
+# Copyright (c) 2006-2015 Varnish Software AS
+# All rights reserved.
+#
+# Author: Poul-Henning Kamp <phk@phk.freebsd.dk>
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL AUTHOR OR CONTRIBUTORS BE LIABLE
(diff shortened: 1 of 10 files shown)


nigoroll:
I do not feel ready for a comprehensive review yet, but just one simple question:
Why did you chose to parse the builtin vcl with its own `struct vcc`? Would this not lead to more complications looking forward?
I would have thought we could just take note of when we enter the builtin vcl and not error out for the reserved sub symbols from that point on?

Dridi:
That's because we need to know ahead of time what non-method subroutines the built-in contains, and the less intrusive approach is to parse the built-in first. Otherwise you need to synchronize a table somewhere of the allowed subroutines. With this approach, the `builtin.vcl` file becomes authoritative in that regard.

nigoroll:
but use before define already works for subs?

Dridi:
I tried that first, and it was very complicated to re reconcile because you need to deny any `vcl_*` subroutine that is not in the built-in. Having a builtin-only symbol table simplifies the process.

Dridi:
I expanded the split of the built-in VCL, see 0d2441991a for more details. I realized that my patch allows the built-in to declare subs out of order, which might make the split even easier to follow. At least for me.

bin/varnishd/builtin.vcl
```diff
@@ -1,78 +1,108 @@
...
-        /* We only deal with GET and HEAD by default */
-        return (pass);
-    }
-    if (req.http.Authorization || req.http.Cookie) {
-        /* Not cacheable by default */
-        return (pass);
-    }
-    return (hash);
+sub vcl_req_host {
+	if (req.http.host) {
```
nigoroll, line 57: `if (req.http.host ~ "[A-Z]") ` would save a copy for most cases

bin/varnishd/builtin.vcl
```diff
@@ -1,78 +1,108 @@
...
+	if (req.method != "GET" &&
+		req.method != "HEAD" &&
+		req.method != "PUT" &&
+		req.method != "POST" &&
+		req.method != "TRACE" &&
+		req.method != "OPTIONS" &&
+		req.method != "DELETE" &&
+		req.method != "PATCH") {
+		# Non-RFC2616 or CONNECT which is weird.
+		return (pipe);
```
nigoroll, line 82: do we want to change to `return (synth(405))` now or later?

nigoroll (approved):
the `builtin.vcl` lgtm.
I would like to still try an idea on the vcc changes, but that might not be important.

Dridi:
General ideas discussed during bugwash:

- add "doc" comments to the built-in
- "doc" comments are lines starting with `##`
  - easily recognizable `\n##` pattern
  - contains RST code?
- generate sphinx/man documentation from the built-in inline docs
- strip doc comments from `vcl.show -v`
  - including from user VCL
- show doc comments with `vcl.show -v -d`

Dridi:
Bugwash summary:

- keep the inline documentation for later (see https://github.com/varnishcache/varnish-cache/pull/3503#issuecomment-767004116)
- we are generally in favor of the split as it is so far in this PoC
- split `vcl_recv` and `vcl_backend_response` (@dridi)
- get live traffic through (@nigoroll and hopefully others)

Dridi:
As requested in yesterday's intwash I have pushed a split of the two main candidates and added some insights in the commit log so please make sure to read that and not simply look at the diff.

One thing that is not obvious is the reason why I called the subroutines the way I called them. For example `vcl_req_method` that is called by `vcl_recv`. Why not `vcl_recv_method` instead you ask? Because once we introduce `vcl_begin` that is in charge of checking the method we can keep the same name if it's tied to `req` instead of `recv`.

The patch series in its current state (94af15e8ed...34eb12a85a) does not update the documentation.

bsdphk:
Ok with me.

nigoroll (commented):
This is now the third time I look at this PR and I am still very much in favor. Yet, it is only now that I noticed a variation which could improve the usefulness of the change:

Consider `vcl_beresp_stale {}` as an example. If defined and used it like this:

```
sub vcl_beresp_stale {
    call vcl_beresp_hitmiss;
}

sub vcl_backend_response {
    # ...
    if (beresp.ttl <= 0s) {
        call vcl_beresp_stale;
    }
    # ...
```

users could selectively _prepend_ their custom handling to `vcl_beresp_stale` whenever they want to handle stale objects.

The same goes for several other of the proposed split-outs like `vcl_req_authorization`, `vcl_req_cookie` and all `vcl_beresp_*`.

Dridi:
That's an interesting twist, and while that removes the consistency of having all "assistant vcl_ subs" (coined by phk) containing the condition and default effect, I can see the advantage of pulling the condition out again.

There's one sub where I would feel strongly against doing it, and that's `vcl_beresp_control`.

So for the simple conditions with only one criterion, I'm convinced.

nigoroll:
I can't quite follow the argument on `vcl_beresp_control`, can you elaborate a bit?

Dridi:
After sleeping on it, I now disagree with the idea altogether.

The reason why I feel uncomfortable with the idea of pulling the "cache control" conditions away from the VCL author is that those conditions are very specific. In fact, I think the user should be able to override any condition and not just a subset (pun intended) of them.

The reason why I find the built-in VCL so inconvenient is the fact that it's appended and for the two subroutines that were split in this pull request trying to have an effect on a specific branch requires either complicated hoops or copy-pasta. If we split only the effects and not the conditions we are painting ourselves into the same corner again.

Here is a common example: in a multi-tier system where the backend might be another Varnish server I might be served stale contents that are perfectly fine to cache for future revalidation. The built-in currently prevents 304 on a stale insertion.

If we split the conditions too, we can do this:

```vcl
sub vcl_beresp_stale {
	if (beresp.http.is-varnish-tier == "true" && beresp.ttl + beresp.grace > 0s) {
		return; # cache stale responses from other tiers
	}
}
```

That would work with your idea too, but not so well with complex conditions, and when it works, not in a way that I would find easy to follow.

If `vcl_beresp_stale` has both the condition and the effect it becomes self-contained, and you can see just by looking at it how we treat stale backend responses, client request methods or whatever else just by looking at the sub in isolation. And the split subs look straightforward:

```vcl
sub vcl_split {
	call something;
	call stuff;
	return (default);
}
```

I changed my mind after thinking about it more.

nigoroll:
@Dridi, sure your case exists, but in that case one would just prepend before `vcl_backend_response`?

If the conditions are in the split sub, they become nothing more than literal code snippets of the original sub. That might provide some benefit, too ("even if it makes no difference, by convention do do all _foo_ stuff in `vcl_foo`"), but this seems less useful to me, given that you can still prepend before the built-in subs.

Dridi:
No, otherwise you're back to the current problem. The built-in becomes all-or-nothing and you can't gracefully neuter one branch leaving the others alone.

Note in my example that I used `return;` without a transition, so this is exiting `vcl_beresp_stale` but still going through the other conditions. Maybe an actual non-Varnish backend generated a stale response that contained a `set-cookie` header because that one client getting a cache miss didn't present a cookie and we don't have private headers yet?

nigoroll:
Yes, I overlooked the significance of `return`. The option to satisfy both cases would be to add another sub level, but I am not sure if it's worth it.
I am still ok with keeping the patch as-is.

Dridi:
The last thing to do before it's ready to merge is to revert the b0.vtc hijack and add a test case dedicated to this change.

Docs reviews welcome, I wasn't too inspired.

Dridi:
Bugwash summary: split a subroutine for piped HTTP methods and find an expressive name.

Dridi:
I tried to rebase this change against current master but it is completely broken by #3516. It might be a very simple case of adding an if statement in the right place like this pull request originally does to enable a behavior only while or after processing the built-in VCL in isolation.

But, I neither have the bandwidth nor the confidence to do it correctly right off the bat as we enter the freeze so, to be revisited later.

nigoroll:
continued in #3541

PR #3487, opened 2021-01-05
Add VRT_priv_task_lookup()

nigoroll:
`VRT_priv_task()` inserts a priv_task if it does not exist.

In some cases, it is more efficient to only insert a priv_task under certain conditions. These require an API function to only return a priv_task if it exists. This is `VRT_priv_task_lookup()`.

A TODO comment has been added for a sensible performance optimization, avoiding one VRBT traversal for the `VRT_priv_task()` case. I would address this after this suggestion got accepted, if so.


nigoroll:
bugwash:
* add vrt_priv_dynamic_get() instead of modifying  vrt_priv_dynamic()
* name the new VRT function VRT_priv_task_get()
otherwise ok for merge

PR #3489, opened 2021-01-08
VCL_REGEX

Dridi:
I wrote this patch myself, ended up not reviewing a previous submission.

There are open questions in VRT where part of the API uses `VCL_REGEX` and others use `void *`.

### DIFF
--- a/bin/varnishtest/tests/c00103.vtc
+++ b/bin/varnishtest/tests/c00103.vtc
@@ -1,0 +1,32 @@
+varnishtest "REGEX expressions in VCL"
+
+varnish v1 -vcl {
+	import debug;
+	backend be none;
+	sub vcl_recv {
+		# NB: the REGEX expression below is needlessly complicated
+		# on purpose to ensure we don't treat REGEX expressions
+		# differently.
+		if (req.url ~ (debug.just_return_regex(
+		    debug.just_return_regex("hello")))) {
+			return (synth(200));
+		}
+		return (synth(500));
+	}
+	sub vcl_synth {
+		set resp.reason = regsub(resp.reason,
+		    debug.just_return_regex("OK"), "world");
+	}
+} -start
+
+client c1 {
+	txreq
+	rxresp
+	expect resp.status == 500
+	expect resp.reason == "Internal Server Error"
+
+	txreq -url "/hello"
+	rxresp
+	expect resp.status == 200
+	expect resp.reason == world
+} -run
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -273,6 +273,7 @@ typedef int64_t					VCL_INT;
 typedef const struct suckaddr *			VCL_IP;
 typedef const struct vrt_backend_probe *	VCL_PROBE;
 typedef double					VCL_REAL;
+typedef const struct vre *			VCL_REGEX;
 typedef const struct stevedore *		VCL_STEVEDORE;
 typedef const struct strands *			VCL_STRANDS;
 typedef const char *				VCL_STRING;
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -61,6 +61,10 @@
  *	VRT_ban_string() signature changed
  *	VRT_priv_task_get() added
  *	VRT_priv_top_get() added
+ *	VRT_re_init removed
+ *	VRT_re_fini removed
+ *	VRT_re_match signature changed
+ *	VRT_regsub signature changed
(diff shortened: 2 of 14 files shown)


Dridi:
This pull request introduces a REGEX type for VCL.

Today regular expressions can only be literals, making it impossible to use built-in functions or operators with a regular expression coming from a VMOD. Conversely a VMOD can also take a REGEX from a STRING literal without having to deal with compilation since that would be handled automatically, putting VMOD functions on par with `regsub[all]`.

The first patch introduces a REGEX type and allows regular expressions to be expressions in addition to literals.

The second patch teaches VMODs to interact with the REGEX type and adds some coverage via vmod_debug.

The pull request is still a draft because today we use `void *` to pass VREs around and we might want to unvoid the VRT API.

Dridi:
Bugwash consensus: go all the way and include documentation (I will undraft the pull request then).

Also requested:

- rename `debug.regex()` to `debug.just_return_regex()`
- comment about parentheses in the test case

Dridi:
Rebased against master, with review items taken care off. I went the extra mile and adopted the REGEX type in vmod_cookie, bringing an overall negative diff stat to the whole pull request.

Ready for review.

Dridi:
Force-pushed, I had the wrong priv in d990d0ac5fc6ec492abdf878c8b1aa802f6d26da's commit log. No change in the diff itself.

Dridi:
Ready for review, with an update to the docs this time.

include/vrt.h
```diff
@@ -456,14 +461,6 @@ VCL_BACKEND VRT_DirectorResolve(VRT_CTX, VCL_BACKEND);
 void VRT_acl_log(VRT_CTX, const char *);
 int VRT_acl_match(VRT_CTX, VCL_ACL, VCL_IP);
 
-/***********************************************************************
```
nigoroll, line 459: :-1:  on removing `VRT_re_*` from VRT.
This will break vmods building VREs dynamically.
According to my vmod library, this issue is real.

bin/varnishtest/tests/c00103.vtc
```diff
@@ -0,0 +1,32 @@
...
+
+varnish v1 -vcl {
+	import debug;
+	backend be none;
+	sub vcl_recv {
+		# NB: the REGEX expression below is needlessly complicated
+		# on purpose to ensure we don't treat REGEX expressions
+		# differently.
+		if (req.url ~ (debug.just_return_regex(
+		    debug.just_return_regex("hello")))) {
```
nigoroll, line 11: what happens when a dynamic string is used as a REGEX argument?

nigoroll (changes requested):
see comments.
Overall I am still very much in favor, but there is one show stopper.

bin/varnishtest/tests/c00103.vtc
Dridi: With this, only CSTR tokens can be used, so what happens is a compile-time error.

include/vrt.h
Dridi: In that case you are asking for a panic if the dynamically-built regex has a syntax error.

FWIW you have `vre.h` to work with such regular expressions. That's what vmod_cookie was using and it was not affected by the privatization of that part of the API. I would rather reconsider how those VMODs are operating today.

Unless we don't want to offer VCL scope and compile-time checks to the REGEX type.

nigoroll: the panic can be avoiding by another `VRE_compile()` before the `VRT_re_init()`, but I had a closer look and now agree that the use cases in vmods are [either wrong](https://github.com/varnish/varnish-modules/blob/f26e057727c6e4ff3c042dae19cfd64e75a95236/src/vmod_header.c#L254) or [motivated by convenience](https://github.com/Dridi/libvmod-querystring/blob/06f8206c91842d44f881c1a343c264312d3a36a7/src/vmod_querystring.c#L598) (examples chosen randomly ;) )

So I agree with you now.

bin/varnishtest/tests/c00103.vtc
nigoroll: I was trying to imply that we should have a test ;)

nigoroll: thank you, our comments have crossed.

bsdphk:
OK with me.

Dridi:
Rebased against master to solve a conflict in `vrt.h`. Waiting for CI to complete before merging.

Issue #3498, opened 2021-01-13
PRIV_TOP vmod arguments are broken


nigoroll:
When a `PRIV_TOP` argument is used for a vmod function like `$Function STRING test_priv_top(PRIV_TOP, STRING)`, we currently do not prevent it to be called from a wrong context, yet we still panic if the vcl contains such a call.

Introduced in 1851ae9acdbed813eaaf4729f92bcd3de53819da, shown by the test change below.

Note that also this does not conform to the documentation: https://varnish-cache.org/docs/6.0/reference/vmod.html#private-pointers

> When used from backend VCL subs, a NULL pointer will be passed.

```
diff --git a/bin/varnishtest/tests/v00043.vtc b/bin/varnishtest/tests/v00043.vtc
index 828b06668..6f985b037 100644
--- a/bin/varnishtest/tests/v00043.vtc
+++ b/bin/varnishtest/tests/v00043.vtc
@@ -81,7 +81,7 @@ varnish v1 -cliok "param.set debug +syncvsl" -vcl+backend {
 
        sub vcl_backend_response {
                set beresp.do_esi = true;
-
+               set beresp.http.x1 = debug.test_priv_top("");
        }
 
        sub vcl_deliver {
```

```
***  v1    debug|Error: Child (950) Panic at: Wed, 13 Jan 2021 16:45:07 GMT
***  v1    debug|Wrong turn at cache/cache_vrt_priv.c:243:
***  v1    debug|PRIV_TOP is only accessible in client VCL context
***  v1    debug|version = varnish-trunk revision 06a2e92e74764e05acd7fa77556755a1a24e952c, vrt api = 13.0
***  v1    debug|ident = Linux,4.9.0-14-amd64,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,epoll
***  v1    debug|now = 30026.169678 (mono), 1610556307.524721 (real)
```

nigoroll:
I have pondered how best to solve this:

If we just removed the `WRONG()` and returned `NULL` for `VRT_priv_top()` as before https://github.com/varnishcache/varnish-cache/commit/1851ae9acdbed813eaaf4729f92bcd3de53819da, we would have no useful user feedback at all. Also, the `WRONG()` is appropriate for programmatic use from vmods, which we should require to only call `VRT_priv_top()` from the right context.

[...]

PR #3500, opened 2021-01-18
PoC: vcc: Restrict static PRIV_TOP scope

Dridi:
Following up after #3498.

### DIFF
--- a/bin/varnishtest/tests/v00043.vtc
+++ b/bin/varnishtest/tests/v00043.vtc
@@ -55,6 +55,22 @@ server s1 {
 	txresp
 } -start
 
+varnish v1 -errvcl "Not available in subroutine 'vcl_backend_fetch'" {
+	import debug;
+	backend be none;
+	sub vcl_backend_fetch {
+		debug.test_priv_top("only works on client side");
+	}
+}
+
+varnish v1 -errvcl "Not available in subroutine 'vcl_init'" {
+	import debug;
+	backend be none;
+	sub vcl_init {
+		debug.test_priv_top("only works on client side");
+	}
+}
+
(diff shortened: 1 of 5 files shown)


nigoroll:
nice! this is similar to what I tried and failed with.

Dridi:
Why is this a draft pull request you ask?

I wrote this patch in a hurry and believe that it should rather work with VIP4 in mind. Basically VMOD symbols should start with a complete `r_methods` mask until we implement VIP4. Then upon finding a `PRIV_TOP` argument the symbol can be `&`'d in a idempotent fashion unlike the current conditional. A VMOD that would not `$Restrict` a function making use of `PRIV_TOP` would get the compile error for free, and a libvcc would enforce a proper scope if `$Restrict` was too broad (we might want a warning message).

Making this change will however require that all VMOD symbols are cross-referenced with subroutines where they are used, which I haven't studied yet.

Dridi:
Or maybe the appropriate mask would be `action_mask` rather than `r_methods`?

Dridi:
I think this change is ready now.

nigoroll:
for all purposes I would suggest to finish this PR after #3163 is in

Dridi:
Agreed, conflicts (if any) are probably easier to solve on this smaller patch series.

bsdphk:
Looks OK.

Dridi:
Merging because it doesn't conflict with #3163, after trying a clean rebase on my system.

nigoroll:
that was not the my point.
my point, which I cou^Wshould have made clearer was that this might need additional consideration with VCL_SUB

Dridi:
We can discuss it offline, I'm not worried about that since for me both changes are orthogonal.

PR #3503, opened 2021-01-22
vcl: New vcldoc syntax for comments

Dridi:
This is a spinoff from the built-in VCL split PoC (https://github.com/varnishcache/varnish-cache/pull/3484#issuecomment-749008751) that focuses on introducing a new "vcldoc" syntax inspired by how other languages tool chains allow to embed documentation into code.

Here is a javadoc example taken from [wikipedia](https://en.wikipedia.org/wiki/Javadoc):

```java
/**
 * Short one line description.
 * <p>
 * Longer description. If there were any, it would be
 * here.
 * <p>
 * And even more explanations to follow in consecutive
 * paragraphs separated by HTML paragraph breaks.
 *
 * @param  variable Description text text text.
 * @return Description text text text.
 */
public int methodName (...) {
    // method body with a return statement
}
```

We can see that it:

- uses C-style comments with a double star at the beginning (`"/**"`)
- deals with a `" * "` margin on line breaks
- embeds HTML code
- contains magic `@tags`

For the VCL tool chain I had something simpler in mind:

- the line starts with `"## "`
- following the space, vcldoc is made of raw text to the end of the line
- if the space is missing, that generates a newline character

The goal, in preparation for a VCL built-in split is to make the builtin authoritative:

- it would contain both FSM and non-FSM subroutines
- the documentation could live alongside as we make changes

I anticipated that having a lot of docs would probably bloat the `vcl.show` output, it would at least increase significantly in the case of `vcl.show -v` that includes a copy of the builtin. So libvcc keeps track of vcldoc comments so they can be omitted by default from the output. It can still be displayed with a new `-d` option.

Next in the tool chain is the extraction of vcldoc comments, and this is done by `varnishd -D -f <vcl_file>` similarly to how the `-C` option works for C code. With that extraction in place some of the online sphinx-only docs were moved and slightly adapted to `builtin.vcl` and the embedded RST docs are extracted and included in the `vcl(7)` manual at build time.

To get a feel for how vcldoc comments would work, I suggest to start with the following patches:

- doc: Introduce the concept of vcldoc comments
- doc: Make builtin.vcl document itself

The former documents how vcldoc works and the latter makes `builtin.vcl` the second VCL to make use of them (the first one is in a test case). They are of course the last patches since I needed a working tool chain first.


Dridi:
Thanks to @daghf I was able to see why this fails with the Debian Buster job:

```
# cat doc/sphinx/include/vcl_builtin.rst   
Message from vcldoc:
-----------------------------------------------------
Suppressions used:
  count      bytes template
      3        141 STV_Config
      2      16424 vcc_
-----------------------------------------------------

Built-in VCL
============

The built-in VCL is always appended at the very end of a loaded VCL
as if it had been included. It contains default rules for a sane HTTP
```

This is the problem of `varnishd -C` writing its output to `stderr` and `varnishd -D` inherits this behavior for consistency. The problem here is that the `stderr` output is polluted. I'l try to update both `varnishd -[CD]` to output to `stdout` (a breaking change I can live with, it has always annoyed and puzzled me).

Dridi:
Changing the output to `stdout` didn't help, `VSUB_run()` redirects the fork's `stderr` to its `stdout` and there's no way to distinguish actual errors from sanitizers diagnostics.

bsdphk:
"here's no way to distinguish actual errors from sanitizers diagnostics."

Doesn't the exit(3) code do that ?

Dridi:
Not here because those are only suppression notices.

I think the way forward would be to have the output of `varnishd -[CD]` on `stdout` anyway, because POLA. And we need a `VSUB_something()` that doesn't mix it with `stderr`. Specifically because in this case the output is our deliverable, not a shared library on disk.

mbgrydeland:
Having thought a little more about this after the bugwash discussion, I think it would be a misstake to implement any type of comment filtering on outputs. Source code really should be left as is, if not messages containing line numbers become ambigous.

Dridi:
> Having thought a little more about this after the bugwash discussion, I think it would be a misstake to implement any type of comment filtering on outputs. Source code really should be left as is, if not messages containing line numbers become ambigous.

This wasn't really about removing comments, but filtering out by default (so not always) comments that were explicitly meant for embedding documentation to be extracted on its own. For VCL line numbers, is there any case where we report them besides `VCL_trace` logs \*after\* the compilation succeeded? If you are playing with traces I assume you are savvy enough to work on the unfiltered VCL, there is otherwise nothing to filter out if your VCL didn't compile in the first place.

During bugwash @bsdphk made it clear that if we filter anything out, he would rather we removed all comments. In which case I would agree this is not a good idea.

Anyway, the bugwash direction is to:

- do without vcldoc comments for #3484 for now
- possibly make libvcc available to more than just varnishd
- revisit the docs integration later

On the question of `varnishd -C` outputting to `stderr`:

- we might have other programs than `varnishd` linking to libvcc
- the `varnishd -C` option might move to a different program
- changing the output to `stdout` would be for 7.0 anyway

I will later open new issues to track both changes and close this pull request.

Dridi:
Registering an insight here, the built-in VCL should be documented in its own manual page: the VCL manual in this pull request is a bit bloated.

bsdphk:
I think a builtin-vcl(7) man-page is a good idea.

(I have slim'ed down vcl(7) by moving variables, backends and probe out to separate pages.)

PR #3506, opened 2021-01-24
Add AGEZ()

nigoroll:
Assert Greater or Equal to Zero.

Trivial suggestion to counter Coverity issues 1472401 and 1472402: Simply wrap a signed value which we require to be positive.

The Coverity issue basically complains about assigning an `ssize_t` to `vz.avail_in` which is unsigned.

If accepted, we should probably promote this to a vdef.h?

### DIFF
--- a/bin/varnishtest/vtc_gzip.c
+++ b/bin/varnishtest/vtc_gzip.c
@@ -54,6 +54,13 @@ vtc_report_gz_bits(struct vtclog *vl, const z_stream *vz)
 }
 #endif
 
+static size_t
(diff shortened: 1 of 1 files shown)


Dridi (commented):
+0

bsdphk:
Obligatory bikeshedding: "AGE" being an actual but entirely unrelated word is a bit distractin, and the `size_t` specialization is non-obvious.

For now, until we spot more uses, keep it as a local #define in vtc_gzip.c.

PR #3508, opened 2021-01-25
Introduce VCL_SUB and VCL_SUB_DYNAMIC

rezan:
This PR introduces `VCL_SUB` and `VCL_SUB_DYNAMIC`. Its based on #3163 and implements @Dridi's idea to split the type into 2 so we can have both compile time and runtime checking behavior. This is done as 3 additional commits to #3163. The usecases for these types range from function aliasing, dynamic calls, callbacks, and looping.

As I mentioned before, I believe that this split actually gives us a much stronger design than previously implemented. The best way to explain this is to address what this combined PR actually does:

* Adds a new `SUB` vmod type. This type is identical in behavior to the existing VCL based `sub`/`call` syntax. Touching a `sub` anywhere will result in a compile time context check. The usecase is that the `SUB` is immediately called by the vmod (callbacks and looping).
* Adds new `VRT_*` functions to support calling subs in vmod code. These functions have additional runtime checks, which add more safety. This comes directly from #3163.
* You **cannot** return a `SUB` from a vmod function. The `SUB` type is restricted to VCL based definitions that can be compile time checked. VMODs can only accept them, its a one way street.
* Introduces `SUB_DYNAMIC`. This is a completely new type which happens to share its C code core struct with `SUB`. This means the `VRT_*` functions with runtime checks work on both types since its the same C struct. However, VCC sees them as different animals. Because its a new type, existing compile time VCL checks simply don't exist and aren't done.
* Allow `SUB` types to be cast into `SUB_DYNAMIC`. This bridges the above VCC gap and allows VCL based subs to be inputted in as unchecked `SUB_DYNAMIC` vmod types. This solves the aliasing and dynamic call usecase.
* You **cannot** cast a `SUB_DYNAMIC` back to a `SUB`. The enforces the rule that all `SUB` types must be VCL based and compile time checked. Subs returned from vmods can be anything, so they can only be `SUB_DYNAMIC` (see previous point blocking `SUB` as a return type). If you want to pass a SUB between 2 vmods, `SUB_DYNAMIC` is how that is done.

To me, this is a very logical separation between the 2 types while sharing common behavior and code. The takeaway here is that all of the behavior from #3163 is maintained while adding a new type which brings back the existing compile time checking behavior.

As to why? I will try and keep this brief, but we (Varnish Software) have found that compile errors can be easily understood and remedied by Varnish users themselves (assuming the compiler messages are good). However, runtime errors usually require extra support from us or this project and can be labor intensive. Runtime errors require going thru logs and piecing together behavior, which seems to something not all users can do. This idea is echoed throughout the software project ecosystem where more and more compile time checks are being done to prevent runtime problems. The Rust language is an innovator in this space. I think we would all love to have our C bugs spotted ahead of time by the compiler than dealing with the aftermath here as a Github issue... This is the motivation for keeping the compile time checks for the VCL `SUB` type.

### DIFF
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -740,6 +740,9 @@ vcc_expr5(struct vcc *tl, struct expr **e, vcc_type_t fmt)
 				vcc_expr_tostring(tl, e, STRINGS);
 				ERRCHK(tl);
 			}
+			if (fmt == SUB) {
+				vcc_AddCall(tl, tl->t, sym);
+			}
 			return;
 		}
 		VSB_printf(tl->sb,
@@ -1343,8 +1346,8 @@ vcc_expr0(struct vcc *tl, struct expr **e, vcc_type_t fmt)
 	t1 = tl->t;
 	if (fmt->stringform)
 		vcc_expr_cor(tl, e, STRINGS);
-	else if (fmt == REGEX)
-		vcc_expr4(tl, e, REGEX);
+	else if (fmt == REGEX || fmt == SUB)
+		vcc_expr4(tl, e, fmt);
 	else
 		vcc_expr_cor(tl, e, fmt);
 	ERRCHK(tl);
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -720,8 +720,10 @@ vcc_expr5(struct vcc *tl, struct expr **e, vcc_type_t fmt)
 			(*e)->instance = sym;
 			return;
 		}
-		if (sym->kind == SYM_FUNC && sym->type == VOID) {
-			VSB_cat(tl->sb, "Function returns VOID:\n");
+		if (sym->kind == SYM_FUNC && (sym->type == VOID ||
+		    sym->type == SUB)) {
+			VSB_printf(tl->sb, "Function returns %s:\n",
+			    sym->type->name);
 			vcc_ErrWhere(tl, tl->t);
 			return;
 		}
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -445,7 +445,7 @@ vcc_do_arg(struct vcc *tl, struct func_arg *fa)
 		vcc_do_enum(tl, fa, PF(tl->t));
 		SkipToken(tl, ID);
 	} else {
-		if (fa->type == SUB)
+		if (fa->type == SUB || fa->type == SUB_DYNAMIC)
 			tl->subref++;
 		vcc_expr0(tl, &e2, fa->type);
 		ERRCHK(tl);
@@ -1382,6 +1382,9 @@ vcc_expr0(struct vcc *tl, struct expr **e, vcc_type_t fmt)
 		vcc_expr_tobool(tl, e);
 		ERRCHK(tl);
 	}
+
+	if ((*e)->fmt == SUB && fmt == SUB_DYNAMIC)
+		(*e)->fmt = SUB_DYNAMIC;
 
 	if (fmt != (*e)->fmt)  {
 		VSB_printf(tl->sb, "Expression has type %s, expected %s\n",
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -445,6 +445,8 @@ vcc_do_arg(struct vcc *tl, struct func_arg *fa)
 		vcc_do_enum(tl, fa, PF(tl->t));
 		SkipToken(tl, ID);
 	} else {
+		if (fa->type == SUB)
+			tl->subref++;
 		vcc_expr0(tl, &e2, fa->type);
 		ERRCHK(tl);
 		assert(e2->fmt == fa->type);
--- a/bin/varnishd/cache/cache_vrt_vcl.c
+++ b/bin/varnishd/cache/cache_vrt_vcl.c
@@ -521,30 +521,55 @@ VCL_##func##_method(struct vcl *vcl, struct worker *wrk,		\
 /*--------------------------------------------------------------------
  */
 
-VCL_VOID
-VRT_call(VRT_CTX, VCL_SUB sub)
+VCL_STRING
+VRT_check_call(VRT_CTX, VCL_SUB sub)
 {
 	struct vbitmap *vbm;
+	VCL_STRING err = NULL;
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	CHECK_OBJ_NOTNULL(sub, VCL_SUB_MAGIC);
 	CHECK_OBJ_NOTNULL(ctx->vcl, VCL_MAGIC);
 	assert(sub->vcl_conf == ctx->vcl->conf);
+	vbm = ctx->called;
+	AN(vbm);
 
 	if ((sub->methods & ctx->method) == 0) {
-		VRT_fail(ctx, "Dynamic call to \"sub %s{}\" not allowed "
-		     "from here", sub->name);
+		err = WS_Printf(ctx->ws, "Dynamic call to \"sub %s{}\""
+		    " not allowed from here", sub->name);
+		if (err == NULL)
+			err = "Dynamic call not allowed and workspace overflow";
+		return (err);
+	}
+
+	if (vbit_test(vbm, sub->n)) {
+		err = WS_Printf(ctx->ws, "Recursive dynamic call to"
+		    " \"sub %s{}\"", sub->name);
+		if (err == NULL)
+			err = "Recursive dynamic call and workspace overflow";
+	}
+
+	return (err);
+}
+
+VCL_VOID
+VRT_call(VRT_CTX, VCL_SUB sub)
+{
+	struct vbitmap *vbm;
+	VCL_STRING err;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(sub, VCL_SUB_MAGIC);
+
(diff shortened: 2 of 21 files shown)


rezan:
I guess I should put up some examples. With this PR and using the `SUB` type, this is a compile error:

```
sub header_check {
    set beresp.http.check = "true";
}

sub vcl_recv {
    vmod.callback(header_check); // ERROR, CANNOT USE header_check here
}
```

This works:

```
sub header_check {
    set beresp.http.check = "true";
}

sub vcl_backend_response {
    vmod.callback(header_check); // OK
}
```

And if you use the `SUB_DYNAMIC` type, both examples work as the compile time checks are relaxed.

Dridi:
FYI `m00054.vtc` failed in CI with an unexpected panic: https://travis-ci.org/github/varnishcache/varnish-cache/jobs/756156119.

lib/libvcc/vcc_expr.c
```diff
@@ -1345,6 +1348,8 @@ vcc_expr0(struct vcc *tl, struct expr **e, vcc_type_t fmt)
 		vcc_expr_cor(tl, e, STRINGS);
 	else if (fmt == REGEX)
 		vcc_expr4(tl, e, REGEX);
+	else if (fmt == SUB)
+		vcc_expr4(tl, e, SUB);
```
Dridi, line 1354:
```
else if (fmt == REGEX || fmt == SUB)
	vcc_expr4(tl, e, fmt);
```

Dridi (changes requested):
Overall LGTM, I just want to make sure we remove the faulty assert in the right commit instead of merging a broken one (for future `git bisect` usage).

We also need to address the panic failing the test suite on Travis CI.

nigoroll:
> FYI m00054.vtc failed in CI with an unexpected panic: https://travis-ci.org/github/varnishcache/varnish-cache/jobs/756156119.

I do not think this is related to the additions of this PR, but rather it seems the expected panic happens later than the check for it. Yes, we need to fix that

lib/libvcc/vcc_expr.c
rezan: Fixed

nigoroll:
I stand by what I have said before: If additional compile time restrictions are considered useful, I am fine with them.

That said, I would like to point out again that they only provide any benefit when vmod authors cooperate: A static sub can still be saved by a vmod and called from a wrong context and I also do not see how we would or even should prevent that *).

Regarding the static sub return value, I do not understand how it helps to outlaw it, but, again, I see no reasons not to have that for an additional type if considered beneficial.

So, overall, for the previous points I see how, with well behaved VMODs, the additional compile time restrictions can avoid VCL errors for the loop / callback case and I am :+1:  for this patch in that regard.

Where this PR absolutely lost me is renaming the `SUB` type. In my mind, the `SUB` type as proposed in #3163 is generic and the type with additional compile time restrictions is a special animal. I understand that @rezan and me see this from oppososite ends, and I will no longer try to convince him otherwise. So I would go with whatever consensus between @Dridi  and @bsdphk . If however, a tie break was required, my vote was to keep `SUB` as in #3163 and use `SUB_STATIC` for the type suggested in this PR.

*) The best idea I could come up with so far would be to create the static sub on the stack such that vmod authors would likely see a panic if they just saved the pointer. But, again, this would not _prevent_ the "call from wrong context" scenario.

Dridi:
> That said, I would like to point out again that they only provide any benefit when vmod authors cooperate: A static sub can still be saved by a vmod and called from a wrong context and I also do not see how we would or even should prevent that *).

We document it, if VMOD authors don't comply that's their problem.

> Regarding the static sub return value, I do not understand how it helps to outlaw it, but, again, I see no reasons not to have that for an additional type if considered beneficial.

I agree, that prevents use cases like a VMOD function taking a sub and a fallback sub and returning one or the other based on some internal computation. However, it is easier to lift that restriction if we have a use case in the future, enforcing this from day one ensures that relaxing this check does not break existing code (not true the other way around).

> Where this PR absolutely lost me is renaming the `SUB` type.

Agreed, the rationale behind this rename is splitting hair IMHO but I don't feel strongly one way or the other. I'd like however to complain that the rename commit itself is a bit messy.

rezan:
The reason for the rename is that `VCL_SUB` and the existing `SUB` type now share common behavior for compile time checking. This means we can now just reference both as `SUB`. If we were to split it the other way around, the `SUB` type now has 2 different types of behavior, so we now need this logic to figure out if we compile time check: `(type == SUB && sym->kind = SUB) || type == SUB_STATIC`. Its cleaner to just do: `type == SUB`.

nigoroll:
> Its cleaner to just do: `type == SUB`.

I do not find this argument relevant, suggest to stop the debate and leave it with the established decision process.

rezan:
I’m also going to recommend we squash all the commits, I agree it’s a mess.

Dridi:
> I’m also going to recommend we squash all the commits, I agree it’s a mess.

Not necessarily all, but if you have a commit that says it renames a type, it should stick to that. I'm all for incremental changes, not squashing all changes into one patch.

bsdphk:
What if VCL have already called functions involved by the time of the first VRT_Call() ?

    sub foo1 {
        vmod.call_this_now(foo2);
    }

    sub foo2 {
         call foo1;
    }

In VSUB 'infected' VCL's, VGC needs to go through the bitmap song&dance for all calls.

Dridi:
For the static variant it should be treated similarly to the `call` action at compile time, only the dynamic variant has its checks deferred to runtime.

I don't see a test case for this but I would expect it to catch the problem, we'll see.

nigoroll:
> only the dynamic variant has its checks deferred to runtime.

this is not fully accurate. No matter the SUB type, the runtime checks happen always. Only the static case has additional compile time restrictions.

bsdphk:
Maybe this is a better example ?

    sub foo2 {
        vmod.call_stashed();
    }
    
    sub foo1 {
         call foo2;
    }

    sub vcl_recv {
        vmod.stash(foo1);
        call foo1;
    }

VCC has no way of knowing what is called in `foo2`, so unless VGC's calls participate in the bitmap, the recursion detection does not kick in until both `foo1` and `foo2` has been called twice.

nigoroll:
@bsdphk I am not exactly sure if this is what you want to get to, but in my mind your example is exactly the case why I questioned the usefulness of the static sub checks in the first place: We will need the runtime checks anyway and vcc can only restrict the places where a sub is _passed_ into a vmod, but never where it is actually called.
In other words, the static sub vcc checks are useful for one particular scenario, but not in the general case.

bsdphk:
Anything we can detect at compile time, should be detected at compile time.

nigoroll:
@bsdphk I think we do that

nigoroll:
FTR: @rezan and me had a constructive f2f meeting during which we also covered this topics. I do not intend to speak for Reza, but I would summarize that our perspectives have not changed, so I still think the best option is for @Dridi and @bsdphk to make a decision.

rezan:
Yup, great talk with @nigoroll, brainstormed some possible collaboration ideas for 2021.

I spent a bit more time on this today, there will be a larger challenge here and that is we can return a `VCL_SUB`. This gives `SUB` split type behavior, it sometimes can be checked, it sometimes cannot be checked, and the challenge if figuring that out. Basically, if we return a `SUB`, can we use that in a `call`? If we have a `SUB` variable, how do we track what type `SUB` is it? When we cast a `SUB` to a checked `SUB_STATIC`, how do we know if we can perform the check? How do we prevent unchecked `SUB` from being used as checked? The answer here is we need to move more and more to a  weakly typed runtime system (which interestingly enough is halfway implemented here). I will simply link the wikipedia definition of these kinds of type systems: https://en.wikipedia.org/wiki/Strong_and_weak_typing. I think its pretty interesting that we are seeing some of these things here.

That being said, I do have a compromise, and that is to introduce the new types as:

* `VCL_SUB_CHECKED` (this was previously `VCL_SUB`)
* `VCL_SUB_UNCHECKED` (this was previously `VCL_SUB_DYNAMIC`)

Because we have a clean checked `SUB` parent type, then we can cast any `SUB` into each new type safely. Once again `VCL_SUB_CHECKED` is one way, it can only be inputted in, it cannot be returned from a VMOD. And `VCL_SUB_UNCHECKED` cannot be cast back to `SUB`, so anything returned as unchecked stays unchecked.

nigoroll:
The generic `VCL_SUB` as in #3163 can be passed and returned. It can be stashed away and called later. It can only be runtime checked.
The "static sub" suggested here can still be stashed away and called later, so it sill needs to be runtime checked.
Whatever type system we use, saving a function pointer and calling it later is _nothing_ a compiler will ever be able to detect, a point which I have been repeating far too often now, IMHO.
We could actually make the "static sub" a different C type and forbid that be called at runtime from contexts where it was not initially passed as an argument. This could be seen as some kind of weak typing.
But while I could see a point in adding some additional compiler checks for the "static subs" to avoid, by convention, it be used from wrong contexts, I see no point in additionally runtime checking for cases where some programmer has violated the contract.

bsdphk:
My considered opinion:

VCC can tell conclusively that VCL_SUB arguments have been passed to VMODs but has no idea what happens to them past that.

In VCLs where VCL_SUB arguments are used, the VCC compiled `call` needs to participate in the run-time recursion check (ref: my VCL example above.

Since VCC code generation happens in the first (and only) pass, the only way to ensure that, is to always wrap VCC compiled `call` in `VRT_call()` and have the latter examine the `VCL_conf` if the bitmap operations are required.

As far as I can tell, that means VCL_SUB_DYNAMIC, and therefore this PR is surplus to requirements.

I suggest we close her and continue in #3163, which is also not *quite* as shovel-ready as @nigoroll seem to think.

rezan:
Sure

PR #3512, opened 2021-01-30
assert() for VGC and arm assertions on ctx->method

nigoroll:
We wrap `VAS_Fail()` as `VPI_Fail()` and add an `assert()` macro to the Varnish Generated C (VGC) code. This `assert()` definition is basically a copy of that in `vas.h`, but deliberately added to a source file which is only used by `generate.py` for VGC, such that this slightly different definition be only visible to VGC.

I also pondered the option to include VCL source information in `VPI_Fail()`: By using the information updated by `VPI_count()`, we could have that, but as the assert is, for now, intended only to ensure correctness of VCC and core VRT code, I decided against this complication.

With `assert()` in place for VGC, we arm the assertions from 75acb5cc7aa73fcd22b150dd1a3b899c6ec9bc44

### DIFF
--- a/bin/varnishd/cache/cache_vpi.c
+++ b/bin/varnishd/cache/cache_vpi.c
@@ -133,3 +133,10 @@ VPI_vcl_select(VRT_CTX, VCL_VCL vcl)
 	VSLb(ctx->req->vsl, SLT_VCL_use, "%s via %s",
 	    req->vcl->loaded_name, vcl->loaded_name);
 }
+
+void v_noreturn_
+VPI_Fail(const char *func, const char *file, int line,
+    const char *cond)
+{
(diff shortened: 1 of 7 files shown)


nigoroll:
bugwash: Ensure we also dump VPI_count info with a panic

Dridi (approved):

nigoroll:
> bugwash: Ensure we also dump VPI_count info with a panic

see #3515 : Unless I overlooked an easy way, a clean solution is not that simple. I will thus go ahead with this change and look after my promise when the rabbit hole got a bit less deep.

nigoroll:
and FTR, phk had OKed during bugwash...

PR #3514, opened 2021-02-01
vcl: New vcl_lookup subroutine

### LINKED ISSUE #3259: RFC: sub vcl_lookup
Today with VCL we have two subroutines where we can influence the cache lookup..

In `vcl_recv` we can generally set `req.*` variables such as `req.grace` or `req.hash_always_miss` to tweak cache lookups.

In `vcl_hash` we compute the hash, and jump straight to a lookup in core code.

We might need the hash to make meaningful lookup decisions though. With vmod_saintmode for example, we can turn a backend sick for a given hash, and it works in the context of a cluster-type directors that may pick a different backend.

It doesn't work for other purposes and the workflow I'm suggesting is the following:

```vcl
sub vcl_recv {
    set req.backend_hint = my_sm.backend();
}

sub vcl_lookup {
    if (std.healthy(req.backend_hint)) {
        set req.grace = 1m; /* lower maximum grace for healthy backends&objects */
    }
}
```

The `vcl_lookup` step would happen between `vcl_hash` and `vcl_{hit,miss,pass}`, only if `vcl_recv` returned with a `hash` transition, giving one last opportunity to influence the in-core lookup based on `req.*`, _\_including\__ `req.hash`.

Suggested transitions: `go_go_go`, `restart`, `fail`.

Suggested builtin:

```vcl
sub vcl_lookup {
    return (go_go_go);
}
```

Dridi:
This adds an additional step to set req.* variable before the effective
cache lookup. This is useful for anything requiring the request hash to
operate, like for example a circuit breaker based on vmod_saintmode.

Closes #3259

---

This is overall a trivial patch series.


doc/sphinx/users-guide/vcl-built-in-subs.rst
```diff
@@ -139,13 +139,45 @@ of the following keywords:
     see  :ref:`fail`
 
   ``lookup``
-    Look up the object in cache.
+    Control passes to :ref:`vcl_lookup` when coming from a ``hash``
+    return in `vcl_recv`.
```
nigoroll, line 143: Shouldn't that be _when coming from vcl_hash_ ?

nigoroll (commented):
I am :+1: except for a minor doc question

doc/sphinx/users-guide/vcl-built-in-subs.rst
Dridi: Both technically.

gquintard:
Just a thought: this could be an opportunity to make things a bit saner and have the returns match the subroutines a bit better. Currently, `pass`/`synth`/`purge`/`pipe` are fairly ok, but `return(hash)` leads users to believe it's the only one going to `vcl_hash`.

Would it be possible, in the long term, to aim for `return(lookup)` in `vcl_hash` (replacing the current `return(hash)`), and `return(lookup)` (or `done` or `ok`) in `vcl_hash`?

For now, we could introduce them as aliases, and deprecate/obsolete the current ones with the next major.

PR #3520, opened 2021-02-15
lookup: Add a notice for high numbers of variants

Dridi:
As a recurring pitfall sometimes hard to identify, it deserves its own
notice. The accounting of variants reuses the worker's strangelove field
since it doesn't conflict where other parts of the code base use it.


bsdphk:
We are not hardcoding '10' like that, make it a param.

nigoroll (approved):

PR #3536, opened 2021-03-01
Use the REQ priority for incoming connection tasks by the acceptor

mbgrydeland:
When accepting new incoming connections in the acceptor thread, it would
schedule, they would be registered with the VCA priority. This priority is
reserved for the acceptor thread itself, and specifically is not included
in the TASK_QUEUE_CLIENT categorisation. This would interfere with the
thread reserve pools.

t02011.vtc had to be adjusted to account for the new priority
categorisation of the initial request.

### DIFF
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -516,7 +516,7 @@ vca_accept_task(struct worker *wrk, void *arg)
 
 		wa.acceptsock = i;
 
-		if (!Pool_Task_Arg(wrk, TASK_QUEUE_VCA,
+		if (!Pool_Task_Arg(wrk, TASK_QUEUE_REQ,
 		    vca_make_session, &wa, sizeof wa)) {
 			/*
 			 * We couldn't get another thread, so we will handle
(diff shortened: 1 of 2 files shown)


bin/varnishd/cache/cache_acceptor.c
```diff
@@ -516,7 +516,7 @@ vca_accept_task(struct worker *wrk, void *arg)
 
 		wa.acceptsock = i;
 
-		if (!Pool_Task_Arg(wrk, TASK_QUEUE_VCA,
+		if (!Pool_Task_Arg(wrk, TASK_QUEUE_REQ,
```
nigoroll, line 519: wouldn't the failure case a few lines down need the same adjustment?

nigoroll (approved):

bin/varnishd/cache/cache_acceptor.c
Dridi: No, the one below is letting the incoming session take over the thread and rescheduling the acceptor, so only this one is incorrect.

nigoroll: right, thank you

PR #3538, opened 2021-03-01
When streaming, send headers if no body data available yet

nigoroll:
This is a TTFB optimization: I noticed that, when streaming, we would only sometimes send headers while waiting for more body data from the backend, depending on the amount of body data read ahead when reading the backend headers.

The `flush_head` debug bit would avoid the delay, but it leads to an extra flush in all cases, even when we could send some body data down the line with the headers.

This trivial patch peeks into the busy object to determine if the first `ObjWaitExtend()` is likely to block and, if so, issues a flush, which will cause the headers to be sent in addition to any initial data VDPs might have to send.

### DIFF
--- a/bin/varnishd/storage/storage_simple.c
+++ b/bin/varnishd/storage/storage_simple.c
@@ -280,6 +280,8 @@ sml_iterator(struct worker *wrk, struct objcore *oc,
 	l = 0;
 
 	u = 0;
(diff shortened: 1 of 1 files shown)


Dridi (approved):

nigoroll:
ok by phk during bugwash

PR #3540, opened 2021-03-02
Bring back some of the recent hash changes

Dridi:
This brings back the ability to start hashing from `vcl_recv` and the chained hash input. I left out the `vcl_hash` skip, that's the only difference for the commits I brought back.

To be revisited after the freeze.

### DIFF
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -183,19 +183,6 @@ HSH_DeleteObjHead(const struct worker *wrk, struct objhead *oh)
 	FREE_OBJ(oh);
 }
 
-void
-HSH_AddString(struct req *req, void *ctx, const char *str)
-{
-
-	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
-	AN(ctx);
-	if (str != NULL) {
-		VSHA256_Update(ctx, str, strlen(str));
-		VSLb(req->vsl, SLT_Hash, "%s", str);
-	} else
-		VSHA256_Update(ctx, &str, 1);
-}
-
 /*---------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


nigoroll:
My suggestions to go forward:
* turn `hash_data(x)` into `req.hash.add(x)` (see below regarding vcl versions)
* introduce a VRT (runtime) `freeze` type method to prevent changes to objects, e.g.

```
sub vcl_recv {
    req.hash.add("yo");
    req.hash.freeze();
    req.hash.add("this will fail");
    return (synth(200, "unreachable"));
}
```
if accepted, this could be generalized to other types

* make `freeze` only available with `vcl 4.2`
* remove `hash_data()` from `vcl 4.2`. This way, any included vcl still using `hash_data()` will vcc-fail, avoiding the "VCL library wrong hash" silent failure
* introduce a `frozen` runtime method to check if `freeze` had been called and use it in the built-in vcl_hash if `vcl 4.2` is active:
```
sub vcl_hash {
    if (req.hash.frozen()) {
        return (lookup);
    }
    # current vcl_hash
    req.hash.add(req.url);
    if (req.http.host) {
        req.hash.add(req.http.host);
    } else {
        req.hash.add(server.ip);
    }
    return (lookup);
}
```
* keep `req.hash` when switching VCLs to support cache segregation for "virtual hosting"

Regarding the hash chaining, I am not against it, but I also do not see a compelling reason why we need it. It is less efficient and I see no strong benefit.

Dridi:
As I said in https://github.com/varnishcache/varnish-cache/issues/3259#issuecomment-817792117 I think that departing from using the state machines and their steps to map read/write scopes of symbols is a bad idea.

Dridi:
Bugwash: we abandon this idea.

PR #3546, opened 2021-03-08
resp.reason and vcl_synth can race with obj.* pointers

rezan:
When an `obj.*` field is passed into the `resp.reason` in a `return(synth())` expression, under certain conditions the `oc` can be dereferenced early and lead to a use after free in `vcl_synth`. The fix is to do a workspace copy if the value is not local when entering a synthetic response. This will also cover `vcl_backend_error`.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -85,6 +85,14 @@ VRT_synth(VRT_CTX, VCL_INT code, VCL_STRING reason)
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


nigoroll (changes requested):
I do not think this change avoids the issue. The copy of `reason` has to happen before the `HSH_DerefObjCore()` one way or another.

nigoroll (approved):
nevermind, @mbgrydeland helped me notice my misunderstanding

FTR, I am ok with this spot-fix, but we should replace it with a different solution later, which covers all possible cases

PR #3547, opened 2021-03-08
range: Propagate the VDP error for short ranges

Dridi:
And fix the h2_req VDP error handling as per the VDP contract.

Test case inspired by @simonvik. Since this is one of those test cases that
explicitly mix two features I wasn't sure whether I wanted to make this
an h2 test case or a range test case. Since this was ultimately a range
bug I decided to register it in a range test case.

It's not obvious what should have been authoritative here. The range VDP
was rightfully latching an error via SC_RANGE_SHORT that is defined as an
error-type session close reason, but VDP_DeliverObj() doesn't take that
into account. While SC_RANGE_SHORT isn't a session/protocol error for h2
but rather a stream error it is not obvious what VDP_DeliverObj() should
do in the absence of a negative retval and the presence of a non-null
sess_close.

Maybe another way could be to turn enum sess_close into a struct and
embed http1 and h2 specificities directly in struct fields. We already
have somewhat structured information in the sess_close.h table.

Refs 03f71c6e6dae

---

This is for the March release.

### DIFF
--- a/bin/varnishd/cache/cache_range.c
+++ b/bin/varnishd/cache/cache_range.c
@@ -55,8 +55,10 @@ vrg_range_fini(struct vdp_ctx *vdc, void **priv)
 
 	CHECK_OBJ_NOTNULL(vdc, VDP_CTX_MAGIC);
 	CAST_OBJ_NOTNULL(vrg_priv, *priv, VRG_PRIV_MAGIC);
-	if (vrg_priv->range_off < vrg_priv->range_high)
+	if (vrg_priv->range_off < vrg_priv->range_high) {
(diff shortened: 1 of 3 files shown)


Dridi:
> Yes, I think turning sess_close into a struct makes good sense.

I think we should also turn it into a `struct stream_close`, and pack it with enough info for the transport to decide whether the session should be closed too (somewhat similar to the relationship from 5c1d08263a8d161b7318cca13c1778ac955bdb3d but transport-agnostic or maybe rather all-transports-aware). I also think we should require a stream_close on VDP error or provide a generic fallback (the latter would be kind to third-party VDPs).

A discussion for later and a change for after the freeze.

PR #3548, opened 2021-03-09
Introduce vcl_builtin_* subs

gquintard:
Provide a set of predictable subroutines that will be called by the
built-in vcl. Most built-in code now lives in those subroutines, apart
for `vcl_synth` and `vcl_backend_error` which would not be idempotent,
so I stopped at those.

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -36,11 +36,15 @@ vcl 4.0;
 # Client side
 
 sub vcl_recv {
+	call vcl_builtin_recv;
+	return (hash);
+}
+
+sub vcl_builtin_recv {
 	call vcl_req_host;
 	call vcl_req_method;
 	call vcl_req_authorization;
 	call vcl_req_cookie;
-	return (hash);
 }
 
 sub vcl_req_host {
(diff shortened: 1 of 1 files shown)


nigoroll:
@gquintard I do not understand the point about `vcl_synth` and `vcl_backend_error`, what is the issue with moving the body ob these subs?

Dridi:
I think I understand what @gquintard means but I don't see a problem since you already have to actively make use of it by either prepending something to `vcl_builtin_synth` or calling it. Which remains true for any built-in sub.

I don't see a reason to give them a special treatment,

gquintard:
I discussed it with @Dridi and it looks like my expectations were wrong, hence my initial decision. I've now aligned the branch with @Dridi view.

For the record, I was thinking the `vcl_builtin_*` routines should either `return` or be noops, and in this case, the first commit ensured that this three cases would do the same thing:

```
sub vcl_synth {
    <custom logic>
    call vcl_builtin_synth;
}

sub vcl_synth {
    <custom logic>
    call vcl_builtin_synth;
    return (deliver);
}

sub vcl_synth {
    <custom logic>
    return (deliver);
}
```

Dridi (approved):
If it's fine with everyone I will squash and merge myself later today.

nigoroll (approved):
go ahead @Dridi

PR #3553, opened 2021-03-16
std.now() and std.timed_call()

nigoroll:
The first commit of the two is just a doc polish which goes to this PR because of the current freeze.
The second commit is the actual PR:

``std.now()`` fills a gap left by the stable time of the now variable in VCL: Sometimes we need to know the current time after some longer processing in VCL.

``std.timed_call()`` is intended to measure longer operations in VCL. ``std.timestamp()`` exists already to add accurate measurements to the log. ``std.timed_call()`` is for cases where the duration is required in VCL, for example to enrich HTTP requests/responses.

### DIFF
--- a/doc/sphinx/reference/vcl_var.rst
+++ b/doc/sphinx/reference/vcl_var.rst
@@ -1417,6 +1417,15 @@ now
 	When converted to STRING in expressions it returns
 	a formatted timestamp like ``Tue, 20 Feb 2018 09:30:31 GMT``
 
+	``now`` remains stable for the duration of any built-in VCL
+	subroutine to make time-based calculations predictable and
+	avoid edge cases.
+
+	In other words, even if considerable amounts of time are spent
+	in VCL, ``now`` will always represent the point in time when
+	the respective built-in VCL subroutine was entered. ``now`` is
+	thus not suitable for any kind of time measurements. See
+	:ref:`std.timestamp()` in :ref:`vmod_std(3)`.
 
 sess
 ~~~~
--- a/doc/sphinx/reference/vcl_var.rst
+++ b/doc/sphinx/reference/vcl_var.rst
@@ -1425,7 +1425,8 @@ now
 	in VCL, ``now`` will always represent the point in time when
 	the respective built-in VCL subroutine was entered. ``now`` is
 	thus not suitable for any kind of time measurements. See
-	:ref:`std.timestamp()` in :ref:`vmod_std(3)`.
+	:ref:`std.timestamp()`, :ref:`std.now()` and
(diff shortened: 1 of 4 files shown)


nigoroll:
notes from bugwash:
There are requirements for different points in time e.g. for ttl calculations, determining runtimes and timestamps.
We seem to be heading towards more than one timestamp, for example `{req,resp,obj,bereq,beresp}.time` has been suggested.

hermunn:
If we introduce `req.time` and friends, it will be easy to document things like _calculate the time since the request came in_, which is just a difference between the current timestamp and a _read only_ variable.

bsdphk:
We talked about this on bugwash yesterday, and decided to note our views in comments, here is mine:

I think the most sensible and easily understandable semantics is if "now" gets set before we call into the vcl methods, and retains that value for the entire execution of that method.  Other needs for timestamps should be serviced by std.*

nigoroll:
@bsdphk s opinion covers mine completely

Dridi:
I think, just like core code makes the assumption today that significant time may have passed when we produce a timestamp, any attempt at measuring VCL execution by means of `std.timestamp()` should be taken as a signal that significant time may have passed, likely because of a VMOD calling into an external service, and update `now` accordingly.

I´m not against the idea of having a sensible "current subroutine time" semantics, but it should be its own symbol, like for example `sub.time`, something expected to be fixed.

nigoroll:
@Dridi my counter argument is still: Why change semantics and then introduce a new symbol for the old semantics, when you can just introduce a symbol for the new semantics?

Dridi:
Because with what you and @bsdphk suggest we are already changing the behavior and underlying semantics of `now` so we might as well go all the way.

nigoroll:
@Dridi I can follow your point now. In my mind, making sure `now` was always up to date would be a fix, but I agree that this argument is not particularly strong.

nigoroll:
I now have a strong use strong case for `std.now()`: Generating a token after time has passed since the vcl sub was entered.

nigoroll:
FTR: bugwash reached consensus on a related topic:
`now` should consistently become the timestamp when the last "VCL step sub" (newly suggested term for the VGC called from core code) was called.

Dridi:
Instead of adding a new function I would rather build upon what we already have:

```diff
-$Function VOID timestamp(STRING s)
+$Function TIME timestamp([STRING s])
```

You may use `std.timestamp()` to either or both get the current time and log it.

bsdphk:
@nigoroll raised a good point in bugwash:  Having `timestamp` sometimes make a log-record and sometimes not is not good style.

PR #3557, opened 2021-03-19
Fix a race condition segfault in sml_getspace()

mbgrydeland:
The simple stevedores uses VTAILQ_LAST in sml_getspace() on the fetch side with the knowledge that the last storage segment won't be touched by a 'pass' client thread freeing stuff as it sends them. But VTAILQ_LAST and VTAILQ_PREV is implemented as a pointer-to-previous-previous-next dereference, meaning also the next to last storage buffer needs to remain unchanged for this scheme to work.

This problem was spotted while testing out the transit_buffer feature. Studying the problem though points to it not being caused by transit_buffer, but rather transit_buffer making it more likely to happen.

To fix this issue, I suggest to switch to using vstq.h instead for the list of storage buffers. This tail queue implementation is a simple and naive tail queue, but does give the guarantee that if the last element is not changed, then head->last also will not change.

### DIFF
--- a/bin/varnishd/storage/storage.h
+++ b/bin/varnishd/storage/storage.h
@@ -33,6 +33,8 @@
  *
  */
 
+#include "vstq.h"
+
 struct stevedore;
 struct sess;
 struct objcore;
@@ -54,7 +56,7 @@ struct storage {
 #define STORAGE_MAGIC		0x1a4e51c0
 
 
-	VTAILQ_ENTRY(storage)	list;
+	VSTQ_ENTRY(storage)	list;
 	void			*priv;
 
 	unsigned char		*ptr;
--- a/bin/varnishd/storage/storage_persistent_silo.c
+++ b/bin/varnishd/storage/storage_persistent_silo.c
@@ -433,7 +433,7 @@ smp_sml_getobj(struct worker *wrk, struct objcore *oc)
 
 		bad = 0;
 		l = 0;
-		VTAILQ_FOREACH(st, &o->list, list) {
+		VSTQ_FOREACH(st, &o->list, list) {
 			bad |= smp_loaded_st(sg->sc, sg, st);
 			if (bad)
 				break;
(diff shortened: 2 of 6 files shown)


nigoroll:
have you checked `VSTAILQ_*`?

mbgrydeland:
> have you checked `VSTAILQ_*`?

I did consider switching to VSTAILQ, but in sml_trimstore we do a 'remove last element' operation, which is inefficient for that variant. Only remove head can be done quickly.

nigoroll (approved):
I am not too excited about adding another tailqueue implementation, but at this point I agree that is looks like the best option. The other two I pondered are:
- not removing the second last element of the storage list
- keeping another pointer to the second last element such that we could use VSTAILQ and `VSTAILQ_REMOVE_AFTER()`

nigoroll:
Actually, we should probably at least consider the option to switch to `VSTAILQ` and just bite the bullet of walking the list to remove the last element.

bsdphk:
I have reversed the order of segments on the list.

If the diagnosis is correct, that should solve the problem.

mbgrydeland:
I can confirm that the reversal of the tailq does indeed fix the issue for transit buffer, as in no longer able to cause a segfault by stressing the test cases. I believe this proves the analysis of the problem as being the double deref on VTAILQ tail access while not under lock.

As for the chosen solution among the now two proposals (though one seems to already be commited), I think the reversal scores a small minus point for POLA, having to reason in reverse looking at the code. But plus for bringing with it less code I guess. The VSTQ implementation is quite well tested though I think I should add, as we use that extensively internally.

As for the VTAILQ double deref tail/reverse, I still wonder why it is implemented like that. Anyone know? In my opinion it is non-intuitive and makes the code for the macros implementing it very hard to read. My best guess at the benefit from doing it that way would be to improve the chance of CPU cache hits, though I'm not sure I see how that'll be.

No strong opinion on the final chosen solution to this.

mbgrydeland:
Closed in favour of 3fedecadc83d16cfc4a7ac4097c39bd3e9607a2f

PR #3558, opened 2021-03-22
condfetch: Handle failed stale_oc without a boc

Dridi:
The assertion that the stale objcore of a conditional fetch cannot be
failed unless it was streaming is incorrect. Between the moment when
we grab the stale objcore in HSH_Lookup and the moment we try to use
it after vcl_backend_response, the backend fetch may have completed or
failed.

Instead, we need to treat an ongoing fetch and a failed fetch as
separate checks since the latter may happen with or without a boc.


nigoroll (approved):

Dridi:
Rebased and ready to merge, waiting for more potential reviewers.

Dridi:
Approved by @bsdphk too during bugwash today.

Issue #3560, opened 2021-03-24
Early backend close can result in a valid zero length response


rezan:
If a backend connection closes early during streaming, it can result in a *valid* zero length response to the client. This valid response can be accepted by the client and potentially poison any upstream caches or CDNs. The proper behavior is to leave the `Content-Length` unchanged and let the backend connection error mirror to the streaming clients.

VTC (this should fail):

``` vcl
varnishtest "A backend connection error gets returned as a valid short response"

barrier b1 sock 2

server s1 {
	rxreq
	txresp -nolen -hdr "Content-Length: 10"
	barrier b1 sync
	send "12345"
	# Early connection close error
} -start

varnish v1 -vcl+backend {
	import vtc;

	sub vcl_deliver {
		# Make sure we are streaming and give the backend time to error out
		vtc.barrier_sync("${b1_sock}");
		vtc.sleep(1s);
	}
} -start

client c1 {
	txreq
	rxresp
	expect resp.http.Content-Length == "0"
	# Wrong, we should get an invalid response and a client error, failing this VTC
} -run
```

This affects `master` and `6.0`.

PR #3561, opened 2021-03-24
Don't change the C-L if we detect a failed stream

rezan:
Previously we would read the response Content-Length from a failed oc, which would make the error response valid. Now, if this is detected, we don't touch the Content-Length.

The root of the problem is that we race with a `boc` and end up with a failed `objcore`, all before transmitting anything. I guess we should be careful, anytime we check for a `boc` and don't get one, we should then check if the resulting `oc` is valid. Most of the time this is harmless. In this case, we are changing the `Content-Length` to zero and returning a completely valid response.

Fixes https://github.com/varnishcache/varnish-cache/issues/3560

### DIFF
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -417,7 +417,7 @@ cnt_transmit(struct worker *wrk, struct req *req)
 	status = http_GetStatus(req->resp);
 	head = !strcmp(req->http0->hd[HTTP_HDR_METHOD].b, "HEAD");
 
-	if (boc != NULL)
+	if (boc != NULL || (req->objcore->flags & (OC_F_FAILED)))
 		req->resp_len = clval;
 	else
 		req->resp_len = ObjGetLen(req->wrk, req->objcore);
@@ -471,6 +471,11 @@ cnt_transmit(struct worker *wrk, struct req *req)
 	VSLb_ts_req(req, "Resp", W_TIM_real(wrk));
 
 	HSH_Cancel(wrk, req->objcore, boc);
+
+	if (!req->doclose && (req->objcore->flags & OC_F_FAILED))
+		/* The object we delivered failed due to a streaming error.
+		 * Fail the request. */
+		req->doclose = SC_TX_ERROR;
 
 	if (boc != NULL)
 		HSH_DerefBoc(wrk, req->objcore);
--- a/bin/varnishtest/tests/r03560.vtc
+++ b/bin/varnishtest/tests/r03560.vtc
@@ -1,0 +1,29 @@
+varnishtest "A backend connection error gets returned as a valid short response"
+
+barrier b1 sock 2
+
+server s1 {
+	rxreq
+	txresp -nolen -hdr "Content-Length: 10"
+	barrier b1 sync
+	send "12345"
+	# Early connection close error
+} -start
+
+varnish v1 -vcl+backend {
+	import vtc;
+
+	sub vcl_deliver {
+		# Make sure we are streaming and give the backend time to error out
+		vtc.barrier_sync("${b1_sock}");
+		vtc.sleep(1s);
+	}
+} -start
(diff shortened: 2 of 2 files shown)


bin/varnishtest/tests/r03560.vtc
```diff
@@ -0,0 +1,29 @@
...
+		vtc.barrier_sync("${b1_sock}");
+		vtc.sleep(1s);
+	}
+} -start
+
+client c1 {
+	txreq
+	rxresphdrs
+	expect resp.http.Content-Length == "10"
+	recv 5
```
Dridi, line 27: `rxrespbody` and then check that what's been received is indeed "12345".

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -472,6 +472,11 @@ cnt_transmit(struct worker *wrk, struct req *req)
 
 	HSH_Cancel(wrk, req->objcore, boc);
 
+	if (!req->doclose && (req->objcore->flags & OC_F_FAILED))
+		/* The object we delivered failed due to a streaming error.
+		 * Fail the request. */
+		req->doclose = SC_REM_CLOSE;
```
Dridi, line 478: This sounds more like `RX_BAD` or `TX_ERROR`, definitely not "Client Closed".

Dridi (changes requested):

bin/varnishd/cache/cache_req_fsm.c
rezan: I just copied what the `v1d_vdp` would send if we caught this during transmission. No real opinion.

bin/varnishtest/tests/r03560.vtc
rezan: I think its more important to show the `expect_close` at the end. Also, it seems like `rxrespbody` doesn't work on a response like this.

bin/varnishd/cache/cache_req_fsm.c
Dridi: In `v1d_vdp` it looks like a fallback.

bin/varnishtest/tests/r03560.vtc
Dridi: RTFM?

> Receive (part of) a response's body.

rezan: Sorry, just to be crystal clear, `rxrespbody` will always result in a VTC failure since the body is incomplete. Maybe I’m missing something?

bin/varnishd/cache/cache_req_fsm.c
nigoroll: `SC_REM_CLOSE` does not have the `is_err` flag and also semantically it is definitely wrong.
From the existing errors, I would also vote for `SC_TX_ERROR`, but I wonder if we should have `SC_TX_SHORT`? If so, we would need to use it in at last one other place.

bin/varnishtest/tests/r03560.vtc
nigoroll: I agree with the VTC as is

nigoroll (changes requested):

nigoroll:
Well spotted!

nigoroll:
bugwash:
* Please use `SC_TX_ERROR`, we are probably going to overhaul the session close reasons, but for not this should be good. Please add an `XXX session close reason overhaul`.
* The VTC is fine

bin/varnishd/cache/cache_req_fsm.c
rezan: Changed to `SC_TX_ERROR`

rezan:
Changed the `req->doclose` flag to `SC_TX_ERROR`. I left out the XXX comment, not sure I understand what was discussed well enough to make a good comment.

nigoroll (approved):

PR #3562, opened 2021-03-25
Add {req,resp,bereq,beresp,obj}.time

nigoroll:
See https://github.com/varnishcache/varnish-cache/pull/3553#issuecomment-804133604

Reviewers please let me know what think about this: We could have more & better assertions on `(ctx)->where`. If we want them, I would propose to add some utility macros like

```
#define VRT_CHECK_req(ctx) ...
#define VRT_CHECK_oc(ctx) ...
...

#define VRT_ACCESS_req(ctx) ...
#define VRT_ACCESS_oc(ctx) ...
...
```

such that we can generalize macros

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -393,6 +393,7 @@ struct busyobj {
 
 	/* Timers */
 	vtim_real		t_first;	/* First timestamp logged */
+	vtim_real		t_resp;		/* response received */
 	vtim_real		t_prev;		/* Previous timestamp logged */
 
 	/* Acct */
@@ -484,6 +485,7 @@ struct req {
 	vtim_real		t_first;	/* First timestamp logged */
 	vtim_real		t_prev;		/* Previous timestamp logged */
 	vtim_real		t_req;		/* Headers complete */
+	vtim_real		t_resp;		/* Entry to last deliver/synth */
 
 	struct http_conn	*htc;
 	struct vfp_ctx		*vfc;
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -427,7 +427,7 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 	VSLb_ts_busyobj(bo, "Fetch", W_TIM_real(wrk));
 	i = VDI_GetHdr(bo);
 
-	now = W_TIM_real(wrk);
+	bo->t_resp = now = W_TIM_real(wrk);
 	VSLb_ts_busyobj(bo, "Beresp", now);
 
 	if (i) {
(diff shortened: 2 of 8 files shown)


hermunn:
Should there be some text somewhere about what happens for retries and restarts? Maybe something like, _if a retry occurs, bereq.time will be updated again before `sub vcl_backend_fetch` is called for the retry_.

nigoroll:
@hermunn sure we can have that just to be make it clear

nigoroll:
bugwash: @Dridi and me to discuss in light of our timeouts reform proposal

Dridi (changes requested):
I don't think we need to look at this through the timeout lens, and this is a very tidy diff very pleasant to review!

I would like to see small changes:

- @hermunn's suggested clarification regarding restarts and retries
- new symbols limited to VCL >= 4.1
- `beresp.time` in `vcl_backend_error` (same as `resp.time` in `vcl_synth`)

Another suggestion on top of this change, should we consider`obj.age` too?

nigoroll:
todo note from discussion with @Dridi 

>     * @hermunn's suggested clarification regarding restarts and retries

yes

>     * new symbols limited to VCL >= 4.1

please, no

>     * `beresp.time` in `vcl_backend_error` (same as `resp.time` in `vcl_synth`)

yes

> Another suggestion on top of this change, should we consider`obj.age` too?

@bsdphk has already done this for me in 2016 :)  81913d93d978f45bfa49b65321cb49e4f1e8ad88

hermunn (approved):
Looks very good, thanks!

Dridi:
@nigoroll the size increase of `struct req` requires these adjustments:

```diff
--- bin/varnishtest/tests/r02219.vtc
+++ bin/varnishtest/tests/r02219.vtc
@@ -22,7 +22,7 @@ varnish v1 -arg "-p workspace_client=9k" \
 } -start
 
 client c1 {
-       send "PROXY TCP4 127.0.0.1 127.0.0.1 1111 2222\r\nGET /${string,repeat,752,A} HTTP/1.1\r\n\r\n"
+       send "PROXY TCP4 127.0.0.1 127.0.0.1 1111 2222\r\nGET /${string,repeat,744,A} HTTP/1.1\r\n\r\n"
        rxresp
 } -run
 
@@ -39,6 +39,6 @@ ${string,repeat,732,"42 "}
 } -run
 
 client c3 {
-       send "PROXY TCP4 127.0.0.1 127.0.0.1 1111 2222\r\nGET /${string,repeat,756,C} HTTP/1.1\r\n\r\n"
+       send "PROXY TCP4 127.0.0.1 127.0.0.1 1111 2222\r\nGET /${string,repeat,748,C} HTTP/1.1\r\n\r\n"
        rxresp
 } -run
```

Dridi:
I just realized that `bereq.time` is not available in `vcl_pipe`. @nigoroll, was it on purpose?

Dridi:
https://github.com/varnishcache/varnish-cache/pull/3759

PR #3566, opened 2021-04-01
vsl: New "fetch status" field for Hit records

Dridi:
The larger the response we hit, the more likely we are to hit an ongoing
fetch and since we buffer transaction logs it can become really tedious
to infer whether the backend transaction may have already been overrun
or not flushed yet. At least, such a heuristic was impossible with a VSL
query until now.

### DIFF
--- a/bin/varnishd/cache/cache_req.c
+++ b/bin/varnishd/cache/cache_req.c
@@ -72,6 +72,33 @@ Req_AcctLogCharge(struct VSC_main_wrk *ds, struct req *req)
 	memset(a, 0, sizeof *a);
 }
 
+void
+Req_LogHit(struct worker *wrk, struct req *req, struct objcore *oc,
+    intmax_t fetch_progress)
+{
+	const char *clen, *sep;
+
+	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
+	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
+
+	if (fetch_progress >= 0) {
+		clen = HTTP_GetHdrPack(wrk, oc, H_Content_Length);
+		if (clen == NULL)
+			clen = sep = "";
+		else
+			sep = " ";
+		VSLb(req->vsl, SLT_Hit, "%u %.6f %.6f %.6f %jd%s%s",
+		    ObjGetXID(wrk, oc), EXP_Dttl(req, oc),
+		    oc->grace, oc->keep,
+		    fetch_progress, sep, clen);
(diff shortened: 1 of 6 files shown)


Dridi:
I´m having second thoughts.

Today we log this:

```
Hit <vxid> <ttl> <grace> <keep>
```

With this patch we´d log this instead:

```
Hit <vxid> <ttl> <grace> <keep> <fetch_status>
```

Where the fetch status might be `fetching` or `complete`.

I´m wondering whether the cost would be acceptable to log this instead:

```
Hit <vxid> <ttl> <grace> <keep> [<fetched_so_far> [<content_length>]]
```

In other words, a more granular progress of an ongoing fetch when hitting a streaming busyobj.

nigoroll:
> I´m wondering whether the cost would be acceptable to log this instead:

:+1:

Dridi:
Updated to be more granular.

bin/varnishd/cache/cache_req.c
```diff
@@ -72,6 +73,37 @@ Req_AcctLogCharge(struct VSC_main_wrk *ds, struct req *req)
...
+{
+	struct objcore *oc;
+	struct boc *boc;
+	const char *clen, *sep;
+
+	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
+	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	oc = req->objcore;
+	CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
+	boc = HSH_RefBoc(req->objcore);
```
nigoroll, line 87: um. Maybe I did not understand the impact of your earlier question on the whether to log `len_so_far`, but taking another boc ref just for logging sounds like something we should avoid.

Dridi: We could avoid the boc reference but would it be safe to read `len_so_far`? Getting the content length from the objcore should remain safe regardless.

nigoroll: I think we should really avoid it. If this feature required another bocref, I'd rather not have it.

Dridi: > but would it be safe to read `len_so_far`?

Dridi:
Bugwash: let's try to produce the SLT_Hit record in HSH_Lookup where we could grab the progress from the bock under the right lock without requiring a boc reference.

I will update the pull request accordingly.

Dridi:
Updated as per the bugwash suggestion, we no longer need a boc reference to capture its progress.

Dridi:
Rebased against master to solve the VTC conflict.

nigoroll (approved):

PR #3567, opened 2021-04-01
[varnishtest] log the line along the command

gquintard:
This one is a low-hanging fruit that can help with large VTC: printing the line number with each command:

```
**   top   === l1: varnishtest foo
*    top   VTEST foo
**   top   === l3: server s1 {
**   top   === l11: foo "
---- top   Unterminated quoted string in line: foo "
*    top   RESETTING after foo.vtc
```

Thoughts?

### DIFF
--- a/bin/varnishtest/vtc.c
+++ b/bin/varnishtest/vtc.c
@@ -315,7 +315,7 @@ parse_string(struct vtclog *vl, void *priv, const char *spec)
 	struct vsb *token_exp;
 	char *e, *p, *q, *f, *buf;
 	int nest_brace;
-	int tn;
+	int tn, ln = 0;
 	unsigned n, m;
 	const struct cmds *cp;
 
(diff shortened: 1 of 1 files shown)


bin/varnishtest/vtc.c
```diff
@@ -315,7 +315,7 @@ parse_string(struct vtclog *vl, void *priv, const char *spec)
 	struct vsb *token_exp;
 	char *e, *p, *q, *f, *buf;
 	int nest_brace;
-	int tn;
+	int tn, ln = 0;
```
Dridi, line 318: Doesn't `ln` need to start from the spec line number? I'd be surprised if it worked with anything but top comments.

gquintard: ah yes, indeed, my tests appeared to work because because I mishandled whitespaces in a convincing way

gquintard:
turns out this needs a bit more than a 5-minutes hack

Issue #3565, opened 2021-03-31
Alpine is not happy


gquintard:
this is on `master`:

```
99b9eabd78e8:/workspace# make V=1
make  all-recursive
make[1]: Entering directory '/workspace'
Making all in include
make[2]: Entering directory '/workspace/include'
make  all-am
make[3]: Entering directory '/workspace/include'
make[3]: Leaving directory '/workspace/include'
make[2]: Leaving directory '/workspace/include'
Making all in lib
make[2]: Entering directory '/workspace/lib'
Making all in libvarnish
make[3]: Entering directory '/workspace/lib/libvarnish'
/bin/sh ../../libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I../..  -I../../include -I../../include   -DVARNISH_STATE_DIR='"/usr/local/var/varnish"'    -g -O2 -pthread -Wall -Werror -Wno-error=unused-result -fstack-protector-strong -Werror -Wall -Wcast-align -Wcast-qual -Wchar-subscripts -Wempty-body -Wextra -Wformat-y2k -Wformat-zero-length -Wmissing-field-initializers -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wpointer-sign -Wreturn-type -Wshadow -Wstrict-aliasing -Wstrict-prototypes -Wswitch -Wsystem-headers -Wunused-parameter -Wunused-result -Wwrite-strings -Wno-thread-safety -Wno-old-style-definition -Wno-sign-compare -Wno-implicit-fallthrough -Wno-builtin-requires-header -Wno-incomplete-setjmp-declaration -Wno-redundant-decls -Wno-missing-variable-declarations -Wno-missing-field-initializers -O0 -g -fno-inline -MT libvarnish_la-vcli_proto.lo -MD -MP -MF .deps/libvarnish_la-vcli_proto.Tpo -c -o libvarnish_la-vcli_proto.lo `test -f 'vcli_proto.c' || echo './'`vcli_proto.c
[...]

PR #3568, opened 2021-04-02
build: wflags,py attempts for #3565

Dridi:
Pull request meant to exercise travis and cci for #3565.


Dridi:
Giving up on alpine, it will run without developer warnings. The important part is that we don't don't wholesale remove crucial warnings for the benefit of a platform we don't actively support.

PR #3570, opened 2021-04-10
Fix format specifier (causes a compilation failure on macOS)

xordspar0:
While gcc is fine with this either way, clang previouly failed to
compile, complaining that asw->count (a uint64_t, aka unsigned long long
on my system) does not match the format specifier %jd (a long on my
system).

### DIFF
--- a/vmod/vmod_debug_acl.c
+++ b/vmod/vmod_debug_acl.c
@@ -30,6 +30,7 @@
 
 #include "config.h"
 
+#include <inttypes.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


bsdphk:
Good catch.

I would prefer if you cast the argument to `(intmax_t)` instead ?

fgsch:
fwiw, there is a macOS builder in Travis that unfortunately is not check often enough.

xordspar0:
Ok, I can do that. Or would it be better to use `%ju` and `(uintmax_t)`, since asw->count is a 64-bit unsigned int?

xordspar0:
Oops, ignore me. The page was out of date when I submitted my last comment.

PR #3572, opened 2021-04-12
feature: Transit buffer

fwsGonzo:
This PR implements the transit buffer feature (previously called pass buffering).

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -278,7 +278,7 @@ struct boc {
 	void			*stevedore_priv;
 	enum boc_state_e	state;
 	uint8_t			*vary;
-	uint64_t		len_so_far;
+	uint64_t		fetched_so_far;
 };
 
 /* Object core structure ---------------------------------------------
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -279,6 +279,8 @@ struct boc {
 	enum boc_state_e	state;
 	uint8_t			*vary;
 	uint64_t		fetched_so_far;
+	uint64_t		delivered_so_far;
+	uint64_t		transit_buffer;
 };
 
 /* Object core structure ---------------------------------------------
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -666,7 +666,7 @@ vbf_stp_fetch(struct worker *wrk, struct busyobj *bo)
 	if (oc->flags & OC_F_PRIVATE)
 		AN(bo->uncacheable);
 
-	oc->boc->len_so_far = 0;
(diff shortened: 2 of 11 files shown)


nigoroll:
this PR has an infinite complexity/description ratio

fwsGonzo:
Feel free to give some input on what the defaults should be. Zero bytes transit buffer has to be allowed, because that disables the feature.

fwsGonzo:
I rebased against latest and added a commit message. The message is not very long right now so if there's anything you would like  me to add there, I will be happy to do that.

Dridi:
I force-pushed a split of the transit buffer patch into 4 commits. Doing so, I tried to polish the commit log, C code and test cases. I couldn't make c112 fail reliably with transit buffer disabled (it passes reliably when enabled) so I didn't keep it.

nigoroll:
powwow: @nigoroll to propose an implementation without the timeout

nigoroll:
rebased onto master here https://github.com/nigoroll/varnish-cache/tree/transit_buffering
now working on the proposed chane

nigoroll:
`tests/c00110.vtc` (formally `tests/c0009.vtc`)  is not reliable on the rebased but otherwise onmodified a2ca25fc658afe0d6d69d010ff5bfb0b647887a0
```
$ ./varnishtest -in 1000 -j50 tests/c00110.vtc 
...
---- v1    Not true: VBE.vcl1.s1.conn (1) == 0 (0)
```

@fwsGonzo / @Dridi could you look into this?
My plan on this ticket was to implement my proposed change (which I still think should be safe, sound and simple), but this issue seems non trivial and is blocking any other progress...

fwsGonzo:
I am running on an X1 Carbon laptop the same test with lower parallelism, but I am unable to reproduce. Should add that I am extending varnishtest to support my CMake build system, but other than that everything is the same. I rebased using your transit_buffering branch.

nigoroll:
@fwsGonzo can we just have the branch for this PR updated without unrelated commits please?

fwsGonzo:
You can push to it if you want!

anthosz:
Dear all,

Do we have an estimation when it will be merged and for which version?

Really interesting feature that can save memory and optimize stability 😁

fwsGonzo:
@anthosz You can probably just build this PR from source and use it as-is. As long as @nigoroll did not break anything with his improvements, it should be fine. This feature has been running in production in the enterprise version of Varnish for a long time now.

anthosz:
> @anthosz You can probably just build this PR from source and use it as-is. As long as @nigoroll did not break anything with his improvements, it should be fine. This feature has been running in production in the enterprise version of Varnish for a long time now.

Thank you but it's strange that it's not merged in this case 🤔

nigoroll:
Pushed in the context of this PR: caf49d8b42b97c3599e01f2e503d24991a1a0bcf

PR #3575, opened 2021-04-12
[LTS Backport] resp.reason and vcl_synth can race with obj.* pointers

stevendore:
We can incorrectly reference resp.reason from other sources when
jumping into vcl_synth. This also covers passing in a reason in
vcl_backend_error.

Original PR: https://github.com/varnishcache/varnish-cache/pull/3546


rezan (approved):
Good to go

Issue #3535, opened 2021-02-24
Wrong Content-Length sometimes sent to no-gzip clients on miss, causing truncation


yched:
## Setup
- debian 8.8 (jessie) 
- varnish_6.0.7-0+daily+20210106.160410+jessie_amd64
- http_gzip_support=on
- `set beresp.do_gzip = true` in vcl_backend_response (our backends don't send gzipped responses)

## Current Behavior
After switching from Varnish 5.2.1 to 6.0.7 recently, we got reports of API consumers sometimes receiving truncated JSON responses.

Investigating our logs showed that:
- Those were from clients without "Accept-Encoding: gzip" support, 
  thus involving varnish unzipping the response 
- They only happened on cache miss (or very rarely on cache hits that were in fact coalesced with a miss around the same moment).
  About 0.01% of cache misses were affected
- The responses had the correct unzipped body, but had a Content-Length header that was lower, and which matched the length of the truncated JSON string our clients were trying to parse.
  How such a response (content-length < actual body) is interpreted seems to depend on the client: some do read the full body nonetheless, others truncate it to the content-length value.
- Those incorrect (too low) Content-Length values seem fairly random (they are not, for instance, equal to the gzipped size),
  but a couple values come up more often than others : 1280, 1024, 256...
- This mostly seems to affect small responses
  We temporarily mitigated the issue by doing something like "set beresp.do_gzip = true only if beresp.http.content-length < 1Ko", which reduced the occurrences from a couple 100s a day to (still) a couple a day at the moment.

The only similar report we found is https://serverfault.com/questions/953009/varnish-cache-randomly-changing-response-content-length from two years ago. Our case is pretty much the same :
- they saw the bug on uncached endpoints, 
  our API is cached and we see the bug on cache misses
  (they mention the same ratio of about 0.01% of buggy responses though)
- their backends return gzipped beresp, 
  ours do not and we do_gzip instead 
- back then they ran 6.1 and had the same issue with 6.0.2, 
  we run 6.0.7 (and never had the issue on 5.2.1)

## Steps to reproduce 

A bit tricky, given it only seems to affect ~0.01% of cache misses.

I failed to come up with a test showing the bug, since Content-Length is present roughly half of the time (with "Transfer-Encoding: chunked" instead when it's not) : 
```
server s1 -repeat 1000 {
    rxreq
    txresp -hdr "Cache-Control: no-cache" -bodylen 300
} -start
[...]

PR #3576, opened 2021-04-12
[LTS Backport] Guard against OA_GZIPBITS streaming race

stevendore:
Originally PR: https://github.com/varnishcache/varnish-cache/pull/2904.

Fixes: https://github.com/varnishcache/varnish-cache/issues/3535

### DIFF
--- a/bin/varnishd/cache/cache_gzip.c
+++ b/bin/varnishd/cache/cache_gzip.c
@@ -292,6 +292,10 @@ vdp_gunzip_init(struct req *req, void **priv)
 	ssize_t dl;
 	uint64_t u;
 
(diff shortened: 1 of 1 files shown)


yched:
Cool, thanks for backporting this.

Not that I really get the underlying code, but in the [issue](https://github.com/varnishcache/varnish-cache/issues/3535#issuecomment-785056604), @Dridi seemed to think #3187 was also part of the orignal fix that went in master ?

rezan:
Its an alternate way to fix the issue, but it requires taking a reference. The fix in this PR resolves the issue.

rezan (approved):
Looks good

PR #3582, opened 2021-04-12
[LTS Backport] Cond fetch boc race

stevendore:
The assertion that the stale objcore of a conditional fetch cannot be
failed unless it was streaming is incorrect. Between the moment when
we grab the stale objcore in HSH_Lookup and the moment we try to use
it after vcl_backend_response, the backend fetch may have completed or
failed.

Instead, we need to treat an ongoing fetch and a failed fetch as
separate checks since the latter may happen with or without a boc.

 Conflicts:
	bin/varnishd/cache/cache_fetch.c

Original PR: https://github.com/varnishcache/varnish-cache/pull/3558


rezan (commented):
Lets double check this.

rezan (approved):
Good to go

PR #3587, opened 2021-04-13
[LTS Backport] Backend cooling fix

stevendore:
This backport does a few things.
1. makes sure we cooling under the correct lock
2. Properly destroys dynamic backends on failures
3. Add an assert that VRT_delete_backend is only called once


rezan (commented):
Requires extra testing.

rezan (approved):
We cooled a handful of VCLs during our testing, no issues found.

PR #3588, opened 2021-04-13
[LTS Backport] Early backend error triggers zero length response

rezan:
A backend error can trigger Varnish to produce a valid zero length response.

https://github.com/varnishcache/varnish-cache/issues/3560
https://github.com/varnishcache/varnish-cache/pull/3561


rezan:
Tests came back clean, good to go.

PR #3590, opened 2021-04-13
[LTS Backport] H2 thread priority

rezan:
This addresses some things with H2 thread priorities.

https://github.com/varnishcache/varnish-cache/pull/2796
https://github.com/varnishcache/varnish-cache/pull/3536

PR #3591, opened 2021-04-14
[LTS Backport] Watchdog priority

rezan:
Limit watchdog to highest priority workers only.

This should be based on https://github.com/varnishcache/varnish-cache/pull/3590, `r02418.vtc` needs to be removed as this patch also removes that deadlock scenario.

https://github.com/varnishcache/varnish-cache/pull/3537


rezan:
Requires H2 load testing

rezan:
Test came out good

PR #3593, opened 2021-04-15
[LTS Backport] Shard fixes

rezan:
Technically not a backport, these are fixes specific to shard and 6.0.

`canon_point` - This was dropped in master before a big change which added weights. In trying to keep 6.0 stable, we left this in and had to approximate what this value is based on the new algorithm. Previously we used the last point (which sometimes doesn't exist), but upon closer inspection, we should be using the first point.

`shard.reconfigure()` - There is a bug where if you `reconfigure()` and empty shard, future calls of `reconfigure()` will fail. This does not exist anymore in master since `reconfigure()` was made optional and the shared state has been removed. The root of the problem is that we do not cleanup our state when we see there are no backends. I also moved the lock up to cover more operations safely.

### DIFF
--- a/bin/varnishtest/tests/d06001.vtc
+++ b/bin/varnishtest/tests/d06001.vtc
@@ -1,0 +1,30 @@
+varnishtest "Empty shard reconfig"
+
+server s1 {
+	rxreq
+	txresp
+} -start
+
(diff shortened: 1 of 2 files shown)


rezan:
Good to go

Issue #3556, opened 2021-03-19
Varnish not closing backend connection when client disconnects - waiting for first_byte_timeout instead


thetuxkeeper:
Hi,

we have an issue with connections hanging in the backend and waiting for POST data though the client already closed the connection. This is a quite unexpected behavior (for us) and might overload the backend with not yet closed connections.

Is that a known/expected behavior?

## Expected Behavior
Close the connection to the backend at the time varnish knows the client is gone. Then the backend doesn't wait for data without knowing the client is already gone.

## Current Behavior
read error is logged but first_byte_timeout kills the backend connection much later.
Probably it only occurs with reused/keepalive connections.

## Steps to Reproduce (for bugs)
1. Send a too short POST-Request and close the connection (backend part of log)
```
POST /test HTTP/1.1
Host: test-host
Content-Length: 100000
content-type: application/json

0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```
2. You'll see something like that in the log /backend part):
```
*   << BeReq    >> 1058444892
-   Begin          bereq 1058444891 pass
-   VCL_use        reload_20210319_140100_14960
-   Timestamp      Start: 1616162470.789302 0.000000 0.000000
-   BereqMethod    POST
-   BereqURL       /test
-   BereqProtocol  HTTP/1.1
-   BereqHeader    content-length: 100000
-   BereqHeader    content-type: application/json
-   BereqHeader    secure: on
-   BereqHeader    x-forwarded-by: ...
-   BereqHeader    x-forwarded-proto: https
-   BereqHeader    x-forwarded-uri: /test
-   BereqHeader    x-forwarded-host: test-host
-   BereqHeader    x-real-ip: ...
-   BereqHeader    x-unique-id: <uuid>
-   BereqHeader    X-Forwarded-For: ...
-   BereqHeader    X-UA-Device: pc
-   BereqHeader    Host: test-host
-   BereqHeader    Surrogate-Capability: varnish=ESI/1.0
-   BereqHeader    X-Varnish: 1058444892
-   VCL_call       BACKEND_FETCH
-   VCL_return     fetch
-   BackendOpen    501 backend 12.34.56.78 8000 23.45.67.89 53480 reuse
-   FetchError     req.body read error: 104 (Connection reset by peer)
-   FetchError     backend write error: 104 (Connection reset by peer)
-   Timestamp      Bereq: 1616162471.629452 0.840149 0.840149
[...]

PR #3594, opened 2021-04-15
[LTS Backport] Backend fetch stall

stevendore:
Original ticket: varnishcache/varnish-cache#3556
Commits to fix it:
varnishcache/varnish-cache@8976958
varnishcache/varnish-cache@b8a3753


rezan (approved):
Good to go

PR #3595, opened 2021-04-16
varnishncsa: Add -j to format for JSON

### LINKED ISSUE #3297: Impossible to produce JSON output with log-tools
We have found agreement, that we want to ultimately change the internal VSL format so that structured (JSON) output can be produced, see [VIP23 - Refactor VSL...](https://github.com/varnishcache/varnish-cache/wiki/VIP-23:-Refactor-VSL-to-support-extracting-structured-data-from-"binary"-log-payloads). In other words, we want to output sensible structures and not just JSON formatting of the existing everything-is-a-string log entries.
With this background, #2869 has been rejected for a good reason, but I think we missed to salvage another very valid point from this PR: To my knowledge, it is not currently possible to output logs with JSON-escaped fields.

In my mind, we should have an option to at least be able to produce output with escaped strings to not break JSON-formatted output. This would not contraict the long term plan and avoid having to re-format output.
It seems [VS has implemented something like this](https://docs.varnish-software.com/varnish-cache-plus/features/json-logging/).

xordspar0:
Fixes #3297

Co-authored-by: Ben Zvan <ben.zvan@target.com>
Co-authored-by: Jasmine Wang <jasmine.wang@target.com>
Co-authored-by: Jordan Christiansen <jordan.christiansen@target.com>
Co-authored-by: Madhavi Dintakurthy <madhavi.dintakurthy@target.com>

### DIFF
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -146,6 +146,9 @@ static struct ctx {
 	struct vsb		*vsb;
 	unsigned		gen;
 	VTAILQ_HEAD(,format)	format;
+	int			quote_how;
+	char			*missing_string;
+	char			*missing_int;
 
 	/* State */
 	struct watch_head	watch_vcl_log;
@@ -199,9 +202,9 @@ vsb_fcat(struct vsb *vsb, const struct fragment *f, const char *dflt)
 {
 	if (f->gen == CTX.gen) {
 		assert(f->b <= f->e);
-		VSB_quote(vsb, f->b, f->e - f->b, VSB_QUOTE_ESCHEX);
+		VSB_quote(vsb, f->b, f->e - f->b, CTX.quote_how);
 	} else if (dflt)
-		VSB_quote(vsb, dflt, -1, VSB_QUOTE_ESCHEX);
+		VSB_quote(vsb, dflt, -1, CTX.quote_how);
 	else
 		return (-1);
 	return (0);
@@ -247,7 +250,7 @@ format_fragment(const struct format *format)
 	if (format->frag->gen != CTX.gen) {
 		if (format->string == NULL)
 			return (-1);
-		VSB_quote(CTX.vsb, format->string, -1, VSB_QUOTE_ESCHEX);
+		VSB_quote(CTX.vsb, format->string, -1, CTX.quote_how);
 		return (0);
 	}
 	AZ(vsb_fcat(CTX.vsb, format->frag, NULL));
@@ -345,13 +348,13 @@ format_auth(const struct format *format)
 	    CTX.frag[F_auth].e)) {
 		if (format->string == NULL)
 			return (-1);
-		VSB_quote(CTX.vsb, format->string, -1, VSB_QUOTE_ESCHEX);
+		VSB_quote(CTX.vsb, format->string, -1, CTX.quote_how);
 		return (0);
 	}
 	q = strchr(buf, ':');
 	if (q != NULL)
 		*q = '\0';
-	VSB_quote(CTX.vsb, buf, -1, VSB_QUOTE_ESCHEX);
+	VSB_quote(CTX.vsb, buf, -1, CTX.quote_how);
 	return (1);
 }
 
@@ -507,7 +510,7 @@ addf_hdr(struct watch_head *head, const char *key)
 	AN(f);
 	f->func = format_fragment;
 	f->frag = &w->frag;
-	f->string = strdup("-");
+	f->string = strdup(CTX.missing_string);
 	AN(f->string);
 	VTAILQ_INSERT_TAIL(&CTX.format, f, list);
 }
@@ -530,7 +533,7 @@ addf_vsl(enum VSL_tag_e tag, long i, const char *prefix)
 		assert(w->prefixlen > 0);
 	}
 	VTAILQ_INSERT_TAIL(&CTX.watch_vsl, w, list);
-	addf_fragment(&w->frag, "-");
+	addf_fragment(&w->frag, CTX.missing_string);
 }
 
 static void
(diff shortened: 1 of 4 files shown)


xordspar0:
This is a summary of the decisions we made in this change: https://github.com/varnishcache/varnish-cache/issues/3297#issuecomment-820752983

Please leave feedback on the `-j` option vs `%{...}j` and whether encoding as valid UTF-8 is required before merging.

bin/varnishncsa/varnishncsa_options.h
```diff
@@ -78,6 +78,9 @@
 	VOPT("E", "[-E]", "Show ESI requests",				\
 	    "Show ESI requests, implies client mode."			\
 	)
+#define NCSA_OPT_j                                          \
+	VOPT("j", "[-j]", "Print logs to stdout as JSON", \
+		 "Print logs to stdout as JSON.")
```
Dridi, line 83: Pedantic police:

- this isn´t printing as JSON, but rather escaping format components as JSON strings.
- whitespace violations :police_officer:

bin/varnishtest/tests/u00016.vtc
```diff
@@ -0,0 +1,34 @@
...
+	rxresp
+	txreq -hdr "x-request-id: ∃" -hdr "test-unicode:"
+	rxresp
+} -run
+
+delay 1
+
+# ASCII string with -j is valid JSON
+shell -match "^YES$" { 
+	varnishncsa -n ${v1_name} -d -F '{"xrid": "%{X-REQUEST-ID}i"}' -j -q ReqHeader:test | python3 -c 'import sys, json; print(json.load(sys.stdin)["xrid"])'
```
Dridi, line 23: You can break the line after the `|` character to avoid very long lines.

Dridi (changes requested):
LGTM, I only have nit-picks.

> Please leave feedback on the `-j` option vs `%{...}j` and whether encoding as valid UTF-8 is required before merging.

I don´t see a reason to bother you with the UTF-8 encoding, we don´t do that with the CLI today, so that may be handled separately, possibly directly at the VSB level.

We can have both `-j` and `%{...}j` if needed, and `-j` probably covers most of the JSON output needs already. So let´s stick to the current patch for which I only have tiny complaints.

Dridi:
In `enum frag`, the first integer-type fragment is `F_b` and if you follow its trail you will see that its fallback representation is `-` which is probably not what we want for a JSON representation where you´d rather expect `0`  or `null`. The same goes for other integer fragments.

bin/varnishncsa/varnishncsa_options.h
xordspar0: Oops, our mistake. Thanks for catching this.

xordspar0:
> you´d rather expect 0 or null

Good point, we didn't think of that.

xordspar0:
We made changes based on all review comments. I also added documentation, so there is more to review now.

> In general, -j makes the replacements JSON-compatible for all inputs.

I'm not sure I'm as confident as the commit message suggests, especially since we only found the integers being replaced with `-` issue in review. Please let me know of any tests or code I can add to make it more robust.

rezan:
Ping @stevendore (he implemented the Enterprise version of this)

bin/varnishtest/tests/u00016.vtc
```diff
@@ -0,0 +1,56 @@
...
+	rxresp
+} -run
+
+delay 1
+
+# ASCII string with -j is valid JSON
+shell -match "^YES$" {
+	varnishncsa -n ${v1_name} -d -q 'ReqHeader:test eq "ascii"' -j \
+		-F '{"xrid": "%{X-REQUEST-ID}i"}' \
+	| python3 -c 'import sys, json; print(json.load(sys.stdin)["xrid"])'
```
Dridi, line 27: You don't need to break a shell line with `\` before a `|`, the `|` character itself is a natural line breaker:

```sh
#!/bin/sh

echo helloworld |
grep -o l
```

stevendore:
You seem to have covered everything that we had tested in enterprise. Particularly the empty string going from `"-"` to `""`. I did notice you are changing types when a number is unavailable though (`"null"` being a different type than a number).  It is best to keep the same type for comparison, graphing and logging. For example, comparing `"null"` with `5` can cause an error in a JSON parser.

See: https://github.com/varnishcache/varnish-cache/blob/f578fcf77fdddccca8c9330427aa2c300f05c2db/bin/varnishncsa/varnishncsa.c#L1135-L1139

nigoroll:
my vote is to make the null replacement configurable. If we enforced the `null -> 0` / or `null -> ""` replacement, it became impossible to differentiate the cases where you'd need to know (e.g. "was this header set or empty?").
I think we should have a `--json-null` option or, even more fine granined, `--nullstring=xyz` and `--nullnumber=xyz`

xordspar0:
> For example, comparing `"null"` with `5` can cause an error in a JSON parser.

Just to be clear, it will be the value `null`, not a string that contains the word `"null"`, given the following format string:

    {"bytes": %b}

I agree with @nigoroll that is is preferable to be able to distinguish between zero and not present, but we also can't know ahead of time which programs are able to store in their internal data structures a value that can be either null or a number. In that case maybe a configurable replacement string is best.

bin/varnishtest/tests/u00016.vtc
xordspar0: Fixed

Dridi (commented):
> my vote is to make the null replacement configurable. If we enforced the `null -> 0` / or `null -> ""` replacement, it became impossible to differentiate the cases where you'd need to know (e.g. "was this header not present or empty?").
> I think we should have a `--json-null` option or, even more fine grained, `--nullstring=xyz` and `--nullnumber=xyz`

I still think that `%{...}j` could better solve this, especially since d5d70a657, we could imagine something like `%{format:fallback}j` that would even work with braced formats:

 ```
%{%{msec}t:null}j
```

We can revisit the `%{...}j` option if we ever see evidence that such granilarity is needed and so we keep a simple behavior for the `-j` option?

File: bin/varnishncsa/varnishncsa.c [hunk 4]
```diff
@@ -140,6 +140,9 @@ VTAILQ_HEAD(vsl_watch_head, vsl_watch);
 static struct ctx {
 	/* Options */
 	int			a_opt;
+	int			quote_how;
+	char			*empty_string;
+	char			*empty_int;
 	char			*w_arg;
```
Dridi, line 146: I think "missing" is more accurate than "empty" for those fields.

xordspar0:
> We can revisit the %{...}j option if we ever see evidence that such granilarity is needed and so we keep a simple behavior for the -j option?

This looks good. In the mean time should we replace missing ints with 0?

Dridi (approved):

PR #3598, opened 2021-04-19
synth: obj.fetch_{failed,abandoned}

Dridi:
This is scratching an old itch, the inability to know in `vcl_synth` whether we come from a failed fetch, and whether that failed fetch was abandoned.


Dridi:
Bugwash: conflicting with #3113, closing. I still suggest we rename `OC_F_ABANDON` to something less ambiguous like `OC_F_CLIENTGONE`.

PR #3599, opened 2021-04-21
Director reference counts

bsdphk:
Replace the director cooling list with reference counts, the cheapest way I could find.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -413,7 +413,7 @@ vbe_dir_event(const struct director *d, enum vcl_event_e ev)
 			VBP_Control(bp, 0);
 		VRT_VSC_Hide(bp->vsc_seg);
 	} else if (ev == VCL_EVENT_DISCARD) {
-		VRT_Assign_Backend(&bp->director, NULL);
+		VRT_DelDirector(&bp->director);
 	}
 }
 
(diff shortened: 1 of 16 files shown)


nigoroll:
This patch takes the cost of each `set {req.backend_hint,bereq.backend}` in VCL from a pointer assignment to two or four mutex operations. This cost was the primary motivation for the extension of the coollist idea which was the basis for #2725. [The criticism which killed that PR](https://github.com/varnishcache/varnish-cache/pull/2725#issuecomment-438645504) was that it required two mutex operations per task, and my plan was to move acquisition of the coollist reference to under the `vcl_mtx`, which should basically eliminate any additional overhead for normal VCL execution.
I agree that the coollist concept brings with it additional complexity, so I would first like to come to a decision which of the two concepts to pursue.

If we were to go in the direction of this PR, I wonder if we needed to consider this scenario: We must destroy all directors from a vmod before unloading that vmod. Because of cross-vmod director references, I think that directors also need to gain a reference on their defining vmod, which is returned when the director refcount goes to zero. That, in turn, might require us rechoreograph the vmod fini callback -> `VRT_DelDirector()` dance.

nigoroll:
pow-wow: maybe we could get out of the complexity vs. efficiency dilemma by introducing a director scope. References would only need to be taken for scopes smaller than the VCL.

nigoroll:
It looks like this should work. I would like to play a little with actual dynamic backends.

bsdphk:
This has now been committed.

PR #3602, opened 2021-04-26
varnishncsa: Add %{sec,msec,usec,msec_frac,usec_frac}t specification

nigoroll:
... following the example of Apache HTTP Server, see http://httpd.apache.org/docs/current/mod/mod_log_config.html

Note on this limitation:

> The extensions can not be combined with each other or strftime(3) in the same specification. Use multiple ``%{X}t`` specifications instead.

I had actually finished an implementation without this limitation when I realized that being compatible with Apache probably makes more sense than the presumably less limited implementation. Also, I was not convinced that `%{%FT%T.%{usec_frac}}t` was better than `%{%FT%T}t.%{usec_frac}t`


nigoroll:
bugwash: dridi to review & decide

Dridi (approved):

PR #3579, opened 2021-04-12
[LTS Backport] HTTP1 minimal response assert

stevendore:
Original ticket: https://github.com/varnishcache/varnish-cache/issues/3415
Original PR: https://github.com/varnishcache/varnish-cache/pull/3430

PR #3430, opened 2020-10-08
Consistent use of VTCP_Assert

mbgrydeland:
This PR adds EAGAIN/EWOULDBLOCK as an accepted errno for operations on sockets. This fixes #3415.

It also adds VTCP_Assert around other operations on sockets throughout the code. This is for consistency.

Issue #3415, opened 2020-09-21
Assert error in http1_minimal_response(), http1/cache_http1_fsm.c line 208:   Condition(VTCP_Check(1)) not true.


simonvik:
```
Panic at: Fri, 18 Sep 2020 23:27:29 GMT
Assert error in http1_minimal_response(), http1/cache_http1_fsm.c line 208:
  Condition(VTCP_Check(1)) not true.
version = varnish-6.5.0 revision b79e940e716474ab5186336db2a86461b32f2413, vrt api = 11.0
ident = Linux,4.15.18+20190618.1109.03d21a5f.bionic,x86_64,-junix,-smalloc,-sdefault,-hcritbit,epoll
now = 39611894.752254 (mono), 1600471631.662630 (real)
Backtrace:
  0x5564da0c7eaf: /usr/sbin/varnishd(+0x51eaf) [0x5564da0c7eaf]
  0x5564da13a043: /usr/sbin/varnishd(VAS_Fail+0x43) [0x5564da13a043]
  0x5564da0f77cb: /usr/sbin/varnishd(+0x817cb) [0x5564da0f77cb]
  0x5564da0f8466: /usr/sbin/varnishd(+0x82466) [0x5564da0f8466]
  0x5564da0ec8e7: /usr/sbin/varnishd(+0x768e7) [0x5564da0ec8e7]
  0x5564da0ecdd4: /usr/sbin/varnishd(+0x76dd4) [0x5564da0ecdd4]
  0x7f1d26ac06db: /lib/x86_64-linux-gnu/libpthread.so.0(+0x76db) [0x7f1d26ac06db]
  0x7f1d267e9a3f: /lib/x86_64-linux-gnu/libc.so.6(clone+0x3f) [0x7f1d267e9a3f]
errno = 11 (Resource temporarily unavailable)
pthread.self = 0x7f1cbdd20700
pthread.name = (cache-worker)
pthread.attr = {
  guard = 4096,
  stack_bottom = 0x7f1cbdcf1000,
  stack_top = 0x7f1cbdd21000,
  stack_size = 196608,
}
thr.req = 0x7f14ab736020 {
  vxid = 620399327, transport = HTTP/1 {
    state = HTTP1::NewReq
  }
  step = R_STP_TRANSMIT,
  req_body = NULL,
  err_code = 400, err_reason = (null),
  restarts = 0, esi_level = 0,
  sp = 0x7f081d7df520 {
[...]

PR #3603, opened 2021-04-26
[LTS Backport] Catch up VTCP fixes

stevendore:
Original ticket: #3415
Original PR: #3430
Continuation of #3579

### DIFF
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -43,10 +43,6 @@
 #include "vav.h"
 #include "vcli_serve.h"
 
-#if !defined(__has_feature)
-#define __has_feature(x)	0
-#endif
-
 struct plist {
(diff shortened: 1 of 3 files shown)


rezan:
This needs to run thru the tests. I think we can skip the ASAN tests for this and just focus on load and stability.

Dridi:
> I think we can skip the ASAN tests for this and just focus on load and stability.

I added this commit precisely not to lose the ability to use ASAN on the stable branch with what we have in tree (suppressions and all) so I would recommend to check that sanitizer builds didn't break on the stable branch as far as the test suite goes as a minimal baseline.

rezan:
ok, we can rerun the ASAN tests

stevendore:
ASAN test and plain load test are running.

rezan (approved):
Tests came back clean, but be advised we only test on Linux.

PR #3608, opened 2021-05-06
introduce a "startswith" operator for bans

gquintard:
Here's another dumb and unannounced quick PR: adding yet another operator to bans, because:
- a lot of the URL-focused bans look like `req.url ~ "^/something/something/"`
- people rarely realize that the string they feed the ban may contain dots or other special characters that will be interpreted in a regex context
- regex are slow and painful on big caches

there's possibly an argument to add `contains` to the operator set too, but I don't want to push my luck


Dridi:
Why not borrow an actual operator from the regex syntax?

```
req.url ^ "/something/something"
```

gquintard:
I thought of it, but it felt wrong as it's already largely associated with XOR and that felt wrong. That being said, the name is unimportant to me

bin/varnishd/cache/cache_ban.c
```diff
@@ -578,6 +578,12 @@ ban_evaluate(struct worker *wrk, const uint8_t *bsarg, struct objcore *oc,
 			    0, 0, NULL, 0) >= 0)
 				return (0);
 			break;
+		case BANS_OPER_START:
+			if (strlen(arg1) < strlen(bt.arg2) &&
+			    memcmp(arg1, bt.arg2, strlen(bt.arg2))) {
```
bsdphk, line 583: using strlen(bt2.arg) for comparison when arg1 is shorter seems wrong ?!

bsdphk (changes requested):

mbgrydeland:
Discussed during bugwash.

* The ^ operator suggestion was in disfavor, better to spell it out.
* Concerns about breaking the byte code, as that will have implications for persisted caches.
* Suggestion to instead implement this using regexp, as a foo ~ "^\Q...\E", using VRE_quote to help. This would solve the persistent cache concerns too.

bin/varnishd/cache/cache_ban.c
gquintard: ahem, yes, absolutely dumb

nigoroll:
bugwash:
* We should add `re.quote()` (or a similar vmod name
* With that, vcl will be able to create _startswith_  ban-regexen as `"^" + re.quote(req.http.ban)`
* For bans via the cli, I would argue that it is rarely used directly by users not familiar with regexen and any api client programs can easily implement startswith

we suggest to close this ticket

gquintard:
closing works for me. Quoting can be done in `vmod_std` but `re` seems to hint at something grander, any thoughts about it?

nigoroll:
Re @gquintard 
> `re` seems to hint at something grander, any thoughts about it?

I plan to suggest https://code.uplex.de/uplex-varnish/libvmod-re for inclusion after

* having refactored it to `VCL_REGEX`
* #3616 is in
* `vmod_re` is adjusted to the API changes from the pcre2 switch

gquintard:
great to know, thanks @nigoroll !

PR #3609, opened 2021-05-06
Improve vcl caching

nigoroll:
#3245 continued, which, for some reasons, I cannot reopen any more

@bsdphk I agree with your analysis that proper VCL caching with labels is non trivial, yet this PR merely makes an opportunistic attempt to improve it.

Nevertheless, this PR proposes other improvements which are mostly relevant for the non-label case (which is still the most common one, at least in my world):

* The first commit enables vcl reference caching also for backend requests and ESI subrequests
* The third commit collapses `VCL_Recache()` into `VCL_Rel()`, cutting the number `vcl_mtx' operations in half for the former `VCL_Recache()` case

As always, I will accept your veto (if it is one), but in this case I would like to make sure that we do not throw away what I think is sensible optimization potential.

### DIFF
--- a/bin/varnishd/cache/cache_acceptor.c
+++ b/bin/varnishd/cache/cache_acceptor.c
@@ -446,7 +446,7 @@ vca_accept_task(struct worker *wrk, void *arg)
 
 	/* Dont hold on to (possibly) discarded VCLs */
 	if (wrk->wpriv->vcl != NULL)
-		VCL_Rel(&wrk->wpriv->vcl);
+		VCL_Rel(&wrk->wpriv->vcl, NULL);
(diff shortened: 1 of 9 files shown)


nigoroll:
during a discussion with @bsdphk we found my misconception: in `VBO_GetBusyObj()`, `wrk` is still the client worker, so chances of hitting a cached vcl there are marginal if existent at all. An obvious solution would be to keep a reference count and transfer a reference to the busyobject, but that is a complication which we should only take on if there is proof for congestion on `vcl_mtx`.
As this PR was for an opportunistic optimization, this and other options are beyond my intention.

PR #3610, opened 2021-05-10
Implement non-parametric VCL macro facility

bsdphk:
This implements trivial non-parameterized VCL macros.


bsdphk:
There seems to be little taste for this idea.

PR #3613, opened 2021-05-12
vcc: Allow global symbols to be defined after use

### LINKED ISSUE #3555: SUB arguments use-before-define
SUB arguments currently can only be used after being defined. E.g. this works:

```
sub foo { }
sub caller {
  some.vmod(foo);
}
```

but this does not:

```
sub caller {
  some.vmod(foo);
}
sub foo { }
```

Dridi:
A side effect is slightly more accurate error messages.

Closes #3555

### DIFF
--- a/bin/varnishtest/tests/r00916.vtc
+++ b/bin/varnishtest/tests/r00916.vtc
@@ -5,7 +5,7 @@ server s1 {
 	txresp -body "FOO"
 } -start
 
-varnish v1 -errvcl {Symbol not found: 's-1'} {
+varnish v1 -errvcl {Undefined backend s-1, first reference:} {
 	backend b { .host = "${localhost}"; }
 	sub s1 {
 	}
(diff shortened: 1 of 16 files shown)


bsdphk:
OK with me.

Dridi:
IRC discussion:

> ```
> 15:29 < dridi> i found a flaw in 3613
> 15:30 < dridi> i added some coverage to see the result
> 15:30 < dridi> **** v1    CLI RX|Undefined backend foo, first reference:
> 15:30 < dridi> **** v1    CLI RX|('<vcl.inline>' Line 5 Pos 40)
> 15:30 < dridi> **** v1    CLI RX|                set req.backend_hint = foo.bar;
> 15:30 < dridi> **** v1    CLI RX|---------------------------------------###-----
> ```

I will address this later.

Dridi:
Rebased against master with the introduction of functions to peek at the next symbol that highlighted a bit of duplicated logic and a lot of missing null checks for invalid input (but this being in libvcc for incomplete VCL is not harmful).

That however served to centralize some code and avoid the partial symbol pitfall I mentioned in my previous comment.

nigoroll (approved):
I have not done a thorough review, but I guess we would find out soon if thinks broke.

PR #3616, opened 2021-05-19
build: Migration to PCRE2

### LINKED ISSUE #3559: Move to pcre2
Check that it's OK for all the platforms we care about.

Dridi:
The process can be summarized as running `sed 's/pcre/\02/gi'` in the source tree, performing the PCRE2 migration until it compiles, and fixing migration mistakes until the test suite passes.

This migration could have been free of breaking changes if hadn't renamed the following parameters:

- pcre_match_limit to pcre2_match_limit
- pcre_match_limit_recursion to pcre2_depth_limit

I decided to make it clear that moving to PCRE2 was not anecdotal and a good-enough reason to break things. For this reason, in addition to the VRE_has_jit earlier retirement I introduced one more breakage in struct vre_limits by renaming the match_recursion field to depth.

Speaking of breaking changes, PCRE2 seems to be stricter and rejects unknown escape sequences. There may be other breaking changes to the syntax and it should be noted in the release notes that VCL may fail to compile because of that. We should however be able to count on helpful error messages from libvcc.

There are outstanding issues with this migration that need scrutiny. It focuses on source code migration, and non of the RST documentation (like the aforementioned release notes) were updated to reflect the change. In particular, installation documentation.

The other outstanding issues are related to correctness and performance: the ban code doesn't care about what matched, only whether a subject matched or not. PCRE2 insists on allocating a pcre2_match_data object to store the "ovector" where the match offsets are written. The solution for bans was to share a dummy pcre2_match_data, but it might be unsafe. This was done to avoid heap allocation in what is essentially a critical path.

In VRE_exec() we have the same problem, we can't take the caller's ovector and we need to pass a pcre2_match_data to pcre2_match(). Worse, it is no longer a vector of int so in order to preserve VRE_exec()'s signature a copy/conversion of the PCRE2_SIZE vector is needed in addition to the pcre2_match_data allocation. Unlike struct vre_limit that is primarily used by our parameter machinery, I decided to preserve VRE_exec()'s API and avoid leaking the PCRE2_SIZE type outside of VRE.

I am however not sure we can avoid the heap allocation that now occurs in VRE_exec() that can be executed from a critical path.

Closes #3559

---

I decided to keep the findings and improvements made in the process as part of the pull request, some of them are debatable anyway.


Dridi:
CI failed spectacularly, of course I forgot to update the configuration.

Dridi:
There will be no way around breaking `VRE_exec()` with this change. My attempt at dealing with the fact that `pcre2_get_error_message()` wants a destination buffer by adding one in `vre_t` turns compile error messages into systematic use-after-free. This is particularly frustrating because as far as the code is concerned in pcre2's trunk, it is effectively copying a `static const char[]` into the buffer without any form of contextual enrichment or formatting, and returning a `const char *` would have been perfectly fine.

Dridi:
Rebased to remove the SQUASHME noise from the git log, I haven't found more outstanding problems.

My suggestions going forward regarding error messages:

- preserve the API
  - keep VRE_exec() but replace pcre2 errors with a generic "something happened"
  - add a VRE_match() that takes a buffer for responses
- break the API
  - replace VRE_exec() with the VRE_match() suggestion

ingvarha:
varnish pcre2 99bbc57 test build on debian sid/ppc32 : https://gist.github.com/ingvarha/ad2079369bafac42845f48369d1e1c69

ingvarha:
varnish pcre2 99bbc57 test builds:
FAILED: epel-7-x86_64 https://kojipkgs.fedoraproject.org//work/tasks/7834/68327834/build.log
FAILED: epel-7-ppc64le https://kojipkgs.fedoraproject.org//work/tasks/7835/68327835/build.log
OK: epel-8-aarch64 https://kojipkgs.fedoraproject.org//work/tasks/8784/68328784/build.log
OK: epel-8-ppc64le https://kojipkgs.fedoraproject.org//work/tasks/8785/68328785/build.log
OK: epel-8-s390x https://kojipkgs.fedoraproject.org//work/tasks/8786/68328786/build.log
OK: epel-8-x86_64 https://kojipkgs.fedoraproject.org//work/tasks/8787/68328787/build.log
OK: rocky linux 8rc x86_64 https://gist.github.com/ingvarha/fdf0855de5d38aed0b5bf7cbe9b0e6c1
OK: All rawhide fedoras (armv7hl, i686, x86_64, aarch64, ppc64le, s390x) https://koji.fedoraproject.org/koji/taskinfo?taskID=68329490

Dridi:
Same failure for both epel7 builds:

```
**** top   shell_out|-I: Regex error at position 4 ()
---- top   shell_expect not found: ("-I: Regex error at position 4 (missing closing parenthesis)")
```

That's the use-after-free problem being mitigated. That's also probably why cci fails its centos_7 jobs, and I was able to reproduce it locally. Until `VRE_match()` is implemented and the fate of VRE_exec() discussed we can forget about this bug.

Thanks for the testing!

Dridi:
Bugwash:

- @nigoroll to review db265fd03495cda0adc2a9c45724e4f5c7c312a8...b578667008d4d7144b6f1f0c6803936302eba6b2 to merge them sooner
- @nigoroll to attempt dodging the heap allocation during a match
- @dridi to change the VRE API as suggested, breaking hard
- @dridi to migrate the ban code to VRE
- @bsdphk wants this completed as soon as possible before the next release

nigoroll (commented):
I am not so happy with the `VRE_has_jit` removal:

* pcre2 has a replacement, quoting the docs:

> A program can tell if JIT support is available by calling pcre2_config() with the PCRE2_CONFIG_JIT option.

* r01576.vtc serves an important purpose

* The argument 

> we would figure at build time whether pcre was itself built with JIT compilation support, which could be different than the pcre that would be used at run time

is not sound: It should be possible to check for JIT also at runtime

> If JIT support is not available, a call to pcre2_jit_compile() does nothing and returns PCRE2_ERROR_JIT_BADOPTION.

Also I would argue that using a pcre2 build at runtime with a different JIT setting to the one used at build time was a packaging / distribution error

So IMHO we should improve how we detect if JIT is enabled rather than removing it and adjust the VTC.

nigoroll:
cherry-picked 0e1a69991be4db453d7c315ac4eb707d2d9e8c16 and 59fceb72d49bcfdf5d47cf8235f4bf99685a6e02 because they seem to make sense even without pcre2

Dridi:
With the latest changes, even the sanitizers stopped complaining. I had to change the `VRE_compile()` signature to cope with how pcre2 delivers error messages.

The next outstanding item is porting the ban facility over to VRE, but for that I'm going to make JIT compilation optional. This is requested by @nigoroll and needed by the ban facility, at least in spirit since it currently doesn't attempt JIT compilation and I don't want to change that as part of the migration.

Dridi:
The ban migration to VRE is implemented with 0350bba8c...e92caa86b.

Dridi:
Discussed with @bsdphk, I will first update the VRE API according to the current one in this pull request:

https://github.com/varnishcache/varnish-cache/blob/e92caa86bfbc08b1addc073cfe4b36bbe068e2fb/include/vre.h#L40-L66

Then we can perform the migration to pcre2 in a self-contained `vre.c`.

PR #3617, opened 2021-05-20
Increase default stack space

hermunn:
Note: This is a PR into 6.0, our LTS branch.

A default stack space of 48k will cause a stack overflow in _some_ of the test cases (VMOD `blob`, specifically) on _some_ Linux distributions. The increase to 64k seems to be enough to solve this.

### DIFF
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -646,9 +646,9 @@ MCF_InitParams(struct cli *cli)
 	MCF_ParamConf(MCF_MINIMUM, "thread_pool_stack", "%jdb", (intmax_t)low);
 
 #if defined(__SANITIZER) || __has_feature(address_sanitizer)
-	def = 92 * 1024;
+	def = 128 * 1024;
 #else
-	def = 48 * 1024;
+	def = 64 * 1024;
 #endif
 	if (def < low)
 		def = low;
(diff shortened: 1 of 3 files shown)


bsdphk:
We should undoubtedly reconsider parameters like the stacksize for 7.0, so the real question is probably if 64K is the right value ?

(No, I'm *not* going to ask if it is enough for everybody :-)

Dridi:
> A default stack space of 48k will cause a stack overflow in _some_ of the test cases (VMOD `blob`, specifically) on _some_ Linux distributions. The increase to 64k seems to be enough to solve this.

It sounds to me that if your problem is the test suite you should rather adjust the test cases before changing a default value on a stable branch.

> We should undoubtedly reconsider parameters like the stacksize for 7.0, so the real question is probably if 64K is the right value ?

Reminder, we have `v00004.vtc`.

rezan:
I agree with Dridi here, not sure if this warrants changing the default stack size for everyone using LTS. I think we should better understand why you are hitting the stack limit. This has never come up in any of my tests.

nigoroll:
bugwash lite: 
* Reconsider stack sizes for trunk / 7.0 if/when there is a concrete reason to.
* The 6.0 gatekeepers will make a decision for 6.0

Dridi:
FYI 7.0 increases the default `thread_pool_stack` to 80k (64k for 32bit) to accommodate the PCRE2 jit compiler (or to be more accurate the code it produces).

hermunn:
I did a simple rebase which also introduces the 6.0.9 heading. Unless someone protests very firmly, I will just merge this.

mbgrydeland (approved):

PR #3619, opened 2021-05-21
Add std.strftime()

nigoroll:
add a vmod function for `strftime()` formatting with an additional argument to select gmtime or localtime.

As `localtime_r()` requires `tzset()`, that is added to `child_main()` rather than the vmod initialization to avoid possible duplication for other vmods possibly using `localtime_r()`.

Test coverage is integrated with the existing m20.vtc. One of the test dates is changed to fall into the european daylight savings time to expose the two hour time difference between UTC and MET.

### DIFF
--- a/bin/varnishd/cache/cache_main.c
+++ b/bin/varnishd/cache/cache_main.c
@@ -346,6 +346,7 @@ child_main(int sigmagic, size_t altstksz)
 #if defined(__FreeBSD__) && __FreeBSD_version >= 1000000
 	malloc_message = child_malloc_fail;
 #endif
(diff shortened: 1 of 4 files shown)


bsdphk:
It used to be, ~2007,  that calling `tzset()` slowed down time/date functions, costing performance in parsing and production of `Date:` headers.

A similar effect was seen with `setlocale(3)` and `ctype(3)`.

I have not tried to measure if that is still the case.

nigoroll:
bugwash: Remove the `tz=local` (`localtime()`) case, otherwise accepted

PR #3621, opened 2021-05-27
Compile VCL 4.0  from 4.1

### LINKED ISSUE #3368: Can't include 4.0 from 4.1
This has been tested on `master` on an up-to-date Arch linux. It's also present on `6.4`

`man vcl` [says](https://github.com/varnishcache/varnish-cache/blob/master/doc/sphinx/reference/vcl.rst#versioning):

> Multiple versions of the VCL syntax can coexist within certain constraints.
>
> The VCL syntax version at the start of VCL file specified with -f sets the hard limit that cannot be exceeded anywhere, and it selects the appropriate version of the builtin VCL.
> ...
> The version of Varnish this file belongs to supports syntax 4.0 and 4.1.

So if I understand it correctly, we should be able to use a `vcl 4.0` included from a `vcl 4.1` file, and this should work:

```
varnishtest "include 4.0 from 4.1"

shell {
	cat << EOF > ${tmpdir}/inc.vcl
	vcl 4.0;
	sub vcl_recv {
		set req.esi = false;
	}
EOF
}

varnish v1 -syntax 4.1 -vcl {
	include "${tmpdir}/inc.vcl";
} -start
```

but it fails with:

```
**** v1   CLI RX|Message from VCC-compiler:
**** v1   CLI RX|Symbol not found: 'req.esi' (Only available when VCL syntax <= 4.0)
**** v1   CLI RX|At: ('/tmp/vtc.361713.6d37478f/inc.vcl' Line 3 Pos 21) -- (Pos 27)
**** v1   CLI RX|                set req.esi = false;
**** v1   CLI RX|--------------------#######---------
**** v1   CLI RX|
**** v1   CLI RX|Running VCC-compiler failed, exited with 2

```

dridi:
Despite the apparent compatibility in terms of versions between 4.0 and 4.1, we can have 4.0 syntax in a 4.1 VCL: this is essentially what the built-in VCL does by having a 4.0 version but being appended even to 4.1 code. However, this breaks down once the 4.0 code uses symbols that were removed from VCL 4.1 like `req.esi`.

It can be problematic with includes, as one may need to maintain two almost identical reusable VCL files when there's only a tiny difference.

It's in essence the problem described in #3368 and even though we decided that it would be simpler to make this an explicit syntax I found after addressing https://github.com/varnishcache/varnish-cache/pull/3613#issuecomment-845958677 that I had almost all I needed to make this possible.

I took the work I had started in February, rebuilt it on top of #3613 and solved the conflicts that had cropped up, and that resulted in the last 5 commits of this pull request. The rest comes from #3613 itself. The implementation makes the push/pop operations implicit.

I purposefully left one check out because of time constraints: checking that at the end of the parsing there is nothing left on the stack. See individual commits for more information.

Fixes #3368

### DIFF
--- a/bin/varnishtest/tests/r03368.vtc
+++ b/bin/varnishtest/tests/r03368.vtc
@@ -1,0 +1,27 @@
+varnishtest "include vcl 4.0 from vcl 4.1"
+
+shell {
+	cat >foo.vcl <<-EOF
+	vcl 4.0;
+
+	import std;
+
+	sub esi_status {
+		# this symbol is 4.0 only
+		std.log("ESI:" + req.esi);
+	}
+	EOF
+}
+
+varnish v1 -syntax 4.1 -errvcl {Variable is read only.} {
+	backend be none;
+
+	include "${tmpdir}/foo.vcl";
+
+	# should be in 4.1 syntax
+	sub vcl_recv {
+		call esi_status;
+		# this symbol is read only since 4.1
+		set req.proto = "HTTP/0.9";
+	}
+}
--- a/bin/varnishtest/tests/v00021.vtc
+++ b/bin/varnishtest/tests/v00021.vtc
@@ -168,6 +168,15 @@ varnish v1 -syntax 4.1 -errvcl {(Only available when VCL syntax <= 4.0)} {
 		set req.esi = false;
 	}
 }
+
+varnish v1 -syntax 4.1 -vcl {
+	backend be none;
+	vcl 4.0; # drop effective version for legacy symbol
+	sub vcl_recv {
+		set req.esi = false;
+	}
+}
+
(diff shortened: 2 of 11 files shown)


Dridi:
Rebased against master now that #3613 got merged to remove noise from the commit log.

Dridi:
Rebased against master to resolve conflicts from 9b1f1d170781ffa8a881b46455327ca9d7d5c514.

nigoroll (approved):

Dridi:
We'll try a different approach where we insert actual `vcl x.y;` tokens at the end of an include, if it contained a `vcl` statement. If that results in a syntax error, we can figure that those tokens don't come from a source at all and change the error message to reflect the limitation that this would introduce: an include that defines a VCL version MUST NOT end in the middle of a statement.

PR #3622, opened 2021-06-02
Support ACL as a vmod return type

nigoroll:
to the extent that the match operator `~` will actually accept vmod function/method calls returning it.

Because VMODs also need an invalid ACL value to signal error, we stop panicking for a NULL ACL and trigger a VCL failure instead.

The use case is not (yet) to generate dynamic ACLs, but rather to store and recall global ACL symbols.

This implementation was particularly simple thanks to Dridis work on #3555 and VCL_REGEX.

### DIFF
--- a/bin/varnishtest/tests/v00016.vtc
+++ b/bin/varnishtest/tests/v00016.vtc
@@ -89,7 +89,7 @@ varnish v1 -errvcl {resolves to too many addresses} {
 	}
 }
 
-varnish v1 -errvcl {Symbol 'foo' has wrong type (instance)} {
+varnish v1 -errvcl {Expression has type directors.shard, expected ACL} {
 	import directors;
 	backend b { .host = "${localhost}"; }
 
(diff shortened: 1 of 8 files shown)


bin/varnishtest/tests/v00016.vtc
```diff
@@ -103,7 +103,7 @@ varnish v1 -errvcl {Symbol 'foo' has wrong type (instance)} {
 	}
 }
 
-varnish v1 -syntax 4.0 -errvcl {Undefined acl foo} {
+varnish v1 -syntax 4.0 -errvcl {Expression has type directors.shard, expected ACL} {
```
nigoroll, line 106: Does anyone^Wany @Dridi see any issue with this change?
I understand that preserving support VCL 4.0 syntax is important to you, but do we _really_ need to keep the error messages identical? The complications would not seem justified to me.

nigoroll: pow wow agreed that error messages are not part of the API and may be subject to improvement

nigoroll:
It seems me doing the pugh to master before the rebase to this PRs branch has caused GH to show it as closed rather than merged. But it has been merged.

PR #3626, opened 2021-06-14
req: New hash_ignore_vary flag

Dridi:
It works like the hash_ignore_busy flag: it is disabled by default
and survives a restart, but as the name implies will skip vary checks
for object candidates.


nigoroll:
Can you motivate why `unset beresp.http.Vary` does not work?

nigoroll (approved):

Dridi:
Bugwash: OK but should eventually be documented in the "VCL design patterns" section.

Dridi:
33188acfa13f0abe5f025c63a6c181433888dd5e

PR #3627, opened 2021-06-15
RFC: varnishtest: New ${repeat:<uint>:<str>} macro

Dridi:
When we need long strings it can become quickly unwieldy, and a
dedicated macro would help when those strings follow a repeating
pattern.

This is incomplete work in progress to be discussed.


bsdphk:
I like this concept, not the naming: "repeat" sounds like a flow-control thing to me, not string production.

Maybe `${string:<uint>:<str>}` ?

or  `${string:<uint>*<str>}` ?

If we want to nail down a generic splitter-char for future ${...} use, we should pick one VAV understands:

`${string,<uint>,<string>}`

Dridi:
The problem with VAV is that it won't allow commas in `<string>`, unless we teach it to stop splitting argument passed a maximum number.

bsdphk:
Do we need commas in the strings ?

bsdphk:
Actually, if we use VAV, you can:  `${string,<uint>,"this is a string with commas and spaces"}`

Dridi:
I pushed 114062dea directly after reviewing it with @bsdphk and then reorganized macro parsing to treat the text inside the `${...}` delimiters as an argument vector.

With the current patch series (3ebcd6d11...5ada93e1c) the `${date}` macro is now registered in the main process, and the `${repeat,...}` macro is added similarly. All the error handling and documentation is missing on purpose, to see whether this is a direction we want for varnishtest macros.

The macro is still called `${repeat,...}` because I would like to sort out the underlying technical details before we look at the user interface.

bsdphk:
Looks good.

Considering how much this improves the VTC's now I dont understand why we didnt do this much earlier.

Thanks a lot!

Dridi:
So, regarding the name, I don't like `${string,<uint>,<str>}` but something like `${string,<action>,...}` would work for me:

- `${string,repeat,<uint>,<str>}`
- `${string,generate,<uint>}` (essentially what `-bodylen` does)
- etc

> Considering how much this improves the VTC's now I dont understand why we didnt do this much earlier.

All it took was me engaging in a pcre2 migration and getting frustrated by r1576 :)

bsdphk:
Take it away maestro!

nigoroll (approved):
FTR, I only checked casually

Dridi:
Ready for review.

Dridi:
We should probably use the `ARGV_NOESC` flag here.

Dridi:
bugwash: OK

PR #3629, opened 2021-06-17
vcc: Give the REGEX treatment to ACL expressions

### LINKED ISSUE #3628: if-statement with multiple ACL comp. fails with  '||' must be preceeded by BOOL, found ACL
The following test fails in commit https://github.com/varnishcache/varnish-cache/commit/badee4616bc83361c52e056bea7521b638f2b235 

The following commit seems to have broken it : 40ca3c2911c7c85d8025d5527c0739eea95b08b1

with :  `'||' must be preceeded by BOOL, found ACL. `

```
varnishtest logicoraclfail

server s1 -repeat 1 {
        rxreq
        txresp
} -start

varnish v1 -vcl+backend {
        acl testacl {
                "127.0.0.1";
        }

        sub vcl_recv {
                if(client.ip ~ testacl || client.ip ~ testacl){
                        return(synth(200, "Varnish should not fail?"));
                }
        }
} -start

```

Dridi:
It seems that the common point between both is that we technically start
a brand new expression (expr0) in the RHS of an ongoing expression
instead of using an RHS-specialized function (like expr4 or expr_add).

In practice, we do recursively start a brand new expression, so we need
the subsequent conversion work performed by expr0. Except that expr0
starts with a boolean OR parsing which we don't want here.

Instead of adding another special case to expr0, we can instead split
the conversion work to its own function and parse the RHS of a REGEX or
an ACL match expression as expr4+expr_convert.

Fixes #3628

### DIFF
--- a/bin/varnishtest/tests/v00021.vtc
+++ b/bin/varnishtest/tests/v00021.vtc
@@ -151,6 +151,15 @@ varnish v1 -errvcl {Symbols named 'vcl_*' are reserved.} {
 	}
 }
 
+# 3628
+varnish v1 -vcl {
+	backend be none;
+	acl foo { }
+	sub vcl_recv {
+		if (client.ip ~ foo || server.ip ~ foo) { }
+	}
+}
+
 varnish v1 -errvcl {Expression has type BOOL, expected ACL} {
 	sub vcl_recv {
 		if (client.ip ~ true) { }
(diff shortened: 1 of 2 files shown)


bsdphk:
This patch seems to be a very complicated way of writing:

    if(e2->fmt != {ACL|REGEX}) { issue targeted error message }

What am I missing ?

`vrt_convert()` is not ever going to be able to turn another type into an ACL or a REGEX, is it ?   So all calling it does is testing we have what we want and if not emitting the generic "Expression has type %s, expected %s" error ?

Dridi:
The original patch from https://github.com/varnishcache/varnish-cache/issues/3628#issuecomment-862697174 is a two-liner.

But I came to the conclusion that  using vcc_expr0 for this case is wrong, but we need the followup work from vcc_expr0. So instead, I split that conversion work out of vcc_expr0 so we can use the more appropriate vcc_expr4 entry point to parse the RHS of regex and acl matching.

bsdphk:
Much better: Go for it!

PR #3630, opened 2021-06-19
VRE API revamp

Dridi:
See https://github.com/varnishcache/varnish-cache/pull/3616#issuecomment-864041455.

This revision of the VRE API allows the VRE facade to be the only place in the code where we actually touch pcre and offers a better abstraction for our usage. VMODs interested in more pcre features can as usual use it directly.

We can then perform the pcre2 migration (#3559) in a single `vre.c` file.

### DIFF
--- a/bin/varnishd/Makefile.am
+++ b/bin/varnishd/Makefile.am
@@ -159,7 +159,6 @@ nobase_pkginclude_HEADERS = \
 vcldir=$(datarootdir)/$(PACKAGE)/vcl
 
 varnishd_CFLAGS = \
-	@PCRE_CFLAGS@ \
 	@SAN_CFLAGS@ \
 	-DNOT_IN_A_VMOD \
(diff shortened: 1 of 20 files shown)


Dridi:
Force-pushed to remove facd84797cbbe455e159a94e11c34202302d8b7d after realizing that VRT provides its own interface on top of VRE...

bsdphk:
This is pretty far from my memory, but I took a read and the only two things which stood out for me was that I would prefer if `VRE_error()` and `VRE_sub()` worked on VSB's instead.

We have `WS_VSB_New()` now so the extra overhead is very low, and I expect it might simplify the code in `VRE_sub()` if we are lucky.

Dridi:
Rebased against master using VSBs as requested.

Dridi:
For a pcre2 migration on top of this pull request, see #3635.

Dridi:
bugwash: OK

PR #3631, opened 2021-06-21
[ncsa] -j will escape control characters

gquintard:
open question: does the trailing character count as part of the message, and should it be printed too?

### DIFF
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -572,7 +572,7 @@ addf_vsl(enum VSL_tag_e tag, long i, const char *prefix)
 
 	ALLOC_OBJ(w, VSL_WATCH_MAGIC);
 	AN(w);
-	if (VSL_tagflags[tag])
+	if (VSL_tagflags[tag] && CTX.quote_how != VSB_QUOTE_JSON)
 		VUT_Error(vut, 1, "Tag %s can contain control characters",
 		    VSL_tags[tag]);
 	w->tag = tag;
@@ -921,7 +921,6 @@ process_vsl(const struct vsl_watch_head *head, enum VSL_tag_e tag,
 {
 	struct vsl_watch *w;
 	const char *p;
-
 	VTAILQ_FOREACH(w, head, list) {
 		CHECK_OBJ_NOTNULL(w, VSL_WATCH_MAGIC);
 		if (tag != w->tag)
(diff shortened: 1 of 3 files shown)


bin/varnishncsa/varnishncsa.c
```diff
@@ -966,7 +965,8 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 		skip = 0;
 		while (skip == 0 && 1 == VSL_Next(t->c)) {
 			tag = VSL_TAG(t->c->rec.ptr);
-			if (VSL_tagflags[tag])
+			if (VSL_tagflags[tag] &&
```
nigoroll, line 968: don't we need to check the flags?
I would think this lead to filtered tags being printed also.

nigoroll: have not checked though

nigoroll:
my 2¢: I think it makes sense to either print the null or not emit it in the first place, but not printing it if contained in a binary record makes no sense.
Bugwash seems to agree on this

bin/varnishncsa/varnishncsa.c
gquintard: "filtered" as in "masked"? If so, we are good as the masked records are just not pushed into VSL.

In terms of tags, we have `UNSAFE` and `BINARY` which should be escaped, and `UNUSED` which is only useful in `vsl2rst.c`, so I didn't feel the need to be too granular, but I can change it if you prefer

gquintard:
added a commit to avoid truncating the record if there are flags

gquintard:
we good? ;-)

doc/changes.rst
```diff
@@ -51,6 +51,8 @@ Varnish Cache 7.x.x (2021-09-15)
 
 * ACLs default to `pedantic` which is now a per-ACL feature flag.
 
+* `varnishncsa -j` won't refuse to print fields with control characters anymore.
+
```
bsdphk, line 55: I really hate this kind of double negatives in documentation...

gquintard: I don't see nothing wrong with that, but I'll change that and merge

gquintard:
squashed and rebased, I'll merge early next week unless someone yells

Dridi (approved):

nigoroll (approved):

PR #3637, opened 2021-06-29
RFC: Parse strict standard bytes units

Dridi:
There is an overlap between BYTES and DURATION units, `'m'` can mean either minutes or mebibytes. This patch series contains 2 breaking changes and 1 improvement:

- 9b238bbb removes dead weight
- b0bbf32f0 adopts case-sensitive bytes units (B kB KiB MB MiB GB ...)
  - please note that duration units are case-sensitive
  - that's the visible breaking change, hence the number of files touched
- b3334596f cuts some loose ends

Of course, this is quite the intrusive change, trading some convenience for better semantics. Since there are two competing standards for bytes units, this can be considered to loosely belong to #3246.

Obviously, this is going to break both the CLI and VCL syntax, but since we are heading towards 7.0 the former is in theory acceptable.

Regarding a breakage of the VCL syntax, we have several other changes that could justify a major VCL bump and the retirement of the 4.x versions:

- anything affecting VCL behavior from #3246
- using pcre2's official replacement string syntax for the substitution case (see 1b3d51a4c256904e227d786aef94b46c0bc3be00 and 29eed4b4d26484b39f7a7cf3d8e9a87558d05e4c from #3616)
  - this means less code for us to maintain and access to more capabilities
- `sub vcl_begin` and friends

So maybe not something for the September release, but maybe the next?


bin/varnishd/mgt/mgt_param_tweak.c
```diff
@@ -239,17 +239,17 @@ fmt_bytes(struct vsb *vsb, uintmax_t t)
 	const char *p;
 
 	if (t == 0 || t & 0xff) {
-		VSB_printf(vsb, "%jub", t);
+		VSB_printf(vsb, "%juB", t);
 		return;
 	}
-	for (p = "kMGTPEZY"; *p; p++) {
```
nigoroll, line 245: Do we really want to remove exa/zeta/yotta?
With #2929 this is probably right, but it still triggers my #3723 itch

nigoroll:
_if_ we want to tackle this, should we also address a unit for (m)onths?

Dridi:
bugwash: let's not break VCL.

Issue #3379, opened 2020-08-07
header names can not be numbers


nigoroll:
fallout from something unrelated:

```
**** v1   CLI RX|Message from VCC-compiler:
**** v1   CLI RX|Symbol 'resp.http' has wrong type (none): 
**** v1   CLI RX|At: ('<vcl.inline>' Line 16 Pos 21) -- (Pos 29)
**** v1   CLI RX|                set resp.http.123 = "abc";
**** v1   CLI RX|--------------------#########-------------
```

bsdphk:
I'm inclined to invoke the inverse-postel principle and just say "too bad"

Dridi:
This is the fallout of #2573 that resulted in header names following the rules of VCL symbol names, without a syntax for otherwise valid HTTP header names that aren't valid for symbols.

My proposal has always been `resp.http[123]` because `[` and `]` are HTTP separators and not valid for header names, and this syntax might be familiar to other programming languages that support associative arrays.

nigoroll:
also kind-of duplicate to #3103 as pointed out by @Dridi

nigoroll:
has been discussed again, vmods can do the job, I will docfix and then close

PR #3638, opened 2021-06-30
vcl: Allow header names to be quoted (compliance improvements)

Dridi:
Because we funnel HTTP header names through the symbol table they have
to be valid VCL identifiers. It means that we can't support all valid
header names, which are tokens in the HTTP grammar. To finally close this
loophole without the help of a VMOD we allow header names to be quoted:

    req.http.regular-header
    req.http."quoted.header"

However we don't want to allow any component of a symbol to be quoted:

    req."http".we-dont-want-this

So we teach the symbol table that wildcard symbols may be quoted. There
used to be several use cases for wildcards but it is now limited to HTTP
headers.

Refs #3246
Refs #3379

### DIFF
--- a/lib/libvcc/vcc_symb.c
+++ b/lib/libvcc/vcc_symb.c
@@ -141,21 +141,24 @@ vcc_symtab_new(const char *name)
 }
 
 static struct symtab *
-vcc_symtab_str(struct symtab *st, const char *b, const char *e)
+vcc_symtab_str(struct symtab *st, const char *b, const char *e, unsigned tok)
 {
 	struct symtab *st2, *st3;
 	size_t l;
 	int i;
 	const char *q;
 
+	assert(tok == ID || tok == CSTR);
 	if (e == NULL)
 		e = strchr(b, '\0');
+	q = e;
 
 	while (b < e) {
-		for (q = b; q < e && *q != '.'; q++)
-			continue;
-		AN(q);
-		l = q - b;
+		if (tok == ID) {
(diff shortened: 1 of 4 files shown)


Dridi:
bugwash: OK, but add "don't do this" red tape in the manual.

Dridi:
Rebased against master with a "don't do this" note in the documentation.

lib/libvcc/vcc_symb.c
```diff
@@ -297,7 +306,9 @@ VCC_SymbolGet(struct vcc *tl, vcc_ns_t ns, vcc_kind_t kind,
 		if (tn1->tok != '.')
 			break;
 		tn1 = vcc_PeekTokenFrom(tl, tn1);
-		if (tn1->tok != ID)
+		if (tn1->tok == CSTR && sym->wildcard == NULL)
```
bsdphk, line 309: Flexelint complains that `sym` is possibly NULL here.  Add an assert to make sure.

bsdphk (commented):
If you try to use `req.http."foobar"` from `vcl 4.0` you get a /really/ confusing error message.

Maybe we should make a generic facility in VCC for saying "this feature is only available from vcl X.Y and up" ?

Dridi:
Rebased against current master, with 4.0 test coverage and the missing null check (it shouldn't be an assertion).

> Maybe we should make a generic facility in VCC for saying "this feature is only available from vcl X.Y and up" ?

I'm leaving this outside of the scope of this change but I generally like the idea.

PR #3639, opened 2021-07-01
circleci: Re-enable Alpine latest

xordspar0:
Alpine 3.14 includes a newer version of musl that uses the newer
faccessat2 syscall, which is not yet allowlisted by the seccomp filter
in older Docker versions. Docker 20.10.0+ allows this new syscall.

In this patch, we configure Circle to set up a Docker 20.10+ environment
where we can run Alpine 3.14 builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -169,15 +169,24 @@ jobs:
         default: ""
         type: string
     docker:
-      - image: << parameters.dist >>:<< parameters.release >>
+      - image: centos:7
     working_directory: /workspace
     steps:
+      - setup_remote_docker:
+          version: 20.10.6
       - run:
-          name: Possible << parameters.dist >>:<< parameters.release >> extra repos
+          name: Install docker
+          command: yum install -y docker
+      - checkout
+      - run:
+          name: Extract and distcheck
           command: |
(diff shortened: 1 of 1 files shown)


gquintard:
@xordspar0 thanks a bunch, that has been on my plate for some time and I was unable to find time for it. The change looks sane and while I'm not a fan of the big `bash -c` block, it works.

xordspar0:
> I'm not a fan of the big bash -c block

Yeah, same. An alternative would be to move it into  the Dockerfile.{alpine,centos,debian,*} files, but that is a bigger diff and would break apart the `if` blocks (maybe good) and share less code (maybe bad). Another option would be to move the script to a dedicated distcheck.sh script. That would be functionally the same as it is now but more organized maybe.

Dridi:
It's not so much that `faccessat2` was blocked, and more like it was not allowed in the default profile. The syscall was added to Linux 5.8 which is arguably fairly recent, so maybe it was not explicitly allowed on account of not existing. It doesn't look like circleci has a configuration facility to add system calls to the seccomp profile...

I'll let @gquintard decide how to proceed.

xordspar0:
That's correct. I updated the commit message to make that more clear.

gquintard:
I'm happy that it works, and don't feel like nitpicking on a Friday, this has been merged

PR #3640, opened 2021-07-05
vav: Tighten arguments parsing

Dridi:
Commas unlike spaces are hard separators, so a trailing comma leads to a
last empty parameter.

A comma may appear after an argument's "trailing" spaces and should not
result in an additional parameter. In <foo  , bar> we should expect two
fields, not three.

Comments are only treated as such at arguments boundaries: <foo #bar>
parses one field <foo> and <foo#bar> parses one field <foo#bar>, taking
the shell word splitting as the model, cementing what was the existing
VAV behavior in the first place.

Unlike the shell, we don't expect quotes to start in the middle of a
token so <foo"bar> is invalid unless escaping was disabled.

Fields that are quoted need a separator: <"foo""bar"> is therefore
invalid unless escaping was disabled.

### DIFF
--- a/lib/libvarnish/vav.c
+++ b/lib/libvarnish/vav.c
@@ -52,12 +52,20 @@
 #include "vas.h"
 #include "vav.h"
 
-int
-VAV_BackSlash(const char *s, char *res)
+static int
+vav_backslash_txt(const char *s, const char *e, char *res)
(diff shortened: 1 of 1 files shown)


Dridi:
bugwash: OK

PR #3642, opened 2021-07-10
kevent netbsd build fix proposal.

devnexen:
netbsd still uses the older kevent api flavor.


bsdphk:
I fixed this in a more direct way with less #ifdef polution.

Please let me know if this does not work for you.

PR #3643, opened 2021-07-12
doc: Fix formatting in some examples

xordspar0:
In RST, there needs to be an empty line after ::, otherwise the whole
block is interpreted as a normal paragraph where newlines are not
preserved.


gquintard:
thanks @xordspar0 !

Issue #3600, opened 2021-04-23
More general fix to 3546 required


nigoroll:
taking note of an issue which was almost forgotten: #3546 is just a special case of a broader issue: We should either keep object references until the end of the task or implement some other solution like copying to workspace.
Options need to be pondered.

nigoroll:
I want to prepare a writeup of the broader topic

bsdphk:
This is going to be long-ish.

First, there are two different issues:

A) Call argument lifetimes.

B) Return value lifetimes.

Second, our viewfinder here should encompas both VCC calling VMODs, VMODs calling VRT and VCC calling VRT, and everybody calling object methods:  We want one unified architecture, not three different architectures.

Call Arguments
--------------

If we give call arguments "task lifetime", we have thereby decided that arguments to `VRT_SetHdr()` cannot live on the stack.  That is a total non-starter to me, as matter of programmer sanity.

Call arguments will only be guaranteed to live for the duration of the call, and if the callee wants to stash a copy, it needs to perform whatever song&dance we settle on, to make that copy stable for its own task lifetime.

(Apart from programming languages which premptively copy all function arguments, none of which are performant, this is how all programming languages I know do things.)

I can only see that lead to two viable architectures:

I) Refcounting, where caller is responsible for holding a ref on the argument for (at least) the duration of the call, and the callee is responsible for gaining a ref, if it wants to use the argument longer than that.  This restricts what the caller can subsequently do to `refcount!=1` objects.

II) "Ask the owner", where all the relevant objects have a method, tentatively called `.copy()` where they will either return the object, if its lifetime is sufficient for the present task, or return a object copied to stable storage for the present task.

Under II) the object owner can use the .copy() function to implement refcounting, but only if Somebody™ also calls another method (`.endcopy()` ?) when the task ends.

The advantage of II) over I) is that we do not waste time refcounting object which, ipso-facto, have a guaranteed sufficient lifetime, for instance backends with VCL lifetime.

Return Values
-------------

Return values are much tricker than call arguments, because there is no equally obvious lifetime.

As I see it three viable options exist:

i) Refcounting.  The return value comes with a refcount which must be released by the caller, this means that VMODs calling a VRT are responsible for deref'ing the return value and that VCC will need to learn to do so as well.

ii) Demand the return value lives for the duration of the current task.  All copies to a different task causes a full copy to happen.

[...]

Issue #3550, opened 2021-03-12
resp vs. rollback


nigoroll:
for later, ENOTIME:

https://github.com/varnishcache/varnish-cache/commit/86fc6706ea0277cd6949ab102f958a17a3d83d93 made me realize that for a rollback we also need to re-init `resp.http`, as it could have been modified to contain non-obj data.
when getting to it, also re-check `reresp` in that regard

PR #3644, opened 2021-07-13
ws_emu: Introduce the workspace emulator

Dridi:
The goal of the workspace emulator is to replicate the regular workspace behavior with individual allocations and make it work transparently.

It's the successor of the workspace sanitizer from #3320 with notable differences:

- enabled at configure time instead of run time
- in a separate source file instead of mixed in
- using sparse allocations instead of built-in red zones

This means that the workspace emulator can be combined with regular sanitizer, in particular asan and lsan. If available, asan's public interface is used to mitigate the possible overflow of a reservation after some of it was released.

Even without sanitizers, the fact that we integrate with jemalloc by default and enable its abort and junk options in varnishtest is enough to detect a use-after-free in some cases.

With sanitizers though, the workspace emulator can observe #3550.

One drawback is that the logic is split in two files, and some functions are identical in the two files. It might be possible to split `cache_ws.c` into something like `cache_ws_alloc.c` and `cache_ws_util.c` for example.

Closes #3320
Refs #3550
Refs #3600

---

All the changes leading up to the workspace emulator were either agreed upon already or necessary for the system to work.


Dridi:
It just occurred to me that the workspace sanitizer pull request has a test
case overflowing a write on purpose with the help of vmod_debug. I don't
remember doing it in this pull request.

TODO for later.

(from my phone)

Dridi:
All the original commits except the ones touching pipelining or the emulator were approved and merged already: 40471b963...5d7511cf2

bsdphk:
I have stared at the pipeline stuff, and I see no obvious flaws in it.

OK with me.

Dridi:
> I have stared at the pipeline stuff, and I see no obvious flaws in it.

I did find a not obvious flaw.

I have not understood yet why the pipeline stuff breaks r02679.vtc for a 32bit build.

Dridi:
Rebased against master, it should fail the new shiny 32bit job in circleci.

PR #3647, opened 2021-07-14
Docfix 3634

hermunn:
This patch is a bit wordy, but since the `NULL` string is not explained well, I think it is justified. Comments welcome.

### DIFF
--- a/vmod/vmod_cookie.vcc
+++ b/vmod/vmod_cookie.vcc
@@ -150,8 +150,7 @@ Example::
 
 $Function STRING get(PRIV_TASK, STRING cookiename)
 
-Get the value of ``cookiename``, as stored in internal vmod storage. If
-``cookiename`` does not exist an empty string is returned.
+Get the value of ``cookiename``, as stored in internal vmod storage.
 
 Example::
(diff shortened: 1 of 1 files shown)


vmod/vmod_cookie.vcc
```diff
@@ -161,6 +160,38 @@ Example::
...
+		cookie.parse(req.http.Cookie);
+		if (cookie.isset("a_cookie")) {
+			set req.http.X-tmp = cookie.get("a_cookie");
+			# do something with the X-tmp header...
+		}
+	}
+	# if necessary, do something when a_cookie is not there
+	if (!req.http.X-tmp) {
+		# fallback case
+	}
```
Dridi, line 193: I would avoid nesting here:

```vcl
if (req.http.Cookie) {
	cookie.parse(req.http.Cookie);
}

if (cookie.isset("a_cookie")) {
	# do something with cookie.get("a_cookie")
} else {
	# fallback case
}
```

Dridi (approved):

bsdphk:
OK with me.

PR #3648, opened 2021-07-14
Increase vsl_buffer and workspaces

hermunn:
On many busy production systems, the VSL buffer for transactions often
fills up, triggering a flush to the VSM. If such a transaction takes a
long time, it can happen that the flushed data is overwritten before
the whole transaction completes. The result is that these transactions
are missed by varnishncsa and other tools. Increasing the vsl_buffer
does the trick, at some cost in workspace usage.

The cost implies that one should increase the workspaces, too. Since
the old defaults of 64k each are too low for many common use cases
today, it is natural to increse everything in one patch.

### DIFF
--- a/bin/varnishtest/tests/r01038.vtc
+++ b/bin/varnishtest/tests/r01038.vtc
@@ -45,7 +45,7 @@ server s1 {
 	txresp -body "foo8"
 } -start
 
-varnish v1 -arg "-p workspace_backend=10k" -vcl+backend {
+varnish v1 -arg "-p workspace_backend=10k -p vsl_buffer=4k" -vcl+backend {
 	sub vcl_backend_response {
 		if (bereq.url == "/") {
 			set beresp.do_esi = true;
--- a/bin/varnishtest/tests/r01120.vtc
+++ b/bin/varnishtest/tests/r01120.vtc
@@ -7,7 +7,7 @@ server s1 {
 	txresp -hdr "Vary: Foo" -body "yyyyy"
 } -start
 
-varnish v1 \
+varnish v1 -arg "-p vsl_buffer=4k" \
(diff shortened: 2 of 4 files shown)


hermunn:
It seems tests are failing. I will investigate and update this PR.

Dridi:
A bit more context maybe, we often see backend applications with rather long URLs, so we also often see `vsl_reclen` bumped to something like 4kB. At this point, just receiving a regular client request can mean one flush for the `req.url` alone and one more for `req.http.cookie`, before we even call `vcl_recv`.

That kind of early flushes of course shifts a lot of work towards VSL clients with `vxid` or more grouping, which defeats the purpose of `vsl_buffer`.

Speaking of `vsl_reclen`, we should bump its maximum value to 8kB to be able to fully log `req.url` based on the RECOMMENDED limit:

https://datatracker.ietf.org/doc/html/rfc7230#section-3.1.1

> Various ad hoc limitations on request-line length are found in
> practice.  It is RECOMMENDED that all HTTP senders and recipients
> support, at a minimum, request-line lengths of 8000 octets.

Dridi:
We don't need to touch `vsl_reclen`, I convinced myself that its maximum was around 4kB but it's actually a dynamic maximum of `vsl_buffer - 12 bytes` which sounds about right considering the current  default value for `vsl_buffer`....

hermunn:
I have updated the test cases and the commit message. But should I add some of @Dridi 's explanation about grouping in VSL clients?

Dridi:
> But should I add some of @Dridi 's explanation about grouping in VSL clients?

I primarily meant to provide context to reviewers, no opinion on if or how this should be documented.

Dridi (approved):

nigoroll (approved):

PR #3650, opened 2021-07-16
HTTP case sensitivity hardening (compliance improvements)

Dridi:
This change is in two parts: string comparison macros for the cache process and a general sweep through the code base for other places dealing with HTTP syntax.

The macros are meant to convey and abstract the comparison of components of the HTTP grammar. They are also used as anchors to reference what the RFCs have to say about case sensitivity for comparisons. In other words, avoiding the direct use of `strcmp()`-type functions.

This should hopefully signal that we aren't merely comparing strings, but specific kinds of strings, and prevent regressions in that area.

Some parts I left alone on purpose, sometimes adding XXX comments for reviewers are h2 and VTIM. I wanted to see how this first batch would be received and didn't want to limit myself to a mere sweep through the code like I did for example in varnishtest (where for example that level of detail didn't seem critical enough).

As usual see individual commits.

Refs #3246

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -53,7 +53,7 @@ sub vcl_req_host {
 	}
 	if (!req.http.host &&
 	    req.esi_level == 0 &&
-	    req.proto ~ "^(?i)HTTP/1.1") {
+	    req.proto == "HTTP/1.1") {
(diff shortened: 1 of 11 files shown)


Dridi:
bugwash: add a locale-free set of `vct_str*cmp()` functions, and revisit somehow the `http_*()` macros for comparisons.

PR #3656, opened 2021-08-02
vcl: Change the order of sess.* variables

### LINKED ISSUE #3564: 6.0: Unknown variable: 'sess.timeout_idle' (Only available when VCL syntax)
This does not happen in trunk, but it happens in the 6.0 LTS branch currently.

It's not possible to use `sess.timeout_idle` with the 4.0 syntax, despite its lack of such restriction. I suspect that the symbol `sess.xid` declared just before and restricted to 4.1 syntax might "leak" this requirement via the intermediate `sess` symbol.

I haven't researched this, so this is purely speculative.

Dridi:
This is just the order of their declaration in the VCL manual.

As a side effect it works around a bug where the sess.xid syntax
requirements would prevent sess.timeout_idle to be used in VCL 4.0,
which is less intrusive than a proper fix.

The bug was fixed in trunk without being noticed in the first place
after many heavy changes to libvcc. For a stable branch this is less
risky than a back-port since there are only two sess.* symbols.

Fixes #3564

### DIFF
--- a/bin/varnishtest/tests/r03564.vtc
+++ b/bin/varnishtest/tests/r03564.vtc
@@ -1,0 +1,31 @@
+varnishtest "sess.* symbols and vcl syntax"
+
+varnish v1 -vcl {
+	backend be none;
+
+	sub vcl_deliver {
+		set resp.http.sess-xid = sess.xid;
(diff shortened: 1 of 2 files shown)


mbgrydeland (approved):
LGTM

PR #3657, opened 2021-08-02
How would we handle the pcre2_code typedef in VRE_unpack()?

nigoroll:
Any better ideas? @Dridi


nigoroll:
Here is how vmod_re looks like with direct pcre2 access:
https://code.uplex.de/uplex-varnish/libvmod-re/tree/vre_pcre2_unpack

see also https://github.com/varnishcache/varnish-cache/pull/3655#issue-701422026 for discussion

Dridi (approved):

PR #3658, opened 2021-08-02
Use pcre2_match_context

nigoroll:
@Dridi IIUC you once mentioned that the limits vtc would never report limits being hit, could this be the reason?

### DIFF
--- a/lib/libvarnish/vre.c
+++ b/lib/libvarnish/vre.c
@@ -201,7 +201,7 @@ vre_match(const vre_t *code, const char *subject, size_t length, size_t offset,
 	}
 
 	matches =  pcre2_match(re, (PCRE2_SPTR)subject, length, offset,
-	    options, data, NULL);
+	    options, data, code->re_ctx);
 
 	if (datap != NULL && matches > VRE_ERROR_NOMATCH)
 		*datap = data;


Dridi:
Good catch, with this we can also make another change I forgot during the migration:

```diff
--- bin/varnishd/cache/cache_vrt_re.c
+++ bin/varnishd/cache/cache_vrt_re.c
@@ -64,6 +64,8 @@ VPI_re_fini(vre_t *rep)
 VCL_BOOL
 VRT_re_match(VRT_CTX, const char *s, VCL_REGEX re)
 {
+	struct vsb vsb[1];
+	char errbuf[VRE_ERROR_LEN];
 	int i;
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
@@ -73,8 +75,13 @@ VRT_re_match(VRT_CTX, const char *s, VCL_REGEX re)
 	i = VRE_match(re, s, 0, 0, &cache_param->vre_limits);
 	if (i >= 0)
 		return (1);
-	if (i < VRE_ERROR_NOMATCH )
-		VRT_fail(ctx, "Regexp matching returned %d", i);
+	if (i < VRE_ERROR_NOMATCH ) {
+		VSB_init(vsb, errbuf, sizeof errbuf);
+		AZ(VRE_error(vsb, i));
+		AZ(VSB_finish(vsb));
+		VSB_fini(vsb);
+		VRT_fail(ctx, "Regexp matching failed: %s", errbuf);
+	}
 	return (0);
 }
```

And then we can revive r1576 with a slight tweak:

```diff
--- bin/varnishtest/tests/r01576.vtc
+++ bin/varnishtest/tests/r01576.vtc
@@ -40,7 +38,7 @@ client c1 {
 # PCRE_ERROR_RECURSIONLIMIT  (-21)
 # PCRE_ERROR_JIT_STACKLIMIT  (-27)
 logexpect l1 -v v1 {
-	expect * * VCL_Error "Regexp matching returned -2[71]"
+	expect * * VCL_Error "matching depth limit exceeded"
 } -start
 
 # This should fail with default params and JIT/no-JIT
```

Of course r1576 should be further "migrated" to pcre2 and take advantage of #3627, and possibly match the error message either limit instead of the one I ran into.

PR #3659, opened 2021-08-03
vre: New VRE_capture() function

Dridi:
It populates an array of txt provided by the caller with the groups that were captured on a successful match, and returns the number of such groups or a VRE error. Because the underlying `vre_capture()` function is covered by `VRE_sub()` we can probably spare ourselves `VRE_capture()` coverage via vmod_debug.

Doing so, this patch series makes `txt` available more widely in the code base.

Alternative to #3655.

### DIFF
--- a/lib/libvarnish/vre.c
+++ b/lib/libvarnish/vre.c
@@ -245,6 +245,29 @@ VRE_match(const vre_t *code, const char *subject, size_t length,
 }
 
 int
+VRE_capture(vre_t *code, const char *subject, size_t length, int options,
+    txt *groups, size_t count, const volatile struct vre_limits *lim)
+{
+	int i;
+
+	CHECK_OBJ_NOTNULL(code, VRE_MAGIC);
+	AN(subject);
+	AZ(options & (~VRE_MASK_MATCH));
+	AN(groups);
+	AN(count);
+
+	if (length == 0)
+		length = PCRE2_ZERO_TERMINATED;
+	vre_limit(code, lim);
+	i = vre_capture(code, subject, length, 0, options,
+	    groups, &count, NULL);
+
+	if (i <= 0)
+		return (i);
+	return (count);
+}
+
+int
 VRE_sub(const vre_t *code, const char *subject, const char *replacement,
     struct vsb *vsb, const volatile struct vre_limits *lim, int all)
 {
--- a/lib/libvarnish/vre.c
+++ b/lib/libvarnish/vre.c
@@ -249,12 +249,11 @@ VRE_sub(const vre_t *code, const char *subject, const char *replacement,
     struct vsb *vsb, const volatile struct vre_limits *lim, int all)
 {
 	pcre2_match_data *data = NULL;
-	PCRE2_SIZE *ovector;
-	uint32_t nov;
-	int i, l;
+	txt groups[10];
+	size_t count;
+	int i, offset = 0;
 	const char *s;
 	unsigned x;
-	int offset = 0;
 
 	CHECK_OBJ_NOTNULL(code, VRE_MAGIC);
 	CHECK_OBJ_NOTNULL(vsb, VSB_MAGIC);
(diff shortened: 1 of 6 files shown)


lib/libvarnish/vre.c
```diff
@@ -200,9 +203,24 @@ vre_match(const vre_t *code, const char *subject, size_t length, size_t offset,
...
 
-	matches =  pcre2_match(re, (PCRE2_SPTR)subject, length, offset,
+	matches = pcre2_match(re, (PCRE2_SPTR)subject, length, offset,
 	    options, data, code->re_ctx);
 
+	if (groups != NULL) {
+		AN(count);
+		AN(*count);
+		ovector = pcre2_get_ovector_pointer(data);
+		nov = pcre2_get_ovector_count(data);
```
nigoroll, line 213: I am aware that I suggested using `pcre2_get_ovector_count()`, but later realized that we should also be able to use `matches`.

lib/libvarnish/vre.c
```diff
@@ -200,9 +203,24 @@ vre_match(const vre_t *code, const char *subject, size_t length, size_t offset,
 		AN(data);
 	}
 
-	matches =  pcre2_match(re, (PCRE2_SPTR)subject, length, offset,
+	matches = pcre2_match(re, (PCRE2_SPTR)subject, length, offset,
 	    options, data, code->re_ctx);
```
nigoroll, line 208: I think we should check for match errors here

lib/libvarnish/vre.c
```diff
@@ -200,9 +203,24 @@ vre_match(const vre_t *code, const char *subject, size_t length, size_t offset,
...
+		AN(*count);
+		ovector = pcre2_get_ovector_pointer(data);
+		nov = pcre2_get_ovector_count(data);
+		if (nov > *count)
+			nov = *count;
+		for (g = 0; g < nov; g++) {
+			groups->b = subject + ovector[2 * g];
+			groups->e = subject + ovector[2 * g + 1];
+			groups++;
+		}
```
nigoroll, line 220: Should we `NULL` remaining groups (for `nov < *count`) or leave that to the caller?

lib/libvarnish/vre.c
```diff
@@ -222,40 +240,66 @@ VRE_match(const vre_t *code, const char *subject, size_t length,
...
+
+int
+VRE_capture(vre_t *code, const char *subject, size_t length, int options,
+    txt *groups, size_t count, const volatile struct vre_limits *lim)
+{
+	int i;
+
+	CHECK_OBJ_NOTNULL(code, VRE_MAGIC);
+	AN(subject);
+	AZ(options & (~VRE_MASK_MATCH));
```
nigoroll, line 255: Not sure if we should really ban all options.

lib/libvarnish/vre.c
```diff
@@ -222,40 +240,66 @@ VRE_match(const vre_t *code, const char *subject, size_t length,
...
+	AN(groups);
+	AN(count);
```
nigoroll, line 257: for ease of use, do we really need to enforce groups and count?
and, um, yes, I really still think we could just have one _cough_ exec _cough_ function for both.

nigoroll (commented):
I agree that the `txt` interface is more varnishy, other than that, in my mind, this still is `VRE_exec()`, basically. But, again, if it makes everybody else more happy, I am fine with it :)

lib/libvarnish/vre.c
Dridi: That code was originally from `VRE_sub()`, and while they should be equivalent/off by one (because we create the match data based on the pattern) I decided not to for clarity: we see a clear relationship between `ovector` and `nov`.

lib/libvarnish/vre.c
Dridi: I wouldn't bother, but that may be my laziness speaking on my behalf.

lib/libvarnish/vre.c
Dridi: You were on board when I added option smuggling prevention measures, to enforce what `vre.h` tries to convey: we define the options we want to support.

Otherwise #3657 gives you the escape hatch.

lib/libvarnish/vre.c
Dridi: Agreed.

lib/libvarnish/vre.c
Dridi: I'd rather cross the exec bridge. We don't need the ovector/txt anywhere in the code base, except in one call site for our regsub implementation. Since we end up with a very different signature if we move to txt, we might as well acknowledge that this is a different function altogether. Even though, I agree, it fulfills the same need.

lib/libvarnish/vre.c
nigoroll: ack

lib/libvarnish/vre.c
nigoroll: I am on board, and I agree that `VRE_MASK_MATCH` can be adjusted.

lib/libvarnish/vre.c
```diff
@@ -183,11 +183,14 @@ VRE_export(const vre_t *code, size_t *sz)
 }
 
 static int
-vre_match(const vre_t *code, const char *subject, size_t length, size_t offset,
-    int options, pcre2_match_data **datap)
+vre_capture(const vre_t *code, const char *subject, size_t length,
+    size_t offset, int options, txt *groups, size_t *count,
+    pcre2_match_data **datap)
```
bsdphk, line 188: FlexeLint tells me that `code` could be const ?

bsdphk (approved):

nigoroll (approved):

PR #3661, opened 2021-08-04
Master h2 rxbuf padding

mbgrydeland:
This PR solves two H/2 related bugs:

* Adds a buffer for incoming data frames between the session handling thread and the individual H/2 stream threads. This avoids having the session handling thread block while waiting for a stream thread to come around and request the incoming data, which would block most progress on the H/2 session.
* Implements handling of padding bytes in incoming data frames. Previously we would ingore their presence and treat all the bytes as data bytes.

The patch set also improves on panic output for H/2.

### DIFF
--- a/bin/varnishd/http2/cache_http2_panic.c
+++ b/bin/varnishd/http2/cache_http2_panic.c
@@ -88,6 +88,15 @@ h2_sess_panic(struct vsb *vsb, const struct sess *sp)
 		VSB_printf(vsb, "t_window = %jd, r_window = %jd,\n",
 		    r2->t_window, r2->r_window);
 
+		if (!PAN_dump_struct(vsb, r2->rxbuf, H2_RXBUF_MAGIC, "rxbuf")) {
+			VSB_printf(vsb, "stvbuf = %p,\n", r2->rxbuf->stvbuf);
+			VSB_printf(vsb,
+			    "{size, tail, head} = {%u, %ju, %ju},\n",
+			    r2->rxbuf->size, r2->rxbuf->tail, r2->rxbuf->head);
+			VSB_indent(vsb, -2);
+			VSB_cat(vsb, "},\n");
+		}
+
 		VSB_indent(vsb, -2);
 		VSB_cat(vsb, "},\n");
 	}
--- a/bin/varnishd/http2/cache_http2_panic.c
+++ b/bin/varnishd/http2/cache_http2_panic.c
@@ -36,6 +36,15 @@
 #include "cache/cache_transport.h"
 #include "http2/cache_http2.h"
 
+static const char *
+h2_panic_error(const struct h2_error_s *e)
+{
+	if (e == NULL)
+		return ("(null)");
+	else
+		return (e->name);
+}
+
 void
 h2_sess_panic(struct vsb *vsb, const struct sess *sp)
 {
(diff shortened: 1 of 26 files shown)


bsdphk:
What is the math behind not just using malloc buffers ?

Dridi:
> What is the math behind not just using malloc buffers ?

My assumption is that we use mempools for the duration of the task and in this case it's more opportunistic (it has to be h2 and there needs to be a request body). It's also a form of transient storage since this is a request body.

Not sure why @mbgrydeland chose this, I only assumed.

mbgrydeland:
> What is the math behind not just using malloc buffers ?

It is the unbounded outside influence memory usage that is the concern, and why I wanted to tie it into the stevedores. That way the stevedore's may purge content using LRU as the result of clients asking us to buffer request body data. Each H/2 stream may ask us to buffer 64k of data, so the total sum may become significant easily.

Though this uses the Transient stevedore, which by default is unbounded. So out of the box there will be no limit.

bsdphk:
I can see the point in reusing the stevedore machinery.

But re-purposing Transient is wrong: The limits I would want for this are very different from the limits, if any, I'd want for Transient.

One solution: Invent a new "magic" stevedore for this.

Other solution: Make which stevedore is used a parameter.

I probably prefer the latter.

bin/varnishd/storage/stevedore.c
```diff
@@ -96,6 +96,75 @@ STV_NewObject(struct worker *wrk, struct objcore *oc,
 
 /*-------------------------------------------------------------------*/
 
+struct stv_buffer {
+	unsigned		magic;
+#define STV_BUFFER_MAGIC	0xf39cb6c2
+	const struct stevedore	*stv;
+	size_t			size;
+	uintptr_t		priv;
+	uint8_t			buf[];
```
bsdphk, line 105: The `buf[]` member seems to be unused ?

bin/varnishd/storage/stevedore.c
```diff
@@ -96,6 +96,75 @@ STV_NewObject(struct worker *wrk, struct objcore *oc,
...
+		return (NULL);
+	if (stv->allocbuf == NULL)
+		return (NULL);
+
+	buf = stv->allocbuf(wrk, stv, size + PRNDUP(sizeof *stvbuf), &priv);
+	if (buf == NULL)
+		return (NULL);
+
+	assert(PAOK(buf));
+	stvbuf = TRUST_ME(buf);
```
bsdphk, line 128: This PR adds a lot of `TRUST_ME()` calls, and I am not sure I understand why ?

bin/varnishd/http2/cache_http2_panic.c
```diff
@@ -49,21 +72,53 @@ h2_sess_panic(struct vsb *vsb, const struct sess *sp)
 	h2 = (void*)*up;
 	if (PAN_dump_struct(vsb, h2, H2_SESS_MAGIC, "h2_sess"))
 		return;
-	h2 = (void*)*up;
+	VSB_printf(vsb, "refcnt = %d, bogosity = %d, error = %s\n",
+	    h2->refcnt, h2->bogosity, h2_panic_error(h2->error));
+	VSB_printf(vsb,
+	    "open_streams = %d, highest_stream = %u,"
+	    " goaway_last_stream = %u,\n",
+	    h2->open_streams, h2->highest_stream, h2->goaway_last_stream);
```
bsdphk, line 80: h2->open_streams is unsigned, should use %u

bsdphk (commented):
This review is from a quick flexelint run

There are also a number of complaints about "if (size > cache_param->fetch_maxchunksize)" constructs mixing signed/unsigned.

bin/varnishd/storage/stevedore.c
mbgrydeland: Yeah, it didn't surve much purpose. I've removed it.

bin/varnishd/storage/stevedore.c
mbgrydeland: I've replaced the TRUST_ME's with normal (void *) casts. I guess the TRUST_ME macro adds cast away of const as well, and yes that we don't need for this usage.

bin/varnishd/http2/cache_http2_panic.c
mbgrydeland: Fixed

mbgrydeland:
> FWIW we don't use "stevedore" in the varnishd manual, I would prefer "h2_rxbuf_storage".

Valid point. I will reword things accordingly.

bsdphk:
I think I'm OK with this now.

mbgrydeland:
Rebased, squashed and reordered for a better bisect experience

mbgrydeland:
> > FWIW we don't use "stevedore" in the varnishd manual, I would prefer "h2_rxbuf_storage".
> 
> Valid point. I will reword things accordingly.

Renamed to "h2_rxbuf_storage"

nigoroll:
I do not find time for an appropriate review at the moment and do not want to hold back this PR.

Dridi:
FWIW, we could make it a regular string parameter with a simple change:

```diff
diff --git i/include/tbl/params.h w/include/tbl/params.h
index 2365cc418..90e6a34ee 100644
--- i/include/tbl/params.h
+++ w/include/tbl/params.h
@@ -1551,11 +1551,23 @@ PARAM_THREAD(
  * String parameters
  */
 
-#  define PARAM_STRING(nm, pv, def, ...) \
-	PARAM(, , nm, tweak_string, pv, NULL, NULL, def, NULL, __VA_ARGS__)
+#  define PARAM_STRING(nm, tw, pv, def, ...) \
+	PARAM(, , nm, tw, pv, NULL, NULL, def, NULL, __VA_ARGS__)
+
+PARAM_STRING(
+	/* name */	h2_rxbuf_storage,
+	/* tweak */	tweak_h2_rxbuf_storage,
+	/* priv */	&mgt_stv_h2_rxbuf,
+	/* def */	"Transient",
+	/* descr */
+	"The name of the storage backend that HTTP/2 receive buffers"
+	" should be allocated from.",
+	/* flags */	MUST_RESTART
+)
 
 PARAM_STRING(
 	/* name */	cc_command,
+	/* tweak */	tweak_string,
 	/* priv */	&mgt_cc_cmd,
 	/* def */	VCC_CC,
 	/* descr */
@@ -1568,6 +1580,7 @@ PARAM_STRING(
 
 PARAM_STRING(
 	/* name */	vcl_path,
+	/* tweak */	tweak_string,
 	/* priv */	&mgt_vcl_path,
 	/* def */	VARNISH_VCL_DIR,
 	/* descr */
@@ -1582,6 +1595,7 @@ PARAM_STRING(
 
 PARAM_STRING(
 	/* name */	vmod_path,
+	/* tweak */	tweak_string,
 	/* priv */	&mgt_vmod_path,
 	/* def */	VARNISH_VMOD_DIR,
 	/* descr */
@@ -1657,28 +1671,10 @@ PARAM_PCRE2(
 	" messages."
 )
 
-/*--------------------------------------------------------------------
- * Custom parameters with separate tweak function
- */
-
-#  define PARAM_CUSTOM(nm, pv, def, ...) \
-	PARAM(, , nm, tweak_ ## nm, pv, NULL, NULL, def, NULL, __VA_ARGS__)
-
-PARAM_CUSTOM(
-	/* name */	h2_rxbuf_storage,
-	/* priv */	&mgt_stv_h2_rxbuf,
-	/* def */	"Transient",
-	/* descr */
-	"The name of the storage backend that HTTP/2 receive buffers"
-	" should be allocated from.",
-	/* flags */	MUST_RESTART
-)
-
 #  undef PARAM_ALL
 #  undef PARAM_PCRE2
 #  undef PARAM_STRING
 #  undef PARAM_VCC
-#  undef PARAM_CUSTOM
 #endif /* defined(PARAM_ALL) */
 
 #undef PARAM_MEMPOOL
```

Dridi:
I suppose we could also call the tweak function "tweak_storage", it doesn't really have anything specific to the h2 rxbuf.

PR #3662, opened 2021-08-04
[6.0] vcc: Insert the built-in source last

Dridi:
In the output of vcl.show -v, it means that the least useful file (in the
sense that it is common to every single vcl.load) is now printed last.

This change originates from a larger and more intrusive refactoring.

It also helps get rid of spurious Wstring-contatenation warnings from
clang 12 in the test suite, instead of disabling it altogether.

Refs c8174af68956206115972e72e75ceddbd758116e


Dridi:
The same change was reviewed and approved on a private branch.

PR #3667, opened 2021-08-11
[build] add a shebang to tests

gquintard:
on alpine/ppc64le, shell tests will fail without shebang (they'll throw an `applet not found` message)


Dridi:
I don't think we should generate those static files or use `echo -e`. I will try to pull them out of the makefiles and slap shebangs in the process.

Dridi:
Does the current master branch build on alpine/ppc64le after c85c4ecb4...5884e2f2f ?

gquintard:
looks like we're happy with `master`, I'm closing this

PR #3668, opened 2021-08-13
vct: Strict HTTP/1 CRLF parsing

Dridi:
We should avoid loose parsing of HTTP/1 header line delimiters. This
could potentially create a smuggling vector if Varnish was stacked with
a different server that would interpret the delimiters differently.

Refs #3246


nigoroll:
I agree that our parsers should be strict by default, but regarding introduction of the change, I think we need the option to revert to the old behaviour, at least for some time.

Dridi:
I don't know how I missed this in my first read:

> Although the line terminator for the start-line and header fields is
> the sequence CRLF, a recipient MAY recognize a single LF as a line
> terminator and ignore any preceding CR.

Issue #3645, opened 2021-07-14
sscanf needs null-terminated input


fgsch:
Revision: `450961a019d1c1955ca1851d51940ff2c87bdc98`

When build with `--enable-asan --enable-ubsan` in macOS, `make check` fails with:

```
=================================================================
==58812==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000000614 at pc 0x000108a8a954 bp 0x7ffee7aa5850 sp 0x7ffee7aa5010
READ of size 4 at 0x602000000614 thread T0
    #0 0x108a8a953 in wrap_strlen+0x183 (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x19953)
    #1 0x7fff203e3cb4 in vsscanf_l+0x7f (libsystem_c.dylib:x86_64+0x4fcb4)
    #2 0x108a93f09 in wrap_sscanf+0xd9 (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x22f09)
    #3 0x10815d64d in vav_backslash_txt vav.c:106
    #4 0x10815ea2c in VAV_ParseTxt vav.c:209
    #5 0x108160f0a in test_run vav.c:426
    #6 0x10816073d in main vav.c:481
    #7 0x7fff204daf5c in start+0x0 (libdyld.dylib:x86_64+0x15f5c)

0x602000000614 is located 0 bytes to the right of 4-byte region [0x602000000610,0x602000000614)
allocated by thread T0 here:
    #0 0x108ab94c0 in wrap_malloc+0xa0 (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x484c0)
    #1 0x108160c0f in test_run vav.c:423
    #2 0x10816073d in main vav.c:481
    #3 0x7fff204daf5c in start+0x0 (libdyld.dylib:x86_64+0x15f5c)

SUMMARY: AddressSanitizer: heap-buffer-overflow (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x19953) in wrap_strlen+0x183
Shadow bytes around the buggy address:
  0x1c0400000070: fa fa fd fa fa fa fd fa fa fa fd fd fa fa fd fa
  0x1c0400000080: fa fa fd fa fa fa fd fa fa fa fd fa fa fa fd fa
  0x1c0400000090: fa fa fd fa fa fa fd fa fa fa fd fa fa fa fd fa
[...]

PR #3673, opened 2021-08-23
vrg: Consistency check of backend responses (compliance improvements)

Dridi:
That is, only when `http_range_support` is on, which is the default.

Refs #3246

---

The test coverage is by no means comprehensive, and only the last commit changes behavior.

### DIFF
--- a/bin/varnishd/cache/cache_http.c
+++ b/bin/varnishd/cache/cache_http.c
@@ -821,29 +821,42 @@ http_GetHdrField(const struct http *hp, hdr_t hdr,
 
 /*--------------------------------------------------------------------*/
 
+static ssize_t
+http_parse_uint(const char *b, const char **e)
+{
+	ssize_t u;
+	unsigned n;
+
+	u = 0;
+	if (!vct_isdigit(*b))
+		return (-1);
+	for (; vct_isdigit(*b); b++) {
+		if (u > (SSIZE_MAX / 10))
+			return (-1);
+		u *= 10;
+		n = *b - '0';
+		if (u > (SSIZE_MAX - n))
+			return (-1);
+		u += n;
+	}
+
+	*e = b;
+	return (u);
+}
+
 ssize_t
 http_GetContentLength(const struct http *hp)
 {
 	ssize_t cl;
 	const char *b;
-	unsigned n;
 
 	CHECK_OBJ_NOTNULL(hp, HTTP_MAGIC);
 
 	if (!http_GetHdr(hp, H_Content_Length, &b))
 		return (-1);
-	cl = 0;
-	if (!vct_isdigit(*b))
+	cl = http_parse_uint(b, &b);
+	if (cl < 0)
 		return (-2);
-	for (; vct_isdigit(*b); b++) {
-		if (cl > (SSIZE_MAX / 10))
-			return (-2);
-		cl *= 10;
-		n = *b - '0';
-		if (cl > (SSIZE_MAX - n))
-			return (-2);
-		cl += n;
-	}
 	while (vct_islws(*b))
 		b++;
 	if (*b != '\0')
--- a/bin/varnishd/cache/cache_http.c
+++ b/bin/varnishd/cache/cache_http.c
@@ -864,6 +864,70 @@ http_GetContentLength(const struct http *hp)
 	return (cl);
 }
 
+ssize_t
+http_GetContentRange(const struct http *hp, ssize_t *lo, ssize_t *hi)
+{
+	ssize_t tmp, cl;
+	const char *b;
+
+	CHECK_OBJ_NOTNULL(hp, HTTP_MAGIC);
+
+	if (lo == NULL)
+		lo = &tmp;
+	if (hi == NULL)
+		hi = &tmp;
+
+	*lo = *hi = -1;
+
+	if (!http_GetHdr(hp, H_Content_Range, &b))
+		return (-1);
+
+	if (strncasecmp("bytes", b, strlen("bytes")))
+		return (-1);		// Unknown range unit, ignore
+	b += strlen("bytes");
+
+	if (!vct_islws(*b))
+		return (-1);		// Unknown range unit, ignore
+	while (vct_islws(*b))
+		b++;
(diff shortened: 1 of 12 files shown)


bin/varnishd/cache/cache_http.c
```diff
@@ -821,36 +821,158 @@ http_GetHdrField(const struct http *hp, hdr_t hdr,
 
 /*--------------------------------------------------------------------*/
 
+static ssize_t
+http_parse_uint(const char *b, const char **e)
+{
```
bsdphk, line 826: This looks like something which belongs in VNUM ?

Dridi: It does, but this patch dates back to February and then I hadn't considered reuse beyond `http_GetContentStuff()`.

On a similar note: https://github.com/varnishcache/varnish-cache/issues/3645#issuecomment-880538864

bin/varnishd/cache/cache_http.c
```diff
@@ -821,36 +821,158 @@ http_GetHdrField(const struct http *hp, hdr_t hdr,
...
+
+	u = 0;
+	if (!vct_isdigit(*b))
+		return (-1);
+	for (; vct_isdigit(*b); b++) {
+		if (u > (SSIZE_MAX / 10))
+			return (-1);
+		u *= 10;
+		n = *b - '0';
+		if (u > (SSIZE_MAX - n))
```
nigoroll, line 838: to avoid the duplicate checks for SSIZE_MAX, shouldn't we just have
```
n = *b - '0';
if (u > (SSIZE_MAX - n))
    return (-2);
u *= 10;
u += n;
```

bin/varnishd/cache/cache_http.c
```diff
@@ -821,36 +821,158 @@ http_GetHdrField(const struct http *hp, hdr_t hdr,
...
+{
+	ssize_t u;
+	unsigned n;
+
+	u = 0;
+	if (!vct_isdigit(*b))
+		return (-1);
+	for (; vct_isdigit(*b); b++) {
+		if (u > (SSIZE_MAX / 10))
+			return (-1);
```
nigoroll, line 835: by how this function is used ("too big" cases), this should be `return (-2)`?
But see next comment.

Dridi:
Bugwash:

- waiting for @nigoroll's review
- if approved I will:
  - solve conflicts and merge
  - move uint parsing to vnum
  - add hex parsing to vnum

bin/varnishd/cache/cache_http.c
```diff
@@ -821,36 +821,158 @@ http_GetHdrField(const struct http *hp, hdr_t hdr,
...
+
+	if (lo == NULL)
+		lo = &tmp;
+	if (hi == NULL)
+		hi = &tmp;
+
+	*lo = *hi = -1;
+
+	if (!http_GetHdr(hp, H_Range, &b))
+		return (NULL);
```
nigoroll, line 947: hm. Should we really return "success" for no range header?

nigoroll (commented):
overall, very much :+1:

besides the comments on some details, I am not convinced that using `struct http` is the best choice for the interface of the `http_Get*()` functions.
For example, we now have to do two `VRT_GetHdr()` calls from `vrg_range_init()` and the return value NULL signifying both "no header" and "no error" does not make much sense to me. I think just passing the header value as before made more sense.

bin/varnishd/cache/cache_http.c
Dridi: Think of this as its own function, if there is no range, then low and high arguments will be negative. It may be used by code other than `vrg_range_init()`, and `vrg_range_init()` has its own assertion that it is only ever called when there is a range header.

bin/varnishd/cache/cache_http.c
Dridi: Good catch!

Dridi:
> For example, we now have to do two `VRT_GetHdr()` calls from `vrg_range_init()` and the return value NULL signifying both "no header" and "no error" does not make much sense to me. I think just passing the header value as before made more sense.

Actually, the "second" call from the assertion should go away:

```
varnishtest "range filter without range header"

server s1 {
        rxreq
        txresp
} -start

varnish v1 -vcl+backend {
        sub vcl_deliver {
                set resp.filters = "range";
        }
} -start

client c1 {
        txreq
        rxresp
} -run
```

Because the trunk outcome is:

```
***  v1    debug|Assert error in vrg_range_init(), cache/cache_range.c line 263:
***  v1    debug|  Condition(http_GetHdr(req->http, H_Range, &r)) not true.
```

No should this be a VCL failure or a no-op VDP setup?

bin/varnishd/cache/cache_http.c
Dridi: Did you mean `if (u > (SSIZE_MAX - n) / 10)` ?

nigoroll:
> should this be a VCL failure or a no-op VDP setup?

For a bad range request, the fallback is a 200, so the filter should noop, IMHO.
For a failed If-Range, we already send a 416.

Speaking of, I think it would make sense for a filter to veto being added to the filter list. IIUC, at the moment, we can only succeed or fail (see `VCL_StackV?P()' )

bin/varnishd/cache/cache_http.c
nigoroll: that is the one I think we do not need if we have `if (u > (SSIZE_MAX - n))`

Dridi:
> Speaking of, I think it would make sense for a filter to veto being added to the filter list.

It's what happens once you remove the assertion, see 4690d7c.

bin/varnishd/cache/cache_http.c
Dridi: Something for after it has moved to VNUM?

Dridi:
Rebased https://github.com/varnishcache/varnish-cache/pull/3673/commits/80f91f77897b7f91406c2c11d815e77bad210306 onto current master, squashing followup commits in the process.

bsdphk:
ok with me. (also move to vnum)

PR #3674, opened 2021-08-23
circleci: 32bit build on the oldest Ubuntu

Dridi:
Submitted as a pull request only to make sure I'm not breaking the yaml thing.


gquintard:
looks ok to me, but why target the old version rather than a more recent one?

Dridi:
> why target the old version rather than a more recent one?

I rolled a die.

PR #3675, opened 2021-08-24
build: Revisit coverage and sanitizer setup

Dridi:
Summary:

- new `--enable-coverage` configure flag
- `GCOVING` becomes `ENABLE_COVERAGE`
- `__SANITIZER` becomes `ENABLE_SANITIZER`
- new individual `ENABLE_*SAN` macros
- `DONT_DLCLOSE_VMODS` is now inferred
- new `coverage` and `*san` feature checks in `varnishtest`

As usual see individual commits for the details.

I'm not pushing directly despite an approval of the plan because vtest client running GCOV will need a restart  and I only tested this on my development environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,6 +125,10 @@ cscope.*out
 /lib/libvarnishapi/vsl_glob_test
 /lib/libvarnishapi/vxp_test
 
+# GCOV droppings
+*.gcda
+*.gcno
+
 # vtc-bisect.sh default vtc
 /bisect.vtc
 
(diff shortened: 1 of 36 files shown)


gquintard:
do we want (me) to add `gcov` reports to `circleci`?

Dridi:
We already have https://varnish-cache.org/gcov/ so I'm not sure cci would help here.

gquintard:
ah, forgot about that. It would be redundant inded

bsdphk:
OK with me.

bsdphk:
Btw: Should distclean target remove *.gcda and *.gcno files ?

Dridi:
> Btw: Should distclean target remove *.gcda and *.gcno files ?

Good point, I took care of that while squashing e52fe750064a5e581e6c182bb1a431b64ac9023c in its rightful commit.

PR #3676, opened 2021-08-25
Never set a cold VCL as the active VCL

AlveElde:
When loading a VCL in mgt, only set active_vcl if the VCL is warm. Do not start the child process unless active_vcl is set. This guarantees that the child will have an active warm VCL after successful startup.

When loading a VCL in child, do not set vcl_active. vcl_active should only be set by vcl.use, to prevent races with workers started during the warmup event.

---

### More context:

To reproduce this bug, I added a `VCL_Refresh()` to the probe worker under a debug flag. I also added a warmup failure mode to vmod_debug based on VCL name, to make the test reliable. The probe worker temperature check is racing against the child startup procedure, so the worker can still hit this assert without any vmods failing:

https://github.com/varnishcache/varnish-cache/blob/dbc7fa74b52976b0ddffa7457936c056d2e01fa8/bin/varnishd/cache/cache_vcl.c#L362

Without this fix, the included test case hits this assert:

https://github.com/varnishcache/varnish-cache/blob/dbc7fa74b52976b0ddffa7457936c056d2e01fa8/bin/varnishd/cache/cache_vcl.c#L938

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -748,12 +748,6 @@ VSLb_ts_busyobj(struct busyobj *bo, const char *event, vtim_real now)
 /* cache_vcl.c */
 const char *VCL_Name(const struct vcl *);
 
-#define ASSERT_VCL_ACTIVE()					\
-	do {							\
-		assert(vcl_active == NULL ||			\
-		    vcl_active->temp->is_warm);			\
-	} while (0)
-
 /* cache_wrk.c */
 
 typedef void *bgthread_t(struct worker *, void *priv);
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -748,6 +748,12 @@ VSLb_ts_busyobj(struct busyobj *bo, const char *event, vtim_real now)
 /* cache_vcl.c */
 const char *VCL_Name(const struct vcl *);
 
+#define ASSERT_VCL_ACTIVE()					\
+	do {							\
+		assert(vcl_active == NULL ||			\
+		    vcl_active->temp->is_warm);			\
+	} while (0)
+
 /* cache_wrk.c */
 
 typedef void *bgthread_t(struct worker *, void *priv);
--- a/bin/varnishd/mgt/mgt_vcl.c
+++ b/bin/varnishd/mgt/mgt_vcl.c
@@ -285,7 +285,11 @@ int
 mgt_has_vcl(void)
 {
 
-	return (!VTAILQ_EMPTY(&vclhead) && active_vcl != NULL);
+	if (active_vcl == NULL)
+		return (0);
+
+	AZ(VTAILQ_EMPTY(&vclhead));
+	return (1);
 }
 
 /*
--- a/bin/varnishd/mgt/mgt_vcl.c
+++ b/bin/varnishd/mgt/mgt_vcl.c
@@ -285,7 +285,7 @@ int
 mgt_has_vcl(void)
 {
 
-	return (!VTAILQ_EMPTY(&vclhead));
+	return (!VTAILQ_EMPTY(&vclhead) && active_vcl != NULL);
 }
 
 /*
@@ -458,7 +458,7 @@ mgt_new_vcl(struct cli *cli, const char *vclname, const char *vclsrc,
 	AZ(C_flag);
 	vp->fname = lib;
 
-	if (active_vcl == NULL)
+	if (active_vcl == NULL && vp->warm)
 		active_vcl = vp;
 
 	if ((cli->result == CLIS_OK || cli->result == CLIS_TRUNCATED) &&
--- a/bin/varnishtest/tests/r03674.vtc
+++ b/bin/varnishtest/tests/r03674.vtc
@@ -1,0 +1,55 @@
+varnishtest "The active VCL must always be warm"
+
+server s1 { } -start
+
+shell {
+	cat >${tmpdir}/f1_good <<-EOF
+
+	vcl 4.1;
+	backend default none;
+
+	EOF
+}
+
+shell {
+	cat >${tmpdir}/f2_bad <<-EOF
+
+	vcl 4.1;
+	import debug;
+	backend default none;
+
+	probe p1 {
+		.interval = 1s;
+	}
+
+	sub vcl_init {
+		new d1 = debug.dyn("${s1_addr}", "${s1_port}", probe = p1);
+	}
+
+	EOF
+}
+
+shell {
+	cat >${tmpdir}/f3_ugly <<-EOF
+
+	vcl 4.1;
+	import debug;
+	backend default none;
+
+	EOF
+}
+
+varnish v1 -cliok "param.set debug +vclrefresh"
+varnish v1 -clierr 300 "start"
+
+varnish v1 -cliok "vcl.load vcl1 ${tmpdir}/f1_good cold"
+varnish v1 -clierr 300 "start"
+
+varnish v1 -cliok "vcl.load vcl2 ${tmpdir}/f2_bad warm"
+varnish v1 -cliok "start" -cliok "stop"
+
+varnish v1 -cliok "vcl.load debug_warm ${tmpdir}/f3_ugly warm"
+varnish v1 -clierr 300 "start"
+
+varnish v1 -cliok "vcl.discard debug_warm"
+varnish v1 -cliok "start" -cliok "stop"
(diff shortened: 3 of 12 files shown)


include/tbl/debug_bits.h
```diff
@@ -53,6 +53,7 @@ DEBUG_BIT(PROCESSORS,		processors,	"Fetch/Deliver processors")
 DEBUG_BIT(PROTOCOL,		protocol,	"Protocol debugging")
 DEBUG_BIT(VCL_KEEP,		vcl_keep,	"Keep VCL C and so files")
 DEBUG_BIT(LCK,			lck,		"Additional lock statistics")
+DEBUG_BIT(VCLREFRESH,		vclrefresh,	"Rapid VCL refresh")
```
Dridi, line 56: Nitpick: `vcl_refresh`.

nigoroll: agree

vmod/vmod_debug.c
```diff
@@ -442,6 +442,9 @@ event_load(VRT_CTX, struct vmod_priv *priv)
 	if (cache_param->nuke_limit == 42) {
 		VSB_cat(ctx->msg, "nuke_limit is not the answer.");
 		return (-1);
+	} else if (!strcmp(VCL_Name(ctx->vcl), "debug_load")) {
```
nigoroll, line 445: we should try to get to a test case not requiring magic vcl names

bin/varnishtest/tests/s00002.vtc
```diff
@@ -43,7 +43,7 @@ varnish v1 -vcl {
 		.port = "${s1_port}";
 		.probe = {
 			.url = "/";
-			.timeout = 30ms;
+			.timeout = 1s;
```
nigoroll, line 46: why this?

nigoroll (changes requested):

bin/varnishtest/tests/s00002.vtc
AlveElde: This change can be dropped.

We saw the probe time out because the probe worker would refresh the VCL and sometimes see `vcl_active` as NULL. That would put the worker to sleep for 100ms, and as the `VCL_refresh()` happened within the timed codepath, time out the probe in this test case. As the `VCL_refresh()` added in this changeset happens outside of the timed codepath (and under debug flag), this is not be an issue.

vmod/vmod_debug.c
AlveElde: Taking another look at it, I think we can simplify this test case a lot. It boils down to validating the assumption of "active VCL is always warm", and we don't acutally need a worker to do that for us. A simple assert in vmod_debug's `event_warm()` should suffice:

~~~
assert(vcl_active == NULL || vcl_active->temp->is_warm);
~~~

It may be the case that this assert should be placed somewhere in cache_vcl.c, but I am not seeing any obvious places to put it.

nigoroll (approved):
LGTM, but I will pass to @Dridi and maybe @mbgrydeland 
please squash, rebase, force push

Dridi (changes requested):
I have a problem with the following scenario:

- we load a bunch of VCLs
- we successfully start the cache
  - a VCL became active
- we stop the cache
  - presumably the VCL is still active?
- we restart the cache
  - what happens if a cold VCL is loaded before the active VCL?
  - what happens if the active VCL fails to warm up?

Dridi:
One more consideration, is it a good time to get rid of the vmod_debug cryptic `42` features and add proper `vmod_fail_{load,warm}_event` debug flags?

bin/varnishd/mgt/mgt_vcl.c
```diff
@@ -285,7 +285,7 @@ int
 mgt_has_vcl(void)
 {
 
-	return (!VTAILQ_EMPTY(&vclhead));
+	return (!VTAILQ_EMPTY(&vclhead) && active_vcl != NULL);
```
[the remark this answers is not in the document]
AlveElde: Good point, I flipped it around to eliminate the brackets

vmod/vmod_debug.c
```diff
@@ -436,6 +437,7 @@ event_load(VRT_CTX, struct vmod_priv *priv)
 	struct priv_vcl *priv_vcl;
 
 	AN(ctx->msg);
+	assert(vcl_active == NULL || vcl_active->temp->is_warm);
```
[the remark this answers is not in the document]
AlveElde: Seems like a good idea to me. I have removed the assertions from `vmod_debug` and inserted them where relevant as `ASSERT_VCL_ACTIVE()`.

bin/varnishtest/tests/r03674.vtc
```diff
@@ -0,0 +1,17 @@
...
+	vcl 4.1;
+	import debug;
+	backend default none;
+
+	EOF
+}
+
+varnish v1 -cliok "vcl.load vcl1 ${tmpdir}/f1 cold"
+varnish v1 -cliok "vcl.load vcl2 ${tmpdir}/f1 warm"
+varnish v1 -cliok "start"
```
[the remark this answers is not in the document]
AlveElde: I removed `vmod_debug` from the test, replaced by `ASSERT_VCL_ACTIVE()`.

The test case has been improved to address your concerns and nitpicks :)

AlveElde:
@Dridi, regarding your scenario:

Note that `active_vcl` and `vcl_active` are different things.

> we load a bunch of VCLs

At least one of these must be auto/warm for the cache to start. Mgt will set `active_vcl` to the first auto/warm VCL.

> we successfully start the cache 

Mgt will push the VCLs, then ask the chlild to `vcl.use` the VCL pointed to by `active_vcl`. This will prompt the child to set its `vcl_active` for the first time. 

> we stop the cache 
> - presumably the VCL is still active?

Correct, `active_vcl` is still set. 

> we restart the cache 
> what happens if a cold VCL is loaded before the active VCL?

In the provided test case, this is exactly what happens. But as before, `vcl_active` is first set when the child receives a `vcl.use`.

> what happens if the active VCL fails to warm up?

Then the VCL push fails, the child does not receive a `vcl.use`, and `vcl_active` is not set.

bin/varnishtest/tests/r03674.vtc
```diff
@@ -0,0 +1,40 @@
...
+varnish v1 -cliok "vcl.load vcl_auto ${tmpdir}/f1 warm"
+varnish v1 -cliok "vcl.use vcl_auto"
+varnish v1 -cliexpect "available *warm *warm *0 *vcl_warm" "vcl.list"
+varnish v1 -cliexpect "active *warm *warm *0 *vcl_auto" "vcl.list"
+
+# Cool the previous active VCL.
+varnish v1 -cliok "vcl.state vcl_warm cold"
+varnish v1 -cliexpect "available *cold *cold *0 *vcl_warm" "vcl.list"
+
+# Restart the cace.
```
nigoroll, line 36: typo

bin/varnishd/cache/cache.h
```diff
@@ -748,6 +748,12 @@ VSLb_ts_busyobj(struct busyobj *bo, const char *event, vtim_real now)
 /* cache_vcl.c */
 const char *VCL_Name(const struct vcl *);
 
+#define ASSERT_VCL_ACTIVE()					\
+	do {							\
+		assert(vcl_active == NULL ||			\
+		    vcl_active->temp->is_warm);			\
+	} while (0)
```
Dridi, line 755: Something for `cache_varnishd.h` instead, but if you are no longer using it in vmod_debug `cache_vcl.h` could maybe work too.

Dridi (changes requested):
> Note that `active_vcl` and `vcl_active` are different things.

We would probably be better off renaming them to `mgt_active_vcl` and `cache_active_vcl` or similar to avoid confusion. I'm not suggesting it for this pull request...

Dridi (approved):

mbgrydeland (approved):
LGTM

bsdphk:
I think the proposed change is wrong.

Currently the logic is "The first VCL you load becomes active VCL" and along the way we made it possible for that VCL to be loaded cold.

The solution is not to pick the non-cold first VCL we see as active VCL, the solution is to not panic if the active VCL was loaded cold.

As far as I can tell, that means refusing to start the child process in this one particular case, I cannot right here think of any other situations that might arise, but this should be checked.

We have come a long way since the "first VCL is active" logic was chosen, and in particular we now have labels, where it is almost guaranteed that you do not want the first VCL to become active.  I think we should reconsider the logic to become instead:

If there is a -b or a single -f argument, that VCL becomes active.

In all other cases, an explicit `vcl.use` will be required.

nigoroll:
superseded by #3737

PR #3678, opened 2021-08-30
tools: New script to check magic numbers

Dridi:
Only from a git clone, and only from the root directory, which should
leave alone distcheck and builds from release archives.

Fix offending magic numbers and add `make check` coverage.


bsdphk:
OK with me.

Dridi:
Force-pushed to add a copyright notice to the script.

PR #3680, opened 2021-08-31
stevedore: Set up nuke_limit for buffer allocations

Dridi:
This looks like an oversight based on the test case expectations.


mbgrydeland (approved):
This was indeed an omission. Thank you @Dridi for fixing it.

bsdphk:
OK with me

PR #3689, opened 2021-09-09
param: Cleanup of cc_command (and collaterals)

Dridi:
See the current trunk docs for `cc_command`:

https://varnish-cache.org/docs/trunk/reference/varnishd.html#cc-command

Currently it says:

> Default: exec clang -g -O2 -Wall -Werror -Wno-error=unused-result -fstack-protector-strong -Werror -Wall -Wcast-align -Wcast-qual -Wchar-subscripts -Wempty-body -Wextra -Wformat -Wformat-y2k -Wformat -Wformat-zero-length -Wmissing-field-initializers -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wpointer-sign -Wreturn-type -Wshadow -Wstrict-aliasing -Wstrict-prototypes -Wstring-plus-int -Wswitch -Wunused-parameter -Wunused-result -Wwrite-strings -Wno-old-style-definition -Wno-sign-compare -Wno-implicit-fallthrough -Wno-missing-variable-declarations -Wno-nullability-completeness -fstack-protector -Wno-missing-field-initializers -pthread -fpic -shared -Wl,-x -o %o %s

On the environment building the online docs, `param.show cc_command` would report the same default value.

After merging this pull request it would say, in the docs:

> Default: exec c99 -lpthread -shared -fPIC -o %o %s

The default value reported by `param.show cc_command` for the environment building the online docs would become:

> Default: exec clang -g -O2 -pthread -fpic -shared -Wl,-x -o %o %s

The collaterals are `vcl_path` and `vmod_path`, see individual commits for more details.

I would like to target 7.0 with this change.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -654,9 +654,11 @@ gl_LD_VERSION_SCRIPT
 # idiocy where write is marked as warn_unused_result, causing build
 # failures.
 
+WFLAGS=
+
 AX_CHECK_COMPILE_FLAG([-Wall],
      [CFLAGS="${CFLAGS} -Wall"
-      OCFLAGS="${OCFLAGS} -Wall"])
+      WFLAGS="${WFLAGS} -Wall"])
 
 if test "$SUNCC" = "yes" ; then
     SUNCC_CFLAGS=" \
@@ -665,27 +667,27 @@ if test "$SUNCC" = "yes" ; then
 	"
     AX_CHECK_COMPILE_FLAG([${SUNCC_CFLAGS}],
 	[CFLAGS="${CFLAGS} ${SUNCC_CFLAGS}"
-	 OCFLAGS="${OCFLAGS} ${SUNCC_CFLAGS}"])
+	 WFLAGS="${WFLAGS} ${SUNCC_CFLAGS}"])
 else
     AX_CHECK_COMPILE_FLAG([-Werror],
 	[CFLAGS="${CFLAGS} -Werror"
-	 OCFLAGS="${OCFLAGS} -Werror"])
+	 WFLAGS="${WFLAGS} -Werror"])
 fi
 
 case $target in
     *-*-darwin*)
 	AX_CHECK_COMPILE_FLAG([-Wno-expansion-to-defined],
 	[CFLAGS="${CFLAGS} -Wno-expansion-to-defined"
-	 OCFLAGS="${OCFLAGS} -Wno-expansion-to-defined"])
+	 WFLAGS="${WFLAGS} -Wno-expansion-to-defined"])
 	;;
 esac
 
 AX_CHECK_COMPILE_FLAG([-Werror=unused-result],
     [CFLAGS="${CFLAGS} -Wno-error=unused-result"
-     OCFLAGS="${OCFLAGS} -Wno-error=unused-result"],
+     WFLAGS="${WFLAGS} -Wno-error=unused-result"],
     [AX_CHECK_COMPILE_FLAG([-Wunused-result],
      [CFLAGS="${CFLAGS} -Wno-unused-result"
-      OCFLAGS="${OCFLAGS} -Wno-unused-result"])])
+      WFLAGS="${WFLAGS} -Wno-unused-result"])])
 
 # This corresponds to FreeBSD's WARNS level 6
 DEVELOPER_CFLAGS=`$PYTHON $srcdir/wflags.py`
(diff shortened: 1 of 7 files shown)


bsdphk:
This has been annoying me too, but not enough to do something about it, because I feel that people almost never should muck about with the `cc_command` setting.

My idea for dealing with it was to add a helper-shell-script and hide the insanity there.

I'm a little bit loath to have a parameter which says "the value is this (except it really isn't)"

Dridi:
> I'm a little bit loath to have a parameter which says "the value is this (except it really isn't)"

Except that the values in our online docs today document defaults that aren't (`cc_command` and  `{vcl,vmod}_path`). And the NB formulation makes it clear that it depends on how Varnish is built, which is not as misleading as you imply.

Dridi:
> My idea for dealing with it was to add a helper-shell-script and hide the insanity there.

Well, since those parameters don't have static defaults, being build-dependent, I would think a dynamic default and a note is appropriate just like any other parameter with a dynamic property.

To me, the insanity is gone once the only way to see a `cc_command` with a gazillion `-W` options is to run `param.show` after building with the `--enable-developer-warnings` configure option. We don't need a shell script indirection out of the box.

nigoroll (commented):
I see a point in removing the long list of warnings from cc_command and I like @bsdphk s suggestion to add a separate parameter for the warnings only.
I can not follow on the suggestion to use canned defaults in the documentation. In my mind, the documentation for a specific build reflecting the actual defaults has a lot of value.
If the online docs are an issue, we should fix those specifically, IMHO.

Dridi:
> I see a point in removing the long list of warnings from cc_command and I like @bsdphk s suggestion to add a separate parameter for the warnings only.

I suggested `%w` in the `cc_command` for the warnings, similar to the existing `%o` and `%s`, which I think would be less bothersome than an entirely separated parameter.

> I can not follow on the suggestion to use canned defaults in the documentation. In my mind, the documentation for a specific build reflecting the actual defaults has a lot of value.

One disagreement was on the dynamic nature of those default values. What I call "dynamic" properties (min/max/def) for parameters are properties that may change without touching the source code. For example, `thread_pool_stack`'s minimum is queried during `varnishd`'s startup.

In this case, those parameters default values depend on `./configure --options` and may thus change without editing the source code, so they aren't static defaults, that's why I would like to give them dynamic defaults instead.

One suggestion was to give `vcl_path` a dynamic default of `${sysconfdir}/varnish:${datarootdir}/varnish/vcl` to have really neutral docs (something similar for `vmod_path`).

nigoroll:
my opinion and what I took from bugwash:

* split `%w` from `cc_command`: 👍🏽 
* canned texts: 👎🏽 
* add `BUILD_OPTIONS_TEXT`: 👍🏽 
* Only `-Wall -Werror` in `%w`, unless developer flags used

Dridi:
> add `BUILD_OPTIONS_TEXT`: 👍🏽

There is already one...

nigoroll:
> There is already one...

The comment was to support its addition, yes.

Dridi:
Ok, I thought there was a misunderstanding here. I'm waiting for suggestions to rephrase the `BUILD_OPTIONS_TEXT` since it was apparently not enough.

Dridi:
Updated patch series cb446a38d7...8f962f757f:

- new `cc_warnings` parameter
- new `%w` expansion in `cc_command`
- new `%d` expansion in `cc_command`
  - expands to the default value of `cc_command` itself
- neutral defaults for the documentation

Dridi:
Rebased against master, with two additional changes after discussing it with @bsdphk and @nigoroll:

- new `%D` expansion in `cc_command`
  - expands to an expansion of the default value of `cc_command`
- better cc_command description

Dridi:
Rebased against master, and I forgot to mention a `%n` expansion in `cc_command`.

That should be it for the collateral changes that came up for a project I'm working on.

include/tbl/params.h
```diff
@@ -1530,11 +1530,42 @@ PARAM_STRING(
 	/* priv */	&mgt_cc_cmd,
 	/* def */	VCC_CC,
 	/* descr */
-	"Command used for compiling the C source code to a "
-	"dlopen(3) loadable object.  Any occurrence of %s in "
-	"the string will be replaced with the source file name, "
-	"and %o will be replaced with the output file name.",
-	/* flags */	MUST_RELOAD
+	"The command used for compiling the C source code to a "
```
nigoroll, line 1533: 👍🏽 very good improvement

configure.ac
```diff
@@ -804,24 +806,33 @@ else
 	*-*-solaris*)
 		case $PTHREAD_CC in
 		*gcc*)
-			VCC_CC="$PTHREAD_CC $OCFLAGS $PTHREAD_CFLAGS -fpic -shared -o %o %s"
```
nigoroll, line 807: I would not be surprised to see fallout from some of the shell variables containing `%`, but I guess we can handle it when it happens?

nigoroll (approved):

configure.ac
Dridi: I think so too, but I also did 60143b401d1be5020f5bd4d6e607407b7d8ed2b3 when I noticed, even though those percent sequences were no-ops.

Dridi:
Rebased against master with the squashme commit folded in the commit where it belongs. Waiting for CI to complete before merging.

PR #3691, opened 2021-09-09
varnishadm: Only print the status in interactive mode

### LINKED ISSUE #3687: `varnishadm vcl.list  -j` produces not parsable JSON with status code on the first line
## Expected behavior

`varnishadm vcl.list  -j` should produce a valid json, but the output contains status code `200` on the first line, which is not parsable.

The status code is also added to the first line in the text output mode without -j parameter. 

Run server:
```
docker run --rm --name varnish varnish:6.6
```

Run varnishadm
```
docker exec -it varnish varnishadm vcl.list  -j 
```

output:

```json
200
[ 2, ["vcl.list", "-j"], 1631021700.200,
  {
    "status": "active",
    "state": "auto",
    "temperature": "warm",
    "busy": 0,
    "name": "boot"
}
]
```
It affects only latest varnish version, output from `varnishadm vcl.list  -j` in older `varnish-6.5.1` returns valid json, there is no status on the first line.

## Possible Solution
Remove 200 status from json output

## Your Environment
docker, macosx

Dridi:
Or when the -p option is used.

Fixes #3687

### DIFF
--- a/bin/varnishadm/varnishadm.c
+++ b/bin/varnishadm/varnishadm.c
@@ -76,6 +76,10 @@
 		exit(status); \
 	} while (0)
 
+enum pass_mode_e {
+	pass_script,
+	pass_interactive,
+};
 
 static double timeout = 5;
 static int p_arg = 0;
@@ -161,7 +165,7 @@ cli_sock(const char *T_arg, const char *S_arg)
 }
 
 static unsigned
-pass_answer(int fd)
+pass_answer(int fd, enum pass_mode_e mode)
 {
 	unsigned u, status;
 	char *answer = NULL;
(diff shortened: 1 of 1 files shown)


bsdphk:
No opinion, but sounds sensible.

nigoroll (approved):
sounds like a sensible fix for release

Dridi:
Bugwash: approved for the 7.0 release. I may add VTC coverage before merging if not too complicated, otherwise test coverage will come separately.

Dridi:
https://github.com/varnishcache/varnish-cache/commit/2ac661197a84c4ae0242e3a277a659f50dc9cec1

PR #3693, opened 2021-09-09
vcl: Introduce req.uncacheable

Dridi:
The current way to bypass the cache is to either return(pass) in
vcl_recv or to trigger an HfP/HfM on the backend side.

The former forces you to interrupt the request processing, such as
header normalization, or to set a variable/header to remember your
choice so that you can return(pass) once processing is done.

The latter is nicer but has to deal with request coalescing and it
doesn't really make sense to wait that long before making this decision.

The new req.uncacheable mirrors beresp.uncacheable but really just
transforms a return(hash) into a return(pass). This means that it
takes precedence over req.hash_always_miss but also operate slightly
earlier, at the end of vcl_recv rather than at the end of vcl_hash.

Signed-off-by: Dridi Boukelmoune <dridi.boukelmoune@gmail.com>

---

I found this patch from @gquintard in my inbox, this is a followup to #3457  but with a batter name for the flag that matches the `uncacheable` property of `bereq`, `beresp` and `obj`.

I think this patch is in much better shape than #3457 and with the VCL built-in split in mind, I think we should set this flag instead of initiating a pass transition in `vcl_req_method`, `vcl_req_authorization` and `vcl_req_cookie`.

Despite the improvements of the built-in VCL thanks to the split, as it is today a POST request processed by the built-in VCL would break the flow before calling `vcl_req_cookie` and getting a chance for example to normalize and clean up cookies even for uncacheable requests.

Yes, I know, one can actively `call vcl_req_cookie;` but I'm talking about turnkey behavior that would mirror what `vcl_backend_response` does and flow more naturally.

### DIFF
--- a/doc/sphinx/reference/vcl_var.rst
+++ b/doc/sphinx/reference/vcl_var.rst
@@ -393,6 +393,27 @@ req.hash_always_miss
 	This is useful to force-update the cache without invalidating
 	existing entries in case the fetch fails.
 
+req.uncacheable
+
+	Type: BOOL
+
+	Readable from: client
+
+	Writable from: client
+
+	Default: ``false``.
+
+	Force a cache pass for this request, even if perfectly
+	good matching objects are in the cache.
+
+	This allows to mark a request as uncacheable in ``vcl_recv``
+	without immediately returning from the subroutine with a
+	``pass`` transition.
+
+	Clearing the variable has no effect and will log the warning
+	"Ignoring attempt to reset req.uncacheable".
+
+
 req.is_hitmiss
 
 	Type: BOOL
@@ -593,9 +614,9 @@ bereq.uncacheable
 	Readable from: backend
 
 
-	Indicates whether this request is uncacheable due to a
-	`pass` in the client side or a hit on an hit-for-pass object.
-
+	Indicates whether this request is uncacheable due to a ``pass``
+	transition or ``req.uncacheable`` being set on the client side
+	or a hit on an hit-for-pass object.
 
 bereq.connect_timeout
 
@@ -1125,8 +1146,8 @@ obj.uncacheable
 
 	Readable from: vcl_deliver
 
-	Whether the object is uncacheable (pass, hit-for-pass or
-	hit-for-miss).
+	Whether the object is uncacheable (pass, ``req.uncacheable``,
+	hit-for-pass or hit-for-miss).
 
 
 obj.storage
--- a/bin/varnishd/cache/cache_req.c
+++ b/bin/varnishd/cache/cache_req.c
@@ -282,6 +282,7 @@ Req_Cleanup(struct sess *sp, struct worker *wrk, struct req *req)
 	req->t_req = NAN;
 	req->req_body_status = NULL;
 
+	req->uncacheable = 0;
 	req->hash_always_miss = 0;
 	req->hash_ignore_busy = 0;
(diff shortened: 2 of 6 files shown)


doc/sphinx/reference/vcl_var.rst
```diff
@@ -393,6 +393,27 @@ req.hash_always_miss
 	This is useful to force-update the cache without invalidating
 	existing entries in case the fetch fails.
 
+req.uncacheable
+
+	Type: BOOL
+
+	Readable from: client
+
+	Writable from: client
```
Dridi, line 402: It should maybe be writable only in `vcl_recv` and `vcl_hash`, and become immutable once a lookup is initiated.

nigoroll:
I can see the use case, but just in general I am not a big friend of adding another special "return (hash) eventually results in a cache lookup, except when req.uncacheable is true, in when it results in a pass" case

gquintard:
>  "return (hash) eventually results in a cache lookup, except when req.uncacheable is true, in when it results in a pass"

I get that sentiment, but the current situation isn't great either. First `req.hash_always_miss` already exists, so things aren't as pure as we'd hope. And `return(pass)` on the other end of the spectrum forces us to return immediately. The dual stance is pretty confusing.

Personally, I'd like for VCL to more closely mirror C with `vcl_recv` only allowing `return(ok)` (and checking `req.{uncacheable,hash_always_miss,ttl,grace}`) and `return(synth)`. That would possibly alleviate the confusion that we still go though `vcl_hash` even if we don't `return(hash)`.

I do realize that's a longer term view, but that felt like the right issue to bring it up.

nigoroll:
> First `req.hash_always_miss` already exists

Yes. I am coming from a position of regret that we get more and more of these flags.

Dridi:
Today I ran into the need to check something like `req.uncacheable` in `vcl_hash`.

bsdphk:
Discussed at bugwash, but did not reach quorum.

As a general principle, we are trying to avoid "magic settings" and invisible state in VCL and prefer relying on the code flow to do make it clear what happens.

If we want to do something like this, (that is: if there actually is a problem to fix), a better way would be to introduce an explicit return-variable so something like:

    if (foo) {
        set retval = pass;
    }
    [...]
    return (retval);

But again: First requirement is that there is a problem (this or others) to be fixed.

gquintard:
I'm probably going to repeat myself, but at the very least, I will try to be concise.

To me the problem is real, and actually exists at 2.5 different levels:
1. ergonomics + advocated logic: it's very common to know early what `return()` you want for a request, but you still want that request to go through all the sanitation steps, notably through the built-in ones, that we always recommend to go through. Forcing the user to `return()` is an obstacle to this.
2. API consistency: there's already `req.hash_always_miss` which is equivalent to a deferred `return(miss)` and that is great, but there's no matching deferred `return(pass)`. I feel that the arguments for both are exactly the same, so either we should have a way to defer `pass`, or we should clean the API and prune `req.hash_always_miss`.
3. (hot take bonus) Confusing returns/steps: The fact that `return(pass)` doesn't go directly go to `vcl_pass` trips a lot of people (I'm not just throwing that out baselessly, I've taught a few varnish classes), and while I hear the argument about "magic settings", I personally feel the logic would be a lot saner if we had a `req.handling` enum and `return(ok)/return(error)` (yes, basically what the C does).

PR #3700, opened 2021-09-18
Fixed build on macOS 11

catap:
`%jd` and `%ju` should be used only for `int64_t` and `uint64_t`. `intmax_t` and `uintmax_t` not always is this type.

The right way is using macros `PRId64` and `PRIu64` which is points to the right format on this system.


mbgrydeland:
I opted to cast to (u)intmax_t instead for readability in 20e007a5b17c1f68f70ab42080de384f9e192900. Closing this PR.

PR #3701, opened 2021-09-20
Build VSC generated sources in lib/libvsc

mbgrydeland:
This is to resolve the chicken and egg problem of the VSC headers being
generated only once bin/varnishd is built, while they are used in several
VMODs and libraries.

This moves the generation of the VSC .c and .h header files to a directory
in lib that can be built earlier than its dependencies.


nigoroll (approved):
I have not seen any build issues on sunos, tested in-tree and out-of-tree build

Dridi (changes requested):
Please update `.gitignore` with the new paths.

mbgrydeland:
> Please update `.gitignore` with the new paths.

Done

Dridi (approved):

PR #3703, opened 2021-09-22
[vnum] more lenient test

### LINKED ISSUE #3697: vnum_c_test fails on alpine x_86
while trying to build packages for `7.0.0` I hit this:
```
==================================================
   Varnish 7.0.0: lib/libvarnish/test-suite.log
==================================================
# TOTAL: 6
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
.. contents:: :depth: 2
FAIL: vnum_c_test
=================
./vnum_c_test: VNUM_Duration() wrong, 31556908.800 delta = 8.658390e-10
FAIL vnum_c_test (exit status: 1)
```

This is on `alpine`, so using `muslc`, and the arch is `x86`, full log here: https://gitlab.alpinelinux.org/gquintard/aports/-/jobs/487996

gquintard:
fixes #3697

@Dridi, I know we discussed `VTIM_EPSILON` but `VNUM_EPSILON` kept things in vnum land, and we use it to test `VNUM_duration`, so it made sense?

### DIFF
--- a/include/vnum.h
+++ b/include/vnum.h
@@ -31,6 +31,7 @@
  */
 
 /* from libvarnish/vnum.c */
+#define VNUM_EPSILON (1e-9d)
 double VNUM(const char *p);
 vtim_dur VNUM_duration_unit(vtim_dur r, const char *b, const char *e);
 vtim_dur VNUM_duration(const char *p);
--- a/lib/libvarnish/vnum.c
+++ b/lib/libvarnish/vnum.c
@@ -693,9 +693,10 @@ main(int argc, char *argv[])
 		++ec;
 	}
 	d1 = VNUM_duration(" 365.24219d ");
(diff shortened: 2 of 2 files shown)


include/vnum.h
```diff
@@ -31,6 +31,7 @@
  */
 
 /* from libvarnish/vnum.c */
+#define VNUM_EPSILON 1e-9
```
Dridi, line 34: Maybe something like `#define VNUM_EPSILON (1e-9d)` ?

Dridi (changes requested):
Of course, `VNUM_EPSILON` makes more sense.

include/vnum.h
gquintard: there you go

PR #3704, opened 2021-09-22
build: Simply require that SO_???TIMEO are present

### LINKED ISSUE #3692: RFC: Requiring socket timeouts to work?
After I accidentally spotted e4a04d593 it occurred to me that we have `PLATFORM_DEPENDENT` flags for timeout parameters but that's incomplete.

At least `between_bytes_timeout` is set up with both `poll(2)` and `SO_RCVTIMEO` (the latter for HTTP/1 when we pull the beresp body from the VFP chain).  Ideally it should be documented that it could be partially implemented.

Should we instead pull the plug on a non-working socket timeouts? 

The reason why we have this is:

> Older Solaris versions define SO_{RCV,SND}TIMEO, but do not implement them.

I think OpenSolaris has been dead long enough for us not to care anymore, and I'd be surprised if OpenSolaris descendants didn't have socket timeouts support today.

Dridi:
Closes #3692


nigoroll (approved):
have not tried it, lgtm

bsdphk:
distcheck pass on FreeBSD

nigoroll:
distcheck pass on SunOS

PR #3705, opened 2021-09-28
vdef: implement vmin[_t], vmax[_t] and vlimit[_t] 

asadsa92:
This should make the code easier to read and stop us from being inconsistent.

Signed-off-by: Asad Sajjad Ahmed <asadsa@varnish-software.com>

### DIFF
--- a/include/vdef.h
+++ b/include/vdef.h
@@ -165,6 +165,55 @@ int __llvm_gcov_flush(void);
 #define RUP2(x, y)  (((x)+((y)-1))&(~((uintptr_t)(y)-1UL))) /* PWR2(y) true */
 
 /**********************************************************************
+ * Find the minimum or maximum values.
+ * Only evaluate the expression once and perform type checking.
+ */
+
+/* ref: https://stackoverflow.com/a/17624752 */
+
+#define VINDIRECT(a, b, c)	a ## b ## c
+#define VCOMBINE(a, b, c)	VINDIRECT(a, b, c)
+
+#if defined(__COUNTER__)
+#	define VUNIQ_NAME(base)	VCOMBINE(base, __LINE__, __COUNTER__)
+#else
+#	define VUNIQ_NAME(base)	VCOMBINE(base, __LINE__, 0)
+#endif
+
+/* ref: https://gcc.gnu.org/onlinedocs/gcc/Typeof.html */
+
+#define _vmin(a, b, _va, _vb)						\
+({									\
+	typeof (a) _va = (a);						\
+	typeof (b) _vb = (b);						\
+	(void)(&_va == &_vb);						\
+	_va < _vb ? _va : _vb;						\
+})
+
+#define _vmax(a, b, _va, _vb)						\
+({									\
+	typeof (a) _va = (a);						\
+	typeof (b) _vb = (b);						\
+	(void)(&_va == &_vb);						\
+	_va > _vb ? _va : _vb;						\
+})
+
+#define vmin(a, b)		_vmin((a), (b), VUNIQ_NAME(_vmina),	\
+    VUNIQ_NAME(_vminb))
+#define vmax(a, b)		_vmax((a), (b), VUNIQ_NAME(_vmaxa),	\
+    VUNIQ_NAME(_vmaxb))
+
+#define vmin_t(type, a, b)	vmin((type)(a), (type)(b))
+#define vmax_t(type, a, b)	vmax((type)(a), (type)(b))
+
+/**********************************************************************
+ * Clamp the value between two limits.
+ */
+
+#define vlimit(a, l, u)		vmax((l), vmin((a), (u)))
+#define vlimit_t(type, a, l, u)	vmax_t(type, (l), vmin_t(type, (a), (u)))
+
+/**********************************************************************
  * FlexeLint and compiler shutuppery
  */
 
--- a/bin/varnishd/cache/cache_backend_probe.c
+++ b/bin/varnishd/cache/cache_backend_probe.c
@@ -633,8 +633,7 @@ vbp_set_defaults(struct vbp_target *vt, const struct vrt_backend_probe *vp)
 	if (vt->initial == ~0U)
 		vt->initial = vt->threshold - 1;
 
-	if (vt->initial > vt->threshold)
-		vt->initial = vt->threshold;
+	vt->initial = vmin(vt->initial, vt->threshold);
 }
 
 /*--------------------------------------------------------------------
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -663,8 +663,7 @@ ved_gzgz_bytes(struct vdp_ctx *vdx, enum vdp_action act, void **priv,
 		/* The main body of the object */
 		dl = foo->last / 8 - foo->ll;
 		if (dl > 0) {
-			if (dl > len)
-				dl = len;
+			dl = vmin(dl, len);
 			if (ved_bytes(foo->ecx, act, pp, dl))
 				return (-1);
 			foo->ll += dl;
@@ -686,8 +685,7 @@ ved_gzgz_bytes(struct vdp_ctx *vdx, enum vdp_action act, void **priv,
 		/* Last block */
 		dl = foo->stop / 8 - foo->ll;
 		if (dl > 0) {
-			if (dl > len)
-				dl = len;
+			dl = vmin(dl, len);
 			if (ved_bytes(foo->ecx, act, pp, dl))
 				return (-1);
 			foo->ll += dl;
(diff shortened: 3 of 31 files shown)


asadsa92:
Currently, we have a mix of both for finding the maximum value:
```C
if (a < b)
        a = b;
```
and
```
if (b > a)
        a = b;
```

Both are equivalent, but make the code base inconsistent.
This PR is meant to simplify things through macros.

Reference:
https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html#Statement-Exprs
https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Typeof.html#Typeof        

The current set of location were found using a construct of this sort:
`git grep -nA1 "if \(.* < .*\)" -- ':!/lib/libvgz/' | sed -n 'N; /if (\(.*\) < \(.*\)).*\1 = \2;/p'`

Dridi:
> The current set of location were found using a construct of this sort:
> `git grep -nA1 "if \(.* < .*\)" -- ':!/lib/libvgz/' | sed -n 'N; /if (\(.*\) < \(.*\)).*\1 = \2;/p'`

Have a look at coccinelle for structural refactoring using semantic patches.

We have some in the source tree:

https://github.com/varnishcache/varnish-cache/tree/master/tools/coccinelle#readme

asadsa92:
Cool, I will look into it.
For EL7 things are not looking pretty, so EL7 has old GCC version not supporting `__auto_type`.
`typeof` exists, but has problems with overshadowing variables when macros are used inside each other, like for `vclamp`.

include/vdef.h
```diff
@@ -164,6 +164,37 @@ int __llvm_gcov_flush(void);
...
 #define RUP2(x, y)  (((x)+((y)-1))&(~((uintptr_t)(y)-1UL))) /* PWR2(y) true */
 
+/**********************************************************************
+ * Find the minimum or maximum values.
+ * Only evaluate the expression once and perform type checking.
+ */
+
+#define vmax(a, b)							\
+({									\
+	__auto_type _va = (a);						\
```
nigoroll, line 174: I checked the C Standard (ISO/IEC 9899 4th edition 2018-07) and it has no hits on `__auto_type`. In fact it seems to be a [gcc extension](https://gcc.gnu.org/onlinedocs/gcc/Typeof.html), and the example given in the documentation is `max(a,b)`.
In Varnish-Cache, we try to adhere to standard C for compatibility with compilers, linters and other code quality tools, so unless there is a compelling argument and at least alternative code for where some language extension is not available, I would rather want to avoid such changes.
I can see that min/max macros can help clarity for some, but in my mind the improvement is not _that_ relevant. For example, the pattern
```
  if (a > x) x = a;
```
is clearly identifiable as a _max_ semantics and I personally do not find `x = max(x, a)` a relevant improvement.

Dridi: The benefit is actually a couple lines below:

```
(void)(&_va == &_vb);
```

This is going to trigger an error if:

- `a` and `b` expressions have different types
- varnish was built with `--enable-developer-warnings`

The alternative would be `typeof` but I think it's also a GNU extension.

Dridi: FWIW, the `({ ... })` expansion of macros is also not standard.

nigoroll: > The benefit is actually a couple lines below:

Yes, it looks like `typeof()` is an extension also.

Is the type mismatch an error which we would not otherwise notice? I would have expected at least flexelint to complain.

bsdphk:
We have quenched a fair bit of complaints from FlexeLint about type mismatches, most of them, to my judgement a bit too stringent for our code style.

include/vdef.h
asadsa92: Yes, even `typeof()` is a GNU extension, but there is ongoing work to make it part of `C2X`:
- https://thephd.dev/_vendor/future_cxx/papers/C%20-%20typeof.html
- https://en.wikipedia.org/wiki/C2x

I think it is fine to go with a solution without `typeof()`, like FreeBSD does:
- http://ftp.jp.freebsd.org/pub/NetBSD/misc/joerg/GENERIC/src/src/sys/external/bsd/ipf/netinet/ip_compat.h.html#532
- http://ftp.jp.freebsd.org/pub/NetBSD/misc/joerg/GENERIC/src/src/sys/external/bsd/common/include/linux/kernel.h.html#_M/max_t

But, keep in mind that you will run into problems with evaluating the expression twice. (OK to say that it is the user of the macros fault)

We can even go as far to say that the type has to be explicitly provided, but not sure if this would trigger the proper warning.

About the use case, I disagree that `vmin[_t], vmax[_t] and vclamp[_t]` is not bringing anything to the table. Just the inconsistent use:  `if (a > x) x = a;` vs.   `if (x < a) x = a;` or `if (a < x) x = a;` vs.  `if (x > a) x = a;` does not read as easily as: `x = vmax(x, a);` or `x = vmin(x, a)`.  `x = vmax_t(int, x, a);` and `x = vmin_t(int, x, a)` also add useful semantics to the code rather than having to add the casting yourself.

Not to talk about `x = vclamp(x, 0, y);` vs.:
```
if (x < 0) x = 0;
if (x > y) x = y;
```
or

```
if (0 > x) x = 0;
if (x > y) x = y;
```

or

```
if (x < 0) x = 0;
if (y > x) x = y;
```

or

```
if (0 > x) x = 0;
if (y > x) x = y;

```
or

```
if (x > y) x = y;
if (x < 0) x = 0;
```
etc.

nigoroll: Evaluating expressions twice is simply not an option for me. The consequences of missing to special case arguments with side effects can be serious, let alone the potential performance impact.
Regarding the legibility argument, I agree that it is a matter of personal preference. I respect your argument that the proposed macros improve clarity _for you_ (and/or others) and all I said is "I _personally_ do not find `x = max(x, a)` a relevant improvement. Personally, the same argument applies to clamping, with the detail added that at least my mind is conditioned so much on VCL that I _still_ read `vclamp()` as `vcl_amp()` every time I see it.

So if it was only for the clarity argument, I would abstain or pass my vote.

Yet, unfortunately, without an implementation covered by current standards (and, I missed to raise that point previously, supported by the compilers which we actually use), I would vote 👎🏽 .

asadsa92: Understood, I respect your standpoint so I will not try to convince you.
BTW, `typeof` is well supported by the current set of compilers even thought it is not strictly part of the C standard yet.
Just like `strdup(3) is NOT part of the C standard yet, but is a candidate for C2X (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2353.htm).
So, I am not buying the argument about following the current C standard, as VC as is does not follow it.

nigoroll: Let's not confuse C and libraries. The fact that C might adopt certain functions into the language has nothing to do with the prior discussion.
Regarding C compilers, a specific example to mention is that we compile with the nowadays quite exotic [SunCC compiler](http://varnish-cache.org/vtest/), which has already helped spot some issues missed by the other compilers.
Anyway, I think all is said.

asadsa92: Yes, I am not trying pushing against, but just trying to explore what option we have.
FYI, https://docs.oracle.com/cd/E37069_01/html/E37074/bjapp.html#OSSCGbjaqo
So, I think SunCC has support for `typeof`, but I do not have it installed so I can not test.

nigoroll: A trivial test on SunCC works:

https://gist.github.com/nigoroll/2f9833e5b0635787608e7a45a3a37622

nigoroll:
FTR: bugwash decision IIUC:
* uppercase the macros
* use `VLIMIT()` for `vclamp()`

Otherwise, as `typeof()` seems to be supported on all relevant platforms and the macros have been rewritten to avoid `__auto_type()` we seem to be good.
I still expect some backscatter, but we can handle it when it happens.

include/vdef.h
asadsa92: Thanks for running the test.
Could you please also paste the output of the preprocessing phase and do a test with vclamp/vlimit?
I just want to make sure `__COUNTER__` is defined and used.

nigoroll: FWIW, I did not upcase the macros in the test, and I wonder if now, that you got rid of the double evaluation with just typeof(), we would actually need to any more?

asadsa92: The `__COUNTER__` does not seems to work at suncc (you see 0 postfix which is the else branch of the "if defined"). 
This means you will have problems with variable shadowing (-Wshadow) if the line number is the same.

Dridi:
Should we consider to "single-evaluate" arguments from other `vdef.h` macros like `RUP2` for example?

asadsa92:
Rebased, and should now be ready for merge.

nigoroll:
Reviewed all changes and did not spot any issues.

PR #3707, opened 2021-09-28
Add a debug storage engine

nigoroll:
For now, the purpose is to regression test fc1ce6daf6a966df3aa8b7519228f60b040ff072 with a minimal amount of code.

### DIFF
--- a/bin/varnishd/storage/storage_debug.c
+++ b/bin/varnishd/storage/storage_debug.c
@@ -1,0 +1,58 @@
+/*-
+ * Copyright 2021 UPLEX - Nils Goroll Systemoptimierung
+ * All rights reserved.
+ *
+ * Author: Nils Goroll <nils.goroll@uplex.de>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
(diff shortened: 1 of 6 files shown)


bin/varnishd/storage/storage_debug.c
```diff
@@ -0,0 +1,95 @@
...
+	memcpy(parent, &sma_stevedore, sizeof *parent);
+	parent->ident = ident;
+	parent->name = smd_stevedore.name;
+
+	methods = malloc(sizeof *methods);
+	AN(methods);
+	memcpy(methods, &SML_methods, sizeof *methods);
+	parent->methods = methods;
+
+	av = malloc(sizeof *av * (aac + 1));
```
Dridi, line 72: calloc()

bin/varnishd/storage/storage_debug.c
```diff
@@ -0,0 +1,95 @@
...
+	for (i = 0; i < aac; i++) {
+		if (aav[i] == NULL) {
+			av[ac++] = NULL;
+			continue;
+		}
+		if (! strcmp(aav[i], "lessspace")) {
+			methods->objgetspace = smd_lsp_getspace;
+			continue;
+		}
+		av[ac++] = strdup(aav[i]);
```
Dridi, line 83: REPLACE()

bin/varnishd/storage/storage_debug.c
nigoroll: makes sense

bin/varnishd/storage/storage_debug.c
nigoroll: not so much, I think.

Dridi: Why not? At the very least shouldn't we `AN(the_dup)`?

nigoroll: ✔️

PR #3711, opened 2021-10-12
Alternate #3676 fix

bsdphk:
This is my suggested fix.

### DIFF
--- a/bin/varnishd/mgt/mgt.h
+++ b/bin/varnishd/mgt/mgt.h
@@ -231,7 +231,7 @@ void mgt_vcl_init(void);
 void mgt_vcl_startup(struct cli *, const char *vclsrc, const char *origin,
     const char *vclname, int Cflag);
 int mgt_push_vcls(struct cli *, unsigned *status, char **p);
-int mgt_has_vcl(void);
+const char *mgt_has_vcl(void);
 extern char *mgt_cc_cmd;
 extern const char *mgt_vcl_path;
 extern const char *mgt_vmod_path;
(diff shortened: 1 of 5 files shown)


Dridi:
> (note the difference to active_vcl which is mgt's notion of the current active VCL)

Note: nobody objected to renaming `active_vcl` to `mgt_vcl_active` after #3676 or this is merged, I'm planning to take care of it.

> A criticism of #3676 has been that on the mgt side it will grab the first VCL loaded that isn't specifically requested loaded cold as the active_vcl when active_vcl is NULL. Having thought more about this, I think this is a fair criticism and can take users by surprise. This patch attempts to fix that issue by introducing a new state, where the active_vcl can be non-NULL and cold, but only on startup and then refusing child startup while in this state. While the asserts already in place could be fixed up to allow this, it adds complexity.
>
> I believe it would be better to instead implement the logic that mgt's active_vcl is only set for the very first loaded VCL as long as that is loaded as warm. If it was loaded cold, mgt's active_vcl remains NULL. Which is already a managed and known state, so no changes should be needed except the very_first_load_temp_check. It also has the added benefit of keeping the invariants for mgt's active_vcl and the child's vcl_active the same, that is if non-NULL it points to a warm VCL and once set it can not become NULL again.

There is a difference between state and temperature: inside the cache process we only have temperatures cold and warm dictated by mgt and applied by mgt, and inside mgt we have states cold, warm and auto. I'm leaving the label state aside on purpose.

What if it was the auto state that would imply a `vcl.use` if there is no active VCL yet?

mbgrydeland:
> Note: nobody objected to renaming `active_vcl` to `mgt_vcl_active` after #3676 or this is merged, I'm planning to take care of it.

:+1: 

> There is a difference between state and temperature: inside the cache process we only have temperatures cold and warm dictated by mgt and applied by mgt, and inside mgt we have states cold, warm and auto. I'm leaving the label state aside on purpose.
> 
> What if it was the auto state that would imply a `vcl.use` if there is no active VCL yet?

I'm not sure I understand fully what you mean here. Is it that, if we had the rule that the very first VCL loaded should become the mgt active_vcl only if it was warm, then that should also be executed if it was set to AUTO? If so, I agree. The auto state should then make it both warm and active.

nigoroll:
superseded by #3737

PR #3718, opened 2021-10-25
cli: New debug.shutdown.delay command

Dridi:
For the purpose of simulating a workload where the cache process takes significant time to shut down.


Dridi:
Approved during bugwash today.

PR #3722, opened 2021-10-26
Macros to simplify VCL_STRANDS use

nigoroll:
As much as `VCL_STRANDS` has it's merits, it is tedious to work with for simple cases like passing arguments to VRT functions.

These macro attempt to make life a little easier for vmod authors.

Motivated by https://github.com/kenshaw/libvmod-dns/pull/15/files

Usage examples:
```c
VCL_STRING on_ws = VRT_STRANDS_string(ctx, VSTR(some_function()));
VCL_STRANDS on_stack = VSTRS(4, "one", two_func(), "3", four_func());
```

BTW: I could not find a nice/clean way to count the number of arguments to a variadic macro, so I do not think we get around the requirement to declare the number of strings to pass.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -54,6 +54,7 @@
 #include "common/vsmw.h"
 #include "proxy/cache_proxy.h"
 
+// NOT using TOSTRANDS() to create a NULL pointer element despite n == 0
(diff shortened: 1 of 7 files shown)


bsdphk:
I think it is a good idea, but the macros needs to have "STRAND" in their name to give people a clue about what happens.

MKSTRAND[S] or TOSTRAND[S] or ${BIKESHED}_STRAND[S]

PR #3724, opened 2021-10-28
Whitelist some pcre2 match options

nigoroll:
which I think vmods will require eventually.

I would actually need `PCRE2_NOTBOL|PCRE2_NOTEOL` right now, but hope to get around it...


nigoroll:
As discussed with @Dridi recently, the options masks are a nice idea but do limit VRE users in a significant way, unless we kept them up to date.

PR #3727, opened 2021-10-29
VCL_REGEX: Support concatenation of constant strings

### LINKED ISSUE #3726: VCL_REGEX vs. STRAND-ish string
Having migrated [vmod_re](https://code.uplex.de/uplex-varnish/libvmod-re) to [VCL_REGEX](https://code.uplex.de/uplex-varnish/libvmod-re/commit/0ee2f4aed195f08df5f5cf61d9aa94d3db2fe64a), I noticed this regression in behavior:

Using a pattern constructed with the `+` operator (a STRAND-ish string) does not work any more as it did with a `STRING` argument:

```
new re = re.regex("foo" + "bar");
```

only test-compiles "foo" and then complains about `REGEX + REGEX` not being possible

nigoroll:
For the `REGEX` type, we now peek ahead after the first CSTR token if a `+` operator follows and, if yes, create the pattern as a concatenation.

Fixes #3726

### DIFF
--- a/lib/libvcc/vcc_utils.c
+++ b/lib/libvcc/vcc_utils.c
@@ -51,19 +51,38 @@
 void
 vcc_regexp(struct vcc *tl, struct vsb *vgc_name)
 {
+	struct vsb *pattern;
 	char buf[BUFSIZ];
 	vre_t *t;
 	int error, erroroffset;
+	struct token *t1;
 	struct inifin *ifp;
 
-	assert(tl->t->tok == CSTR);
+	pattern = VSB_new_auto();
+	AN(pattern);
 
-	t = VRE_compile(tl->t->dec, 0, &error, &erroroffset, 0);
+	assert(tl->t->tok == CSTR);
+	VSB_cat(pattern, tl->t->dec);
+
+	t1 = vcc_PeekToken(tl);
+	AN(t1);
+	while (t1->tok == '+') {
+		vcc_NextToken(tl);
+		SkipToken(tl, '+');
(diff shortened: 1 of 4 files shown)


lib/libvcc/vcc_utils.c
```diff
@@ -74,9 +92,10 @@ vcc_regexp(struct vcc *tl, struct vsb *vgc_name)
 	Fh(tl, 0, "static struct vre *%s;\n", buf);
 	ifp = New_IniFin(tl);
 	VSB_printf(ifp->ini, "\tVPI_re_init(&%s, ",buf);
-	EncToken(ifp->ini, tl->t);
+	VSB_quote(ifp->ini, VSB_data(pattern), -1, VSB_QUOTE_CSTR);
 	VSB_cat(ifp->ini, ");");
 	VSB_printf(ifp->fin, "\t\tVPI_re_fini(%s);", buf);
+	VSB_destroy(&pattern);
```
Dridi, line 98: If `t` was null, you don't destroy `pattern`.

Dridi (approved):

lib/libvcc/vcc_utils.c
nigoroll: thank you! 👍🏽

nigoroll:
This sounds like a sensible optimization which I think we did already have at some point in the past.

Dridi:
First steps in this direction: 0c96fc6597c3ba6da057bb43acb7f677e6b776f8...1636a31a316d1634aef4977da419ab0bc500a80a

PR #3730, opened 2021-11-05
vre pcre2 update proposal: using fast code path pcre2_jit_match for p…

devnexen:
…erformance bump.

pcre2_match calls internally pcre2_jit_match but does quite some sanity check
 work. calling pcre2_jit_match gives a noticeable performance gain (observed doing unit tests)
but only doing so if pcre2_jit_compile succeeds.

### DIFF
--- a/lib/libvarnish/vre.c
+++ b/lib/libvarnish/vre.c
@@ -51,6 +51,7 @@
 
 struct vre {
 	unsigned		magic;
+	int                     re_jit;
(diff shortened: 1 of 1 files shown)


nigoroll:
the idea sounded reasonable under the precondition that there actually _is_ a performance benefit, so why did you retract it?

devnexen:
it did not really work until I realize I forgot one detail about vre_capture :)

nigoroll:
-> #3731

Issue #3734, opened 2021-11-09
Cannot discard labeled VCL program xxx


michbsd:
Varnish 6.6.1 running on FreeBSD 12.2

I use labels, when starting varnishd  I load basic VCL before loading VCLs and assigning labels - once loaded, I switch to the VCL. 

Startup VCL (basic): default.vcl:

```
vcl 4.0;

/* We have to have a backend, even if we do not use it */
backend default { .host = "127.0.0.1"; }
```

Once started, I run the following script:

```
#!/bin/sh

/usr/local/bin/varnishadm vcl.load xxx_1 /usr/local/etc/varnish/xxx.vcl
/usr/local/bin/varnishadm vcl.load yyy_1 /usr/local/etc/varnish/yyy.vcl
/usr/local/bin/varnishadm vcl.load zzz_1 /usr/local/etc/varnish/zzz.vcl

/usr/local/bin/varnishadm vcl.label lbl_xxx xxx_1
/usr/local/bin/varnishadm vcl.label lbl_yyy yyy_1
/usr/local/bin/varnishadm vcl.label lbl_zzz zzz_1

/usr/local/bin/varnishadm vcl.load main_1 /usr/local/etc/varnish/main.vcl
/usr/local/bin/varnishadm vcl.use main_1

# remove temp "default VCL"
/usr/local/bin/varnishadm vcl.discard boot
```

The last command fails, I cannot discard the "boot" vcl. I get: 

```
varnish> vcl.discard boot
300
Cannot discard labeled VCL program xxx_1:
	lbl_xxx
```

```
varnish> vcl.list
200
available auto   cold       0  boot
available auto   warm       0  xxx_1         <-  (1 label)
available auto   warm       0  yyy_1    <-  (1 label)
available auto   warm       0  zzz_1         <-  (1 label)
available label  warm       0  lbl_xxx       ->  xxx_1 (1 return(vcl))
available label  warm       0  lbl_yyy  ->  yyy_1 (1 return(vcl))
available label  warm       0  lbl_zzz       ->  zzz_1 (1 return(vcl))
active    auto   warm       0  main_1
```

```
varnish> vcl.deps
200
boot
xxx_1
yyy_1
zzz_1
lbl_xxx        xxx_1
lbl_yyy   yyy_1
lbl_zzz        zzz_1
main_1         lbl_xxx
main_1         lbl_yyy
main_1         lbl_zzz
```

[...]

Issue #3721, opened 2021-10-26
7.0.1 planning


nigoroll:
collection of suggestions of tickets for inclusion in 7.0.1

Candidates to merge in the 7.0 branch:

* [x] #3706 46bbfe2584380dec4c9036cfdc4a28d6c3d1ab9c 1996366df42bf5c89e7f2e74a8871f247e4693a3
* [x] #3719 f5798fd1f9362e2a56933c457ead9b85e9d97c79
* [x] #3720 2fd71469f60e56683723b9fddb6ae34ed9b79997
* [x] #3726 
  * prep: 464dad5ebfb787d4723a20b6cbf51c0a5a7ce12b 5e1589bf469be312e9af6ee98053936d55cfb7ce
  * fix: 8a4a7cc612253267b5ff97d3d52130953618c3c2
* [x] #3738
* [x] #3737
* [x] #3732

left out because ENOTREADY

* #3725
  * #3724 maybe also to complement pcre2 support?

bsdphk:
We should consider adding #3737 to the list.

nigoroll:
Note on #3732 :
it seems there was some misunderstanding yesterday about that merge, both @bsdphk and @mbgrydeland just said they want it in.

nigoroll:
@mbgrydeland will look after remaining steps to release, thank you

PR #3738, opened 2021-11-12
Use correct list var name in mgt_vcl_discard_depcheck

mbgrydeland:
Using the wrong list variable would cause the dependency check to consider
VCLs that were not attempted discarded.

Fixes: #3734


Dridi (approved):
I'm adding this to #3721.

PR #3717, opened 2021-10-25
Offer a configure --with-unwind switch (6.0)

Dridi:
```
Conflicts:
	.travis.yml
	bin/varnishd/cache/cache_panic.c
```
---

Test cases m35 to m38 are currently failing on my machine, two of the failures seem to be caused by a segmentation fault in `funlockfile()`. As *usual* (emphasis on usual) libunwind made this much easier to figure out than the libexecinfo back-trace, so I'm submitting a back-port of @gquintard's patch.

PR #3739, opened 2021-11-12
build: Use libunwind if available at configure time

Dridi:
One can still use the --without-unwind configure flag.

Refs #3717


Dridi:
I force-pushed a rename of `has_unwind` to `have_unwind` for consistency, the patch is otherwise identical.

mbgrydeland (approved):
LGTM

PR #3743, opened 2021-11-23
Rework HSH_Purge()

mbgrydeland:
HSH_Purge() suffers from a problem when there are more objcores on a given
objhead than we can hold temporary references for on the available
workspace, and multiple purges for the same objhead are issued at the same
time.

The previous algorithm would use a flag on the objcores (OC_F_PURGED) in
order to keep track of how far in the list we had gotten. This becomes a
problem when multiple threads are purging the same objhead at the same
time. When a new purge is started, the flags would be reset, which would
cause any existing purge thread to restart. If a steady stream of purges
for a given objhead is issued and enough objcores are present, none would
finish.

This rework uses OC references in the list as bookmarks, in order to know
how far into the list we were when releasing the mutex partway through and
want to resume again. This relies on the list not being reordered while we
are not holding the mutex. The only place where that happens is in
HSH_Unbusy(), where an OC_F_BUSY OC is moved first in the list. This does
not cause problems because we skip OC_F_BUSY OCs.

The OC_F_PURGED flag is removed as it is no longer needed.

The existing r02372.vtc test case exercises this code.

### DIFF
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -672,82 +672,101 @@ unsigned
 HSH_Purge(struct worker *wrk, struct objhead *oh, vtim_real ttl_now,
     vtim_dur ttl, vtim_dur grace, vtim_dur keep)
 {
-	struct objcore *oc, **ocp;
-	unsigned spc, ospc, nobj, n, n_tot = 0;
-	int more = 0;
+	struct objcore *oc, *oc_nows[2], **ocp;
+	unsigned i, j, n, n_max, total = 0;
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
-	ospc = WS_ReserveAll(wrk->aws);
-	assert(ospc >= sizeof *ocp);
-	/*
-	 * Because of "soft" purges, there might be oc's in the list that has
-	 * the OC_F_PURGED flag set. We do not want to let these slip through,
-	 * so we need to clear the flag before entering the do..while loop.
-	 */
-	Lck_Lock(&oh->mtx);
-	assert(oh->refcnt > 0);
-	VTAILQ_FOREACH(oc, &oh->objcs, hsh_list) {
-		CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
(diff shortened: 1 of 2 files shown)


bin/varnishd/cache/cache_hash.c
```diff
@@ -672,82 +672,103 @@ unsigned
...
+	n_max = WS_ReserveLumps(wrk->aws, sizeof *ocp);
+	if (n_max < 2) {
+		/* No space on the workspace. Give it a stack buffer of 2
+		 * elements, which is the minimum for the algorithm
+		 * below. */
+		ocp = oc_nows;
+		n_max = 2;
+	} else {
+		assert(PAOK(wrk->aws->f));
+		ocp = (void *)wrk->aws->f;
```
Dridi, line 690: You should do this instead:

```c
ocp = WS_Reservation(wrk->aws);
```

Dridi (changes requested):
Only a minor change request in this branch, we try not to access `struct ws` fields directly outside of `cache_ws*.c`.

bin/varnishd/cache/cache_hash.c
mbgrydeland: Thanks, fixed

Dridi (approved):

bin/varnishd/cache/cache_hash.c
```diff
@@ -672,82 +672,101 @@ unsigned
...
-		assert(oc->objhead == oh);
-		oc->flags &= ~OC_F_PURGED;
-	}
-	Lck_Unlock(&oh->mtx);
 
-	do {
-		more = 0;
-		spc = ospc;
-		nobj = 0;
+	n_max = WS_ReserveLumps(wrk->aws, sizeof *ocp);
```
nigoroll, line 681: We should have plenty of stack space here because of the de-facto minimum of [32k for pcree jit](https://pcre.org/current/doc/html/pcre2jit.html#stackcontrol), so we could either just use the stack, or at least use more stack elements as a fallback

nigoroll (approved):

bin/varnishd/cache/cache_hash.c
nigoroll: FWIW, `wrk->aws` _is_ the stack also...

bsdphk:
I see no problems with this, so: OK.

Issue #3740, opened 2021-11-18
Failing u00006 leaves varnishlog processes running


bsdphk:
Not sure if this is something varnishtest can do anything about (See also #2469) or if the test needs to be rethought.

Dridi:
That's because we start the leaking varnishlog as a daemon, so it exits the test's process group.

Workaround, lower the leak window:

```diff

--- bin/varnishtest/tests/u00006.vtc
+++ bin/varnishtest/tests/u00006.vtc
@@ -10,11 +10,6 @@ varnish v1 -vcl+backend {} -start
 process p1 {
        exec varnishlog -n ${v1_name} -g raw -k 3 -w ${tmpdir}/vlog -A
 } -start
-shell {
-       exec varnishlog -n ${v1_name} -D -P ${tmpdir}/vlog.pid \
-           -w ${tmpdir}/vlog.bin -R 10/s \
-           -i RespStatus -i ReqURL -i BereqURL
-}
 
 shell -match "Usage: .*varnishlog <options>" \
        "varnishlog -h"
@@ -124,6 +119,14 @@ shell {grep -q "0 CLI" ${tmpdir}/vlog}
 shell -match "0 CLI[ ]+- Wr 200 [0-9]+ PONG" \
        {varnishlog -n ${v1_name} -d -g raw -X "Wr 200 [0-9]+ [^P]"}
 
+shell {
+       exec varnishlog -n ${v1_name} -D -P ${tmpdir}/vlog.pid \
+           -w ${tmpdir}/vlog.bin -R 10/s \
+           -i RespStatus -i ReqURL -i BereqURL
+}
+
+delay 0.1
+
 client c1 {
        txreq -url /foo
        rxresp
```

Maybe we could detect child processes daemonizing and keeping track of them per test process to act accordingly, timeout or not?

Dridi:
```
**** top   shell_out|         0 Debug          - "sockopt: Not setting unmodified SO_LINGER for a0=127.0.0.1:62816%00"
**** top   shell_out|         0 Debug          - "sockopt: Not setting unmodified SO_KEEPALIVE for a0=127.0.0.1:62816%00"
**** top   shell_out|         0 Debug          - "sockopt: Setting SO_SNDTIMEO for a0=127.0.0.1:62816%00"
**** top   shell_status = 0x0000
**   top   === shell {grep -q "0 CLI" ${tmpdir}/vlog}
[...]

PR #3744, opened 2021-11-23
vut: Write to stdout with a "-w -" option

Dridi:
As suggested in the https://github.com/varnishcache/varnish-cache/issues/3740#issuecomment-975240592 summary, I'm adding "-" meaning `stdout` for the `-w` option for parity with the `-r` option. Because VUTs daemonize closing the standard streams, it is now treated as an error to both daemonize and read or write from `stdin` or `stdout`.

Instead of adding errors as suggested for `-A` without `-w`, it is documented as a no-op in this situation, and the same is done for `-a` and `-u`.

### DIFF
--- a/bin/varnishlog/varnishlog.c
+++ b/bin/varnishlog/varnishlog.c
@@ -72,9 +72,12 @@ openout(int append)
 {
 
 	AN(LOG.w_arg);
-	if (LOG.A_opt)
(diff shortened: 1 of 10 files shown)


nigoroll (approved):

Dridi:
Bugwash: OK

Dridi:
Rebased against master, I will merge once CI completes.

PR #3745, opened 2021-11-26
Fix unnecessary log buffer flushing

Dridi:
When vsl_reclen (maximum length of a record) is close to vsl_buffer,
the vsl buffer is flushed to the main shared memory log for every
record. This is because we test if there is space for vsl_reclen, not
the actual size of the formatted log record (which likely is much
smaller).

Fix this by doing the formatting in up to two stages. If the first
stage shows that we overflowed, but would not have overflown
vcl_reclen if there was more room in the buffer, then flush and redo
the formatting. This limits the extra work of formatting twice only to
the one record that happened to overflow the buffer.

### DIFF
--- a/bin/varnishd/cache/cache_shmlog.c
+++ b/bin/varnishd/cache/cache_shmlog.c
@@ -409,6 +409,8 @@ VSLbv(struct vsl_log *vsl, enum VSL_tag_e tag, const char *fmt, va_list ap)
 	const char *u, *f;
 	unsigned n, mlen;
 	txt t;
+	va_list ap2;
+	int first;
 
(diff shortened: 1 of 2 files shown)


nigoroll:
why not bprintf into a stack buffer of size `vsl_reclen`?

Dridi:
First, `bprintf()` is fatal when overflowed, it would force an additional copy inside the VSL buffer despite most log records being below `vsl_reclen`, then that would force us to dynamically allocate that space on the stack:

```c
char buf[cache_param->vsl_reclen];
```

Bugwash consensus: @bsdphk OK, @nigoroll on board, to be merged.

PR #3746, opened 2021-11-26
std.ip: Always provide some form of fallback

Dridi:
Otherwise valid code can panic on workspace exhaustion:

    std.ip(req.http.X-Real-IP, std.ip(req.http.X-Client-IP, client.ip))

If the nested std.ip() call runs out of workspace, it will return a null
ip instead of the fallback, and since the outer std.ip() call is getting
a fallback argument, it will panic upon checking the suckaddr sanity.

Refs a3b26d0ba7dd8763a4e34d9e18d92fc24f1c291c

### DIFF
--- a/vmod/vmod_std_conversions.c
+++ b/vmod/vmod_std_conversions.c
@@ -200,17 +200,23 @@ vmod_ip(VRT_CTX, struct VARGS(ip) *a)
 {
 	uintptr_t sn;
 	void *p;
-	VCL_IP retval = NULL;
+	VCL_IP retval = NULL, fb = bogo_ip;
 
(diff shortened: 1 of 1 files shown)


bsdphk:
I think this should be a `VRT_fail()` ?

Dridi:
We could.

In the scenario I got where the nested `std.ip()` call runs out of workspace, the nested call `VRT_fail`s but the parent call panics upon getting a null IP.

nigoroll (approved):

Dridi:
I will submit a patch for the 6.0 branch too.

Dridi:
https://github.com/varnishcache/varnish-cache/pull/3748

PR #3748, opened 2021-11-29
std.ip: Always provide some form of fallback (6.0)

Dridi:
Otherwise valid code can panic on workspace exhaustion:

    std.ip(req.http.X-Real-IP, std.ip(req.http.X-Client-IP, client.ip))

If the nested std.ip() call runs out of workspace, it will return a null
ip instead of the fallback, and the outer std.ip() call will panic upon
checking the suckaddr sanity.

Refs #3746


hermunn:
Bugwash consensus: Merge it!

PR #3753, opened 2021-12-13
param: Infrastructure for deprecated aliases

Dridi:
With this change, we can formalize the renaming of a parameter while
maintaining the old name temporarily for compatibility.

A deprecated alias can be set with either `param.set` or the `-p` option,
but won't be listed by:

- `param.show [-j]`
- `param.show [-j] changed`
- `param.show -l`

Only an explicit `param.show` for the name of the alias will provide a
minimal documentation with a deprecation notice and the current value.
In the manual, there is only a deprecation notice. The rationale is
that administration tools shouldn't pick them up when enumerating the
parameters.

Since we currently don't have deprecated parameters, this can only be
tested manually, for example:

    PARAM_ALIAS(vcl_dir, vcl_path)

To ensure that we don't break this, we could consider having a perpetual
deprecated parameter.

### DIFF
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -101,10 +101,12 @@ static const char ONLY_ROOT_TEXT[] =
 	"NB: This parameter only works if varnishd is run as root.";
 
 static const char NOT_IMPLEMENTED_TEXT[] =
-	"This parameter depends on a feature which is not available"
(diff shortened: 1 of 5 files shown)


nigoroll (approved):
have not tried the patch, but it sounds like a good idea

Dridi:
Added test coverage, I will merge once CI completes.

PR #3754, opened 2021-12-13
vmod_cookie docs: name and value are separated by equal sign

ciencia:
Cookie name and value are separated by equal sign, not colon.

Also, returned cookies are always terminated by a semicolon.

See tests


Dridi:
While this docfix is technically correct, based on what the VMOD does, I didn't catch where in the RFC a trailing semi-colon is mentioned:

https://datatracker.ietf.org/doc/html/rfc6265#section-5.4

Dridi:
See https://github.com/varnishcache/varnish-cache/pull/3791 for a comprehensive fix+docfix.

PR #3756, opened 2021-12-14
cli: param.show -u shows the actual user input

Dridi:
It can be challenging to have a consistency check of certain parameters
because of the way they are displayed. For example bytes parameters are
pretty-printed when possible in plain text, and plainly-printed in JSON
mode.

The new -u option for param.show adds an \<Input is: XXX> line in plain
text and an \<"input": "XXX"> field in JSON mode where XXX is the exact
user input for the current value of the parameter. Consistency checks
can use the user input to verify expectations and update parameters
accordingly, without changing the usual param.show output.

### DIFF
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -522,7 +522,7 @@ MCF_ParamProtect(struct cli *cli, const char *args)
 void
 MCF_ParamSet(struct cli *cli, const char *param, const char *val)
 {
-	const struct parspec *pp;
+	struct parspec *pp;
 
 	pp = mcf_findpar(param);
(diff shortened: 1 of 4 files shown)


nigoroll (approved):

Dridi:
Alternative suggestion:

```
param.set -j default_ttl 100
200
[ 1, ["param.set", "-j", "default_ttl", "100"], timestamp,
  {
    "old": 120.000,
    "new": 100.000,
  }
]
```

Dridi:
bugwash: let's use `param.set -j` to report the value as shown by `param.show` for subsequent consistency checks.

I will come up with something.

Dridi:
See #3771 for the `param.set` alternative.

PR #3758, opened 2021-12-16
vcc: New SYM_ALIAS kind for deprecated symbols

Dridi:
Their sole purpose is to resolve to another symbol. The script expects aliases to be defined like this in the RST docs:

    my.alias ``VCL <= 4.1``

        Type: DEPRECATED

        Alias of: something.else

        A little description is still necessary.

This should facilitate renaming variables in the future. One thing we could do is to have libvcc warn about deprecated aliases (subject to a new `vcc_err_alias` parameter too) to make this visible not just in the manual.

There is no `DEPRECATED` type in addition to the new `SYM_ALIAS` kind, this is a purely cosmetic suggestion.

Deprecated aliases should probably have a high VCL version limit like in the example above, `generate.py` doesn't enforce it so far.

Aliases have zero runtime cost, since ultimately the aliased symbol is used at compile time instead of the alias itself. Renaming a variable and keeping the old name as an alias will result in VRT breakage.

---

This is the VCL counterpart of #3753, and the first variable to be renamed this way is `req.esi_level`.

### DIFF
--- a/bin/varnishtest/tests/r01332.vtc
+++ b/bin/varnishtest/tests/r01332.vtc
@@ -15,6 +15,7 @@ varnish v1 -vcl+backend {
 	sub vcl_deliver {
 		objx.enum(martin);
 		set resp.http.foo = objx.foo("");
+		set resp.http.bar = objx.bar("");
 	}
 } -start
 
@@ -23,4 +24,5 @@ client c1 {
 	rxresp
 	expect resp.status == 200
 	expect resp.http.foo == "BOO"
+	expect resp.http.bar == "BOO"
 } -run
--- a/include/tbl/symbol_kind.h
+++ b/include/tbl/symbol_kind.h
@@ -33,6 +33,7 @@
 
 VCC_KIND(NONE,		none)
 VCC_KIND(RESERVED,	reserved)
+VCC_KIND(ALIAS,		alias)
 VCC_KIND(ACL,		acl)
 VCC_KIND(ACTION,	action)
 VCC_KIND(BACKEND,	backend)
--- a/lib/libvcc/generate.py
+++ b/lib/libvcc/generate.py
@@ -169,18 +169,22 @@ def varproto(s):
         varprotos[s] = True
 
 class vardef(object):
-    def __init__(self, nam, typ, rd, wr, wu, vlo, vhi):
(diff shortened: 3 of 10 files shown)


Dridi:
bugwash:

- prepare the esi to sub rename as a VIP
- introduce SYM_ALIAS once we have actionable aliases

Dridi:
It was brought to my attention today that there are VMODs we maintain that could benefit from deprecated aliases, so vmodtool could also grow an `$Alias` stanza. As it turns out, even our vmod_cookie could use this system since we forgot to polish the API before its first in-tree release.

Dridi:
Rebased against current master. I split the commits into more logical chunks and added an alias to vmod_cookie instead of the premature `req.sub_level`.

nigoroll (approved):
Generally 👍🏽 but I think a segregation of Aliases in the generated docs would be helpful.
Also I can see a point in phks suggestion to have `$Alias <newname> <existing>`

Dridi:
> Generally 👍🏽 but I think a segregation of Aliases in the generated docs would be helpful.
> Also I can see a point in phks suggestion to have `$Alias <newname> <existing>`

I'm looking at it and trying to figure how to do this. It's not very difficult, but then I get this awkward manual structure:

```
NAME

SYNOPSIS

DESCRIPTION

bar()

    description

qux()

    description

... more symbols ...

DEPRECATED

``foo()`` is a deprecated alias for ``bar()``

``baz()`` is a deprecated alias for ``qux()``

...

SEE ALSO

COPYRIGHT
```

With the current approach `foo` is documented as such right after `bar`, which makes it both simpler to implement and easier to follow. With a disjoint deprecated section, unlike the one in vmod_std for example that describes new symbols of their own, it becomes really inconvenient outside of sphinx docs, in a terminal.

Also, unlike actual deprecated functions, that should be demoted in their own section, aliases aren't supposed to refer to deprecated symbols. Only the old name is deprecated.

Dridi:
I discussed this with @bsdphk yesterday and we still want to go for `$Alias <alias> <symbol>` so in the VCC file the syntax changed to these:

```
$Alias alias function

[Optional description]

$Alias .alias obj.method

[Optional description]
```

Only `vmodtool.py`, `vmod_cookie.vcc` and `vmod_debug.vcc` had to change, the rest remains identical.

PR #3766, opened 2021-12-30
param: Optionally pass a duration unit to timeouts

Dridi:
Allowing units is much more convenient and keeping them optional
maintains compatibility. On the other hand, changing the defaults
for default_ttl and default_grace has no impact on the generated
RST.

Keeping track of the user input could help since default values
are "washed" like user input, without resorting to a dynamic
default.

Refs #3756

### DIFF
--- a/bin/varnishd/mgt/mgt_param_tweak.c
+++ b/bin/varnishd/mgt/mgt_param_tweak.c
@@ -52,34 +52,44 @@ const char * const JSON_FMT = (const char *)&JSON_FMT;
  * Generic handling of double typed parameters
  */
 
+static double
+parse_decimal(const char *p, const char **err)
+{
+	double v;
+
+	v = SF_Parse_Decimal(&p, 0, err);
+	if (errno == 0 && *p != '\0') {
+		errno = EINVAL;
+		*err = "Invalid number";
+	}
+	return (v);
+}
+
 static int
 tweak_generic_double(struct vsb *vsb, const char *arg, const struct parspec *pp,
     const char *fmt)
 {
 	volatile double u, minv = VRT_DECIMAL_MIN, maxv = VRT_DECIMAL_MAX;
 	volatile double *dest = pp->priv;
-	const char *p, *err;
+	const char *err;
 
 	if (arg != NULL && arg != JSON_FMT) {
 		if (pp->min != NULL) {
-			p = pp->min;
-			minv = SF_Parse_Decimal(&p, 0, &err);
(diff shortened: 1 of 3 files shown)


nigoroll (commented):
not looked at the details of the code yet, but 👍🏽 on the interface change

bin/varnishd/mgt/mgt_param_tweak.c
```diff
@@ -67,29 +69,29 @@ parse_decimal(const char *p, const char **err)
 
 static int
 tweak_generic_double(struct vsb *vsb, const char *arg, const struct parspec *pp,
-    const char *fmt)
+    parse_double_f parse, const char *fmt)
```
bsdphk, line 72: I would have made the parse argument default to parse_decimal if NULL.

bin/varnishd/mgt/mgt_param_tweak.c
```diff
@@ -437,8 +434,10 @@ tweak_poolparam(struct vsb *vsb, const struct parspec *par, const char *arg)
 			    par->dyn_max_reason);
 			if (retval)
 				break;
-			retval = tweak_generic_double(vsb,
-			    &px.max_age, av[3], "0", "1000000", "%.0f");
+			pt.priv = &px.max_age;
+			pt.min = "0";
+			pt.max = "1000000";
+			retval = tweak_generic_double(vsb, av[3], &pt, "%.0f");
```
bsdphk: So this is pretty much exactly why I made those arguments explicit to `tweak_generic_double` in the first place.

I'm not a fan of this ephemeral parspec, but I can live with it.

Dridi:
Rebased against master to solve the merge conflict after having pushed e2220de36d2e67d523a6b4cfc834756e7510dfee directly.

bin/varnishd/mgt/mgt_param_tweak.c
Dridi: I made a u-turn here because the function signature was requiring more effort than the temporary parspec.

bin/varnishd/mgt/mgt_param_tweak.c
Dridi: Do we really need a default for a static function with 3 call sites?

bsdphk:
OK with me.

Dridi:
@nigoroll apologies, I merged thinking you had already approved but no.

Let me know if you find something. I made sure to formally request your review on other tickets approved by @bsdphk.

PR #3767, opened 2021-12-31
esi: Implement <esi:include onerror="continue"/>

Dridi:
This changes the default behavior of includes, matching the ESI language specification. The onerror attribute seems to only accept one value, so in its absence a failed ESI fragment delivery aborts the top request delivery.

In addition, it breaks the ESI object attribute format for persisted caches.

---

This is OK for trunk to change the default behavior and objattr format, but for the 6.0 LTS branch we would need to gate this behind a feature flag. I'm wondering whether this is fine to have such an `esi_include_onerror` flag in 6.0 that does not exist in trunk.

### DIFF
--- a/bin/varnishd/cache/cache_esi.h
+++ b/bin/varnishd/cache/cache_esi.h
@@ -39,7 +39,8 @@
 #define	VEC_S1	(0x60 + 1)
 #define	VEC_S2	(0x60 + 2)
 #define	VEC_S8	(0x60 + 8)
-#define	VEC_INCL	'I'
+#define	VEC_INCL_ABRT	'I'
+#define	VEC_INCL_CONT	'i'
 
 typedef ssize_t vep_callback_t(struct vfp_ctx *, void *priv, ssize_t l,
     enum vgz_flag flg);
--- a/bin/varnishd/cache/cache_esi.h
+++ b/bin/varnishd/cache/cache_esi.h
@@ -39,8 +39,8 @@
 #define	VEC_S1	(0x60 + 1)
 #define	VEC_S2	(0x60 + 2)
 #define	VEC_S8	(0x60 + 8)
-#define	VEC_INCL_ABRT	'I'
-#define	VEC_INCL_CONT	'i'
+#define	VEC_IA	(0x40 + 9)
+#define	VEC_IC	(0x60 + 9)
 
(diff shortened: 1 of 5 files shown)


bin/varnishd/cache/cache_esi.h
```diff
@@ -39,7 +39,8 @@
 #define	VEC_S1	(0x60 + 1)
 #define	VEC_S2	(0x60 + 2)
 #define	VEC_S8	(0x60 + 8)
-#define	VEC_INCL	'I'
```
nigoroll, line 42: swap `i` and `I`?

nigoroll (approved):

Dridi:
bugwash: also gate this feature behind a feature flag in trunk.

Dridi:
Rebased against master, added a feature flag.

bin/varnishd/cache/cache_esi.h
Dridi: I'm tempted to use `A` and `C` and retire `I` altogether.

bsdphk: 'A' is nogo, it is already used for VEC_V1 (0x40+1)

I would prefer we stuck to hex definitions throughout to avoid such issue going forward.

bin/varnishd/cache/cache_esi_deliver.c
```diff
@@ -899,4 +904,9 @@ ved_deliver(struct req *req, struct boc *boc, int wantbody)
 		req->doclose = SC_REM_CLOSE;
 
 	req->acct.resp_bodybytes += VDP_Close(req->vdc);
+
+	if (i && ecx->abrt) {
+		req->top->topreq->vdc->retval = -1;
+		req->top->topreq->doclose = req->doclose;
```
bsdphk, line 910: This looks deeply suspect to me.  Doesn't the VDP stack not already ensure this ?

Dridi: See test case, apparently not.

Dridi:
bugwash: OK once all markers are in the `(HEX + offset)` form.

Dridi:
For reference: 582ded6a2d6ae1a4467b1eb500f2725b42888016

Issue #3765, opened 2021-12-29
getnameinfo = -12 Unknown error


jlecour:
Hi

I've been using Varnish in production for a while and recently, I've notices some logs that I can't figure out the cause :

```
Dec 28 10:02:38 my-hostname varnishd[31113]: Info: Child (31126) said getnameinfo = -12 Unknown error
Dec 28 10:02:38 my-hostname varnishd[31113]: Child (31126) said getnameinfo = -12 Unknown error
```

I get many of them, randomly throughout the day.
The host doesn't seem to have any issue resolving DNS.

I didn't find a way to reproduce with a basic configuration yet, but I might find some time to do so in the coming weeks.

The bug appears in this environment : 

```
# uname -a
Linux boost-par00 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux

# cat /etc/debian_version 
10.11

# dpkg -l varnish | grep varnish
ii  varnish        6.1.1-1+deb10u1 amd64        state of the art, high-performance web accelerator
```

Thanks for any help or idea to help figuring this out.

Dridi:
On my system it's this one:

```c
# define EAI_OVERFLOW     -12   /* Argument buffer overflow.  */
```

bricewge:
More specifically, the [manual](https://www.mankier.com/3/getnameinfo#Return_Value-EAI_OVERFLOW) says it is an overflow of the second or third pair of arguments:
> EAI_OVERFLOW
>     The buffer pointed to by host or serv was too small.

[...]

PR #3768, opened 2022-01-05
vcc: Teach HEADER symbols to accept constant strings

Dridi:
Static strings are unfortunately copied into workspace simply because
they don't belong to it. Since it is safe to reference a static string,
or at the very least a string that is guaranteed to outlive a VCL
transaction since in this case it has the same scope as the VCL shared
object, we can bypass the regular STRANDS assignment.

When a transaction runs out of workspace, fails and goes to vcl_synth,
we don't want to fail again because of a static VCL assignment of the
content-type and retry-after headers from the built-in VCL.

Refs #3765

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -634,6 +634,32 @@ VRT_SetHdr(VRT_CTX, VCL_HEADER hs, const char *pfx, VCL_STRANDS s)
 	http_SetHeader(hp, b);
 }
 
+VCL_VOID
+VRT_SetHeader(VRT_CTX, VCL_HEADER hs, const char *hdr)
+{
+	VCL_HTTP hp;
+	unsigned wl, hl;
+
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	AN(hs);
+	AN(hs->what);
+	AN(hdr);
+
+	hl = strlen(hdr);
+	wl = hs->what[0];
+	assert(hl > wl);
+	AZ(strncasecmp(hs->what + 1, hdr, wl));
+	hp = VRT_selecthttp(ctx, hs->where);
+	CHECK_OBJ_NOTNULL(hp, HTTP_MAGIC);
+
+	if (FEATURE(FEATURE_VALIDATE_HEADERS) && !validhdr(hdr + wl)) {
+		VRT_fail(ctx, "Bad static header %s", hdr);
+		return;
+	}
+	http_Unset(hp, hs->what);
+	http_SetHeader(hp, hdr);
+}
+
 /*--------------------------------------------------------------------*/
 
 VCL_VOID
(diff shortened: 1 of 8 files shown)


Dridi:
I added cef6f007b722a7358d2ba083343cc9b956bf4f1e to show that `vcc_ExprEdit()` can be useful for more than just constant HEADER assignments, it can also help simplify libvcc.

The general pattern is:

```diff
-       Fb(tl, 1, "BEFORE\n");
-       tl->indent += INDENT;
-       vcc_Expr(tl, FOO);
-       tl->indent -= INDENT;
+       vcc_ExprEdit(tl, FOO, "BEFORE\v+\n\v1\v-AFTER\n", NULL);
        ERRCHK(tl);
-       Fb(tl, 1, "AFTER\n");
```

There may be other types of symbols that could benefit from a constant assignment, but HEADER symbols seem to be the only obvious one with on one hand one less STRANDS detour and one less workspace allocation.

nigoroll:
I came back here after we found agreement on #4269. I am not sure if I expressed my thoughts on this ticket clearly enough, so here is an attempt to make good on that.

My understanding of the premise of this ticket is that when we have `set x.http.a = "foo"`, we can can generate a header string `"a: foo"` and basically pass this as is (via some intermediate functions) to `http_SetH()`. This is certainly a good idea.

The additional idea I mentioned is that we can also optimize `set x.http.a = y.http.a` for the same reason, because `y.http` already has the `a: <whatever>` header, we just don't have the API to return it.

My hope was that, if we taught VCC to handle the `"name: value"` string instead of `HEADER` and `"value"`, this would more or less fall out as a byproduct, but looking at this PR again, it now seems to me that the implementation is indeed less general than I initially thought when I commented on it in #4269.

So now my question is: Is it worth it? I can see that the optimization for VCL constants makes sense, in particular for avoiding out-of-workspace conditions on error paths, but the `set x.http.a = y.http.a` case is probably an exotic edge case.

The other case which we _have_ optimized with #4269 is `set resp.reason = resp.http.foo`, but that did not need any VCC changes.

----

(This section might be going to far, please do not take it as a concrete proposal...)

Another question is, if we are going to touch this area anyway, are there better options, still?

One obvious idea would be to represent headers as

```c
struct hdnv {
	unsigned namelen, valuelen;	// 1, 3
	const char *name;		// "a: "
	const char *value;		// "foo"
}
```

Compared to our ``txt``, we would increase storage for header pointers and the number of I/O vectors needed to send HTTP/1 headers by 50% [^1], but we could easily avoid copying in "all" cases, because we could set name and value independently.

[^1]: Currently we send `txt + "\r\n"`, so 50% would be the worst case addition because we could detect the common `name + namelen + 2 == value` and coalesce, but probably kernels do this already anyway?

nigoroll:
Just for completeness, again, I do not want to push this topic:

> I'm not up for touching how we use this format to pack headers in storage.

The storage representation would not change, it would still be "header: value". What would change is how our API looks at it. At the moment, it has a txt double pointer, the idea mentioned here is instead one "name" and one "value" pointer, both with a length. For headers from storage, the "value" pointer would point to the same string as "name".

PR #3771, opened 2022-01-05
cli: Make param.set -j output the updated parameter

Dridi:
It uses the same output as `param.show`, so applications may be able to reuse an existing parsing facility out of the box.

Closes #3756


bsdphk:
OK with me.

nigoroll (approved):

PR #3772, opened 2022-01-10
filter: Pass a VRT_CTX to the init methods

Dridi:
This allows VDPs and VFPs to pass state from a PRIV_TASK to respectively
a vdp_entry or a vfp_entry. When a filter is set up by a VMOD, there is
otherwise no way to configure a filter on a per-task basis.


bsdphk:
OK with me.

nigoroll (approved):

PR #3780, opened 2022-02-03
Ci update

gquintard:
move things around a bit, but most importantly removes `centos:8` in favor of `fedora:latest`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,14 +176,15 @@ jobs:
                     ;;
             esac
 
-            set -x
-            echo "FROM $IMG" > Dockerfile
-            echo "ADD make-$EXT-packages.sh /usr/bin/" >> Dockerfile
-            echo "CMD make-$EXT-packages.sh" >> Dockerfile
-
             cp .circleci/make-$EXT-packages.sh .
-            docker build -t docker_img .
-            docker run --rm -it -e PARAM_DIST=$(echo "<< parameters.platform >>" | cut -d: -f1) -e PARAM_RELEASE=$(echo "<< parameters.platform >>" | cut -d: -f2) -v$(pwd):/varnish-cache docker_img
+            docker run \
+              --rm \
+              -it \
+              -e PARAM_DIST=$(echo "<< parameters.platform >>" | cut -d: -f1) \
+              -e PARAM_RELEASE=$(echo "<< parameters.platform >>" | cut -d: -f2) \
+              -v$(pwd):/varnish-cache \
+              $IMG \
+              /varnish-cache/make-$EXT-packages.sh
       - run:
           name: List created packages
           command: find ./packages -type f
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,8 +160,8 @@ jobs:
             mkdir -p packages
             case "<< parameters.platform >>" in
                 debian:*|ubuntu:*)  EXT=deb ;;
-                centos:*)         EXT=rpm ;;
-                alpine:*)         EXT=apk ;;
+                centos:*|fedora:*)  EXT=rpm ;;
+                alpine:*)           EXT=apk ;;
                 *)
                     echo "unrecognized platform: << parameters.platform >>"
                     exit 1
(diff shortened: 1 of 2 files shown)


gquintard:
it will need https://github.com/varnishcache/pkg-varnish-cache/pull/153 (or something like it) for the packages to build though

.circleci/config.yml
```diff
@@ -176,14 +176,15 @@ jobs:
...
-            docker build -t docker_img .
-            docker run --rm -it -e PARAM_DIST=$(echo "<< parameters.platform >>" | cut -d: -f1) -e PARAM_RELEASE=$(echo "<< parameters.platform >>" | cut -d: -f2) -v$(pwd):/varnish-cache docker_img
+            docker run \
+              --rm \
+              -it \
+              -e PARAM_DIST=$(echo "<< parameters.platform >>" | cut -d: -f1) \
+              -e PARAM_RELEASE=$(echo "<< parameters.platform >>" | cut -d: -f2) \
+              -v$(pwd):/varnish-cache \
+              $IMG \
+              /varnish-cache/make-$EXT-packages.sh
```
Dridi, line 187: Can't we skip the copy (technically tainting the local git clone) and run this instead?

```
[...] \
    $IMG \
    /varnish-cache/.circleci/make-$EXT-packages.sh
```

.circleci/config.yml
```diff
@@ -353,9 +351,13 @@ workflows:
...
-          name: distcheck_centos_8
-          dist: centos
-          release: "8"
+          name: distcheck_fedora_stable
+          dist: fedora
+          release: latest
+      - distcheck:
+          name: distcheck_fedora_rawhide
+          dist: fedora
+          release: rawhide
```
Dridi, line 356: I would prefer `distcheck_fedora_latest` to match the image tag.

.circleci/make-rpm-packages.sh
```diff
@@ -13,13 +13,9 @@ elif [ -z "$PARAM_DIST" ]; then
...
 
-yum install -y epel-release
 
 if [ "$PARAM_DIST" = centos ]; then
-  if [ "$PARAM_RELEASE" = 8 ]; then
-      dnf install -y 'dnf-command(config-manager)'
-      yum config-manager --set-enabled powertools
-  fi
+    yum install -y epel-release
 fi
```
Dridi, line 19: This adjustment should happen before fedora's introduction, modulus el8 removal.

Dridi (changes requested):
I pushed the arch changes and the machine executor update, but please leave the centos:8 removal out for now.

.circleci/config.yml
gquintard: That would be smarter indeed

gquintard:
Are you sure you want to keep CentOS:8? The repos seem to have been failing for some time now

Dridi:
> Are you sure you want to keep CentOS:8? The repos seem to have been failing for some time now

I want to handle centos-8 separately, and I'm considering centos-stream (and/or centos-stream[89]) as a replacement.

gquintard:
tweaked and force-pushed. I left this change https://github.com/varnishcache/varnish-cache/pull/3780/commits/e39438ba919cdda73b7b738f5a93dd2d92717f89#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L223 because it works now, and I don't feel like shooting in the dark trying to get compatibility with a distribution I can't test

.circleci/config.yml
Dridi: Why mark this one as resolved?

.circleci/config.yml
```diff
@@ -176,15 +176,14 @@ jobs:
...
-            cp .circleci/make-$EXT-packages.sh .
             docker run \
               --rm \
               -it \
               -e PARAM_DIST=$(echo "<< parameters.platform >>" | cut -d: -f1) \
               -e PARAM_RELEASE=$(echo "<< parameters.platform >>" | cut -d: -f2) \
               -v$(pwd):/varnish-cache \
               $IMG \
-              /varnish-cache/make-$EXT-packages.sh
+              /varnish-cache/.circleci/make-$EXT-packages.sh
```
Dridi, line 186: I see, you did that in a followup commit instead.

PR #3782, opened 2022-02-14
Invalidate vxid caches when the vxid wraps

### LINKED ISSUE #3775: VSL client panics on AN(keep) in vslq_next()
An application that uses the VSL API has intermittently experienced panics in production due to [this assertion](https://github.com/varnishcache/varnish-cache/blob/master/lib/libvarnishapi/vsl_dispatch.c#L1372) in vslq_next(), called by VSLQ_Dispatch().

The site is actually running Varnish 6.6, but the relevant part of vslq_next() appears to be unchanged in master:

```
        vtx = vtx_lookup(vslq, vxid);
        keep = tag != SLT_Begin || vslq_candidate(vslq, ptr);
        if (vtx == NULL && tag == SLT_Begin && keep) {
                vtx = vtx_add(vslq, vxid);
                AN(vtx);
        }
        if (vtx != NULL) {
                AN(keep);
                r = vtx_append(vslq, vtx, &c->rec, len);
                if (r == vsl_more)
                        vtx_scan(vslq, vtx);
        }
```
It seems possible to me that `keep` may be zero in the `vtx != NULL` clause if `tag` is SLT_Begin and vslq_candidate() returned false. The VSL client uses request grouping (VSL_g_request), which narrows down the possible reasons for vslq_candidate() to return false, none of which seem to be out of the question, as far as I can tell.

The panic is rare. The client reads millions of transactions per day, and the problem has been observed a handful of times in recent months. Unfortunately I don't know a way to reproduce it.

I don't see an obvious way for the VSL client to avoid this situation. The motivations for the assertion are not quite clear to me, but if `keep == 0` indicates an intractable problem, would it be possible for vslq_next() (and hence VSLQ_Dispatch()) to return an error status when it happens, rather than an assertion? A panic leads to a fairly significant loss of data, whereas an error for the transaction would only lead to loss of data for that transaction.

nigoroll:
To avoid vxid clashes, we add a vxid generation number which is increased every time the global vxid counter wraps.

Any vxid caches from a previous generation are invalidated.

This patch should fix #3775 if the latest hypothesis is correct that transaction inconsistencies in the vsl reader are caused by vxid clashes due to vxid caching on the varnishd side.

Implementation notes:

- The global generation counter value is deliberately read outside the lock to keep the performance impact minimal. It should not matter if an outdated value is read momentarily, because the time scale of vxids is some orders of magintudes larger than that of SMP cache inconsistencies.

- This implementation invalidates vxid caches earlier than it absolutely needed to, but I decided to go with this idea for its simplicity in contrast to calculating a distance between the global vxid base and the cached value.

### DIFF
--- a/bin/varnishd/cache/cache_main.c
+++ b/bin/varnishd/cache/cache_main.c
@@ -182,24 +182,29 @@ THR_Init(void)
  */
 
 static uint32_t vxid_base;
-static uint32_t vxid_chunk = 32768;
+static uint32_t vxid_chunk = 32768, vxid_gen = 0;
 static struct lock vxid_lock;
 
 uint32_t
 VXID_Get(const struct worker *wrk, uint32_t mask)
 {
 	struct vxid_pool *v;
+	uint32_t vxid_new;
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(wrk->wpriv, WORKER_PRIV_MAGIC);
 	v = wrk->wpriv->vxid_pool;
 	AZ(VXID(mask));
 	do {
-		if (v->count == 0) {
+		if (v->gen != vxid_gen || v->count == 0) {
 			Lck_Lock(&vxid_lock);
 			v->next = vxid_base;
 			v->count = vxid_chunk;
-			vxid_base = (vxid_base + v->count) & VSL_IDENTMASK;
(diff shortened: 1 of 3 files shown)


Dridi:
This change is made obsolete by plans to move to 64bit VXIDs, but I would like to keep it open until I port it to the 6.0 branch.

nigoroll:
I am going to flag it with r=6.0. Please close when done

PR #3784, opened 2022-02-14
circleci: Add fedora jobs and migrate centos:8 jobs

Dridi:
Closes #3780 #3702

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,8 +160,8 @@ jobs:
             mkdir -p packages
             case "<< parameters.platform >>" in
                 debian:*|ubuntu:*)  EXT=deb ;;
-                centos:*)         EXT=rpm ;;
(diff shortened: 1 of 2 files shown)


gquintard:
> from the messy https://github.com/varnishcache/varnish-cache/pull/3780

how dare you! But also, yes.

> If someone with cci privileges could rerun this job, I would appreciate the help.

I don't have those right anymore :-(

Regarding the packaging jobs, it probably doesn't make sense anymore, they are fast, and parallelism has been greatly enhanced, so we don't need to pick which job should start first. I say run everything every time

Dridi:
I think I'm fine with the nightly scheme, if only to avoid the job explosion on a daily basis.

gquintard:
the explosion doesn't really affect committers, except if they push repetitively within 20 minutes, and that's not really the norm, but up to you

PR #3785, opened 2022-02-15
circleci: Update packages on archlinux

Dridi:
We see linking failures sometimes, so hopefully that solves those.


Dridi:
I looked at the manual suspecting I could do a single `pacman -Syu <list of packages>` but couldn't find a definitive answer.

gquintard:
you definitely can, there's no harm in doing it in two lines though

PR #3789, opened 2022-03-11
Fold bits parameters into the main parameters table

Dridi:
I wasn't actively thinking about it, and this solution just popped in my head, and it worked on my machine.


bsdphk:
OK with me.

nigoroll (approved):

PR #3791, opened 2022-03-14
vmod_cookie: Avoid generating a trailing semi-colon

Dridi:
The separator between cookies in a cookie header is "; " and we should
not generate it in a way that adds a spurious ';' at the end of the
cookie list.

This also bundles the docfix from #3754, further docfixed.

Closes #3754


doc/changes.rst
```diff
@@ -31,6 +31,14 @@ http://varnish-cache.org/docs/trunk/whats-new/index.html and via
 individual releases. These documents are updated as part of the
 release process.
 
+===============================
+Varnish Cache NEXT (2022-09-15)
+===============================
+
+* Cookie headers generated by vmod_cookie no longer have a spurious trailing
```
nigoroll, line 38: 👍🏽

nigoroll (approved):

PR #3793, opened 2022-03-17
circleci: Revamp Ubuntu jobs

Dridi:
Use their version number as the image tag instead of the code name for
clarity. Add a packaging job for the incoming Ubuntu 22.04, and add
distcheck jobs for Ubuntu 20.04 and 22.04 with a comment explaining
which Ubuntu is arbitrarily selected for 32bit builds.

Also, stop pretending that CentOS is a dpkg-based distribution, we are
not fooling anyone.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -393,10 +393,15 @@ workflows:
           name: distcheck_fedora_rawhide
           dist: fedora
           release: rawhide
+      # latest debian uses sanitizers
       - distcheck:
           name: distcheck_debian_buster
           dist: debian
           release: buster
+      - distcheck:
+          name: distcheck_debian_bullseye
+          dist: debian
+          release: bullseye
           extra_conf: --enable-asan --enable-ubsan --enable-workspace-emulator
       # oldest ubuntu goes 32bit
       - distcheck:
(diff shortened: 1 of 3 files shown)


Dridi:
@mbgrydeland pointed out that this changes package versions and may jeopardize their upgrade path. So instead I'm only adding missing jobs and additional routine maintenance.

.circleci/config.yml
```diff
@@ -210,6 +210,10 @@ jobs:
       release:
         description: the release name (stretch|buster|bionic|focal)
         type: string
+      check:
```
mbgrydeland, line 213: To avoid confusion, maybe we should rename the job from 'distcheck' to just 'build', and call the param 'buildtarget', and edit the 'name' below from "Extract and distcheck" to "Build (<< param.buildtarget >>)"

.circleci/config.yml
```diff
@@ -385,24 +389,40 @@ workflows:
           name: distcheck_almalinux_8
           dist: almalinux
           release: "8"
+      # fedora is our witness
       - distcheck:
           name: distcheck_fedora_latest
```
mbgrydeland, line 394: rename to 'witness_fedore_latest'?

Dridi: Not if we rename them all to `build_<dist>_<release>`?

Dridi:
Rebased against master:

- renamed jobs from `distcheck_*` to `build_*`
- added jobs on almalinux 9

Dridi:
https://github.com/varnishcache/varnish-cache/pull/3793/commits/57de4298e7a54d955342434e76661e4853c4b369 came out all green, I will clean up the branch and merge.

PR #3798, opened 2022-04-11
std.cache_req_body(BYTES size, BOOL partial = 0)

Dridi:
Last week my employer organized among other things a hackathon day (thanks!) and I made this seemingly work. I revisited the code and reorganized the commits the next day during my flight, but didn't find much to say in an area of the code base not too familiar to me. I guess it's time to throw more eyeballs at it.

Quoting from the main patch of the series:

> req.body: New BS_PARTIAL body status for caching
> 
> Partial caching allow VMODs to fetch and inspect part of the body. This
> gives the same properties as BS_CACHED, but allows the request body to
> be larger than what is stored. If a bereq send failures happens within
> the partial cached body delivery, it also becomes eligible to retries.
> 
> This enables a more robust caching and retry policy where not knowing
> the complete body size (chunked or h2 without content-length) or on the
> other hand knowing that a body is larger (content-length) does not lead
> to VCL failures.
> 
> It can also be valuable to inspect the beginning of a request body. For
> example an image upload service could parse images metadata directly in
> Varnish to enforce rules such as specific file formats or how large in
> pixels an image may be.
> 
> One limitation left on purpose is that BS_PARTIAL objects will be sent
> with chunked encoding to the backend, even when we know the length.

See individual commits for more details.


nigoroll:
Because `Transfer-Encoding: chunked` is uncommon for request bodies, I would expect some backends to choke on it and would thus think a mode to foward the client's `Content-Length` would be good. On IRC, @bsdphk rightly pointed out the need for extra care and attention due to smuggling potential, and I agree that we should look into how to handle, for example, the case when a client fails to send the announced length.

Dridi:
This morning I replaced https://github.com/varnishcache/varnish-cache/pull/3798/commits/c8ce29fc47e1a323335ee5252379b0b5c1ab0afa with 99f21e5c98...fc0a497200, then I added more coverage in the test case.

Basically:

- BS_PARTIAL is replaced by a `req_body_cached` flag
- another `req_body_partial` flag is added

bin/varnishd/cache/cache_req_body.c
```diff
@@ -43,6 +43,47 @@
...
+		assert(req_bodybytes == l0);
+		assert(req_bodybytes == l);
+		AZ(http_GetHdr(req->http0, H_Transfer_Encoding, NULL));
+		AZ(http_GetHdr(req->http, H_Transfer_Encoding, NULL));
+	} else {
+		/* We must update also the "pristine" req.* copy */
+		AZ(http_GetHdr(req->http0, H_Content_Length, NULL));
+		assert(l0 < 0);
+		http_Unset(req->http0, H_Transfer_Encoding);
+		http_PrintfHeader(req->http0, "Content-Length: %ju",
```
nigoroll, line 78: This problem existed before the proposed change:
I think this conflicts with [Req_Rollback()](https://github.com/varnishcache/varnish-cache/blob/142c3a012935ceb68ffd6fe753f5498eedd0c8ed/bin/varnishd/cache/cache_req.c#L220), because the header is created on the rolled back part of the client workspace.
Because the case is likely not particularly relevant and given that [a variant of the problem already exists](https://github.com/varnishcache/varnish-cache/blob/142c3a012935ceb68ffd6fe753f5498eedd0c8ed/bin/varnishd/cache/cache_req_body.c#L303), I think we might just veto rollback with a cached body.

bin/varnishd/cache/cache_req_body.c
```diff
@@ -43,6 +43,47 @@
...
+	AZ(req->req_body_cached);
+	assert(req_bodybytes >= 0);
+
+	l0 = http_GetContentLength(req->http0);
+	l = http_GetContentLength(req->http);
+
+	assert(req->req_body_status->avail > 0);
+	if (req->req_body_status->length_known) {
+		assert(req_bodybytes == l0);
+		assert(req_bodybytes == l);
```
nigoroll: As `req.http.Content-Length` can be modified from VCL, I think we should rather either fix the header or fail the request rather than panicking.

bin/varnishd/cache/cache_req_body.c
```diff
@@ -43,6 +43,47 @@
...
+		AZ(http_GetHdr(req->http0, H_Transfer_Encoding, NULL));
+		AZ(http_GetHdr(req->http, H_Transfer_Encoding, NULL));
```
nigoroll, line 72: same ⬆️

bin/varnishd/cache/cache_req_body.c
```diff
@@ -61,11 +61,16 @@ vrb_cached(struct req *req, ssize_t req_bodybytes)
 
 	assert(req->req_body_status->avail > 0);
 	if (req->req_body_status->length_known) {
-		assert(req_bodybytes == l0);
-		assert(req_bodybytes == l);
+		if (req->req_body_partial) {
+			assert(req_bodybytes < l0);
+			assert(req_bodybytes < l);
+		} else {
+			assert(req_bodybytes == l0);
```
nigoroll, line 68: see comment on previous commit: we should rather not panic

bin/varnishd/cache/cache_req_body.c
```diff
@@ -224,12 +258,17 @@ ssize_t
...
+	unsigned allow_partial;
 	int i;
 
+	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	AN(vsl);
 	AN(func);
 
-	if (req->req_body_cached) {
+	allow_partial = (req->vsl == vsl) && (func != httpq_req_body_discard);
```
nigoroll, line 269: I am not sure about these semantics.
Sure we want to blacklist `httpq_req_body_discard`, but other than that the caller should be in control if they want to process the body partially or fully.
But maybe this is for another PR...

nigoroll (approved):
Generally 👍🏽 
Some of my comments should probably be addressed before merging...

bin/varnishd/cache/cache_req_body.c
Dridi: Different VSL simply means different thread in this context. If you are in a backend task you can only consume the entire body.

simonvik:
The following test that i stole from vmod_bodyaccess can assert varnish with "Condition((bo->req->body_oc) != 0) not true." 

```varnishtest "chunked test"

server s1 {
    rxreq
    txresp -hdr "OK: yes"
    rxreq
    txresp -hdr "OK: yes"
    rxreq
    txresp -hdr "OK: yes"
} -start

varnish v1 -cli "param.set vsl_mask +Hash"

varnish v1 -vcl+backend {
    import std;

    sub vcl_recv {
        if(req.method == "POST"){
                std.cache_req_body(std.bytes(req.http.cache, 10KB));
        }
    }

    sub vcl_deliver {
         if (obj.hits > 0) {
             set resp.http.X-Cache = "HIT";
         } else {
             set resp.http.X-Cache = "MISS";
         }
         set resp.http.x-len = req.http.x-len;
     }

} -start

client c1 {
    txreq -req POST -body "BANANE"
    rxresp
    expect resp.http.ok == "yes"
    expect resp.http.X-Cache == "MISS"

    txreq
    rxresp
    expect resp.http.ok == "yes"
    expect resp.http.X-Cache == "MISS"
} -run

client c2 {
    txreq -req POST -nolen -hdr "Transfer-encoding: chunked"
    chunked {BANA}
    delay .2
    chunked {NE}
    delay .2
    chunkedlen 0
    rxresp
    expect resp.http.ok == "yes"
    expect resp.http.X-Cache == "HIT"
} -run

```

Not 100% sure why but someone probably has an idea

Dridi:
```diff
--- bin/varnishd/cache/cache_req_fsm.c
+++ bin/varnishd/cache/cache_req_fsm.c
@@ -893,6 +893,7 @@ cnt_recv_prep(struct req *req, const char *ci)
                req->hash_always_miss = 0;
                req->hash_ignore_busy = 0;
                req->hash_ignore_vary = 0;
+               req->req_body_cached = 0;
                req->client_identity = NULL;
                req->storage = NULL;
        }
```

We lack coverage of keepalive after a cached req body for HTTP/1.

bsdphk:
How does this work with H2 ?

I guess it causes head-of-line-blocking, since I see nothing tweaking the H2 windows to stop the client from pushing ahead ?

Do we care ?

Dridi:
I don't think we need to care about h2, the rxbuf from #3661 is here to prevent head of line blocking.

FWIW, I plan to address the panic reported by @simonvik like this:

```diff
--- i/bin/varnishd/cache/cache_req_body.c
+++ w/bin/varnishd/cache/cache_req_body.c
@@ -125,6 +125,7 @@ vrb_pull(struct req *req, ssize_t maxsize, unsigned partial,
                AN(oa_len);
                req_bodybytes = oa_len;
                AZ(HSH_DerefObjCore(req->wrk, &req->body_oc, 0));
+               req->req_body_cached = 0;
                req->req_body_partial = 0;
        }
 
@@ -345,10 +346,14 @@ VRB_Free(struct req *req)
 
        CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 
-       if (req->body_oc == NULL)
+       AZ(req->req_body_partial);
+       if (req->body_oc == NULL) {
+               AZ(req->req_body_cached);
                return;
+       }
 
        r = HSH_DerefObjCore(req->wrk, &req->body_oc, 0);
+       req->req_body_cached = 0;
 
        // each busyobj may have gained a reference
        assert (r >= 0);
diff --git i/bin/varnishd/cache/cache_req_fsm.c w/bin/varnishd/cache/cache_req_fsm.c
index de2a4b8584..37840ccd14 100644
--- i/bin/varnishd/cache/cache_req_fsm.c
+++ w/bin/varnishd/cache/cache_req_fsm.c
@@ -895,6 +895,8 @@ cnt_recv_prep(struct req *req, const char *ci)
                req->hash_ignore_vary = 0;
                req->client_identity = NULL;
                req->storage = NULL;
+               AZ(req->req_body_cached);
+               AZ(req->req_body_partial);
        }
 
        req->is_hit = 0;
```

The idea is to clear the `req_body_cached` flag when `req->body_oc`'s reference is dropped, something that used to be inherent to the assignment of `req_body_status`. I'll try to find some time to carefully review this aspect of the change.

Dridi:
Rebased against current master:

- flags are now cleared consistently and asserted
- test coverage added in 706844a18395eb9d8fdbab0eee018a157a17062b

bin/varnishd/cache/cache_req_body.c
```diff
@@ -43,15 +43,63 @@
...
+		AZ(http_GetHdr(req->http, H_Transfer_Encoding, NULL));
+	} else if (!req->req_body_partial) {
+		/* We must update also the "pristine" req.* copy */
+		AZ(http_GetHdr(req->http0, H_Content_Length, NULL));
+		assert(l0 < 0);
+		http_Unset(req->http0, H_Transfer_Encoding);
+		http_PrintfHeader(req->http0, "Content-Length: %ju",
+		    (uintmax_t)req_bodybytes);
+
+		AZ(http_GetHdr(req->http, H_Content_Length, NULL));
```
nigoroll, line 81: Also FTR:
The `req.http0` changes are logged, which confused me at first thinking that the `vrb_pull()` might be called twice:
```
**** v1    vsl|       1004 ReqUnset        c Transfer-encoding: chunked
**** v1    vsl|       1004 ReqHeader       c Content-Length: 110
**** v1    vsl|       1004 ReqUnset        c Transfer-encoding: chunked
**** v1    vsl|       1004 ReqHeader       c Content-Length: 110
```

nigoroll:
24fe3772c520700b32711eb3b466dc1737327546 exposes that we have an issue with chunked encoding: When the body cache size exactly matches the length, we fail caching.

The trivial options are:

* ignore (and probably docfix) that, for chunked encoding, the buffer needs to be at least one byte larger than the actual length
* make the buffer one byte larger than specified

But this is only the symptom of another, possibly more relevant issue: the v1f chunked vfp currently can not send `VFP_END` with the last bytes. To solve this issue, we need to peek into the next chunk header before returning `VFP_OK`. I am currently checking how this could look like in the code.

nigoroll:
> To solve this issue, we need to peek into the next chunk header before returning `VFP_OK`.

This is #3809

I would prefer to get it done before continuing here.

nigoroll:
My continuation of this PR to implement VMOD access to partially cached request bodies is not available at https://github.com/nigoroll/varnish-cache/tree/partial_req_body_vrb_iterate2

nigoroll:
Updated https://github.com/nigoroll/varnish-cache/tree/partial_req_body for the modified #3809

bin/varnishd/cache/cache_req_body.c
nigoroll: You have my sympathies if you are saying this from a feeling of frustration.
But I actually think we should not not have any panics which can be triggered from VCL.

Also I wanted to take note of an idea from bugwash: For the rollback case, can we make the code DTRT even with the unmodified headers?

Dridi: > You have my sympathies if you are saying this from a feeling of frustration.
> But I actually think we should not not have any panics which can be triggered from VCL.

Actually no, I'm genuinely thinking that if we leave dragons in VCL, we can't afford a gentle failure here. If you both mess with framing headers and request body caching things are going horribly wrong and should blow up.

> Also I wanted to take note of an idea from bugwash: For the rollback case, can we make the code DTRT even with the unmodified headers?

One thought I didn't discuss during bugwash is that body status shouldn't be the responsibility of the HTTP connection. We miss some separation of concerns in `struct http_conn` that was OK when HTTP/1 was the only game in town. There are more things to disentangle.

nigoroll: In the course of my [continuation of this work](https://github.com/varnishcache/varnish-cache/pull/3798#issuecomment-1148576467) I replaced the code with [just setting the respective headers](https://github.com/nigoroll/varnish-cache/commit/c371d92c45ac3ff530cca7301c9721adc408248e) because, with filters on the request body (but before caching), we can not make any assumptions on the relation between the content-length before/after caching.

dridi:
Some changes originating from this pull request (but going one step further) were submitted in #4125 and pushed upstream (ce3e446...f0e9df8). Partial request body caching should be revisited once trailers are supported.

PR #3799, opened 2022-04-11
Add -fno-var-tracking-assignments to cc_command if accepted by the compiler

### LINKED ISSUE #3795: gcc variable tracking size limit exceeded
A client has seen this warning from the VCC C-Compiler:

```
Message from C-compiler:
vgc.c: In function ‘VGC_function_recv_5f_branch.part.0’:
vgc.c:21941:1: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
 21941 | VGC_function_recv_5f_branch(VRT_CTX, enum vcl_func_call_e call,
       | ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Should we try if our compiler accepts `-fno-var-tracking-assignments` and, if yes, add that to the default `cc_command`?

`cc_command` is:
```
exec gcc -g -O2  -fdebug-prefix-map=/build/varnish-2HaA1g/varnish-20220326+focal=.  -fstack-protector-strong -Wformat -Werror=format-security %w -pthread   -fpic -shared -Wl,-x -o %o %s
```

nigoroll:
Closes #3795

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -803,21 +803,40 @@ if test "$ac_cv_env_VCC_CC_set" = "set"; then
 	VCC_CC="$ac_cv_env_VCC_CC_value"
 else
 	case $target in
+	*-*-darwin*)
+		VCC_CC="cc $OCFLAGS"
+		;;
+	*)
+		VCC_CC="$PTHREAD_CC $OCFLAGS"
+		;;
+	esac
+
+	save_CFLAGS="$CFLAGS"
+	save_CC="$CC"
+	CFLAGS=
+	CC="$VCC_CC"
+	AX_CHECK_COMPILE_FLAG(
+		[-fno-var-tracking-assignments],
+		[VCC_CC="$VCC_CC -fno-var-tracking-assignments"])
+	CFLAGS="$save_CFLAGS"
+	CC="$save_CC"
+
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -802,22 +802,41 @@ AC_ARG_VAR([VCC_CC], [C compiler command line for VCL code])
 if test "$ac_cv_env_VCC_CC_set" = "set"; then
 	VCC_CC="$ac_cv_env_VCC_CC_value"
 else
+	case $target in
+	*-*-darwin*)
+		VCC_CC="cc $OCFLAGS"
+		;;
```
Dridi, line 808: Why not `$CC`?

nigoroll: > Why not `$CC`?

no idea. I just preserved what was there before:

https://github.com/varnishcache/varnish-cache/pull/3799/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810L817

Dridi (approved):

configure.ac
Dridi: Someone with a vested interest in macos should probably look at this, but for this pull request it's out of scope.

PR #3806, opened 2022-05-14
workaround for non-existent PATH_MAX

janmojzis:
e.g. GNU HURD doesn't define PATH_MAX


bsdphk:
Sorry, just go with a local #ifdef, all the compat gunk adds nothing of value.

PR #3686, opened 2021-09-06
HTTP1: Read data on backend connection reset

Algunenano:
If the backend closes a connection, try to read any available data to see if there is information about it.
This is to deal with common behaviour on some web servers that close the connection as soon as they find an issue with it, for example a large content length.

References https://github.com/varnishcache/varnish-cache/issues/2332

The PR is missing tests (I need to learn how they work and how to recreate this setup within the tests) but I wanted to open it in order to receive feedback as soon as possible as this is my first contribution to the project.

An example with a server that returns `400 Bad Request` based on Content-Length and closes the connection to interrupt the upload:

Before:

```
*              << BeReq    >>   3         
-            3 Begin          b bereq 2 pass
-            3 VCL_use        b boot
-            3 Timestamp      b Start: 1630953265.053809 0.000000 0.000000
-            3 BereqMethod    b POST
-            3 BereqURL       b /v0/datasources?name=stock_prices_1M&mode=append
-            3 BereqProtocol  b HTTP/1.1
-            3 BereqHeader    b Host: localhost:6081
-            3 BereqHeader    b User-Agent: curl/7.78.0
-            3 BereqHeader    b Accept: */*
-            3 BereqHeader    b Authorization: Bearer p.eyJ1IjogImRlMzExNzI4LWY3M2ItNGRmOS05ZWFlLTA2Yzc0Y2VkZGFlNCIsICJpZCI6ICIzYTFiYTQzMC0zYzNlLTRlN2QtYjcyZS0wOTljMjE1NmNlYTQifQ.1kpgmxugmlRk6WYRB38Z6bQskT9d9qnwpEb0BAz01gM
-            3 BereqHeader    b Content-Length: 84248385
-            3 BereqHeader    b Content-Type: application/x-www-form-urlencoded
-            3 BereqHeader    b X-Forwarded-For: 127.0.0.1
-            3 BereqHeader    b X-Varnish: 3
-            3 VCL_call       b BACKEND_FETCH
-            3 VCL_return     b fetch
-            3 Timestamp      b Fetch: 1630953265.053821 0.000012 0.000012
-            3 Timestamp      b Connected: 1630953265.053908 0.000098 0.000086
-            3 BackendOpen    b 26 server1 127.0.0.1 8001 127.0.0.1 56358 connect
-            3 FetchError     b req.body read error: 104 (Connection reset by peer)
-            3 FetchError     b backend write error: 104 (Connection reset by peer)
[...]

Issue #3813, opened 2022-05-31
#3761 collateral damage


nigoroll:
Seen with https://github.com/nigoroll/varnish-cache/tree/partial_req_body rebased onto master

tests/r01772:

```
***  v1    debug|Error: Child (889346) died signal=6
***  v1    debug|Error: Child (889346) Panic at: Tue, 31 May 2022 09:44:53 GMT
***  v1    debug|Assert error in vbe_dir_finish(), cache/cache_backend.c line 246:
***  v1    debug|  Condition(PFD_State(pfd) == PFD_STATE_USED) not true.
***  v1    debug|version = varnish-trunk revision 363dac3df57903b9e5b79128f714f3217aebd8c2, vrt api = 15.0
***  v1    debug|ident = Linux,5.17.0-1-amd64,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,epoll
***  v1    debug|now = 267243.087689 (mono), 1653990293.601019 (real)
***  v1    debug|Backtrace:
***  v1    debug|  0x45f9bd: varnishd() [0x45f9bd]
***  v1    debug|  0x45f710: varnishd() [0x45f710]
***  v1    debug|  0x52b785: varnishd(VAS_Fail+0x55) [0x52b785]
***  v1    debug|  0x41fb25: varnishd() [0x41fb25]
***  v1    debug|  0x433218: varnishd(VDI_Finish+0x288) [0x433218]
***  v1    debug|  0x44061c: varnishd() [0x44061c]
***  v1    debug|  0x445751: varnishd() [0x445751]
***  v1    debug|  0x44176f: varnishd() [0x44176f]
***  v1    debug|  0x4a042e: varnishd() [0x4a042e]
***  v1    debug|  0x49fada: varnishd() [0x49fada]
```

nigoroll:
It's also showing up in [vtest](varnish-cache.org/vtest/) together with f00001 failures

nigoroll:
I am going to revert it for now because the cause is not immediately obvious to me, I need to focus on other tasks and Dridis availability this week is rather limited.

PR #3761, opened 2021-12-17
vbe: Try fetching beresp when sending bereq failed

Dridi:
There is no cancellation mechanism in the middle of an HTTP/1
transaction besides closing the session. If a backend closes
the connection before the end of the bereq delivery, it may
also send a 4XX or 5XX response that could be valuable for the
original client.

In this scenario, we can process the backend response, and we
can use the bereq.send_failed flag to decide what to do about
broken transactions. The beresp.fetch_failed flag can never
be raised in vcl_backend_response but is made available to
enable code sharing with vcl_backend_error as demonstrated in
r02332.

The test case r02332 is a combination of the 3 test cases
from #3686.

There are several things to take into consideration: it is
the responsibility of the backend implementation to raise the
send_failed and fetch_failed flags, if the backend replied
something before closing we shouldn't trigger a timeout, the
connection must not be recycled, and the object must not be
cached.

Fixes #2332
Closes #3686

PR #3814, opened 2022-05-31
vbe: Try fetching beresp when sending bereq failed

### LINKED ISSUE #2332: Varnish loses response when backend returns an error during a POST
Varnish may lose the backend response when a backend immediately returns an error when varnish is still sending  a request body.

Example of this situation:

When a POST is received from varnish, and varnish sends the request to an apache backend with 
 a content-length that exceeds LimitRequestBody, apache will immediately return a 413 (Request Entity Too Large).

```
HTTP/1.1 413 Request Entity Too Large
Date: Tue, 23 May 2017 19:51:33 GMT
Connection: close
Content-Type: text/html; charset=iso-8859-1

<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>413 Request Entity Too Large</title>
</head><body>
<h1>Request Entity Too Large</h1>
The requested resource<br />/app/foo.php<br />
does not allow request data with POST requests, or the amount of data provided in
the request exceeds the capacity limit.
</body></html>
```

At this point, varnish appears to be in a loop reading/writing the request body from the client to the backend, unaware that the backend has sent an error.

Assuming the complete request body has not been sent, apache will wait until the client closes the connection or it hits the default linger timeout of 30 seconds, which at that point will forcefully close the socket.

The forceful shutting of the socket causes varnish to react with:
```
-        65584 Debug          b "Write error, retval = -1, len = 6, errno = Broken pipe"
-        65584 FetchError     b req.body read error: 32 (Broken pipe)
-        65584 FetchError     b backend write error: 32 (Broken pipe)
```

At this point, varnish returns a backend fetch fail error because of the write failure and never reads the response from the backend, and goes straight to vcl_backend_error.

This is particularly a problem with large POSTs from slow spoon feeding clients. The request body cannot be sent to the backend before the linger timeout (30 seconds).

## Expected Behavior
The error message would be read and returned to the client. Ideally, varnish would immediately detect the backend error preventing the long linger timeout to begin with.

## Current Behavior
Varnish attempts to read the request body until fully read, or an error/timeout occurs even if the backend immediately responds with an error.

1. Configure an apache backend LimitRequestBody to a small limit, such as 10485760 
2. Create a 100MB file (dd if=/dev/zero of=largefile.txt bs=1M count=100)
3.  Slowly upload a file to varnish instance with apache backend with slow transfer rate: (curl -v --limit-rate 25K -F foo=@largefile.txt http://varnishendpoint/)

## Your Environment
* Version used: 4.1.6
* Operating System and version: CentOS 7

dridi:
There is no cancellation mechanism in the middle of an HTTP/1
transaction besides closing the session. If a backend closes
the connection before the end of the bereq delivery, it may
also send a 4XX or 5XX response that could be valuable for the
original client.

There are several things to take into consideration: it is
the responsibility of the backend implementation to raise the
bo->send_failed flag, if the backend replied something before
closing we shouldn't trigger a timeout, the connection must not
be recycled, and the object must not be cached.

Fixes #2332
Closes #3686

---

We need to address #3813 before proceeding.

Refs #3761

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -243,7 +243,8 @@ vbe_dir_finish(VRT_CTX, VCL_BACKEND d)
 		AZ(pfd);
 		Lck_Lock(bp->director->mtx);
 	} else {
-		assert (PFD_State(pfd) == PFD_STATE_USED);
+		assert(PFD_State(pfd) == PFD_STATE_USED);
+		AZ(bo->send_failed);
 		VSLb(bo->vsl, SLT_BackendClose, "%d %s recycle", *PFD_Fd(pfd),
 		    VRT_BACKEND_string(d));
 		Lck_Lock(bp->director->mtx);
@@ -268,6 +269,7 @@ vbe_dir_gethdrs(VRT_CTX, VCL_BACKEND d)
 	struct pfd *pfd;
 	struct busyobj *bo;
 	struct worker *wrk;
+	stream_close_t sc;
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	CHECK_OBJ_NOTNULL(d, DIRECTOR_MAGIC);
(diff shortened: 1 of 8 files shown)


nigoroll:
changes.rst which I had written before noticing the revert, dropping here for possible reuse:

```rst
* Previously, when sending a backend request failed, varnish would
  always generate a 503 response and disregard any response that the
  backend might have sent.

  Now a backend response is used if available, falling back to the
  varnish-generated response if not.  (2332_, 3686_)

.. _2332: https://github.com/varnishcache/varnish-cache/pull/2332
.. _3686: https://github.com/varnishcache/varnish-cache/pull/3686
```

nigoroll:
bugwash: we'd like to get this done during this release cycle

vvucetic:
Hey @nigoroll or @Dridi, would you be able to recommend some workaround until this is fixed/released?

In our case, header validation or rate limiting down the request line can terminate requests earlier and that wrongly returns 503 instead of 400 or 429. Pipe solves the issue, but it’s not an option for us.

PR #3815, opened 2022-06-02
Add a utility macro to allocate mini-objects on the workspace

nigoroll:
isn't this something we would have wanted all along?

This simplifies the common pattern to allocate/initialize a miniobj on the workspace to

```c
	WS_TASK_ALLOC_OBJ(ctx, myobj, MYOBJ_MAGIC);
	if (myobj == NULL)
		return (error);
```

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -850,6 +850,15 @@ const char *WS_Printf(struct ws *ws, const char *fmt, ...) v_printflike_(2, 3);
 void WS_VSB_new(struct vsb *, struct ws *);
 char *WS_VSB_finish(struct vsb *, struct ws *, size_t *);
 
+/* WS utility */
+#define WS_TASK_ALLOC_OBJ(ctx, ptr, magic) do {			\
+	ptr = WS_Alloc((ctx)->ws, sizeof *(ptr));		\
+	if ((ptr) == NULL)					\
+		VRT_fail(ctx, "Out of workspace for " #magic);	\
+	else							\
+		INIT_OBJ(ptr, magic);				\
+} while(0)
+
 /* cache_rfc2616.c */
 void RFC2616_Ttl(struct busyobj *, vtim_real now, vtim_real *t_origin,
     float *ttl, float *grace, float *keep);
--- a/vmod/vmod_cookie.c
+++ b/vmod/vmod_cookie.c
@@ -187,13 +187,10 @@ vmod_set(VRT_CTX, struct vmod_priv *priv, VCL_STRING name, VCL_STRING value)
 		return;
 	}
 
-	cookie = WS_Alloc(ctx->ws, sizeof *cookie);
-	if (cookie == NULL) {
-		VSLb(ctx->vsl, SLT_Error,
-		    "cookie: unable to get storage for cookie");
+	WS_TASK_ALLOC_OBJ(ctx, cookie, VMOD_COOKIE_ENTRY_MAGIC);
+	if (cookie == NULL)
 		return;
-	}
-	INIT_OBJ(cookie, VMOD_COOKIE_ENTRY_MAGIC);
+
 	cookie->name = WS_Printf(ctx->ws, "%s", name);
 	cookie->value = WS_Printf(ctx->ws, "%s", value);
 	if (cookie->name == NULL || cookie->value == NULL) {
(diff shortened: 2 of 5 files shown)


bsdphk:
Not against it.

Any idea how many places in the code it could have been used already ?

nigoroll:
> Did you consider a coccinelle patch?

Yes, but I deemed the patterns too different - for the given example https://github.com/varnishcache/varnish-cache/pull/3815/commits/afe8b2ca8fce84043b9467753097d519bbb5817e there are three different ways of error handling.

bin/varnishd/cache/cache.h
```diff
@@ -843,6 +844,16 @@ const char *WS_Printf(struct ws *ws, const char *fmt, ...) v_printflike_(2, 3);
 void WS_VSB_new(struct vsb *, struct ws *);
 char *WS_VSB_finish(struct vsb *, struct ws *, size_t *);
 
+/* WS utility */
+#define WS_TASK_ALLOC_OBJ(ctx, ptr, magic) do {			\
+	assert(sizeof *(ptr) <= UINT_MAX);			\
```
Dridi, line 849: Do we really need this assertion? I fail to see what it is concretely guarding against.

Dridi (dismissed):
Overall, LGTM! One nit.

bin/varnishd/cache/cache.h
nigoroll: `WS_Alloc() takes an `unsigned` argument, `sizeof`, obviously, returns a `size_t`.
The assertion implies that our mini-objects are no larger than `UINT_MAX`, which I consider reasonable (any allocation for an object size even getting near the limit would fail with reasonably sized workspaces).

Dridi: Precisely, on my system `UINT_MAX` is 4GB (minus 1B). What is the point of checking that a single ws allocation is below that?

Dridi: With `sizeof`? At this point you deserve it.

nigoroll: :sigh: whatever.

nigoroll: @Dridi in an effort to get back to a constructive discussion:

The `bytes` argument of `WS_Alloc()` is `unsigned`. For `WS_TASK_ALLOC_OBJ` to work, it obviously needs to cast a `size_t sizeof()` to `unsigned`. The assertion documents that this is intentional. I do not see any reason for not having it because the truth value will be known at compile time, so the assertion will usually be optimized out by the compiler.

But as this bit is controversial, I will remove it in the hope that compilers will detect problematic type conversion. Mine does, at least, as in this example where I made `struct cookie` larger than 4GB:

```
vmod_cookie.c:191:2: error: implicit conversion from 'unsigned long' to 'unsigned int' changes value from 4294967336 to 40 [-Werror,-Wconstant-conversion]
        WS_TASK_ALLOC_OBJ(ctx, cookie, VMOD_COOKIE_ENTRY_MAGIC);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../bin/varnishd/cache/cache.h:849:28: note: expanded from macro 'WS_TASK_ALLOC_OBJ'
        ptr = WS_Alloc((ctx)->ws, sizeof *(ptr));               \
              ~~~~~~~~            ^~~~~~~~~~~~~
1 error generated.
make: *** [Makefile:1112: libvmod_cookie_la-vmod_cookie.lo] Error 1
```

Dridi: I understand the theoretical downcast problem, but I don't see how someone would accidentally declare and use a struct in the GB order of magnitude.

Now let's say your struct has a size of `UINT_MAX` and it doesn't trigger the panic, "chances" are that the workspace allocation will fail anyway. I don't see the point of guarding against `> UINT_MAX` but not the unrealistic struct sizes `<= UINT_MAX`.

If your struct has a size of 4GB+5B and it's downcast to 5B, rounded up to 8B by workspace code, and you don't notice a thing, then it's really questionable whether we should actively help you in this instance.

I hope this one was also more constructive.

bin/varnishd/cache/cache.h
```diff
@@ -38,6 +38,7 @@
 #  error "cache.h included multiple times"
 #endif
 
+#include <limits.h>
```
Dridi, line 41: Is this still needed?

Dridi (approved):

bin/varnishd/cache/cache.h
nigoroll: thank you! 👍🏽

bin/varnishd/cache/cache.h
nigoroll: I agree that the case is not realistic. I have nothing to add wrt my initial intentions.

PR #3826, opened 2022-07-18
vsc: New MAIN.shm_bytes* counters

Dridi:
We have various stat counters for the shared memory log (shmlog),
but nothing to give us the number of bytes used (consumed).  To
address this, add new stat counters to the shmlog:

    shm_bytes
        Total number of bytes used in the shared memory log.
        This includes bytes unused when we wrap.

    shm_bytes_wrap
        Number of bytes unused when we wrap.

Signed-off-by: Dridi Boukelmoune <dridi.boukelmoune@gmail.com>

---

I'm submitting this change on behalf of @drodden from @varnish, and I bundled an additional trivial patch.

### DIFF
--- a/bin/varnishd/cache/cache_shmlog.c
+++ b/bin/varnishd/cache/cache_shmlog.c
@@ -193,6 +193,7 @@ vsl_get(unsigned len, unsigned records, unsigned flushes)
 	VSC_C_main->shm_writes++;
 	VSC_C_main->shm_flushes += flushes;
 	VSC_C_main->shm_records += records;
+	VSC_C_main->shm_bytes += VSL_BYTES(VSL_OVERHEAD + VSL_WORDS(len));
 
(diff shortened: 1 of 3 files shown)


Dridi:
Speaking of misnomers, all `shm_*` counters could be renamed to `vsl_*` considering that they only count things in the VSL segment of our shared memory... The original goal of this change is to add new counters, and I snuck in with the addition of long description of existing counters.

I will update counters descriptions, but renaming other counters or changing their meaning is out of scope for this pull request.

nigoroll (approved):
👍🏽 on doc polish and adding `shm_bytes` only

Dridi:
Followup: 759f38227d26e2bd47a6a0cf9879e19b505911d0

nigoroll:
This PR was wrongly referenced in two commits because of a typo, which should have read #3836

PR #3828, opened 2022-07-22
vsl: Apply VSL_OPT_E to the Begin[4] field

Dridi:
As I was relying on the lack of -E option to discard another type of
non-ESI sub-requests I recently misled myself with this assumption. It
took me a while to realize that I was also seeing transactions I was not
interested in, corrupting the statistics I was deriving from a log dump.

Now that vmod_pesi [1] set the precedent, we should expect new types of
sub-requests to materialize and expecting the reason for the transaction
to be "esi" is no longer appropriate.

We could also track this field as vtx::sub_level for example, to make it
available as VSL query LHS operand. The focus here was to adjust the
behavior of an existing feature, not add a new one.

[1] https://code.uplex.de/uplex-varnish/libvdp-pesi


nigoroll (approved):
nice one!

Dridi:
bugwash: ok

Dridi:
I will happily wait for feedback from @slimhazard before merging, just in case we find something wrong with the change.

slimhazard:
@Dridi and @nigoroll I'm not entirely sure if this was the testing that you had in mind, but here's what I've done -- for the pesi VDP I've added a new branch named 'begin_sub':

https://code.uplex.de/uplex-varnish/libvdp-pesi/commits/begin_sub

pesi has its own function vped_include that mostly does what ved_include does, and it is modified in the branch to add the two new fields in VSL output:
```
$ git diff
diff --git a/src/vdp_pesi.c b/src/vdp_pesi.c
index fad5eec..036ff39 100644
--- a/src/vdp_pesi.c
+++ b/src/vdp_pesi.c
@@ -342,11 +342,13 @@ vped_include(struct req *preq, const char *src, const char *host,
        req->vsl->wid = VXID_Get(wrk, VSL_CLIENTMARKER);
        VSLdbgv(preq, "vped_include: new xid=%d", VXID(req->vsl->wid));
 
-       VSLb(req->vsl, SLT_Begin, "req %u esi", VXID(preq->vsl->wid));
-       VSLb(preq->vsl, SLT_Link, "req %u esi", VXID(req->vsl->wid));
-
        req->esi_level = preq->esi_level + 1;
 
+       VSLb(req->vsl, SLT_Begin, "req %u esi %u", VXID(preq->vsl->wid),
+            req->esi_level);
+       VSLb(preq->vsl, SLT_Link, "req %u esi %u", VXID(req->vsl->wid),
+            req->esi_level);
+
        req->top = preq->top;
 
        HTTP_Setup(req->http, req->ws, req->vsl, SLT_ReqMethod);
```
That's enough for all of pesi's vtc tests to pass when run against Dridi's branch. We have log_expect scattered throughout the tests, so I take that as sufficient evidence that VSL parsing and linking in Dridi's branch also works with suitably modified pesi.

A few of the log_expect tests in pesi match against "Begin ... esi" entries, they just don't match up to end-of-string ($), so they didn't fail with the new field. @nigoroll we could go through and add  `\d+` to the ends of those matches, for completeness. A look at git grep makes me think that there are about 10 or so. But I didn't bother doing that now, maybe when the PR is pushed to master.

I added a test to verify the output of `varnishlog -E`:

https://code.uplex.de/uplex-varnish/libvdp-pesi/blob/begin_sub/src/tests/varnishlog_e.vtc

That sets up some pesi-includes up to esi level 2 (adapted from another vtc test), runs one client request, and then verifies that `varnishlog -E` output does indeed show `Begin` and `Link` lines with levels 1 and 2 in the new field:
```
shell -match {req \d+ esi 1} {varnishlog -n ${v1_name} -d -c -E -I Begin:esi}
shell -match {req \d+ esi 2} {varnishlog -n ${v1_name} -d -c -E -I Begin:esi}
shell -match {req \d+ esi 1} {varnishlog -n ${v1_name} -d -c -E -I Link:esi}
shell -match {req \d+ esi 2} {varnishlog -n ${v1_name} -d -c -E -I Link:esi}
```
So I think that when Dridi's PR is pushed to master, we can merge the branch in the pesi repository to master, and all will be well.

That's enough for me to say LGTM. @Dridi and @nigoroll let me know if that was the testing that you wanted to know about. If so I'll click on the green check mark here in github.

nigoroll:
@slimhazard thank you, sgtm

Dridi:
> ```
> +       VSLb(req->vsl, SLT_Begin, "req %u esi %u", VXID(preq->vsl->wid),
> +            req->esi_level);
> +       VSLb(preq->vsl, SLT_Link, "req %u esi %u", VXID(req->vsl->wid),
> +            req->esi_level);
> ```

I think one more interesting thing to test would be to change the reason from `esi` to `pesi` and see that things work as expected too, only for testing purposes.

slimhazard:
> I think one more interesting thing to test would be to change the reason from esi to pesi and see that things work as expected too, only for testing purposes.

I just did that in a working directory, without pushing anything since we use the reason "esi" in pesi. I had to adjust a few tests where we explicitly match against the string "esi". With that, all of the vtc tests are green. So I'm going to call it for LGTM.

slimhazard (approved):

Dridi:
Thanks a lot @slimhazard!

Dridi:
I was about to merge, and decided I would try to add VTC coverage with the help of @nigoroll's synthetic logging feature from varnishtest.

Dridi:
Thanks to @nigoroll's amazing `vtc.vsl_replay()` helper I was able to add proper coverage, and found that I was going to introduce a regression when `-b -c` options are present but not `-E`.

Ready to merge when CI passes, but I can wait for another review too. I added two trivial commits.

nigoroll:
@Dridi phk is +-0, so OKed

PR #3829, opened 2022-07-26
http: Configure the Via header with server identity

### LINKED ISSUE #3794: Configure/disable default Via header
Hi, I'd like the default `Via` header added by Varnish to be either disabled or configurable. (see context section for the long story of why).

However, Varnish's use of a hardcoded value is getting in the way and will require me to iterate over all Via headers to find the right one to replace, which is rather suboptimal.

## Expected Behavior

`varnishd` supports a parameter/feature/? to disable automatic adding of `Via` header by Varnish

## Current Behavior

The following line adds a hardcoded `Via` header which isn't trivial to just unset+set in case the upstream response also has `Via` headers.

https://github.com/varnishcache/varnish-cache/blob/5424da941529503c724911d8141e6824db756f88/bin/varnishd/cache/cache_req_fsm.c#L170

## Possible Solution

I see 2 possible approaches:

1. as listed above: allow disabling the automatic adding of the header

2. Allow parameterizing automagically, something like `Via: 1.1 ${gethostname} (Varnish/7.1)` maybe

imo 2. sounds like a can of worms and an invitation to a pointless debate on whether the host name is a valid choice, and if not how to specify where to get it from instead. The approach in 1. seems way less opinionated and lets the user pick whatever makes them happy.

N/A

## Context

Our setup has Varnish instances in multiple places, notably both at the edge and within our internal network.

In this case we want to cache 2 types of data:
- ESI sub-responses
- Assembled responses

Given our ESI target service is not deployed at the edge, assembling all from the edge involves paying the high latency cost twice (once for the base request, and once for the ESI request).

A simplified view of our approach (and why we use 2 Varnish tiers) is:

```
                                             /-> base req upstream (short TTL + has a Via header of its own)           
Edge Varnish --- high lat ---> LB -> Varnish
                                             \-> ESI req upstream (short TTL)
```

This allows us to pay the high latency cost only once.

However, in the current situation what we'll end up with is duplicated `Via: 1.1 varnish (Varnish/7.1)` which doesn't help a ton for tracking things as opposed to for example:
- `Via: 1.1 varnish edge (Varnish/7.1)`
- `Via: 1.1 varnish $servicename (Varnish/7.1)`

Additionally, the edge's Varnish sees a Via header added by the internal load-balancer (and potentially some other intermediates), so it can get rather tricky to find out which one was just added and should be replaced vs. which ones were preexisting from upstream and should be preserved.

It's possible to do something with an approach around saving them in something like `beresp.http.X-Original-Via` perhaps but it sounds just overall unnecessarily convoluted.

Finally, a nitpicky argument is to be able to remove the Varnish version as per general best practices when it comes to what headers should disclose or not to end-users for security purposes.

## Your Environment

Mostly irrelevant but a mix for direct install (ie APT) and k8s/docker.

---

Also, apologies if this would have belonged more somewhere else (like the mailing list)

Dridi:
Change made to the Via header:

    -Via: 1.1 varnish (Varnish/x.y)
    +Via: 1.1 ${i_opt} (Varnish/x.y)

It should actually better serve its purpose of informing about the hops
a request (or in this case a response) went through. This is how Via is
defined in HTTP:

    Via = #( received-protocol RWS received-by [ RWS comment ] )

Using the server identity (which defaults to the server host name) is
more accurate for the received-by field. We still advertise Varnish in
the optional comment field, and hopefully that's what analytics services
rely on to recognize a server as a Varnish instance.

A bit of trivia, a proxy MAY add a Via entry to the response and MUST
add one to a forwarded request: Varnish only adds the optional one.

Refs rfc9110/7.6.3.
Fixes #3794


Dridi:
bugwash: "go for it" from @bsdphk.

Tristan971:
After playing around with it a little, it looks good to me (ie by default I get the original Via chain back in responses alongside Varnish's portion, and the hostname part resolves just fine).

Since you asked for feedback, here's a few notes. I appreciate that some are only tangentially related to the changes.

---

> We still advertise Varnish in

the optional comment field, and hopefully that's what analytics services
rely on to recognize a server as a Varnish instance.

That's probably a reasonable concern. If I built such a service I would rather look out for `x-varnish` as a marker than deal with the potential messiness of Via headers (too much pain with combined vs multiple, funky optional part content breaking parsing, etc....). But I can totally see someone building their service in that way indeed.

---

As per @nigoroll's concern
> \- 0.5 for the added info leak

Given that it's meant to be the new default, I have to agree to some extent.

In my case, we specifically have a different internal vs public host naming scheme, for a few reasons:
- Internal hostnames usually contain information like the city, while public ones contain at most the region
- It makes us sure that we don't accidentally leak something one day for a random reason because the hostname changed after a reboot and cloud-init got tricked into picking a new hostname with unexpected information (funky metadata servers are unfortunately a thing...)
- We can ensure consistency over time across a large fleet of VMs

---

Now what's the right source of that data for everyone as default...? no idea. That's why in #3794 I wrote:
> imo [hostname] sounds like a can of worms and an invitation to a pointless debate on whether the host name is a valid choice, [...]. The approach [of keeping the default Via header as 'Varnish' but making it entirely disabled on-demand] seems [simpler].

There's also the general OWASP-and-friends' recommendation to hide software revisions in headers, which this default still contains. I have faith in Varnish being secure and that it will almost always be irrelevant but if I want that checkbox ticked for some reason (think bureaucratic audits for example), that could be an unnecessary problem.

Also the 1.1 vs HTTP/1.1 bit. I am honestly not sure if that's defined in the spec, but since in practice it varies, we do override ours so that it's consistent. It's mostly a cosmetic concern but well:

Dridi:
> As per @nigoroll's concern
> 
> > - 0.5 for the added info leak
> 
> Given that it's meant to be the new default, I have to agree to some extent.

Now you made me realize that this pull request lacks a doc update. The received-by field from the spec can either be a host name or a pseudonym for precisely this concern.

You gave this example in #3794:

```http
Via: 1.1 varnish edge (Varnish/7.1)
```

That one would be problematic because it breaks syntax, so let's assume a `varnish-edge` pseudonym instead, no space:

```http
Via: 1.1 varnish-edge (Varnish/7.1)
```

Then all you need is a `-i varnish-edge` option in your `varnishd` command line. It's simply that `-i` defaults to the host name, that's why I picked server identity: it was the best match for a host name with the ability to fall back to an arbitrary pseudonym.

> Also the 1.1 vs HTTP/1.1 bit. I am honestly not sure if that's defined in the spec

It is :)

See https://www.rfc-editor.org/rfc/rfc9110.html#name-via

Tristan971:
> That one would be problematic because it breaks syntax, so let's assume a varnish-edge pseudonym instead, no space

Oh, didn't realize. Well we'd use dns-compliant (forget the exact name of the spec for those) ones either way so your assumption would be correct

> Then all you need is a -i varnish-edge option in your varnishd command line. It's simply that -i defaults to the host name, that's why I picked server identity: it was the best match for a host name with the ability to fall back to an arbitrary pseudonym.

Yup that totally makes sense to me why you went with that, and -i does indeed do the trick for my case, now that the preexisting Vias are kept. I meant it more in a "if it's to suddenly be default, people might get surprised".

Obviously I'm now personally well-aware from commenting here, but it might surprise others. Idk.

> > Also the 1.1 vs HTTP/1.1 bit. I am honestly not sure if that's defined in the spec
> 
> It is :)

I assume you refer to

> For brevity, the protocol-name is omitted when the received protocol is HTTP.

Interesting... I guess the wording doesn't clearly state that it is a MUST rather than a MAY for 'HTTP', which would explain why it's not entirely consistently applied across various webservers...

~~The intricacies of the HTTP spec are a pleasure as usual /s~~

Dridi:
> I meant it more in a "if it's to suddenly be default, people might get surprised".

Oh yes, this one should get a prominent mention in the release notes.

> ~The intricacies of the HTTP spec are a pleasure as usual /s~

~Definitely~

Dridi:
I pushed another change: document that `-i` influences Via headers and enforce token syntax.

nigoroll (approved):

Dridi:
Followup: d0acf119167149e1296d5e297449d010600e51c2

PR #3835, opened 2022-08-12
h2: Make requests walk away from the waiting list

Dridi:
When a session goes to the cleanup step, it will indefinitely try to
kill all the streams until those stuck in a waiting list reembark a
thread. This can lead to clients timing out, closing the h2 session,
and opening a new h2 session for the exact same requests. This can
artifically inflate the number of dead tasks in a waiting list with
generous backend timeouts.

To avoid that, an h2 stream will keep a list of requests currently
in a waiting list and make them walk away quietly during cleanup.
An h2 stream would normally only have one request running at any
time but this may no longer be true with modules like vmod_pesi.

---

See https://varnish-cache.org/lists/pipermail/varnish-misc/2022-July/027160.html

This is equivalent to what we have in Varnish Enterprise that I
mentioned in this thread. It does not solve the problem described
in the email thread, because we have no path to walk away from
an HTTP/1 request (pun intended).

I believe we could potentially solve the HTTP/1 case if we involved
a waiter that would only care about closed connections, ignoring
incoming data that I believe trigger a wake up today.

I am convinced that I managed a better design than the original
attempt, and I'm looking forward to reviews. Unfortunately, I will
miss the next two bugwash meetings.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -485,11 +485,13 @@ struct req {
 	void			*transport_priv;
 
 	VTAILQ_ENTRY(req)	w_list;
+	VTAILQ_ENTRY(req)	t_list;
 
 	struct objcore		*body_oc;
 
-	/* The busy objhead we sleep on */
+	/* The busy objhead we sleep on, referenced up to twice */
 	struct objhead		*hash_objhead;
+	struct objhead		*transport_objhead;
 
 	/* Built Vary string == workspace reservation */
 	uint8_t			*vary_b;
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -464,6 +464,12 @@ HSH_Lookup(struct req *req, struct objcore **ocp, struct objcore **bocp)
 	CHECK_OBJ_NOTNULL(oh, OBJHEAD_MAGIC);
 	Lck_AssertHeld(&oh->mtx);
 
+	if (req->walkaway) {
+		assert(oh->refcnt > 0);
+		(void)hsh_deref_objhead_unlock(wrk, &oh, 0);
+		return (HSH_WALKAWAY);
+	}
+
 	if (req->hash_always_miss) {
 		/* XXX: should we do predictive Vary in this case ? */
 		/* Insert new objcore in objecthead and release mutex */
@@ -696,6 +702,42 @@ hsh_rush2(struct worker *wrk, struct rush *r)
 		CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 		VTAILQ_REMOVE(&r->reqs, req, w_list);
 		hsh_reembark(wrk, req);
+	}
+}
+
+void
+HSH_WalkAway(struct worker *wrk, struct objhead **ohp, struct req *req)
+{
+	struct objhead *oh;
+	struct rush r[1];
+
+	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
+	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	TAKE_OBJ_NOTNULL(oh, ohp, OBJHEAD_MAGIC);
+
+	INIT_OBJ(r, RUSH_MAGIC);
+	VTAILQ_INIT(&r->reqs);
+
+	if (DO_DEBUG(DBG_WAITINGLIST))
+		VSLb(req->vsl, SLT_Debug, "walking away <%p>", oh);
+
+	Lck_Lock(&oh->mtx);
+	if (req->waitinglist) {
+		assert(oh == req->hash_objhead);
+		assert(oh->refcnt > 1);
+		oh->refcnt--;
+		VTAILQ_REMOVE(&oh->waitinglist, req, w_list);
+		VTAILQ_INSERT_TAIL(&r->reqs, req, w_list);
+		Lck_Unlock(&oh->mtx);
+
+		wrk->stats->busy_killed++;
+		AZ(req->walkaway);
+		req->walkaway = 1;
+		hsh_rush2(wrk, r);
+	} else {
+		AZ(req->hash_objhead);
+		assert(oh->refcnt > 0);
+		(void)hsh_deref_objhead_unlock(wrk, &oh, 0);
 	}
 }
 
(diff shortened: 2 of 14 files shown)


nigoroll:
generally in favor, have not looked closed yet, still :(

bin/varnishd/cache/cache.h
```diff
@@ -466,11 +466,13 @@ struct req {
 	void			*transport_priv;
 
 	VTAILQ_ENTRY(req)	w_list;
+	VTAILQ_ENTRY(req)	t_list;
```
mbgrydeland, line 469: This feels like quite the waste of memory adding 24 bytes for the list entry and the objhead reference flag on struct req. I think I prefer the original way of doing the workspace allocation, perhaps replaced/expanded with a fallback malloc.

bin/varnishd/http2/cache_http2_session.c
```diff
@@ -444,17 +445,59 @@ h2_new_session(struct worker *wrk, void *arg)
...
+	Lck_AssertHeld(&oh->mtx);
+	AN(req->waitinglist);
+	AZ(req->wrk);
+
+	assert(oh->refcnt > 0);
+	oh->refcnt++;
+
+	Lck_Lock(&r2->h2sess->sess->mtx);
+	AZ(req->transport_objhead);
+	req->transport_objhead = oh;
```
mbgrydeland, line 469: The variable `req->transport_objhead` should be renamed `req->toptransport_objhead`, to reflect that it is "owned" by the top transport. As it is it would be trouble if e.g. esi_deliver wanted to use it. Same argument goes for `t_list`.

bin/varnishd/cache/cache_hash.c
```diff
@@ -355,6 +355,63 @@ hsh_insert_busyobj(const struct worker *wrk, struct objhead *oh)
...
+	 * The objhead reference transfers to the sess, we get it
+	 * back when the sess comes off the waiting list and
+	 * calls us again
+	 */
+	req->hash_objhead = oh;
+	req->wrk = NULL;
+	req->waitinglist = 1;
+
+	if (req->top->topreq->transport->waitlist != NULL)
+		req->top->topreq->transport->waitlist(wrk, req);
```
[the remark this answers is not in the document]
Dridi: In the commit where it is introduced, there is a comment stating that it is only for `req_top`'s transport. The reason for the lack of symmetry is that the "current" transport action is the same for all requests: go into the waiting list.

Without this patch series there is already a reembark asymmetry.

What we have done downstream is less idea in my opinion, because we made each transport responsible for checking whether there is a top `waitlist` callback to signal instead of centralizing this check like in the current iteration.

We could have a distinction between the sole `reembark` callback and two separate `top_waitlist` and `top_reembark`. I didn't because I felt it was too heavy-handed.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -561,6 +561,21 @@ cnt_lookup(struct worker *wrk, struct req *req)
 		had_objhead = 1;
 	wrk->strangelove = 0;
 	lr = HSH_Lookup(req, &oc, &busy);
+	if (lr == HSH_WALKAWAY) {
+		VRY_Finish(req, DISCARD);
+		AZ(req->objcore);
+		AZ(req->stale_oc);
+		VSLb_ts_req(req, "Waitinglist", W_TIM_real(wrk));
+		VSLb(req->vsl, SLT_Error, "The client is going away");
+		if (req->esi_level > 0) {
```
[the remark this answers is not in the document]
Dridi: First, we can add a dedicated `HSH_TIMEOUT`. Otherwise I can use the same argument for the error message preventing us to use this branch for future timeouts. I think we will be fine if we dedicated `HSH_WALKAWAY` to the case where the client went away while a request was in a waiting list.

For the timeout case, we  probably want to have different semantics than plainly ending the transaction. Maybe a waiting list timeout would be a deferred `hash_ignore_busy`?

I must admit I don't remember precisely why we did that downstream. I think we needed this to interrupt ESI delivery and prevent it from recursing as the client is gone anyway.

bin/varnishd/http2/cache_http2_session.c
Dridi: Or, maybe we could move `req::t_list` and `req::transport_objhead` to `reqtop::w_list` and `reqtop::hash_objhead`?

bin/varnishd/cache/cache_hash.c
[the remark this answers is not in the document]
mbgrydeland: > We could have a distinction between the sole `reembark` callback and two separate `top_waitlist` and `top_reembark`. I didn't because I felt it was too heavy-handed.

I think I prefer that option.

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
Dridi: There's actually a comment a few lines below explaining why we fail sub-requests:

```c
/* NB: Interrupt delivery upwards to avoid engaging
 * new sub-request tasks.
 */
```

Granted, it doesn't give many details.

bin/varnishd/cache/cache_hash.c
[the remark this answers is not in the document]
Dridi: Done.

Dridi:
Rebased against current master to solve the trivial merge conflict. I switched to more explicit `tansport::top_*()` methods as suggested. Not addressing other review items, yet.

bin/varnishd/cache/cache.h
Dridi: I just remembered why I didn't use the workspace: it is reserved for vary processing.

Dridi: I got some results on the generalization side, but didn't manage to without allocation space somewhere. Since workspace is reserved during waiting list operations, it is either a heap allocation or preallocated storage in struct req (current solution). Another possibility could be to opportunistically allocate an "entry" for transport tracking on the workspace before the vary reservation, if the top transport has waiting list methods.

To be continued.

Dridi: Considering that eventually both h2 and http/1 (100% of client transports) will track requests and sub-requests in waiting lists, it would be much simpler to go with the current design and have the tracking state directly in struct req.

Dridi:
Closing for now because #3992 will pull the objhead rug pretty hard.

PR #3836, opened 2022-08-15
Add vcc_feature trace to control generation of VCL_trace logs

nigoroll:
To potentially emit `VCL_trace` VSL records, we called `VPI_count()` from VGC for, in the extreme, every line of VCL.

For a normal setup, `VPI_count()` would call into VSL only to find out that `VCL_trace` is masked and not have any effect.

Issuing two additional function calls for each line of VCL obviously is something we should avoid unless there is a real benefit.

Under the assumption that the `VCL_trace` facility is used rarely, we add the `trace` vcc feature which must explicitly be enabled before loading a VCL to enable tracing. This should have no relevant impact for common installations and, if tracing is required, re-loading the vcl to enable tracing is quick and easy.

The benefit of this change is that we can turn `VPI_count()`, which tracks an approximation of our position in VCL processing, into a macro, reducing the overhead from the above mentioned two function calls to a single address write (one of the two lines of `VPI_count()` is an assertion on constant values which can (and usually will) be eliminated by compilers).

### DIFF
--- a/bin/varnishd/cache/cache_vpi.c
+++ b/bin/varnishd/cache/cache_vpi.c
@@ -62,14 +62,13 @@ VPI_wrk_init(struct worker *wrk, void *p, size_t spc)
 }
 
 void
-VPI_count(VRT_CTX, unsigned u)
+VPI_trace(VRT_CTX, unsigned u)
 {
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	CHECK_OBJ_NOTNULL(ctx->vcl, VCL_MAGIC);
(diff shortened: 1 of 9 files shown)


Dridi (approved):
Neat!

nigoroll:
bugwash:
* keep trace check at runtime, but use [branch prediction hints](https://github.com/varnishcache/varnish-cache/pull/3020)
* introduce `req.trace` and `bereq.trace`:
  * `req.trace` gets initialized from `vcc_feature trace`
  * `bereq.trace` gets initialized from `req.trace`

nigoroll:
Bugwash: We agree that we prefer #3837 and will not pursue the compile time switch further.

PR #3839, opened 2022-09-01
Polish vary code a bit

nigoroll:
Just polish some bits which I noticed while working in something else.
On hold until after release.
flexelint has been run locally with no new reports.


Dridi (approved):
I'd like @mbgrydeland to have a look too when he returns,

nigoroll:
I am confident that this does not need more reviews, I only stashed it during the freeze

PR #3841, opened 2022-09-06
fix client_req double count with waitlist

LaughingQuoll:
Fixes an issue where client_req counter does not account for waiting list re-embarks.

This issue was discovered by https://github.com/sirn

### DIFF
--- a/bin/varnishd/http1/cache_http1_fsm.c
+++ b/bin/varnishd/http1/cache_http1_fsm.c
@@ -386,10 +386,10 @@ HTTP1_Session(struct worker *wrk, struct req *req)
 		} else if (st == H1PROC) {
 			req->task->func = http1_req;
 			req->task->priv = req;
-			wrk->stats->client_req++;
 			CNT_Embark(wrk, req);
 			if (CNT_Request(req) == REQ_FSM_DISEMBARK)
 				return;
+			wrk->stats->client_req++;
 			AZ(req->top->vcl0);
 			req->task->func = NULL;
 			req->task->priv = NULL;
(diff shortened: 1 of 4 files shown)


Dridi:
I would prefer we revert 2e66e0b93ba6e50fa5ef499cd6a4860d0fa56134 that fails to convey the intent behind the change and then we add `client_req` coverage to existing waiting list test cases.

Also I don't consider the waiting list to be "basic" functionality, so if we really add new test cases I would prefer "^c" complex tests (b64 seems to be an exception, and doesn't look remotely basic).

Dridi:
After some research we reached the conclusion that the original patch was correct (modulus VTC nitpicks).

Here is the sequence of relevant events:

- 2e66e0b93ba6e50fa5ef499cd6a4860d0fa56134 moves `client_req` increment to transport code
  - ESI transport doesn't go through the `R_STP_TRANSPORT` step and wasn't included anyway
- 2c59afe22ba108c7f2ab5ff75b2e30baf4ef9ec7 makes ESI go through `R_STP_TRANSPORT`
  - reverting 2e66e0b93ba6e50fa5ef499cd6a4860d0fa56134 we see ESI requests counted in `client_req` (regression)
- e9c2a488d291b3129838c411cac7312f8801ff1e adds `esi_req` for ESI sub-requests (tangentially related)

So we'll go with @LaughingQuoll's original patch except that coverage will happen in existing waiting list tests.

nigoroll (approved):

hermunn:
As we are very close to a release, @Dridi will merge this manually.

Dridi:
20d356b3d08a458c6f56117a65f9041b5f6fd405

PR #3845, opened 2022-09-15
Filter error and vcl load failure coverage

nigoroll:
Minor improvements:
* Improve reporting and test of duplicate filter errors
* Improve coverage for vcl init errors


nigoroll:
I did things in the wrong order and this this PR appears as closed, when it should appear as merged.
It is merged as: e51197f1df1b0c6bf4c4ab01bf647c75d94b08ab to 
7791c04df7ae422272d92111b7e35da751ee77c5 but the last one was for #3848

PR #3855, opened 2022-09-23
Cleanup and sum up after bgthread termination

nigoroll:
Strictly speaking, these calls were already missing before 3a22f5f0ded31f8c7340ac00a382ae9106c20a8c, but now that bgthreads are officially allowed to terminate, we should really finalize & cleanup the wpriv and stats structs to provide a consistent interface.


nigoroll:
bugwash:

> (15:09:03) phk: yeah, looks OK

mbgrydeland (approved):
LGTM

Issue #3863, opened 2022-11-10
Support for abstract sockets on Linux?


jlecour:
Hi, I've just heard about abstract UNIX sockets on Linux.

I've found a few pieces of software that support them (HAProxy, curl…) so I wondered if I could use one of them for communication between HAProxy and Varnish instead of the typical UNIX sockets that I've been using.

I've tried to simply add an `@` before the `/` at the beginning of tha socket path, but Varnish refuses to start.

Thanks.

bsdphk:
I'm not sure I see much point, but if somebody writes a patch...

nigoroll:
> I've tried to simply add an @ before the / at the beginning of (the) socket path, but Varnish refuses to start.

With the patch linked above you can now do that

jlecour:
Hi,

> I'm not sure I see much point

@bsdphk If I understand correctly, abstract sockets do not involve any filesystem interaction. That's why it can be a good thing to support. I've also heard about some performance gains (probably related to the previous point).

> With the patch linked above you can now do that

I wasn't asking for it, just wondering if this was possible or not.
But thanks a lot @nigoroll for the work !

I am not able to try this patch until it is releases and packaged, so I'll trust you.

If you think it's done, you can close the issue.

PR #3864, opened 2022-11-14
Support abstract sockets for accept and .path to backends

nigoroll:
We use the commonplace @<name> syntax to specify abstract socket names.

Implements #3863


bsdphk:
Bugwash: Needs polish.

nigoroll:
bugwash: polish, then merge:
- isuds() macro
- vtc feature abstract_uds

PR #3866, opened 2022-11-22
http1: Emit ReqStart on bad requests

Dridi:
It makes no sense to emit this just before we end the transaction, except to help varnishncsa find the client IP address.


Dridi:
I wanted to avoid code duplication before merging, and now it's done. I will merge after CI completes.

PR #3869, opened 2022-11-28
shmlog: Avoid allocating vsl_reclen bytes in VSLv()

Dridi:
When vsl_reclen increases to a large value, like 8kB to hold long ReqURL
or *Header:Cookie records, making such an allocation on the stack could
jump over a 4kB guard page. The underlying printf call could also result
in significant stack usage. Both factors could result in hard to debug
segmentation faults.

Instead, we can opportunistically print to a reasonably-sized buffer,
and fall back to printing directly inside the VSL segment when a record
is too large for the buffer. At this point we know the full record size
and we can adjust it to vsl_reclen.

This is inspired by a different patch from @mbgrydeland targeting the
6.0 branch.

### DIFF
--- a/bin/varnishd/cache/cache_shmlog.c
+++ b/bin/varnishd/cache/cache_shmlog.c
@@ -294,13 +294,13 @@ VSLv(enum VSL_tag_e tag, vxid_t vxid, const char *fmt, va_list ap)
 
 	if (strchr(fmt, '%') == NULL) {
 		vslr(tag, vxid, fmt, strlen(fmt) + 1);
-	} else {
-		n = vsnprintf(buf, mlen, fmt, ap);
-		n = vmin_t(unsigned, n, mlen - 1);
-		buf[n++] = '\0'; /* NUL-terminated */
-		vslr(tag, vxid, buf, n);
+		return;
 	}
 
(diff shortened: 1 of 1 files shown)


bsdphk:
This needs a VTC to provoke test-coverage.

nigoroll (commented):
I agree that the patch helps avoid an SO for a deep call tree, but it really looks to me as if it was addressing a secondary problem, and not a primary one:

* For all practical purposes, within varnish-cache, anywhere we could possibly run a regex, we need to have 32k of stack available for JIT anyway. So this should really not be an issue for common varnish-cache code and reasonable `vsl_reclen` sizes.
* If someone runs into this for real, I suspect they are deep into a custom vmod's call stack or using an oversized `vsl_reclen`.
* This does not exclude an SO for a deep call path (we do not try to avoid SOs in general, we rely on the stack to be "just about" large enough). My argument would be that someone who is 92k up the stack, might just as easily end up 96k up the stack.
* Why is anyone using unbuffered VSL at that hypothetical place?

Current impression: We are missing a reasonable optimization for "normal users" in order to _maaaaybe_ reduce stack usage for an exotic case

So no real opinion here.

PR #3870, opened 2022-11-28
vgz: Catch up with trunk (6.0)

Dridi:
Skipping irrelevant commits made to libvgz since the last common patch.


Dridi:
While this catch up is now outdated, I will still proceed to merge it once CI completes with hopefully less test failures. It is technically not needed for security releases, but it was the occasion to reach the bug fix from 03abd3d45eff519b784ae3a9588c92c70860da62.

Going to 1.2.13 would actually solve the C compilation warnings problem for good, but ironically this would also result in merge conflicts in the zlib code I do not want to engage on short notice.

Dridi:
I will merge this now that #4002 is registered.

Issue #1777, opened 2016-03-04
Range requests problem


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/1777

lkarsten:
Backport review: Fixed in 4.1.0-beta1.

Needs 4.0 review.

kanongil:
I'm investigating an issue regarding the fix for this in 121593312daabbef4e290a590b364a099980a8b7. The trac link is however broken, so I'm missing context. Is it possible to access somehow?

kanongil:
Nvm, found it here: https://web.archive.org/web/20160809133121/https://varnish-cache.org/trac/ticket/1777

PR #3859, opened 2022-10-26
range: Move more parsing to http_GetRange()

Dridi:
This came up on the enterprise side, the need to move more parsing free of side effects to `http_GetRange()`. It changes the function signature but I'm not sure whether it qualifies as VRT material.

PR #3872, opened 2022-11-30
Improve streaming range responses

kanongil:
This is a proper fix for #1777 so that it doesn't sometimes deliver `200` responses to range requests. This is achieved by simply not adding a `content-length` header to streaming responses. With this fix the response just contains less bytes than the `content-range` range would indicate, which is a valid response when the [complete-length](https://httpwg.org/specs/rfc9110.html#field.content-range) is unknown as indicated by `/*`.

FYI, returning less bytes than the `content-range` indicates is exploited in [RFC 8673](https://www.rfc-editor.org/rfc/rfc8673.html). Incidentally, this patch makes Varnish partially support such streaming range responses.

I have re-enabled the disabled tests from #1777, and revised all streaming range tests to expect `206` responses.

**Background**

I am using varnish to cache requests to a backend that delivers [Low-Latency HLS](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#appendix-B.1). This includes a mode, where parts are delivered using byte range requests from the containing segments. Besides requiring byte-range support, this mode also requires the server to deliver data using streaming range responses. When I tried to deliver individual parts using byte range requests, Varnish failed to work and delivered a full segment in response to `range` requests (response code `200`).


Dridi (approved):
Very nice and tidy!

kanongil:
FYI this new logic conflicts with the refactor in the pending #3859.

Dridi:
At quick glance it shouldn't be a conflict too hard to solve, thank you for pointing it out.

Dridi (commented):
bugwash: we need to study the effect on HTTP/1 framing closely.

kanongil:
Without a known `resp_len`, the byte stream will be transmitted using `Transfer-Encoding: chunked`.

The potentially new case is that the response processes less bytes than indicated by `range_high`. This case is handled here:

https://github.com/varnishcache/varnish-cache/blob/eec2a6bd8c5a4f1aad3ed75131fe450e35747614/bin/varnishd/cache/cache_range.c#L51-L64

It probably needs to be patched so that the request doesn't fail when `resp_len < 0`, which I guess will cause the HTTP/1.1 connection to be closed.

Dridi:
After a brief look at 535a5f0fe6d40d186cb286786a810531478a42e2, I come to the same nice and tidy conclusion. You even went the extra mile and found a disabled test case that you rehabilitated. I missed this when I first reviewed this directly on Github and I just noticed after a first pass in my proper work environment.

FYI we have lined up 3 reviewers to comb this pull request.

Dridi (approved):
It has taken me several partial attempts but I finally managed an almost complete review. I couldn't find anything wrong and I'm confident this is a solid change. We often get very good contributions but this one checks all my marks and then more.

Thank you again!

The last thing I need to do what I mentioned above in https://github.com/varnishcache/varnish-cache/pull/3872#issuecomment-1332162097 but I would rather adapt my work after merging this one.

bsdphk:
Bugwash: OK.

PR #3873, opened 2022-11-30
gzip: We can't work with a standalone zlib

dridi:
Below is the commit message of the last patch in this series. Reviewing the previous patches may or may not be needed to build enough context to digest this one. Despite this definitive failure at working with a standalone zlib, I think there are a couple patches we can cherry-pick in trunk.

More comments after the commit message.

> Unlike the inflate() API that lands itself nicely to figuring block
> boundaries, I have come to the conclusion that this is impossible for
> its deflate() counterpart.
> 
> I proceeded similarly to inflate():
> 
> - find how and when boundaries are registered
> - port the formula to varnishd's code
> - see what boundaries we operate on outside zlib
> 
> What I found is that we keep track of the start, stop and last bit
> positions deep in the inflate code. The start bit is not a problem for
> us, it is guaranteed to always be 80 bits, the initial 10B gzip header.
> We know for a fact that gzip streams we generate never have anything
> extra. This is verified by an assertion.
> 
> Porting the formula was trivial using the deflatePending() function
> and it turns out to be essential to figure boundaries out.
> 
> The problem I initially suspected was that one call to deflate() could
> result into multiple blocks being output, as the structure of the code
> suggests in the trees.c file. This poor man's tracing confirms that we
> can't achieve from the outside what our patched zlib does.
> 
> Output example:
> 
>     $ ./varnishtest -i tests/c00034.vtc | grep z= | sed 's/.*said//'
>     z_deflate z=0x7fca7d016040 start=80
>     VGZ_UpdateObj z=0x7fca7d016040 start=80 stop=0 last=0 bits=80
>     z__tr_flush_block z=0x7fca7d016040 last=80
>     z__tr_flush_block z=0x7fca7d016040 stop=196
>     VGZ_UpdateObj z=0x7fca7d016040 start=80 stop=196 last=80 bits=264
>     VGZ_UpdateObj z=0x7fca7d016040 start=80 stop=196 last=80 bits=264
> 
> We know that we can hard-code the start bit to 80, and we could maybe
> capture the last bit (80) but unlike the inflate() code there doesn't
> seem to be state we could inspect to inform this need. On the inflate()
> side we can rely on the z_stream::data_type field, but here it seems to
> solely contain Z_BINARY or Z_TEXT (if not Z_UNKNOWN) and no special bits
> like in inflate mode.
> 
> After adding the Z_BLOCK flag in this patch, the situation does not
> improve:
> 
>     $ ./varnishtest -i tests/c00034.vtc | grep z= | sed 's/.*said//'
>     z_deflate z=0x7ff0e6416040 start=80
>     VGZ_UpdateObj z=0x7ff0e6416040 start=80 stop=0 last=0 bits=196
>     z__tr_flush_block z=0x7ff0e6416040 last=196
>     z__tr_flush_block z=0x7ff0e6416040 stop=220
>     VGZ_UpdateObj z=0x7ff0e6416040 start=80 stop=220 last=196 bits=288
>     VGZ_UpdateObj z=0x7ff0e6416040 start=80 stop=220 last=196 bits=288
> 
> First, we get a chance to run VGZ_UpdateObj() after the initial header
> and before the first block is written. So we can first record the start
> at 80, so we gain nothing. Once again, we get a chance to capture the
> last bit at 196 (instead of 80) but have no state to inspect to confirm
> whether we should do that. By the time we come back the stop bit has
> already been recorded and we already went past it. We also can't infer
> the stop bit of the stream as `total_out * 8 - 64` (where 64 is the
> number of bits needed for the CRC32 and ISIZE fields at the end of a
> block) because the compressed payload of a block may not have a bit size
> that is a multiple of 8.
> 
> Even worse, we artifically force the block boundary and end up with a
> larger gzip stream altogether.
> 
> And that is just a simplistic test case when it comes to gzip. Another
> one like v00004.vtc shows an even bleaker picture. And if we aren't even
> close to making a gzip test case pass, there's no hope for ESI assembly
> of multiple fragments compressed (or not) independently.

After it came up one time too many I decided to try @madler's suggestion on how to find block boundaries without patching zlib.

See https://github.com/madler/zlib/issues/189 for some context.

The way it works is that varnishd tries to find the gzip block boundaries we care about, and then compares notes with libvgz (read: asserts). We could get away with a standalone zlib for gunzip operations, but not for compression.

Now that I have a good understanding of what our patched zlib (libvgz) does, and now that I am convinced that doing so without our downstream patch is impossible, I plan to reopen the aforementioned issue to explain precisely what it is that we do. One problem I see with our patch is that the ESI assembly of maybe-compressed fragments is done completely outside of zlib, so adding this block boundaries tracking feature in zlib could be a hard sell without APIs to make use of this tracking.

Regardless of the outcome, we should probably cherry-pick the relevant commits and clearly document why we have libvgz today.

### DIFF
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -645,7 +645,7 @@ ved_gzgz_init(VRT_CTX, struct vdp_ctx *vdc, void **priv, struct objcore *oc)
  * in VDP_bytes(), but ved_bytes() covers it.
  *
  * To avoid unnecessary chunks downstream, it would be nice to re-structure the
- * code to intendify the last block, send VDP_END/VDP_FLUSH for that one and
+ * code to identify the last block, send VDP_END/VDP_FLUSH for that one and
  * VDP_NULL for anything before it.
  */
 
--- a/bin/varnishd/cache/cache_esi_fetch.c
+++ b/bin/varnishd/cache/cache_esi_fetch.c
@@ -97,7 +97,7 @@ vfp_vep_callback(struct vfp_ctx *vc, void *priv, ssize_t l, enum vgz_flag flg)
 		}
 		VGZ_Obuf(vef->vgz, ptr, dl);
 		i = VGZ_Gzip(vef->vgz, &dp, &dl, flg);
-		VGZ_UpdateObj(vc, vef->vgz, VUA_UPDATE);
+		VGZ_UpdateObj(vc, vef->vgz, i);
 		if (dl > 0) {
 			vef->tot += dl;
 			VFP_Extend(vc, dl, VFP_OK);
@@ -145,7 +145,7 @@ vfp_esi_end(struct vfp_ctx *vc, struct vef_priv *vef,
 
 	if (vef->vgz != NULL) {
 		if (retval == VFP_END)
-			VGZ_UpdateObj(vc, vef->vgz, VUA_END_GZIP);
+			VGZ_UpdateObj(vc, vef->vgz, VGZ_END);
 		if (VGZ_Destroy(&vef->vgz) != VGZ_END)
 			retval = VFP_Error(vc,
 			    "ESI+Gzip Failed at the very end");
--- a/bin/varnishd/cache/cache_gzip.c
+++ b/bin/varnishd/cache/cache_gzip.c
@@ -264,7 +264,7 @@ VGZ_Gzip(struct vgz *vg, const void **pptr, ssize_t *plen, enum vgz_flag flags)
 	case VGZ_ALIGN:		zflg = Z_SYNC_FLUSH; break;
 	case VGZ_RESET:		zflg = Z_FULL_FLUSH; break;
 	case VGZ_FINISH:	zflg = Z_FINISH; break;
-	default:		INCOMPL();
+	default:		WRONG("Invalid VGZ flag");
 	}
 	i = deflate(&vg->vz, zflg);
 	if (i == Z_OK || i == Z_STREAM_END) {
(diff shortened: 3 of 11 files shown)


dridi:
After revisiting this topic, I was able to go a little further on the compression side, but it's still not possible to reliably collect the stop bit offset.

Closing in favor of VIP35: https://github.com/varnishcache/varnish-cache/wiki/VIP35:-Retire-libvgz

madler:
See https://github.com/madler/zlib/issues/189#issuecomment-2201755658 .

dridi:
Thank you for making it (almost) possible to work is a regular zlib.

I replied in https://github.com/madler/zlib/issues/189#issuecomment-2205172650 to discuss how `deflateUsed()` is not sufficient in its current state.

dridi:
Thanks to @madler's recent change in zlib, we can almost use a standalone zlib, once released and readily available on platforms we care about, and modulus one zlib change present in this patch series and to be discussed in https://github.com/madler/zlib/issues/189.

We will keep the libvgz fork around for a while, but we should be able to eventually omit it at configure time once `deflateUsed()` becomes widely available.

dridi:
I stand corrected, I can actually get everything working with `deflateUsed()` as it currently works. I still think I shouldn't have to know where those 5 extra bits come from, but at least we know that we can work with the upcoming zlib release, whether or not it includes https://github.com/madler/zlib/pull/989.

dridi:
This was fixed in https://github.com/madler/zlib/commit/884e0c0809e4cc5e88d6545d1c49de7946c1b1ca and https://github.com/madler/zlib/commit/3adaa095a7a6c3f2cc00f73311f9bc8688c21aa7.

I think this is ready to merge and be part of the next release.

Neustradamus:
Good job to all!

PR #3878, opened 2022-12-07
vtc_http2: add stream peer_window tracking 

### LINKED ISSUE #3681: varnishtest: Update h2 streams windows
Looking into a t02020.vtc failure I wanted to find out whether the logic from #3661 could be racy, or whether it was the test case. Trying to add expectations to the test case I was surprised not to find any in the manual, which I dealt with in d2e526ce95d6d8426ec170b4cfc78d41abc07179 but even then, `stream.window` would always be equal to the initial window.

We need to keep track of h2 streams windows in varnishtest, but not enforce them.

walid-git:
Add h2 peer window size tracking in varnishtest
Fixes #3681

### DIFF
--- a/bin/varnishtest/vtc_http2.c
+++ b/bin/varnishtest/vtc_http2.c
@@ -420,7 +420,7 @@ parse_data(struct stream *s, struct frame *f)
 	if (s->id)
 		s->ws -= size;
 
-	s->hp->ws -= size;
+	s->hp->h2_win_self->size -= size;
 
 	if (!size) {
 		AZ(data);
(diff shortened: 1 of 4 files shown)


bin/varnishtest/vtc_http2.c
```diff
@@ -2397,6 +2412,48 @@ cmd_rxpush(CMD_ARGS)
...
+}
+
+/* SECTION: stream.spec.settings_rxsettings rxsettings
+ *
+ * Receive a SETTINGS frame.
+ */
+static void
+cmd_rxsettings(CMD_ARGS)
+{
+	struct stream *s, *_s;
```
Dridi, line 2441: I understand you are just following the code style in this area, but we should avoid naming variables `_s`.

Dridi (approved):
Good job!

bin/varnishtest/vtc_http2.c
Dridi: Let's not do it in this pull request.

Dridi:
Pushed rearranged as 9b42ca376...0bbadc56b.

Issue #3879, opened 2022-12-07
Assert in HSH_Lookup with Varnish 7.2.1


jocel1:
Hi,

I'm using varnish with a malloc allocator and a file allocator for static files.
Settings are the following:

ExecStart=/usr/sbin/varnishd -a :80 -p workspace_session=1024 -p workspace_thread=4096 -p workspace_client=131072 -p workspace_backend=131072 -p listen_depth=16383 -p http_resp_hdr_len=65536 -p http_resp_size=98304 -p workspace_backend=131072 -p thread_pool_min=200 -p thread_pool_max=5000 -p feature=+http2 -P %t/%N/varnishd.pid -T localhost:6082 -f /etc/varnish/default.vcl -s malloc,12288m -s static=file,/tmp/varnish_storage.bin,100G

## Expected Behavior
No crash

## Current Behavior
I figured out varnish was complaining in the systemctl log.
   Active: active (running) since lun. 2022-12-05 20:39:57 CET; 1 day 18h ago

déc. 06 21:25:33 varnishd[2813]: Child (8130) Started
déc. 06 21:25:33 varnishd[2813]: Child (8130) said Child starts
déc. 06 21:25:33 varnishd[2813]: Child (8130) said <jemalloc>: Error in munmap(): Invalid argument
déc. 06 21:25:33 varnishd[2813]: Child (8130) said SMF.static mmap'ed 107374182400 bytes of 107374182400
déc. 07 07:59:26 varnishd[2813]: Child (8130) died signal=6
déc. 07 07:59:26 varnishd[2813]: Child (8130) Panic at: Wed, 07 Dec 2022 06:59:26 GMT
                                                                               Assert error in HSH_Lookup(), cache/cache_hash.c line 426:
                                                                                 Condition((oc)->magic == 0x4d301302) not true....

So the first warning happened a day after starting the server. Then a few hours later, we can see the assert.

This is happening on a ARM architecture (t4g AWS servers) on Amazon Linux 2 + kernel 5.15.75-48.135
I'm using 4 vmod: directors, header, std & xkey
I'm fetching the binary from https://packagecloud.io/install/repositories/varnishcache/varnish72/script.rpm.sh & https://github.com/varnish/varnish-modules/releases/download/0.21.0/varnish-modules-0.21.0.tar.gz

My vcl_hash is not that complicated:

```
sub vcl_hash {
  # Called after vcl_recv to create a hash value for the request. This is used as a key
[...]

Issue #3867, opened 2022-11-23
vgz: Segmentation fault in deflateReset()


Dridi:
https://varnish-cache.org/lists/pipermail/varnish-misc/2022-November/027172.html

```
varnishd[15482]: Child (15492) Panic at: Tue, 22 Nov 2022 16:10:26 GMT

Wrong turn at cache/cache_main.c:284:
Signal 11 (Segmentation fault) received at 0xffff887fe000 si_code 2
version = varnish-6.0.11 revision a3bc025c2df28e4a76e10c2c41217c9864e9963b, vrt api = 7.1
ident = Linux,4.14.290-217.505.amzn2.aarch64,aarch64,-junix,-smalloc,-sfile,-sdefault,-hcritbit,epoll
now = 3120063.050288 (mono), 1669133426.954528 (real)
Backtrace:
  0x43f5b4: /usr/sbin/varnishd() [0x43f5b4]
  0x4a968c: /usr/sbin/varnishd(VAS_Fail+0x54) [0x4a968c]
  0x43a350: /usr/sbin/varnishd() [0x43a350]
  0xffff90fd7668: linux-vdso.so.1(__kernel_rt_sigreturn+0) [0xffff90fd7668]
  0xffff90c8f540: /lib64/libc.so.6(memset+0x100) [0xffff90c8f540]
  0x4b88f4: /usr/sbin/varnishd(deflateReset+0x48) [0x4b88f4]
  0x42e924: /usr/sbin/varnishd(VGZ_NewGzip+0x88) [0x42e924]
  0x42ebc0: /usr/sbin/varnishd() [0x42ebc0]
  0x42df28: /usr/sbin/varnishd(VFP_Open+0x98) [0x42df28]
  0x42b950: /usr/sbin/varnishd() [0x42b950]
thread = (cache-worker)
thr.req = (nil) {
},
thr.busyobj = 0xffff3d040020 {
  end = 0xffff3d050000,
  retries = 0,
  sp = 0xffff3c241a20 {
    fd = 28, vxid = 32945,
    t_open = 1669133426.953970,
    t_idle = 1669133426.953970,
    ws = 0xffff3c241a60 {
      id = \"ses\",
      {s, f, r, e} = {0xffff3c241aa0, +96, (nil), +344},
    },
    transport = HTTP/1 {
      state = HTTP1::Proc
    }
    client = 172.31.47.149 50812 :80,
  },
  worker = 0xffff832326c8 {
[...]

PR #3881, opened 2022-12-08
build: Require an explicit --with-jemalloc to use it

Dridi:
This removes the automatic check made only on Linux systems for jemalloc now that operating systems we care about ship a much more recent version than the 3.6 that is known to work really well with Varnish on x86_64 hardware.

Of the platforms we provide packages for, only Ubuntu 18.04 (bionic) and RHEL 7 (via EPEL) ship jemalloc 3.6.0 and the rest already moved on to a 5.x series.

It appears that jemalloc 3.6 on aarch64 results in unstable Varnish workloads while jemalloc 5 is known to generate a lot of waste with its default configuration with a highly threaded workload like Varnish would operate, based on feedback we have seen over years.

From now on, jemalloc is found via pkg-config, only if it is explicitly requested at configure time. With explicit opt-in, the Linux-only check is gone and we solely rely on pkg-config to find the library. This can of course be overriden as usual at configure time:

    ./configure --with-jemalloc JEMALLOC_CFLAGS=... JEMALLOC_LIBS=...

Or, if you don't like long command lines, via environment variables.

Refs #3867
Refs #3879

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -387,6 +387,7 @@ workflows:
           name: build_centos_7
           dist: centos
           release: "7"
+          extra_conf: --with-jemalloc
       - build:
           name: build_centos_stream
           prefix: quay.io/centos/
--- a/bin/varnishd/Makefile.am
+++ b/bin/varnishd/Makefile.am
@@ -180,8 +180,12 @@ varnishd_LDADD = \
 	$(top_builddir)/lib/libvarnish/libvarnish.la \
 	$(top_builddir)/lib/libvsc/libvsc.la \
 	$(top_builddir)/lib/libvgz/libvgz.la \
-	@JEMALLOC_LDADD@ \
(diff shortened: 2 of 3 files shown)


bsdphk:
We had a bit of a pow-wow about this one, and the result was:

Make it easier to configure with/without jemalloc (and/or other malloc implementations ?)

Disable jemalloc on ARM64 for now, as there are credible indications that it causes trouble.

But we do not want to disable jemalloc across the board.

... In particular because we have no contemporary benchmark results on glibc::malloc.

Dridi:
bugwash: ask issue reporters to try some jemalloc flags.

nigoroll:
> bugwash: we want pkg-varnish-cache to explicitly build with jemalloc support if we merge this change, removing automatic detection.

I recall that at least two of us wanted to keep varnish-cache itself (not pkg-*) build with jemalloc by default on amd64, if available.

jocel1:
FWIW, arm build without jemalloc + x64 build with jemalloc are working perfectly for me (see my comments on #3879)

PR #3884, opened 2022-12-25
backends are in charge of printing headers

gquintard:
the current backend implementation reads the headers all at once, as a big buffer, then manually chops them up, and later on, in `vbf_stp_startfetch`, Varnish loops through all the headers and prints them.

This is inconvenient for custom backends that are most likely going to use `http_SetH` (either directly or via `http_SetHeader`, `http_PrinfHeader` or others), which also prints the headers being added. As a result, those implementations end up logging the headers twice.

This commit works around the issue by pushing the burden of logging the `beresp` headers onto the backend implementation. It does change one test, as now the `Beresp` timestamp appears after the headers in the log instead of before.


Dridi (changes requested):
No objection, except for the useless change in `cache_director.c`.

I actually quite like the idea of getting the beresp timestamp after the initial beresp headers.

gquintard:
> No objection, except for the useless change in cache_director.c.

fixed and force-pushed

Dridi (approved):

bsdphk:
OK with me.

Dridi:
bugwash: we will wait for @nigoroll's review.

Meanwhile, can you please document this new responsibility in `doc/sphinx/reference/directors.rst`?

nigoroll (approved):
fine with me.

gquintard:
done, I can rebase/squash/force-push whenever you say so

Dridi:
I spared you a conflict resolution for the changelog and pushed 7063308676415010da5f7b70f571a22a62c66c4b directly.

PR #3886, opened 2023-01-13
Deal with backends that don't close connections

walid-git:
Add "expect_close" parameter to probe to deal with backends that are considered sick because they don't close the TCP connection after sending the response. 
The new parameter accepts the value true or false, and defaults to true for backward compatibility.

### DIFF
--- a/bin/varnishtest/tests/c00113.vtc
+++ b/bin/varnishtest/tests/c00113.vtc
@@ -1,0 +1,31 @@
+varnishtest "probe expect close"
+
+server s0 {
+	rxreq
+	txresp
+	expect_close
+} -dispatch
+
+varnish v1 -vcl+backend {
+	probe default {
+		.window = 1;
+		.threshold = 1;
+		.timeout = 1s;
+		.interval = 0.1s;
+	}
+} -start
+
+varnish v2 -vcl+backend {
+	probe default {
+		.window = 1;
+		.threshold = 1;
+		.timeout = 1s;
+		.interval = 0.1s;
+		.expect_close = false;
+	}
+} -start
+
+delay 2
+
+varnish v1 -cliexpect sick backend.list
+varnish v2 -cliexpect healthy backend.list
--- a/bin/varnishtest/tests/c00113.vtc
+++ b/bin/varnishtest/tests/c00113.vtc
@@ -12,6 +12,7 @@ varnish v1 -vcl+backend {
 		.threshold = 1;
 		.timeout = 1s;
 		.interval = 0.1s;
+		.expect_close = true;
 	}
 } -start
 
--- a/bin/varnishtest/tests/v00005.vtc
+++ b/bin/varnishtest/tests/v00005.vtc
@@ -32,6 +32,17 @@ varnish v1 -vcl {
 	}
 }
 
+# Check expect_close definition
+varnish v1 -vcl {
+	backend b1 {
+		.host = "${localhost}";
+		.probe = {
+			.url = "/";
+			.expect_close = false;
+		}
+	}
+}
+
 # Check redefinition
 varnish v1 -errvcl {Probe request redefinition at:} {
 	backend b1 {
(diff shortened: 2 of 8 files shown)


Dridi:
bugwash: @bsdphk OK, and I would like to see some changes mentioned above. Maybe @nigoroll wants to have a look too.

Dridi:
This is why I requested a warning in the documentation. It's much simpler to implement "expect_close=false" as a last chance after timeout.

lib/libvcc/vcc_utils.c
```diff
@@ -285,6 +285,28 @@ Resolve_Sockaddr(struct vcc *tl,
...
+
+uint8_t
+vcc_BoolVal(struct vcc *tl)
+{
+	struct symbol* sym;
+
+	if (tl->t->tok != ID) {
+		Expect(tl, ID);
+		return (0);
+	}
```
Dridi, line 300: These lines are superfluous, `VCC_SymbolGet()` will fail just fine on its own.

bin/varnishtest/tests/v00005.vtc
```diff
@@ -32,6 +32,17 @@ varnish v1 -vcl {
...
+# Check expect_close definition
+varnish v1 -vcl {
+	backend b1 {
+		.host = "${localhost}";
+		.probe = {
+			.url = "/";
+			.expect_close = false;
+		}
+	}
+}
```
Dridi, line 44: I just realized that you are testing the happy path here, and what I had requested was invalid code:

```diff
--- bin/varnishtest/tests/v00005.vtc
+++ bin/varnishtest/tests/v00005.vtc
@@ -33,12 +33,12 @@ varnish v1 -vcl {
 }
 
 # Check expect_close definition
-varnish v1 -vcl {
+varnish v1 -errvcl {Expected "true" or "false"} {
        backend b1 {
                .host = "${localhost}";
                .probe = {
                        .url = "/";
-                       .expect_close = false;
+                       .expect_close = faux;
                }
        }
 }
```

bin/varnishtest/tests/c00113.vtc
```diff
@@ -0,0 +1,31 @@
...
+	expect_close
+} -dispatch
+
+varnish v1 -vcl+backend {
+	probe default {
+		.window = 1;
+		.threshold = 1;
+		.timeout = 1s;
+		.interval = 0.1s;
+	}
```
Dridi, line 15: Please add an explicit `.expect_close = true;` for coverage, we already have test cases without this field.

doc/sphinx/reference/vcl-probe.rst
```diff
@@ -95,6 +95,19 @@ The expected HTTP status, defaults to ``200``::
...
+
+Whether or not to expect the backend to close the TCP connection.
+
+Accepts ``true`` or ``false``, defaults to ``true``::
+
+    .expect_close = false;
+
+Note that probing takes more time when backend does not close the connection.
+Even when ``expect_close`` is set to ``false``, probe will still wait until timeout is
+reached before processing the response.
```
Dridi, line 109: Please change this text to:

> Warning: when the backend does not close the connection, setting `expect_close` to `false` makes probe tasks wait until they time out before inspecting the response.

Dridi (changes requested):
I caught a few things I didn't notice in my preview review.

lib/libvcc/vcc_utils.c
walid-git: I agree it will fail, but with a simple assert, whereas this one shows a clearer error message and designates the problematic token in the VCL file (line number and position)

Dridi: I see, then we should also add coverage with `.expect_close = 1`. Good catch!

Dridi:
Reviewed and merged as 908c9b811...6f50b7c8f with @walid-git.

PR #3887, opened 2023-01-18
varnishtest: File limits and timeouts

Dridi:
A small grab bag of changes from downstream, see individual commits for details.

### DIFF
--- a/bin/varnishtest/vtc.h
+++ b/bin/varnishtest/vtc.h
@@ -76,7 +76,7 @@ extern volatile sig_atomic_t vtc_error; /* Error, bail out */
 extern int vtc_stop;		/* Abandon current test, no error */
 extern pthread_t	vtc_thread;
 extern int iflg;
-extern unsigned vtc_maxdur;
+extern vtim_dur vtc_maxdur;
 extern char *vmod_path;
 extern struct vsb *params_vsb;
 extern int leave_temp;
--- a/bin/varnishtest/vtc_http.c
+++ b/bin/varnishtest/vtc_http.c
@@ -496,12 +496,12 @@ http_rxchar(struct http *hp, int n, int eof)
 		pfd[0].fd = hp->sess->fd;
 		pfd[0].events = POLLIN;
 		pfd[0].revents = 0;
-		i = poll(pfd, 1, hp->timeout);
+		i = poll(pfd, 1, hp->timeout * 1000);
 		if (i < 0 && errno == EINTR)
 			continue;
 		if (i == 0) {
 			vtc_log(hp->vl, hp->fatal,
-			    "HTTP rx timeout (fd:%d %u ms)",
+			    "HTTP rx timeout (fd:%d %.3fs)",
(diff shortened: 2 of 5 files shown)


Dridi:
> Why limit fd's

It's in the commit message, with a slightly different CI setup we have seen builds run forever in `VSUB_closefrom()`.

> and why limit them to 1000 ?

The limit was arbitrarily selected and has been good enough for us so far. We can add a command line option to tweak it if needed.

nigoroll:
Standardizing the timeout makes sense to me.
@AlveElde I do not understand why we should set the resource limit in `varnishtest`, when a simple `ulimit -n` in (a/the) shell calling it would do the same and be transparent.

AlveElde:
Looks like downstream is missing https://github.com/varnishcache/varnish-cache/commit/14721f5379ea644cec138dc73958ecbfcee09888, which mitigates the issue by looking for the highest fd in `/proc/<PID>/fd/`. This more or less accomplishes the same goal as https://github.com/varnishcache/varnish-cache/pull/3887/commits/bd103fc4141b8a5a0ca44d2c1820a3d18d3c4a3f, with two caveats:

1. The `varnishtest` process must have read permissions to `/proc/<PID>/fd/`. This is not guaranteed, as the process must have PTRACE_MODE_READ_FSCREDS access mode. I speculate that this could be an issue in containerized environments.
2. The `varnishtest` process must have no numerically high open file descriptors.

I have made https://github.com/AlveElde/varnish-cache/commit/c6c43454ed9b8be1e6c87f6cc5e410b66d5eff31 to address the second point, which tries to only `close()` open file descriptors. But this can still take a long time if you have a lot of files open. How long you ask? Running downstream `make check` with my systems default fd limit (524288), takes nearly double the time of running it with an fd limit of 1000. In more CPU constrained environments, like tiny CI containers, I believe we have seen even larger gains when decreaseing the fd limit.

On systems without `closefrom()`, and without access to `/proc/<PID>/fd/`, lowering the fd limit is pretty much mandatory IMO. We can either keep https://github.com/varnishcache/varnish-cache/pull/3887/commits/bd103fc4141b8a5a0ca44d2c1820a3d18d3c4a3f  (pending a CLI option to tweak the limit), or we can drop it and require affected users to always call `ulimit -n` before running `varnishtest`.

https://github.com/AlveElde/varnish-cache/commit/c6c43454ed9b8be1e6c87f6cc5e410b66d5eff31 also adds a test that will likely time out on systems that hit the `sysconf(_SC_OPEN_MAX);` branch with a high fd limit, letting you know that you should decrease it.

AlveElde:
Does it make sense to move to a single `VSUB_closefrom()` call in varnishtests `main()`? I don't really see a reason for calling it for every test we start, and for every `vtc_varnish` and `vtc_process` we start.

Dridi:
bugwash:
- @dridi to merge the timeout changes
- @AlveElde to follow up with a closefrom change

PR #3888, opened 2023-01-24
heritage: Remove unset params::[gu]id fields

Dridi:
They are shared with the cache process but are never used. Only the VCC
process uses them, but they are never set. This specific fchown(2) call
in the VCC process was probably a no-op in the first place: since the
fields are never set this is transferring ownership to root:root and if
that succeeded the process was already root in the first place. If it
failed, we never see the error message since we lacked root privileges.

When this fchown(2) call is made (multiple times) the files are created
by a VCC process that already dropped privileges, if applicable, via the
jail system. So the VJ_fix_fd() replacement may also be a no-op.

### DIFF
--- a/bin/varnishd/common/common_param.h
+++ b/bin/varnishd/common/common_param.h
@@ -140,9 +140,5 @@ struct params {
 #undef ptyp_vsl_mask
 #undef ptyp_vsl_reclen
 
-	/* Unprivileged user / group */
-	uid_t			uid;
-	gid_t			gid;
-
 	struct vre_limits	vre_limits;
 };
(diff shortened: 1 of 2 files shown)


bsdphk:
I think you are right that the ``VJ_fix_fd()`` is surplus to requirements in jail_unix.

In theory one could imagine a jail-implementation where the file had to be "given away" before CC can access it, but that would be a totally different level of paranoia, so I think we can disregard that.

Dridi:
bugwash: don't bother keeping the fchown(2) call via the jail system, just remove the dead code.

Dridi:
Followup patch: https://github.com/varnishcache/varnish-cache/commit/9b42ca376923b0cbf534c24f280bca477de27767

PR #3889, opened 2023-01-27
Ensure variable is initialized before printing it to resolve compilation error

jawn-smith:
This variable may have been used uninitialized, which was causing this package to fail to build successfully in Ubuntu. I've chosen -1 here for the error number, but would be happy to adjust that if you'd prefer.

### DIFF
--- a/lib/libvarnish/vnum.c
+++ b/lib/libvarnish/vnum.c
@@ -575,7 +575,7 @@ main(int argc, char *argv[])
 {
 	int ec = 0;
 	struct test_case *tc;
-	uintmax_t val;
+	uintmax_t val = 0;
 	const char **p;
 	const char *e;
 	double d1, d2;
@@ -678,8 +678,6 @@ main(int argc, char *argv[])
 		e = VNUM_2bytes(tc->str, &val, tc->rel);
 		if (e != NULL)
 			val = 0;
-		else
-			val = -1;
 		if (e == tc->err && val == tc->val)
 			continue;
 		++ec;
--- a/lib/libvarnish/vnum.c
+++ b/lib/libvarnish/vnum.c
@@ -678,6 +678,8 @@ main(int argc, char *argv[])
 		e = VNUM_2bytes(tc->str, &val, tc->rel);
 		if (e != NULL)
 			val = 0;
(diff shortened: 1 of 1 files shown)


Dridi:
Your patch is breaking continuous integration, could you please share the warning you tried to fix?

jawn-smith:
```
In file included from /usr/include/stdio.h:906,
                 from vnum.c:41:
In function ‘printf’,
    inlined from ‘main’ at vnum.c:688:3:
/usr/include/powerpc64le-linux-gnu/bits/stdio2.h:86:10: error: ‘val’ may be used uninitialized [-Werror=maybe-uninitialized]
   86 |   return __printf_chk (__USE_FORTIFY_LEVEL - 1, __fmt, __va_arg_pack ()); 
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
vnum.c: In function ‘main’:
vnum.c:578:19: note: ‘val’ was declared here 
  578 |         uintmax_t val; 
      |                   ^~~
      ```

I can look into the failing tests.

**Comment by Dridi** [comment]
This warning is not entirely wrong, `val` may not be initialized here:

```c
e = VNUM_2bytes(tc->str, &val, tc->rel);
```

But the `if` statement to which you added an `else` branch solves just that. Maybe `VNUM_2bytes()` should unconditionally assign something to its `r` argument.

Can you please share your `config.log` file?

Do you still get a warning with this diff?

```diff
--- lib/libvarnish/vnum.c
+++ lib/libvarnish/vnum.c
@@ -575,7 +575,7 @@ main(int argc, char *argv[])
 {
        int ec = 0;
        struct test_case *tc;
-       uintmax_t val;
+       uintmax_t val = 0;
        const char **p;
        const char *e;
        double d1, d2;
```

jawn-smith:
The warning is resolved with this diff, so I have updated the PR.

nigoroll (approved):

Dridi:
Cherry-picked as f088fc4d943224b5d781a8c4bb3f7e4f82af3b25, thank you.

PR #3891, opened 2023-02-02
varnishtest: Fork and clean up FD table before running tests

AlveElde:
This PR is a follow-up to #3887. It does two things:

1. On systems without the `closefrom` syscall, we now iterate through `/proc/<PID>/fd/` and individually close each FD.
2. The main `varnishtest` process is now forked, and the FD table is cleaned up, before we start forking out test processes.

The main motivation for these changes is that we have seen big performance gains by reducing the number of file descriptors each test process has to close, especially in CPU constrained environments.


nigoroll:
VDD: Keep the `VSUB_closefrom()` change only, we should not need to fork once again. OK to merge.

PR #3892, opened 2023-02-13
Honor H2 request limits

### LINKED ISSUE #3709: h2: Currently not honoring request limits
The following test case passes, it should fail. Attempting the same request with HTTP/1 definitely fails. I'm not sure how the correct test case should look like, my assumption is that stream 1 should be reset and the connection preserved.

```vcl
varnishtest "h2 req limits"

server s1 {
        rxreq
        expect req.url ~ "(/0123456789){30}"
        expect req.http.foo ~ "bar(, 0123456789){1300}"
        txresp
} -start

varnish v1 -cliok "param.set feature +http2"
varnish v1 -cliok "param.set http_req_size 256b"
varnish v1 -cliok "param.set http_req_hdr_len 40b"

varnish v1 -vcl+backend { } -start

client c1 {
        stream 1 {
                txreq -url "${string,repeat,30,/0123456789}" \
                    -hdr foo {bar${string,repeat,1300,", 0123456789"}}
                rxresp
                expect resp.status == 200
        } -run
} -run
```

walid-git:
This PR honors h2 request limits for http_req_size,  http_req_hdr_len, and h2_max_header_list_size 
Fixes #3709

### DIFF
--- a/bin/varnishd/http2/cache_http2.h
+++ b/bin/varnishd/http2/cache_http2.h
@@ -140,6 +140,7 @@ struct h2_req {
 	VTAILQ_ENTRY(h2_req)		list;
 	int64_t				t_window;
 	int64_t				r_window;
+	uint64_t			req_len;
 
 	/* Where to wake this stream up */
 	struct worker			*wrk;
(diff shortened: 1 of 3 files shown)


nigoroll:
bugwash:
* please improve test coverage - even without this patch it was not too good
* please test with an actual browser
* do we need to reconsider parameter defaults?
* test on a live server if possible

PR #3897, opened 2023-02-28
vtest: Allow to inject varnishd arguments via VTEST_VARNISHD_ADD_ARGS

nigoroll:
This is useful to test extensions with otherwise unaltered varnish test cases, for example:

```
VTEST_VARNISHD_ADD_ARGS='-E/tmp/lib/varnish/vmods/libvmod_slash.so -sfellow=fellow,${tmpdir}/fellow.stv,100MB,1MB,64KB -sTransient=fellow,${tmpdir}/transient.stv,100MB,1MB,64KB' ./varnishtest -i ...
```

These arguments are added and this injection method does not achieve its goal in all cases (e.g. for the example it breaks when other stevedore definitions conflict), but it still reduces the cases requiring manual intervention substantially.

### DIFF
--- a/bin/varnishtest/vtc_varnish.c
+++ b/bin/varnishtest/vtc_varnish.c
@@ -93,6 +93,8 @@ struct varnish {
 
 #define NONSENSE	"%XJEIFLH|)Xspa8P"
 
+#define VARNISHD_ADD_ARGS_ENV_VAR	"VTEST_VARNISHD_ADD_ARGS"
+
 static VTAILQ_HEAD(, varnish)	varnishes =
     VTAILQ_HEAD_INITIALIZER(varnishes);
 
@@ -389,7 +391,7 @@ varnish_launch(struct varnish *v)
 	char abuf[128], pbuf[128];
 	struct pollfd fd[3];
 	enum VCLI_status_e u;
-	const char *err;
+	const char *err, *env_args;
 	char *r = NULL;
 
 	/* Create listener socket */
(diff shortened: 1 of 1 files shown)


nigoroll:
I had originally flagged this freeze before the release, but as it concerns vtest only and there had not been negative feedback, I see no reason against a merge. Let me know if anyone disagrees.

Dridi:
Did we not wash this pull request at all since Feb 28th? I didn't really get a chance to object as I was in a call almost all of this afternoon.

I would have proposed a macro as an alternative, since this is the precedent we have in terms of interface:

```
varnishtest -Dvarnishd_args='-E/tmp/lib/varnish/vmods/libvmod_slash.so -sfellow=fellow,${tmpdir}/fellow.stv,100MB,1MB,64KB -sTransient=fellow,${tmpdir}/transient.stv,100MB,1MB,64KB' ...
```

Of course we would need this change internally:

```diff
-        env_args = getenv(VARNISHD_ADD_ARGS_ENV_VAR);
-        if (env_args)
-                VSB_printf(vsb, " %s", env_args);
-        VSB_printf(vsb, " %s", VSB_data(v->args));
+        VSB_printf(vsb, " %s ${varnishd_args}", VSB_data(v->args));
```

nigoroll:
@Dridi that's a nice alternative, let's do that and thank you!

nigoroll:
It's not going to look quite as nice as in the simple diff because I need macro expansion for the macro, so I will drop another PR.

Dridi:
Hum, can't you do something like this?

```
varnishtest -Dfoo=bar -Dbaz='${foo}' ...
```

Or maybe we don't expand macros when we create them?

nigoroll:
Your proposal adds a macro to the argument string, which gets expanded to whatever the macro contains. We do not (and should not, I think, for the general case) expand macros recursively. And no, they do not get expanded upon definition.

Dridi:
https://github.com/varnishcache/varnish-cache/pull/3919

PR #3898, opened 2023-03-02
Inform about the default working directory in usage information

nigoroll:
I noticed that users, apparently, have no way of finding out what their distribution configured?


Dridi (approved):
Not sure I would use a temp buffer here, but that's the least intrusive quick fix.

nigoroll:
> Not sure I would use a temp buffer here, but that's the least intrusive quick fix.

exactly why I did it this way.

PR #3901, opened 2023-03-06
build: Implement "proper" maintainer programs

Dridi:
There is a bunch of programs that are not strictly needed when building from a release "dist" archive, because we ship the deliverables. For example we should not require `rst2man` because pre-built manual pages are present in the dist archive, and likewise, we shouldn't need sphinx-build since we ship the HTML documentation.

One problem with how we deal with documentation however is that we generate some of them from programs we compile at build time, so we need to avoid updating such targets unless they actually change. This shouldn't be difficult.

There is however a maintainer program that goes one step further: not only do we ship SVGs in dist archives, we also check them in. So even when working from a git clone (as opposed to rebuilding from a dist archive) we don't require a working graphviz installation.

The first maintainer program promoted to the relevant autotools macros is dot(1) since it's the easiest one to deal with. The first actual patch makes a +6-47 diff to result in the same behavior, modulus a slightly different configure command line (when applicable) and a slightly different error message when dot(1) is needed but missing.

I'd like to take care of the other maintainer programs after the March release, meanwhile, please check that it does not just work on my machine. The macros I used are all old enough to satisfy our autoconf and automake minimum versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -405,7 +405,7 @@ workflows:
           name: build_fedora_latest
           dist: fedora
           release: latest
-          make_target: witness
+          make_target: witness.dot
       - build:
           name: build_fedora_rawhide
           dist: fedora
--- a/Makefile.am
+++ b/Makefile.am
@@ -81,7 +81,7 @@ gcov_digest:
 witness.dot: all
 	$(MAKE) check AM_VTC_LOG_FLAGS=-pdebug=+witness
 	$(AM_V_GEN) $(srcdir)/tools/witness.sh witness.dot bin/varnishtest/ \
-		lib/libvmod_*/
+		vmod/
 
 .dot.svg:
 	$(AM_V_GEN) $(DOT) -Tsvg $< >$@
--- a/Makefile.am
+++ b/Makefile.am
@@ -84,19 +84,8 @@ witness.dot: all
 		lib/libvmod_*/
 
 .dot.svg:
-if ! HAVE_DOT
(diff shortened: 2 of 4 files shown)


Dridi:
> * not quite?
> 
> 
> ```
> [uplex@vtest2 ~/src/dridi/varnish-cache]$ pfexec mv /opt/local/bin/dot /opt/local/bin/dot.mv
> [uplex@vtest2 ~/src/dridi/varnish-cache]$ type dot
> -bash: type: dot: not found
> [uplex@vtest2 ~/src/dridi/varnish-cache]$ ./autogen.sh  && ./configure --prefix=/tmp/dridi && make -j40
> ...
> configure.ac:707: warning: AC_CACHE_VAL(ac_cv_have_viz, ...): suspicious presence of an AC_DEFINE in the second argument, where no actions should be taken
> ../../lib/autoconf/general.m4:2031: AC_CACHE_VAL is expanded from...
> ../../lib/autoconf/general.m4:2052: AC_CACHE_CHECK is expanded from...
> configure.ac:707: the top level
> ```

This viz check is related to zlib, not graphviz. I think we can safely rule this warning out.

nigoroll:
> This viz check is related to zlib

oh. confirmation bias. sorry.

Dridi:
> oh. confirmation bias. sorry.

No worries, I also thought you had found something related, I had to check the configure script...

Dridi:
bugwash: Ok, subsequent changes in this area in subsequent pull requests.

PR #3902, opened 2023-03-06
release: Prep work before the next release

Dridi:
This is currently work in progress, but I plan to cover a first sweep of:

- [x] VSL (in)compatibility
- [x] initial release notes
- [x] initial upgrade notes
- [x] VRT history
- [x] libvarnishapi review
- [ ] gc SLTs?

### DIFF
--- a/doc/sphinx/whats-new/changes-trunk.rst
+++ b/doc/sphinx/whats-new/changes-trunk.rst
@@ -164,4 +164,12 @@ Custom backend implementations are now in charge of printing headers, which
 avoids duplicates when a custom implementation relied on ``http_*()`` that
 would also log the headers being set up.
 
+The ``VRT_new_backend*()`` functions take an additional backend argument, the
+optional via backend. It can not be a custom backend implementation, but it
+can be a director resolving a native backend.
+
+There is a new ``authority`` field for via backends in ``struct vrt_backend``.
+
+There is a new ``exp_close`` field in ``struct vrt_backend_probe``.
+
 *eof*
--- a/doc/sphinx/whats-new/changes-trunk.rst
+++ b/doc/sphinx/whats-new/changes-trunk.rst
@@ -25,8 +25,32 @@ Parameters
 
 **XXX changes in -p parameters**
 
+There is a new parameter ``transit_buffer`` disabled by default to limit the
+amount of storage used for uncacheable responses. This is useful in situations
+where slow clients may consume large but uncacheable objects, to prevent them
+from filling up storage too fast at the expense of cacheable resources. When
+transit buffer is enabled, a client request will effectively hold its backend
+connection open until the client response delivery completes.
+
 Other changes in varnishd
 ~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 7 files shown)


Dridi:
I'm afraid not, sorry. But we could maybe build packages with jemalloc support disabled for ARM? cc @hermunn

jocel1:
That would definitely be great, I'm currently using the patch from 3881 to disable jemalloc on AWS ARM EC2 instances (both graviton 2 and graviton 3), and it has been working without any issue for 3 months now.

hermunn:
> I'm afraid not, sorry. But we could maybe build packages with jemalloc support disabled for ARM? cc @hermunn

I think we should use slightly different words here, and simply decide that the default is building without `jemalloc` on ARM. If we do this, then we will ship packages without `jemalloc` wit an entry in the change log about the change. I would love to have more data, of course, but right now I think switching away from `jemalloc` on ARM is a natural choice.

jocel1:
Hi @hermunn!
FYI I've added a few data about the behaviour on arm with and without jemalloc on https://github.com/varnishcache/varnish-cache/issues/3879

Dridi:
For RPMs it could look like this:

```diff
--- redhat/varnish.spec
+++ redhat/varnish.spec
@@ -81,7 +81,11 @@ Varnish Cache is a high-performance HTTP accelerator
 
 
 %build
+%ifnarch aarch64
 %configure --localstatedir=/var/lib --with-contrib
+%else
+%configure --localstatedir=/var/lib --with-contrib --without-jemalloc
+%endif
 %make_build V=1
```

Not sure about other package managers.

Dridi:
These rules are actually broken, we would do this instead:

```diff
--- debian/rules
+++ debian/rules
@@ -3,18 +3,18 @@
 DH_VERBOSE=1
 
 # List of architectures where jemalloc is not available
-DISABLE_JEMALLOC_ARCH_LIST := hppa m68k
+WITHOUT_JEMALLOC_ARCH_LIST := hppa m68k arm64
 
 # Explicitly initialize a variable to select architecture, unless it has been
-# defined before.  This is compared against the DISABLE_*_LIST variables later
+# defined before.  This is compared against the WITHOUT_*_LIST variables later
 # in this makefile
 DEB_HOST_ARCH ?= $(shell dpkg-architecture -qDEB_HOST_ARCH)
 
 # Set local state dir for FHS
 LOCAL_CONFIGURE_FLAGS  = --localstatedir=/var/lib --libdir=/usr/lib
 
-ifneq ($(filter $(DEB_HOST_ARCH),$(DISABLE_JEMALLOC_ARCH_LIST)),)
-LOCAL_CONFIGURE_FLAGS += --disable-jemalloc
+ifneq ($(filter $(DEB_HOST_ARCH),$(WITHOUT_JEMALLOC_ARCH_LIST)),)
+LOCAL_CONFIGURE_FLAGS += --without-jemalloc
 endif
 
 # Ship contrib scripts
```

Dridi:
https://github.com/varnishcache/pkg-varnish-cache/pull/162

PR #3908, opened 2023-03-16
fix bug in abstract sockets

walid-git:
Current behavior: 
```
bin/varnishd/varnishd -a @varnish -f /etc/varnish/default.vcl -n /tmp/varnishd`
ss -lnxp | grep varnish
u_str LISTEN 0      1024   @varnish@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ 2741935            * 0    users:(("cache-main",pid=894543,fd=3),("varnishd",pid=894529,fd=3))  
```
As you can see, the whole 108 bytes of sun_path are used as the socket name/path.
From man UNIX(7):
```
The socket's address in this namespace is given  by  the  additional  bytes  in  sun_path  that are covered by the specified length of the address structure.  (Null bytes in the name have no special significance.)  The name has no connection with filesystem pathnames.  When the address of an abstract socket is returned, the returned addrlen is greater than sizeof(sa_family_t) (i.e., greater than 2), and the name of the socket  is  contained  in  the first (addrlen - sizeof(sa_family_t)) bytes of sun_path.
```
Which means that socklen is used to determine the length of name and must be set to the length of the struct sockaddr_un up to the last character of name, otherwise the 108 characters of sun_path will be treated as the name of the socket, including NULL bytes.


walid-git:
I agree that it can look confusing, but for me `Null bytes in the name have no special significance.` means that NULL bytes do not act as string delimiters, which means that the only other way is to use the length argument.

nigoroll:
So this means we are incompatible with other implementations? Have you actually checked/tested?

walid-git:
Yes I tried to connect as a client with `socat -  ABSTRACT-CLIENT:varnish`, it is unable to connect without this patch

lib/libvarnish/vus.c
```diff
@@ -95,7 +95,12 @@ int
 VUS_bind(const struct sockaddr_un *uds, const char **errp)
 {
 	int sd, e;
-	socklen_t sl = sizeof(*uds);
+	socklen_t sl;
+
+	if (*uds->sun_path)
+		sl = sizeof(*uds);
+	else
+		sl = strchr(uds->sun_path + 1, '\0') - (const char*)uds;
```
nigoroll, line 103: We could add assertions following from `sun_init()`: `assert(uds->sun_path[1] != '\0');`.
and assert that `strchr()` returns non-null for the same reason.
sl should be checked to not exceed `sizeof(*uds)`.
And all this probably qualifies to wrap it in a function for reuse.

PR #3909, opened 2023-03-29
build: Install vmod_*.vcc files

Dridi:
This change adds a new /usr/share/varnish/vcc (or equivalent) directory where the built-in VMOD descriptors are installed. Only vmod_debug.vcc is omitted since vmod_debug is not installed. Shipping VCC files can be useful to end users, for example, to implement VCL autocompletion in a text editor.

This change does not even begin to manage this for third-party VMODs.


gquintard:
I like where this is going,

I'm happy to update `varnish-modules`, the rust ones and to open PRs to `libvmod-dynamic` (i.e. the ones I care about). Should I do that for `vmod-example`, and follow-up question: do we care about it, and/or do we have something better?

Dridi:
> Should I do that for `vmod-example`, and follow-up question: do we care about it, and/or do we have something better?

I may look at this later and experiment with vcdk, but right now the scope is only built-in VMODs.

Since I was in the neighborhood I bundled a cleanup of VSC make rules, please yell if this should be a separate pull request.

Dridi:
> Since I was in the neighborhood I bundled a cleanup of VSC make rules, please yell if this should be a separate pull request.

circleci is yelling in my direction that it only works on Fedora, so I will submit it independently later.

Dridi:
https://github.com/varnishcache/varnish-cache/pull/3910

Dridi:
bugwash: OK, I will come up with third-party support later.

PR #3910, opened 2023-03-29
build: Centralize suffix .c.vsc make rules

Dridi:
The new vsc.am file takes care of most of the logic, and include sites only need to reference sources as such.


nigoroll (approved):

nigoroll (approved):
I am still not the master of make which Dridi is, but I like this improvement.

Dridi:
bugwash: OK, I will come up with third-party support later.

PR #3912, opened 2023-04-01
Add a private pointer to struct vdp

nigoroll:
to facilitate custom filter parametrization:

VMOD filters are likely to use the same init/bytes/fini callbacks for many vcl objects, which differ only in their parameters. It would be helpful to have a place to stash a per VCL private pointer.


nigoroll:
phk ok during bugwash

PR #3914, opened 2023-04-05
bereq: Also unset a cached req body

LaughingQuoll:
From an Enterprise ticket, trying to unset a cached request body currently accomplishes nothing.

The body can be sent after the first request (without Content-Length), and the body bytes become a part of the next request.


Dridi (approved):

nigoroll (approved):

nigoroll:
phk was go during bugwash, but it needs a `changes.rst` entry

PR #3916, opened 2023-04-11
Leave `HSH_Cancel()` to the transport

nigoroll:
This patch enables transports to keep references also to _final_ (`OC_F_PRIVATE | OC_F_HFM | OC_F_HFP`) objects until after the transport's delivery callback returns, effectively making them responsible for deciding when/if they do not need a _final_ object's body any more.

The net effect of the patch should be zero: where `HSH_Cancel()` was called from `cnt_transmit()`, it is now called before happy path returns from transports' deliver callbacks. For error returns, a call remains in `cnt_transmit()`.

The use case is [libvdp-pesi](https://gitlab.com/uplex/varnish/libvdp-pesi): In this VMOD, we currently need to copy _final_ objects because of that single `HSH_Cancel()` call *1). With this patch, it becomes possible to only keep a reference and avoid the copy for one out of two cases *2).
The resulting simplification can be seen here: https://gitlab.com/uplex/varnish/libvdp-pesi/-/merge_requests/2/diffs

----

*1) I had also explored an alternative (`T_FINAL`) node type to block the ESI subrequest until time comes to execute, but that is prone to deadlocks, so I never activated that code to run by default.

*2) The other case is less relevant and could be addressed later by adding, for VDPs, the option veto the `final` argument to `ObjIterate()`.

### DIFF
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -866,19 +866,24 @@ ved_deliver(struct req *req, struct boc *boc, int wantbody)
 
 	CAST_OBJ_NOTNULL(ecx, req->transport_priv, ECX_MAGIC);
 
-	if (wantbody == 0)
+	if (wantbody == 0) {
+		HSH_Cancel(req->wrk, req->objcore, boc);
 		return;
+	}
 
 	if (!ecx->incl_cont &&
 	    req->resp->status != 200 &&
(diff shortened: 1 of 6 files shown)


nigoroll:
bugwash: good idea by @bsdphk : Try if moving `HSH_Cancel()` to `VDP_Close()` would work

nigoroll:
obsoleted by #3927

PR #3917, opened 2023-04-14
cache: It's time for the big quit

Dridi:
When mgt sends a command to the cache process, whether it is a period ping or an actual operation, it must complete within cli_timeout. When the cache fails to meet this requirement, mgt sends a SIGQUIT signal to the cache process. As a result the cache process MAY dump a core file for post-mortem analysis.

When the core file is missing we are left to our own devices.

To mitigate this, a new signal handler is added for SIGQUIT, but since we can't (or don't even try to) guarantee delivery on the CLI thread, we make a last-ditch effort to forward SIGQUIT signals to properly panic from the CLI thread. With a regular panic we may get both a panic report and a core dump.

I didn't add test coverage for this, since we try to avoid intentional core dumps in test cases with the `no_coredump` feature flag that turns SIGQUIT into a SIGKILL signal.

---

This has already proven useful for troubleshooting on a system where core dumps were somehow missing.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -605,7 +605,8 @@ void BAN_Abandon(struct ban_proto *b);
 
 /* cache_cli.c [CLI] */
 extern pthread_t cli_thread;
-#define ASSERT_CLI() do {assert(pthread_equal(pthread_self(), cli_thread));} while (0)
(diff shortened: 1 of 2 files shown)


nigoroll (approved):

Dridi:
Bugwash: OK

PR #3919, opened 2023-04-14
varnishtest: Introduce a ${varnishd_args} macro

Dridi:
> If the ${varnishd_args} macro is defined, it is expanded and
> appended to the varnishd command line, before the command line
> itself is expanded. This enables tweaks to the varnishd command
> line without editing test cases. This macro can be defined using
> the ``-D`` option for varnishtest.

Refs #3897


nigoroll (approved):
👍🏽

nigoroll:
bugwash: I will double check if this does what I need, then merge if so

PR #3924, opened 2023-04-24
Close VDPs also for error conditions and extend debug.pedantic vdp to test

nigoroll:
Context: This issue was spotted while working on #3916 

We currently miss to close VDPs for transport error conditions. This PR adds a (possibly redundant) `VDP_Close()` call to `cnt_transmit()` to ensure that processors' fini method is always called.

The rest of the PR is test infrastructure: the `debug.pedantic` VDP now asserts that the fini method be called by using a `PRIV_TASK` which, when _it_ is finalized, the VDP already was.


nigoroll:
phk ok during bugwash

PR #3929, opened 2023-05-16
build: Generate VSCs .c and .h files separately

### LINKED ISSUE #3928: Make dependency missing in vmod/
### Current Behavior

    critter phk> cd vmod
    critter phk> make clean >& /dev/null
    critter phk> make |& tail -10
    touch vcc_debug_if.c
      CC       libvmod_debug_la-vmod_debug.lo
    vmod_debug.c:45:10: fatal error: 'VSC_debug.h' file not found
    #include "VSC_debug.h"
             ^~~~~~~~~~~~~
    1 error generated.
    *** Error code 1
 
    Stop.
    make: stopped in /critter/Varnish/trunk/varnish-cache/vmod
    critter phk> make -k >& /dev/null
    critter phk> make |& tail -10
      CC       libvmod_debug_la-vmod_debug.lo
      CCLD     libvmod_debug.la

There seems to be neither a dependency on ``VSC_debug.h`` nor any make-rule to produce it, should such a dependency have existed.

### Varnish Cache version

Trunk

### Operating system

FreeBSD-Current.

Dridi:
Fixes #3928


Dridi:
I forgot about that, I will add this rule too.

Dridi:
bugwash: OK, will merge after adding the suggested rule.

Dridi:
RST rules ready for a review.

nigoroll:
I didn't realize you were waiting for me. No objections.

I'll add one question, but that should  not hold us up.

vsc.am
```diff
@@ -1,15 +1,30 @@
...
 
 .vsc.c:
-	$(AM_V_GEN) $(PYTHON) $(VSCTOOL) -ch $<
+	$(AM_V_GEN) $(PYTHON) $(VSCTOOL) -c $<
+
+.vsc.h:
+	$(AM_V_GEN) $(PYTHON) $(VSCTOOL) -h $<
+
+.vsc.rst:
+	$(AM_V_GEN) $(PYTHON) $(VSCTOOL) -r $< >${@}_
```
nigoroll, line 24: Only out of interest: why are you using this pattern for rst and not for the other targets? Is that because for .h and .c, the build will fail, but not for an empty .rst?

nigoroll (approved):

PR #3934, opened 2023-06-05
vrb: Reduce storage pressure from req.body

Dridi:
Instead of trying to allocate storage for the entire request body when the content length is known, let it go in fetch_chunksize steps. This may prevent spurious cache evictions for large request bodies for pass transactions when we are merely streaming the body.

On the other hand, when the content length is known and the goal is to cache the request body, we may still attempt a single allocation and all that entails.

### DIFF
--- a/bin/varnishd/cache/cache_req_body.c
+++ b/bin/varnishd/cache/cache_req_body.c
@@ -60,6 +60,7 @@ vrb_pull(struct req *req, ssize_t maxsize, objiterate_f *func, void *priv)
 	enum vfp_status vfps = VFP_ERROR;
 	const struct stevedore *stv;
 	ssize_t req_bodybytes = 0;
+	unsigned hint;
 
 	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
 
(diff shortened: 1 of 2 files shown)


bin/varnishd/cache/cache_req_body.c
```diff
@@ -122,6 +125,8 @@ vrb_pull(struct req *req, ssize_t maxsize, objiterate_f *func, void *priv)
 			req_bodybytes += l;
 			if (yet >= l)
 				yet -= l;
+			else if (yet > 0)
+				yet = 0;
```
Dridi, line 129: Dead branch spotted by @nigoroll.

Dridi:
bugwash: waiting for @nigoroll's approval.

I will first double check the dead branch, and CI failure for bullseye (job with sanitizers).

bin/varnishd/cache/cache_req_body.c
Dridi: `l` may be increased by `VFP_GetStorage()` and become larger than `yet`, in which case this branch may be reached when the remaining bytes fit entirely in the allocated buffer.

Dridi:
The sanitizer failure also happens on the master branch, and could probably be solved with a revert of 20824dc0f75c71dbcd0a2edc83bc88c1b8293641 modulus a conflict resolution. But this isn't straightforward since this may break other CI jobs (we can always stick a `feature 64bit` with a comment).

Ready to merge with no reserves on my end after @nigoroll's approval.

bin/varnishd/cache/cache_req_body.c
nigoroll: Oh, sorry, I confused the `l` for a `1`.

nigoroll (approved):

PR #3936, opened 2023-06-09
collect errno after pthread_* calls

walid-git:
Commit log: 
> pthread_* calls don't set errno but return it instead. This cocci patch
ensures we always collect the errno value returned by pthread calls
when asserting

First commit adds the coccinelle patch and second one applies it


bsdphk:
I am a little bit worried about the implications of this in terms of generated instructions.

Have anybody looked at two objdump(1) outputs to see what the consequences are?

Dridi:
We have not, but what we have run into is the uncertainty of why certain lock operations failed, especially those that may return EAGAIN.

Dridi:
An alternative could be to have a dedicated thread-local `errno`-like variable to pick up in the panic code. Going with `errno` looked simpler.

nigoroll (approved):
I would have asked about 🐞 , but it's there ;)

walid-git:
Added PTOK macro as suggested by phk during bugwash to avoid editing errno when not needed

include/vas.h
```diff
@@ -90,6 +90,15 @@ do {									\
 	VAS_Fail(__func__, __FILE__, __LINE__, expl, VAS_WRONG);	\
 } while (0)
 
+#define PTOK(call)							\
+do { 									\
+	int _pterr = call;						\
```
nigoroll, line 95: could use `VUNIQ_NAME()`, but that's not too relevant

Dridi: I completely forgot about this macro, we could use it in `REPLACE()` too.

nigoroll: lgtm

nigoroll (approved):

nigoroll:
bugwash OK

PR #3947, opened 2023-06-28
pool: Safely update pool statistics

Dridi:
I ran into a waiting list queuing case recently and the lack of consistent `sess_queued` was really getting in the way.


Dridi:
bugwash: no objections

PR #3954, opened 2023-07-09
Simple storage: Finalize busy object state also for ObjFreeObj()

### LINKED ISSUE #3953: stevedore busy object finalization missing for "object swap" in vbf_stp_error
This issue was originally discovered by @xcir with SLASH/, documented in [slash issue #14](https://gitlab.com/uplex/varnish/slash/-/issues/14):

Here in `vbf_stp_error()`, ...
https://github.com/varnishcache/varnish-cache/blob/046724d6cc2660299e8d98c226b82b4a800c404b/bin/varnishd/cache/cache_fetch.c#L905

...  an existing storage object is freed to be replaced with a new one, potentially from a different storage (usually `-sTransient`).

In this case, `ObjBocDone()` is not called, because the `boc` is to be preserved - from the client thread's perspective, the backend transaction is still in progress, and the client is waiting for the busy object to finish.

At the storage level, however, the process of creation of a new object _is_ finished, and any busy object state which the storage might keep needs to be finalized.

nigoroll:
If `oc->boc` is not `NULL` for an `ObjFreeObj()` call, the stevedore needs to finish any busy object transaction.

Please see bug ticket for a more detailed explanation.

Fixes #3953


Dridi:
How hard would it be to create a test case?

nigoroll:
> How hard would it be to create a test case?

We would need another storage engine which occupies `boc->stevedore_priv`, and then make b20.vtc run into it. This could be done with the `debug` stevedore, but we'd need to wrap/copy some code from `malloc` and I was not sure if it's worth it.

BTW, it's nice that you picked up the `TAKE_OBJ_NOTNULL()` idea. :)

nigoroll:
bugwash participants would like to get @mbgrydeland\ 's opinion, but I actually think we should be fine:

> This is quite contained in varnish-cache, and any stevedore would, if it used stevedore_priv like the in-tree code, hit the same thing. Also, the fix for stevedores does not require the fix in v-c

mbgrydeland:
Reading the ticket description, I guess my noop-argument is false. But my fear of an introduced race persists.

mbgrydeland:
I think a cleaner solution would be to introduce a `ObjBocReset()` function that is called from where core decides to reuse the boc.

nigoroll:
Oh, and I forgot: Thank you for your review, @mbgrydeland !

mbgrydeland:
Thanks @nigoroll , your detailed explanation has peaced my mind. I have no issues with this patch.

Issue #3946, opened 2023-06-25
Add libvarnish.la to varnish-devel packages


neufeind:
### Expected Behavior

Build the slash-vmod mentioned in your news should work using varnish-devel package (on Rocky 9).

### Current Behavior

varnish-devel is missing libvarnish.la. That is created during compilation but not added to varnish-devel during packaging.

### Possible Solution

For now it's possible to build the slash-vmod by hand-compiling the whole varnish. But the docs for building the module on Debian suggest that installing varnish-devel (along with other build-tools) should be sufficient. However the rpm (package used here on Rocky 9) does not contain libvarnish.la.

### Steps to Reproduce (for bugs)

1. Install varnish and varnish-devel (7.3.0) from packagecloud-repos on Rocky 9.
2. Try to build slash-vmod (7.3 branch).

### Context

Trying to build slash-vmod.

### Varnish Cache version

varnishd (varnish-7.3.0 revision 84d79120b6d17b11819a663a93160743f293e63f)

### Operating system

Rocky Linux 9

### Source of binary packages used (if any)

https://packagecloud.io/varnishcache/varnish73/el/9/$basearch

nigoroll:
Hi @neufeind,
I am the author of SLASH/, which is an extension not part of core varnish-cache code. Thus, a better place for this issue would have been https://gitlab.com/uplex/varnish/slash/-/issues

Regarding your suggestion:

* At this point, `libvarnish` is not planned for inclusion in `varnish-devel` because it is only intended to be used by `varnishd` and bundled tools.
* `vmod_slash` itself would not need to statically link to it either.
* but SLASH/ brings a along unit tests and an additional utility (`fellow_log_dbg`), which need functions from `libvarnish` 

I understand that having to compile varnish-cache may be an inconvenience, but [INSTALL.rst](https://gitlab.com/uplex/varnish/slash/-/blob/master/INSTALL.rst) clearly mentions this requirement:

> Build required components from varnish-cache:
> [...]

Note that besides `libvarnish`, SLASH/ also needs `libvsc` from varnish-cache.

neufeind:
[...]

PR #3956, opened 2023-07-10
vfil: Make struct vfil_path usable outside of varnishd

Dridi:
This is a cherry-pick of the uncontroversial parts from #3182 motivated by https://github.com/varnishcache/varnish-cache/issues/3946#issuecomment-1627701080.

If we proceed with exposing libvarnish entirely, I would rather spin a `vpath.h` off `vfil.h` for this purpose.

### DIFF
--- a/include/miniobj.h
+++ b/include/miniobj.h
@@ -30,6 +30,16 @@
 			(to)->magic = (type_magic);			\
 	} while (0)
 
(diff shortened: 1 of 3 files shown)


Dridi:
bugwash: proceed, and extract `VPATH_` symbols from `VFIL_`.

PR #3961, opened 2023-07-21
Add default "Server" and "User-Agent" headers in vtc

walid-git:
As discussed on varnish-dev ML: https://varnish-cache.org/lists/pipermail/varnish-dev/2023-July/009755.html

This adds default "Server: sNAME" and "User-Agent: cNAME" headers to server `txresp` and client `txreq` respectively in vtc, with the ability to disable them with `-noserver` and `-nouseragent`. Also, both can be overridden with `-hdr`

### DIFF
--- a/bin/varnishtest/tests/a00024.vtc
+++ b/bin/varnishtest/tests/a00024.vtc
@@ -1,0 +1,52 @@
+varnishtest "test -noserver"
+
+server s1  {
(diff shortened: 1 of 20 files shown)


bsdphk:
OK with me.

nigoroll (approved):

PR #3966, opened 2023-08-18
vte: Provide a CLI-independent API

Dridi:
Some work on the VTE was requested somewhere else, and it so happened to align with #3946 goals, although, this pull request is not moving anything to libvarnishapi.

The new VTE API is inspired by the VSB API and ultimately avoids one printf per character.

One thing I didn't attempt was to adopt the VTE for the `varnishstat -1` output.

### DIFF
--- a/bin/varnishd/cache/cache_director.c
+++ b/bin/varnishd/cache/cache_director.c
@@ -41,6 +41,7 @@
 #include "cache_director.h"
 
 #include "vcli_serve.h"
+#include "vte.h"
 #include "vtim.h"
 
 /* -------------------------------------------------------------------*/
@@ -309,6 +310,7 @@ struct list_args {
 	int		j;
 	const char	*jsep;
 	struct vsb	*vsb;
+	struct vte	*vte;
 };
 
 static const char *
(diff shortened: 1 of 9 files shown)


bsdphk:
Are queue.h and vcli_serve.h used in vte.c after this ?

bsdphk:
Now that I see what the VALLOC_FLEX_OBJ() is for I'm not very keen on it, it is far too opaque for my taste.

What alternatives do we have ?

Dridi:
> Are queue.h and vcli_serve.h used in vte.c after this ?

Good catch, I just removed them.

> Now that I see what the VALLOC_FLEX_OBJ() is for I'm not very keen on it, it is far too opaque for my taste.
> 
> What alternatives do we have ?

This is not strictly part of this pull request, although this patch series was one of its first consumers. I dropped `offsetof()` as requested during bugwash (9cc2fd874758e4cfcbfe2cb22b46fd629041b3ad) and added `ALLOC_OBJ_EXTRA()` (9c3d28cefcec462ac8cdb413f17428900c0cb08e).

I will merge this once CI completes.

PR #3969, opened 2023-08-22
vte: Add support for left-justified fields

Dridi:
A meeting was pushed today, so I was able to quickly do it.


Dridi:
bugwash: ok

PR #3972, opened 2023-09-01
Rename VSB_QUOTE_JSON and prepare VSB_quote_*() for returning errors

nigoroll:
We rename our existing JSON quoting to make clear that it is not quite JSON.

For standard JSON quoting, we are going to need to signal failures, so prepare VSB_quote*() for that.

### DIFF
--- a/bin/varnishd/cache/cache_conn_pool.c
+++ b/bin/varnishd/cache/cache_conn_pool.c
@@ -559,7 +559,7 @@ VCP_Panic(struct vsb *vsb, struct conn_pool *cp)
 	if (PAN_dump_struct(vsb, cp, CONN_POOL_MAGIC, "conn_pool"))
 		return;
 	VSB_cat(vsb, "ident = ");
-	VSB_quote(vsb, cp->ident, VSHA256_DIGEST_LENGTH, VSB_QUOTE_HEX);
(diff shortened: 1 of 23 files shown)


Dridi:
I find this too intrusive.

In general adding a return value to `VSB_quote*()` is a good idea, but instead of guarding against failure everywhere, we should relax the return value check for flexelint. This is what we do in many places: feed the VSB and wait until `VSB_finish()` to check for errors. Consumers of `VSB_quote*()` may otherwise perform early checks when relevant.

Even worse, the indiscriminate coccinelle patch added assertion to panic code paths, potentially raising a signal inside a signal handler.

I think we should limit the change to `vsb.c`, `vsb.h` and `/flint.lnt`.

I haven't thought about the JSON part of the change.

nigoroll:
Discussed on IRC: @bsdphk does not see standard conforming JSON encoding as a core function of VSB, so together with @Dridi's objections, the API change is dead.
I wonder if we should still do the JSON->VJSON rename to make it obvious that it's not-quite-json.

nigoroll:
> I wonder if we should still do the JSON->VJSON rename to make it obvious that it's not-quite-json.

dismissed on irc

PR #3976, opened 2023-09-06
varnishd: plugable acceptors

asadsa92:
Make the acceptor code plugable and start off with two acceptor modules TCP and UDS.
Adding a new QUIC acceptor module in the future should be easy now.


nigoroll:
👍🏽 but could not look in detail yet

nigoroll (commented):
I am back at this topic and would want to work on it for real now.
The changes LGTM, except for the rename of _VCA_ to _ACC_. What was wrong about _VCA_ that we need to rename, or what is better about _ACC_? (I can see a bit of a point with MAC (which also to me is more https://en.wikipedia.org/wiki/Medium_access_control), but Management ACceptor is not exactly a bad idea...)

nigoroll:
I am currently working on a rebase and just a general comment: Copying code out to new files is really not a good idea to keep changes manageable.

nigoroll:
I will use this note to (hopefully) track additional changes which I deem necessary:
* Add an acceptor private argument b3b96c251dbaafc471a1ccf7fcce77ca37ffd707

nigoroll:
bugwash: merge the rebase up to the point of this PR, rename to VAC

dridi:
One remark I made to @asadsa92 and forgot to add here last week is that we should maybe have a "simple" or "stream" pseudo-acceptor (akin to simple storage) that contains the common code between TCP and UDS, leaving the actual TCP and UDS acceptors deal with their specific needs like socket options.

nigoroll:
> we should maybe have a "simple" or "stream" pseudo-acceptor

I agree.

Also I am not sure if a hard uds/tcp makes too much sense. Based on some WIP, I would think that acceptor implementations should be more concerned about the "how" and not so much about the "what endpoint".

The "what endpoint" stuff (basically argument parsing, opening an fd and naming the socket for VCL) could go into the common code.

nigoroll:
I have edited the commits to replace acc/ACC with vca/VCA where I thought it made sense. This was semi-automatic (a bit of sed and a bit of vi) on patch files, then applied with `git am`.
commits are 965f7162118fcc1011670146268672960597c305..99261483189d462c4609cc9df0a766fb972028b0

PR #3979, opened 2023-09-08
varnishd: introduce a new paramter 'reuseport'

asadsa92:
The reuseport paramter controls whether or not we will create a listen group in the kernel using SO_REUSEPORT (or SO_REUSEPORT_LB for some platforms). This leads to a more efficient load distribution across the thread pools.

For backward compability reasons, leave this off by default.

### DIFF
--- a/lib/libvarnish/vtcp.c
+++ b/lib/libvarnish/vtcp.c
@@ -470,6 +470,82 @@ VTCP_bind(const struct suckaddr *sa, const char **errp)
 	return (sd);
 }
 
+int
+VTCP_bind_reuseport(const struct suckaddr *sa, const char **errp)
+{
+	int sd, val, e;
+	socklen_t sl;
+	const struct sockaddr *so;
+	int proto;
+
+	if (errp != NULL)
+		*errp = NULL;
+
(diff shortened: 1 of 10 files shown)


nigoroll (approved):
👍🏽 . As discussed during bugwash, I would (also) vote for default on.
FWIW, I did not stare at the code for long, but I guess we should find any issues pretty soon, if they exist

lib/libvarnish/vtcp.c
```diff
@@ -470,6 +470,82 @@ VTCP_bind(const struct suckaddr *sa, const char **errp)
...
+	if (setsockopt(sd, SOL_SOCKET, SO_REUSEPORT, &val, sizeof val) != 0) {
+		if (errp != NULL)
+			*errp = "setsockopt(SO_REUSEPORT, 1)";
+		e = errno;
+		closefd(&sd);
+		errno = e;
+		return (-1);
+	}
+#else
+	WRONG("SO_REUSEPORT[_LB] not available (set -p reuseport=off)");
```
nigoroll, line 522: this should not be fatal

PR #3982, opened 2023-09-15
Add Python3.11 and `--with-python=`

catap:


Dridi:
I will look at the other changes, but please remove the `--with-python` option.

You can already provide a specific python command at configure time:

```
./configure PYTHON=python3.11
```

Since we require python, let's not bring back the `--with` integration (it implies `--without` too).

catap:
@Dridi ah; if custom / expected Python can be provided… shall I revert value to default aclocal settings which should be updated one day?

Dridi:
I don't know, I need to sit down and look at it, but thank you for bringing it up.

catap:
See some context here: https://bugs.gnu.org/53530

Dridi:
If only `_AS_PATH_WALK` was not an internal macro...

catap:
Well, `./configure ... PYTHON=/bla/bla/python3.11` works as expected.

Dridi:
FYI: #3990

catap:
Let’s close it when

Issue #3988, opened 2023-09-20
docker varnish


pleaseforget:
### Expected Behavior

works, starts

### Current Behavior

breaks with error

```
Assert error in VSUB_closefrom(), vsub.c line 71:
  Condition((close_range(fd, ~0U, 0)) == 0) not true.
  errno = 1 (Operation not permitted)
Assert error in VSUB_closefrom(), vsub.c line 71:
  Condition((close_range(fd, ~0U, 0)) == 0) not true.
  errno = 1 (Operation not permitted)
```

### Possible Solution

downgrade varnish

### Steps to Reproduce (for bugs)

start docker  containe with empty or minimal config.

### Context

dont starts

### Varnish Cache version

VARNISH_VERSION=7.4.0

### Operating system

ubuntu22.04

### Source of binary packages used (if any)

https://hub.docker.com/_/varnish/tags

gquintard:
works here:

```
$ docker run varnish:7.4.0
Debug: Version: varnish-7.4.0 revision f07a6d615c054cf73c9250f3f09d063ac344dff2
Debug: Platform: Linux,6.2.0-32-generic,x86_64,-jnone,-smalloc,-sdefault,-hcritbit
Debug: Child (22) Started
Child launched OK
Info: Child (22) said Child starts
```

can you post the exact command you are running and the full output please?

gquintard:
(should have been opened in https://github.com/varnish/docker-varnish/issues, but eh, that works too)

simonvik:
I wonder if this is a seccomp rule blocking close_range, it seems like newer docker versions added seccomp to the list : https://github.com/moby/moby/commit/54eff4354b17a9c460b851300f28aed1408a8615. 

But i also think that `AZ(close_range(` might be a bit unsafe here, it should probably use the old methods if close_range fails for some reason?

jirkace:
```
-> lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster

-> docker run varnish:7.4.1
Unable to find image 'varnish:7.4.1' locally
7.4.1: Pulling from library/varnish
a803e7c4b030: Pull complete 
fbeceee01016: Pull complete 
81e456fb3f0e: Pull complete 
Digest: sha256:95587ca949be166e5800bd26d6a03049b6d24b60bd52637f15b415543f892265
Status: Downloaded newer image for varnish:7.4.1
[...]

PR #3989, opened 2023-09-22
Loop over FDs if close_range fails.

simonvik:
`close_range` is still a somewhat new function and its common that the userspace in a container is newer than the kernel.

This fixes https://github.com/varnishcache/varnish-cache/issues/3988

### DIFF
--- a/lib/libvarnish/vsub.c
+++ b/lib/libvarnish/vsub.c
@@ -38,9 +38,7 @@
 #include <stdlib.h>		// Solaris closefrom(3c)
 #include <string.h>
 #include <unistd.h>		// BSD/Linux close_range(2)
-#ifdef HAVE_WORKING_CLOSE_RANGE
-#elif HAVE_CLOSEFROM
-#else
+#ifndef HAVE_CLOSEFROM
 #  include <dirent.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


lib/libvarnish/vsub.c
```diff
@@ -38,10 +38,8 @@
 #include <stdlib.h>		// Solaris closefrom(3c)
 #include <string.h>
 #include <unistd.h>		// BSD/Linux close_range(2)
-#ifdef HAVE_WORKING_CLOSE_RANGE
-#elif HAVE_CLOSEFROM
-#else
-#  include <dirent.h>
+#ifndef HAVE_CLOSEFROM
+#include <dirent.h>
 #endif
```
Dridi, line 43: Please bring back the `#  include` indentation, it's nested inside an `#ifdef`.

lib/libvarnish/vsub.c
```diff
@@ -67,13 +65,14 @@ VSUB_closefrom(int fd)
...
-	return;
-#elif HAVE_CLOSEFROM
+#ifdef HAVE_CLOSEFROM
 	closefrom(fd);
 	return;
 #else
+#ifdef HAVE_WORKING_CLOSE_RANGE
+	if (close_range(fd, ~0U, 0) == 0)
+		return;
+#endif
```
Dridi, line 75: Aren't you leaving a stray `#endif` below?

Dridi (changes requested):

lib/libvarnish/vsub.c
simonvik: No, its nested ifdef to avoid the include of `<dirent.h>` . Readability isn't nice, it would be more readable to always include dirent and just ifdef the `HAVE_CLOSEFROM`

lib/libvarnish/vsub.c
simonvik: Indented

lib/libvarnish/vsub.c
simonvik: Indented to make it more readable

Dridi (approved):

Dridi:
The failure for the ASAN job is a test case taking more than a minute, so not timing out stuck somewhere in the script.

PR #3990, opened 2023-09-26
build: Simplify python discovery at configure time

Dridi:
If the desired python command is newer than the ones known by aclocal, we can use the PYTHON variable to select it:

    ./configure PYTHON=python3.11

This already worked before this change, but now we stop chasing python releases to get them automatically tested.

Refs #3982

---

Keeping as a draft for the moment, considering to hold it until we know the next major version (7.5.0 or 8.0.0) to make sure we don't forget to update the serial.


catap:
❤️

varnish.m4
```diff
@@ -124,13 +124,7 @@ AC_DEFUN([_VARNISH_CHECK_DEVEL], [
...
 # ---------------------
 AC_DEFUN([_VARNISH_CHECK_PYTHON], [
-	m4_define_default([_AM_PYTHON_INTERPRETER_LIST],
-		[python3.10 python3.9 python3.8 python3.7 python3.6 dnl
-		python3.5 python3.4 python3 python])
-	AM_PATH_PYTHON([3.4], [], [
-		AC_MSG_ERROR([Python >= 3.4 is required.])
-	])
-
+	AM_PATH_PYTHON([3.4], [], [AC_MSG_ERROR([Python >= 3.4 is required.])])
```
catap, line 127:
```suggestion
	AM_PATH_PYTHON([3.4], [], [AC_MSG_ERROR([Python >= 3.4 is required. You may suggest used python via PYTHON=/path/to/python])])
```

catap: Not sure that it worth it, but might help someone

PR #3991, opened 2023-09-27
CLI improvements

walid-git:


lib/libvarnish/vcli_serve.c
```diff
@@ -93,9 +95,14 @@ enum cmd_error_e {
 void v_matchproto_(cli_func_t)
 VCLS_func_close(struct cli *cli, const char *const *av, void *priv)
 {
+	struct VCLS *cs;
 
 	(void)av;
 	(void)priv;
+	cs = cli->cls;
+	CHECK_OBJ_NOTNULL(cs, VCLS_MAGIC);
```
nigoroll, line 103: can we use (something like) `ASSERT_CLI()`?

include/Makefile.am
```diff
@@ -12,7 +12,6 @@ nobase_pkginclude_HEADERS = \
 	tbl/beresp_flags.h \
 	tbl/boc_state.h \
 	tbl/body_status.h \
-	tbl/cli_cmds.h \
```
nigoroll, line 15: would it not make sense to keep the commands available to tools?

nigoroll (approved):

bin/varnishd/mgt/mgt_cli.c
```diff
@@ -141,7 +141,20 @@ mcf_askchild(struct cli *cli, const char * const *av, void *priv)
 		    "Type 'help' for more info.");
 		return;
 	}
+
+	for (i = 0; i < ncmds; i++) {
+		if (strcmp(av[1], cmds[i]->request))
+			continue;
+		if (cmds[i]->flags & CLI_F_INTERNAL) {
```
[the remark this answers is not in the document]
Dridi: I think we shouldn't expose more than necessary. We are changing the CLI commands table in this very patch series, and I would personally prefer to keep the ability to change the table as we see fit without being concerned about API stability for end-users (especially in the context of back-ports to older branches).

For third-party utilities, I would refer to `help -j`, like `varnishadm` does since 38506b339af043d64bef12b7c27fbfc7df62e657.

bin/varnishd/mgt/mgt_cli.c
```diff
@@ -142,15 +158,11 @@ mcf_askchild(struct cli *cli, const char * const *av, void *priv)
...
-			continue;
-		if (cmds[i]->flags & CLI_F_INTERNAL) {
-			VCLI_Out(cli, "Unknown request.\nType 'help' for more info.\n");
-			VCLI_SetResult(cli, CLIS_UNKNOWN);
-			return;
-		}
-		break;
+	cmd = mgt_cmd_lookup(av[1]);
+	if (cmd != NULL && cmd->flags & CLI_F_INTERNAL) {
+		VCLI_Out(cli, "Unknown request.\nType 'help' for more info.\n");
```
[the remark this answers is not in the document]
Dridi: Since we are reaching out to the mgt CLI server, I would again refrain from mentioning implementation details from the cache CLI server.

lib/libvarnish/vcli_serve.c
Dridi: This is `libvarnish`, so not really.

include/Makefile.am
Dridi: See previous comment referring to https://github.com/varnishcache/varnish-cache/commit/38506b339af043d64bef12b7c27fbfc7df62e657.

bin/varnishd/mgt/mgt_cli.c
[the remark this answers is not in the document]
walid-git: What is the point of exposing commands that can/should not be used anyway ? Internal commands are meant to be only issued by the MGT process, so in my opinion, nothing else should be aware of it.

lib/libvarnish/vcli_serve.c
```diff
@@ -182,7 +182,7 @@ VCLS_func_help_json(struct cli *cli, const char * const *av, void *priv)
 	VCLI_JSON_begin(cli, 2, av);
 	VTAILQ_FOREACH(clp, &cs->funcs, list) {
 		sep = "";
-		if (clp->auth > cli->auth)
+		if ((clp->desc->flags & CLI_F_AUTH) && !cli->auth)
```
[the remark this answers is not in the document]
walid-git: Fixed

bin/varnishd/cache/cache_cli.c
```diff
@@ -77,7 +77,9 @@ cli_cb_before(const struct cli *cli, struct cli_proto *clp, const char *cmdn)
 	ASSERT_CLI();
 	(void)cmdn;
 
-	if (clp == NULL || !(clp->desc->flags & CLI_F_SENSITIVE))
+	if (clp == NULL ||
```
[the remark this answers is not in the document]
walid-git: fixed

walid-git:
- Added a callback in `struct cli_cmd_desc` that allows sensitive commands to provide their own logging function. If a sensitive command does not provide an implementation, then only the command name followed by `(hidden)` will be logged (no arguments). The command response is also logged as `(hidden)`. 
- Added coverage for internal and sensitive commands by introducing debug commands with the respective flags.
- Also, fixed a bug introduced by cd2cfbd522f7dd2dd143e6c0a191ddc73da8f8d9 where the last command's response would be logged twice in syslog if  cli file doesn't end with a new line.

bin/varnishd/mgt/mgt_cli.c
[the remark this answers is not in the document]
nigoroll: My point is that we should not make it harder than it already is to hunt down errors.

include/tbl/cli_cmds.h
```diff
@@ -42,13 +42,24 @@
...
 
+#ifdef CLI_LOG_FUNCS
+
+#  ifndef VSB_H_INCLUDED
+#    include "vsb.h"
+#  endif
+
+
+#undef CLI_LOG_FUNCS
+#endif
```
Dridi: This looks really out of place and while I had reserves regarding a logging callback, I can very likely be convinced that this is a good idea.

This `#ifdef` block however convinces me that this is not properly articulated.

While the sensitive flag belongs in `cli_cmd_desc`, the callback most likely belongs in `cli_proto` since there may be more than one implementation of the same command (essentially mcf and ccf).

include/tbl/debug_bits.h
```diff
@@ -52,6 +52,7 @@ DEBUG_BIT(PROCESSORS,		processors,	"Fetch/Deliver processors")
 DEBUG_BIT(PROTOCOL,		protocol,	"Protocol debugging")
 DEBUG_BIT(VCL_KEEP,		vcl_keep,	"Keep VCL C and so files")
 DEBUG_BIT(LCK,			lck,		"Additional lock statistics")
+DEBUG_BIT(CLI_SHOW_SENSIITIVE,	cli_show_sensitive, "Log sensitive commands to syslog and VSL")
```
Dridi, line 55:
```diff
-CLI_SHOW_SENSIITIVE
+CLI_SHOW_SENSITIVE
```

bin/varnishd/cache/cache_cli.c
```diff
@@ -76,7 +76,9 @@ cli_cb_before(const struct cli *cli, struct cli_proto *clp, const char * const *
 	ASSERT_CLI();
 	(void)av;
 
-	if (clp != NULL &&  (clp->desc->flags & CLI_F_SENSITIVE)) {
+	if (clp != NULL &&
+	    !DO_DEBUG(DBG_CLI_SHOW_SENSIITIVE) &&
+	    (clp->desc->flags & CLI_F_SENSITIVE)) {
```
Dridi: We could imagine a utility function in libvarnish:

```c
unsigned
VCLS_IsSensitive(struct cli_proto *clp, unsigned show)
{

        if (clp == NULL || show)
                return (0);
        return (clp->desc->flags & CLI_F_SENSITIVE);
}
```

And then this if statement and 3 others can be revisited:

```c
if (VCLS_IsSensitive(clp, [MGT_]DO_DEBUG(DBG_CLI_SHOW_SENSITIVE)) {
        /* do the logfunc thing */
}
```

lib/libvarnish/vcli_serve.c
```diff
@@ -77,6 +78,7 @@ struct VCLS {
 	cls_cbc_f			*before, *after;
 	volatile unsigned		*limit;
 	struct cli_proto		*wildcard;
+	pthread_t			pthread;
```
Dridi, line 81: We should call this field `thr`.

lib/libvarnish/vcli_serve.c
Dridi: Or I guess we could have something like `CHECK_VCLS(cs)` in this file that performs a magic check followed by a thread check?

include/vcli_serve.h
```diff
@@ -47,6 +47,18 @@ enum cli_flags {
...
+
+#define CMD_AUTH(c) _FLG_IS(c, CLI_F_AUTH)
+#define CMD_INTERNAL(c) _FLG_IS(c, CLI_F_INTERNAL)
+#define CMD_SENSITIVE(c) _FLG_IS(c, CLI_F_SENSITIVE)
+#define CMD_DEBUG(c) _FLG_IS(c, CLI_F_DEBUG)
+
+#define CLP_AUTH(c) CMD_AUTH(c->desc)
+#define CLP_INTERNAL(c) CMD_INTERNAL(c->desc)
+#define CLP_SENSITIVE(c) CMD_SENSITIVE(c->desc)
+#define CLP_DEBUG(c) CMD_DEBUG(c->desc)
```
Dridi, line 60: We´d probably be better off with just two macros:

```c
#define VCLS_CMD_IS(cmd, FLAG) ((cmd)->flags & CLI_F_##FLAG)
#define VCLS_PROTO_IS(clp, FLAG) VCLS_CMD_IS((clp)->desc, FLAG)
```

And then you use it like this:

```c
if (VCLS_PROTO_IS(AUTH) && !cli->auth)
        continue;
```

This way you don't need to add new macros for new flags.

And I made sure to namespace them in the suggestion above.

include/tbl/cli_cmds.h
```diff
@@ -48,6 +48,18 @@
...
+{
+
+	(void)cli;
+
+	AN(av);
+	AN(vsb);
+
+	VSB_printf(vsb, "%s %s XXXXX", av[1]!=NULL ? av[1] : "(null)",
+	    av[2] != NULL ? av[2] : "(null)");
+}
```
Dridi, line 62: See previous comment on the suspicious `#ifdef`. This function does not belong here.

Dridi (changes requested):
I understand that the sensitive callback was agreed upon during a bugwash but I don't like how it was articulated.

include/tbl/cli_cmds.h
walid-git: I initially thought of making the callback part of `cli_proto` as you suggest, but the thing is that MGT does not (and should not) have access to implementations of child cli commands. Thus, if MGT receives a command it does not know how to execute, it will be able to figure out that this command is sensitive by looping through the list of `cli_cmd_desc`s, but it won't know how to log it. Also I think that we should avoid having different implementations of the same command now that we have internal commands, we can have two separate commands and making the child one internal.

lib/libvarnish/vcli_serve.c
walid-git: Added `CHECK_VCLS` macro in the latest revision

walid-git:
Moved the command logging callback to `struct cli_cmd_desc` as suggested by dridi so that it can be overridden per command implementation. If the sensitive command is a cache process command, the custom log implementation cannot be used in mgt and the command will be logged with the default logging of sensitive commands (command name with hidden arguments). 
Also addressed other minor items. Ready for another review

include/tbl/cli_cmds.h
walid-git: fixed

include/vcli_serve.h
walid-git: Fixed in latest revision

include/tbl/debug_bits.h
walid-git: fixed

bin/varnishd/cache/cache_cli.c
walid-git: Added in latest revision

include/tbl/cli_cmds.h
walid-git: see my last comment

bsdphk:
I have looked over this PR in relation to the VIPC work I'm doing to enable a future CERT subprocess.

This PR attacks some of the issues we need to solve, but I'm not sure those are the ones we need to or should fix first, nor that they should be fixed this particular way.

In the future world order, we will have multiple "entities" which offer CLI commands: MGT, WRK, CERT but also VMODs and VEXTs, and the MGT needs to know where (and when!) to send commands.

We will also be passing file-descriptors around in association with CLI commands (the "smuggling" facility).

And finally we will have distinct levels of CLI authority:  Being able to load a new CERT is a different authority from loading a new VCL, from adding a BAN and from mucking with a backend (for unspecified values of mucking.)

I'm as of yet undecided if we will also have CLI multiplexing, for instance being able to do a "backend.list" while a new VCL is being compiled, but I'll probably default to "Since we're ripping all this apart anyway…"

As for the logging, that's an entire different kettle of fish.  It is not obvious to me that CLI logging belongs in the subprocesses after this reconstruction, maybe MGT needs to grow it's own logging facility (or just use syslog?) and be responsible for all logging related to CLI, including thinking about which authority logs where.

So I'm inclined to call this PR a very valuable prototype, and throw it away (as Brooks commands us.)

PR #3992, opened 2023-09-28
hash: Transfer waiting list ownership to the objcore

Dridi:
The advantages of the objhead owning the waiting list are that we need less waiting list heads as they are shared by across multiple objcores, we get more frequent opportunities to rush waiting requests, and we avoid one expensive part of the lookup, which is to find an objhead from a request hash.

The downsides are that the requirement to always perform a new lookup, which increases contention on the objhead lock, and prevents expired but valid objects to be considered fresh. This is another way to describe the cache-control no-cache directive: a response considered fresh at fetch time, but immediately stale afterwards.

Having the waiting list on the objcore means that we reenter the cache lookup with the object candidate from the previous lookup, so we may short-circuit the more expensive objhead lookup when the objcore is compatible with the request rushing off the waiting list. The expensive objhead lookup from the request hash is still avoided, the objcore has a reference to the objhead.

This moves the infamous waiting list serialization phenomenon to the vary header match. Knowing that a rushed object is guaranteed to lead to a cache hit allows the rush policy to be applied wholesale, instead of exponentially. Only requests incompatible with the objcore vary header may reenter the waiting list, a scenario no different from the spurious rush wakeups when operating on the objhead.

The waiting list operations are still guarded by the objhead lock.

This is the first step towards implementing no-cache and private support at the header field granularity.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -359,6 +359,7 @@ struct objcore {
 	VTAILQ_ENTRY(objcore)	lru_list;
 	VTAILQ_ENTRY(objcore)	ban_list;
 	VSTAILQ_ENTRY(objcore)	exp_list;
+	VTAILQ_HEAD(, req)	waitinglist;
 	struct ban		*ban;
 };
 
@@ -490,8 +491,8 @@ struct req {
 
 	struct objcore		*body_oc;
 
-	/* The busy objhead we sleep on */
-	struct objhead		*hash_objhead;
+	/* The busy objcore we sleep on */
+	struct objcore		*hash_oc;
 
 	/* Built Vary string == workspace reservation */
 	uint8_t			*vary_b;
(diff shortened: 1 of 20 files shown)


Dridi:
On Friday I took the original patch from this pull request and pushed it to the last mile. When I pushed the new patch series earlier I realized there was actually one more obvious thing to do (1f3b2178a17a0fa283c7fc6352a15202ff22ea55).

Since this was established as a priority last week during the VDD, and since this is not a trivial change, I hope it can be triaged during bugwash today so the review effort can start sooner than later. I will try to join the bugwash but connectivity is terrible in my train.

Dridi:
Found more polish to apply after staring a little harder.

Dridi:
In the absence of a bugwash yesterday I took the liberty of rearranging the patch series one more time.

Dridi:
Converted to draft, I want to revisit this series.

Bugwash: I can just push da96634e406c526a6760a9d045f2b235e4bc1be8 and acfab776f11b7033ff1b1d0867c5333a28c533f8.

nigoroll:
Note from yesterday's bugwash: Dridi wants to take a slightly different approach and will come back. This is going to take some time.

Dridi:
> Note from yesterday's bugwash: Dridi wants to take a slightly different approach and will come back. This is going to take some time.

I started and completed the new iteration yesterday, but didn't take time to submit #4032. I believe that this time I cut all the waiting list loose ends and managed much better trade offs.

PR #3994, opened 2023-10-06
New vcl_backend_refresh method

walid-git:
This PR introduces the new vcl_backend_refresh method discussed during VDD23Q3.
See: https://github.com/varnishcache/varnish-cache/issues/3102#issuecomment-1737355644 for more context.


Dridi:
You need to update `doc/graphviz/cache_*.dot` and rebuild  `doc/graphviz/cache_*.svg` as part of this change.

Dridi:
This is also missing the `obj_stale` variable in `vcl_backend_refresh`. See how `obj` is handled for `vcl_hit` in `doc/sphinx/reference/vcl_var.rst`.

bin/varnishd/builtin.vcl
```diff
@@ -209,6 +209,10 @@ sub vcl_backend_response {
 	return (deliver);
 }
 
+sub vcl_backend_refresh {
+	return (merge);
+}
```
Dridi, line 214: I'm having second thoughts here, and I'm wondering whether we should generalize this subroutine to whenever there is a stale object involved.

```vcl
sub vcl_backend_refresh {
        if (obj_stale.status == 200 && beresp.status == 304) {
                return (merge);
        }
        if (beresp.status == 304) {
                return (error(503)); # until we refine error statuses
        }
        return (beresp);
}
```

Of course, with a pinch of splitting.

nigoroll: @Dridi can you please motivate this suggestion? What else would you do with the stale object besides delivering it as-is (for grace mode) or use it for a refresh?

nigoroll: bugwash: This makes sense.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -483,19 +482,39 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
...
+					break;
+				case VCL_RET_ERROR:
+					/* FALLTHROUGH */
+				case VCL_RET_ABANDON:
+					/* FALLTHROUGH */
+				case VCL_RET_FAIL:
+					/* FALLTHROUGH */
+				case VCL_RET_RETRY:
+					skip_vbr = 1;
+					break;
```
Dridi, line 507: You can simply do this:

```c
case VCL_RET_ERROR:
case VCL_RET_ABANDON:
case VCL_RET_FAIL:
case VCL_RET_RETRY:
	skip_vbr = 1;
	break;
```

It shouldn't trigger a warning.

walid-git: I remember doing that on another PR and getting CCI failing compilation because of `-Wimplicit-fallthrough`

bin/varnishd/cache/cache_vrt.c
```diff
@@ -257,6 +257,8 @@ VRT_selecthttp(VRT_CTX, enum gethdr_e where)
 	case HDR_RESP:
 		hp = ctx->http_resp;
 		break;
+	case HDR_OBJ_STALE:
+		/* FALLTHROUGH */
 	case HDR_OBJ:
 		hp = NULL;
 		break;
```
Dridi, line 264:
```c
	case HDR_OBJ:
	case HDR_OBJ_STALE:
		hp = NULL;
		break;
```

bin/varnishd/cache/cache_fetch.c
asadsa92: I think that only happens if you got statements in-between without a break.

bin/varnishd/builtin.vcl
```diff
@@ -210,7 +210,14 @@ sub vcl_backend_response {
 }
 
 sub vcl_backend_refresh {
-	return (merge);
+	if (beresp.status == 304) {
+		if (!obj_stale.http.Last-Modified &&
+		    !obj_stale.http.ETag) {
```
Dridi, line 215: You are using `obj_stale` before it exists.

bin/varnishd/cache/cache_fetch.c
walid-git: Oh right, I can test it then

walid-git:
PR updated according to @dridi's [comment](https://github.com/varnishcache/varnish-cache/issues/4026#issuecomment-2213645899) with slightly modified order in the builtin VCL (swapped `vcl_refresh_conditions` and `vcl_refresh_status` to allow using a custom return (merge/beresp/stale) while still going through the checks).


> Anyway, as far as this ticket is concerned, I think we are fine with our current 304 handling and we can close it.

As @dridi concluded that we are correctly handling 304, I have now undrafted this PR making it ready for reviews.

walid-git:
Rebased against current master. Ready again for reviews.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -488,14 +478,49 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
...
 
-	if (http_IsStatus(bo->beresp, 304) && vbf_304_logic(bo) < 0)
-		return (F_STP_ERROR);
+	if (http_IsStatus(bo->beresp, 304)) {
+		if (bo->stale_oc != NULL) {
+			retried_stale = bo->retried_stale;
+			VCL_backend_refresh_method(bo->vcl, wrk, NULL, bo, NULL);
+			bo->was_304 = 1;
+			switch (wrk->vpi->handling) {
+					case VCL_RET_MERGE:
```
dridi, line 487: Align `case` indentation with `switch`.

dridi, line 487: Align `case` indentation with `switch`.

doc/sphinx/reference/vcl_var.rst
```diff
@@ -1054,9 +1054,9 @@ beresp.do_esi
 
 	Type: BOOL
 
-	Readable from: vcl_backend_response, vcl_backend_error
+	Readable from: vcl_backend_response, vcl_backend_error, vcl_backend_refresh
 
-	Writable from: vcl_backend_response, vcl_backend_error
+	Writable from: vcl_backend_response, vcl_backend_error, vcl_backend_refresh
```
nigoroll, line 1059: do we want to spread do_XXX? I would have thought that as `vcl_backend_refresh` is called in addition, we should only use to decide on the resulting beresp and keep the other logic to `vcl_backend_response`.

(comment applies to the following, too)

nigoroll (commented):
thank you for the update, I under-estimated this review and need to look again

doc/sphinx/reference/vcl_var.rst
nigoroll: bugwash could not come up with strong reasons for or against.
My worry is that this might be confusing, but it probably is not an important point.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -488,14 +478,49 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
...
+						vbf_304_logic(bo);
+						break;
+					case VCL_RET_BERESP:
+						break;
+					case VCL_RET_OBJ_STALE:
+						HTTP_Decode(bo->beresp, ObjGetAttr(bo->wrk, bo->stale_oc, OA_HEADERS, NULL));
```
nigoroll, line 493: missing error handling

nigoroll (changes requested):
only one bit.
I really like how `vcl_refresh_*` allow overriding the builtin functionality

bin/varnishd/builtin.vcl
```diff
@@ -209,6 +209,47 @@ sub vcl_backend_response {
 	return (deliver);
 }
 
+sub vcl_backend_refresh {
+        call vcl_builtin_backend_refresh;
```
nigoroll, line 213: these change the indentation from tab to space. I'd suggest you make this consistent by using tabs as before.

lib/libvcc/vcc_action.c
```diff
@@ -263,6 +263,23 @@ vcc_act_return_pass(struct vcc *tl)
...
+static void
+vcc_act_return_retry(struct vcc *tl)
+{
+	SkipToken(tl, '(');
+	if (!vcc_IdIs(tl->t, "fetch")) {
+		VSB_cat(tl->sb, "Expected: return (retry(fetch))\n");
+		vcc_ErrWhere(tl, tl->t);
+		return;
+	}
+	Fb(tl, 1, "VRT_retry_fetch(ctx);\n");
```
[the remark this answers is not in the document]
walid-git: We have `ctx` but `struct busyobj` is opaque in vgc, so we can't just set `ctx->bo->retried_stale = 1`

bin/varnishd/cache/cache_vrt.c
```diff
@@ -114,6 +114,15 @@ VRT_synth(VRT_CTX, VCL_INT code, VCL_STRING reason)
...
 }
 
+VCL_VOID
+VRT_retry_fetch(VRT_CTX)
+{
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	CHECK_OBJ_NOTNULL(ctx->bo, BUSYOBJ_MAGIC);
+
+	ctx->bo->retried_stale = 1;
+}
```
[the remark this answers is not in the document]
walid-git: Changed to `VPI_`

bin/varnishd/cache/cache_fetch.c
walid-git: thanks, fixed

bin/varnishd/cache/cache_fetch.c
walid-git: thanks, added

bin/varnishd/builtin.vcl
walid-git: thanks, fixed

nigoroll (approved):
but please don't forget the squashes :)

dridi (commented):
I like where this is going, please rebase your branch onto master and squash commits where they belong unless something catastrophic happens in CI.

It would be nice to add coverage for a "404 revalidation" since that was one of the use cases brought up, to make sure that this new subroutine allows custom revalidation.

walid-git:
I have rebased and squashed the commits.
However, I think refreshing objects that do not have a 200 status is still not ideal because we are only introducing ways of controlling what happens after we get a 304 response, but we the logic that makes varnish send a conditional request is still hardcoded in the core code, and that part relies on the presence of the `OF_IMSCAND` object flag, which we only set on objects having a 200 status. So currently, the following test case fails for two reasons: The server does not receive the expected 
`if-none-match` header, and we `return (error)` from `vcl_refresh_conditions`.

```shell
varnishtest "Test 404 revalidation"

server s1 {
	rxreq
	txresp -status 404 -hdr "Etag: abcd" -hdr "from-bo: true" -bodylen 10
	rxreq
	expect req.http.if-none-match == "abcd"
	txresp -status 304
} -start

varnish v1 -vcl+backend {

	sub vcl_backend_response {
		set beresp.ttl = 0.01s;
		set beresp.grace = 0s;
		set beresp.keep = 10m;
		set beresp.http.was-304 = beresp.was_304;
	}

	sub vcl_refresh_status {
		if (obj_stale.status != 200 && obj_stale.status != 404) {
			call vcl_refresh_retry;
		}
		return;
	}


} -start

client c1 {
	txreq
	rxresp
	expect resp.status == 404
	expect resp.http.was-304 == false
} -run

delay 0.01

client c2 {
	txreq
	rxresp
	expect resp.status == 404
	expect resp.http.was-304 == true
} -run

```

walid-git:
> Can anyone please help me refresh my memory when this was ever a topic?
> 
> I think "404 revalidation" is just completely wrong.

FTR: the use case was mentioned on the mailing list https://varnish-cache.org/lists/pipermail/varnish-misc/2023-July/027282.html

lib/libvcc/vcc_action.c
```diff
@@ -263,6 +263,23 @@ vcc_act_return_pass(struct vcc *tl)
...
+	SkipToken(tl, '(');
+	if (!vcc_IdIs(tl->t, "fetch")) {
+		VSB_cat(tl->sb, "Expected: return (retry(fetch))\n");
+		vcc_ErrWhere(tl, tl->t);
+		return;
+	}
+	Fb(tl, 1, "VPI_retry_fetch(ctx);\n");
+	vcc_NextToken(tl);
+	SkipToken(tl, ')');
+}
```
dridi, line 281: In the commit introducing `return (retry(fetch))` there should be a documentation update.

doc/graphviz/cache_fetch.dot
```diff
@@ -43,9 +43,27 @@ digraph cache_fetch {
...
 		v_b_hdrs -> v_b_e
+		v_b_hdrs -> v_b_refresh [label="\"stale object\""] [style=bold]
+
+		v_b_refresh [
+			    shape=record
+			    label="{vbf_stp_startfetch:|{vcl_backend_refresh\{\}|{bereq.*|beresp.*}}|{<error>error|fail|{retry|{<max>max?|<retry>ok?}}|{retry(fetch)|{<rf_max>max?|<rf_retry>ok?}}|abandon|<merge>merge|<stale>obj_stale|<beresp>beresp}}"
+		]
+		v_b_refresh:error:s -> v_b_e
+		v_b_refresh:retry -> v_b_r_retry [color=purple]
+		v_b_refresh:rf_retry -> v_b_r_retry_fetch [color=purple]
```
dridi, line 59: Maybe don't share the `vcl_backend_response` retry transition to have less arrows crossing others? (Just like `vcl_backend_fetch` has its dedicated retry transition.)

I'm also wondering, is there a specific reason not to allow `return (retry(fetch))` from `vcl_backend_response`?

dridi (commented):
Almost there.

doc/graphviz/cache_fetch.dot
dridi: I meant to also include `vcl_backend_error` too.

I also think we should be fine with a single "retry" box, with two distinct arrows towards "RETRY" and "RETRY_FETCH". Especially since there are two max/ok sub "boxes".

nigoroll:
Notes on bugwash:

walid dug out a [mailinglist post bringing up the "404 revalidation" idea](https://varnish-cache.org/lists/pipermail/varnish-misc/2023-July/027282.html), and I thank him for doing do, but I was fully unaware. This ticket exists now for 1.5 years and the topic was not brought up before today. Walid added that he also reads "by default" in https://github.com/varnishcache/varnish-cache/issues/4026#issuecomment-2213645899 as "but otherwise we do", but hey, really, this is really a bit of a stretch to expect readers to interpret two words in a _different_ ticket as the scope of this PR.

So: I am 👎🏽 on extending the scope of this ticket for this reason alone (more important reasons below) and I am irritated by today's notes on this ticket.

This PR is, I think, ready. If anyone thinks it it not, I want to learn why. But I do not accept "but we do not handle 404 revalidations" as a valid reason, for one, because it was only brought up today, but also because I think the whole idea is highly problematic.

Quote [RFC9110:](https://datatracker.ietf.org/doc/html/rfc9110#status.304)

> The 304 (Not Modified) status code indicates that a conditional GET or HEAD request has been received and would have resulted in a [200 (OK)](https://datatracker.ietf.org/doc/html/rfc9110#status.200) response if it were not for the fact that the condition evaluated to false.

So even IF we sent a conditional GET for a 404 cache object, a 304 would mean that the responses status had changed to 200, so now we would assume an unchanged body for a changed status code? Or we would assume an unchanged body for an unchanged status code and misinterpret the RFCs?

I am not against doing all kinds of weird stuff, but this is really a stretch.

If anyone wants to add support for misinterpreting a 304 response as "original status code has not changed", feel free to make a suggestion, but this is not this ticket. 

And by the way if anyone wants to do conditionals on 404, I guess they can today by saving the original status to a header, changing the status to 200 and then changing it back on the way to the client.

I thank @walid-git for his great work on this ticket and I appreciate @dridi s very cool idea to use "the builtin.vcl is always appended" in smart ways, so I absolutely do not want to play down your contribution, on the contrary, I do very much value it.

But this topic triggers me in so many ways, and the worst is that it is a clear RFC violation, which we _can_ of course support, but we should not argue that a change was not ready because it does not support an rfc validation.

dridi:
> So: I am 👎🏽 on extending the scope of this ticket for this reason alone (more important reasons below) and I am irritated by today's notes on this ticket.

It started with my earlier review, see https://github.com/varnishcache/varnish-cache/pull/3994#pullrequestreview-2660733342. @walid-git found the discussion I was referring to with the 404 coverage.

I said that because I thought this changed had made this use case possible (the use case being revalidation of non-200 when generating the resource is expensive but generating the 304 is not) and when I realized this I told @walid-git on a different channel that we shouldn't try to extend the scope.

I thought we had met the conditions to enable it, that is all. The moment @walid-git reported to me that it couldn't work I asked him to forget about it. I only asked for test coverage because I thought this change made it possible.

Apologies for the trigger, it wasn't meant to sneak a stretch goal at the last minute, only test coverage.

nigoroll:
Thank you @dridi for the clarifications.

doc/graphviz/cache_fetch.dot
dridi: I guess we can generalize it later if there is an actual case for it.

Consider this pull request approved on my end, but please also consider my last review items (including the cosmetic side of this one).

dridi:
Last week we found a few kinks with @walid-git following this discussion (https://github.com/varnishcache/varnish-cache/pull/3994#discussion_r1981675919) after my last review. What started as a request to simplify the state machine diagram turned out to reveal unresolved matters.

> The way I understand this patch, `return (retry(fetch))` specifically exists to be able to retry the backend request without conditional headers. When I first saw it, I was not entirely convinced what it was needed, after all, it could just be `return (retry);`, but it seemed like a simplification for this specific case. The code allows exactly one of these special retries, and I think for a good reason. I do not think this should become a generic mechanism, unless there is a strong argument.

It turns out the way libvcc handles transitions already enables `retry(fetch)` anywhere a plain `retry` is allowed.

My suggestion to solve this would be to replace it with a simple `fetch` transition. A `return (fetch)` transition from either `vcl_backend_fetch` or `vcl_backend_refresh` would lead to the same step in the state machine.

That also solves something we realized with the current patch series: there is no explicit way to distinguish between `retry` and `retry(fetch)` from the logs. It can of course be inferred, but that's it. A `fetch` transition would be explicit.

An interesting property of `retry(fetch)` is that, as a retry, it creates a new transaction. It may be opportune to spawn a new transaction with a `return(fetch)` from `vcl_backend_refresh` to avoid duplicate timestamps from the second fetch. We'd need a new value for the `SLT_Begin` reason and @walid-git suggested "refetch". As I'm writing this comment I thought "refresh" could work too. This transition isn't strictly required to to bump `bereq.retries` since it can't loop forever.

Something we didn't pay attention to with the current `retry(fetch)` transition is the shape of logs in the new transaction. I think we should add `varnishncsa` coverage to make sure that we get POLA access logs with the default format, and `logexpect` coverage to make sure we have all the essentials (timestamps, initial `bereq` etc).

If there are other important aspects we failed to identify, I guess now would be a good time to ponder them.

nigoroll:
After a bugwash discussion, I now think that from the VCL language perspective, `return(retry(fetch))` is significantly cleaner than `return(fetch)` called from `sub vcl_backend_refresh {}` because we cycle back in the state machine.
If we need additional checks in VCC, we should add these.

Adding `return(refetch)` might be a compromise ("a retry for just the fetch step, skipping vcl_backend_fetch"), but I do not think it was particularly clean.

nigoroll:
bugwash: @walid-git will look into fixing `return(retry(fetch))` and also check with @dridi

bsdphk:
After talking with walid about this, and him pointing out the case where the old object got purged under our feet, I think I finally understand the situation.

As I see it, we have three desirable outcomes:

A) Go back into the state machine, so VCL writers can write "Ok, that didn't work..." logic.

B) Do a "OK, that didn't work: Pretend the 304 didn't happen and hit the backend again ASAP without any If-* headers."

C) Try to see if there is another old object we can use.

I can see the purity of A) but I'll be damned if I can imagine any sane use for it,  and one can always fail the fetch and retry it that way.

With B) I think it is important that the VCL syntax is clearly different from normal fetch/refetch logic, precisely because there is no way for VCL to get involved, and my first thought was `return(iznogood)` - which probably says more about me than anybody wants to know.

C) is interesting for hideously expensive objects. Getting a new candidate object requires a new lookup which fail+restart provides.

So all in all I think I come down on the B) team, but I think "fetch" and "refetch" both confuse more than they explain.

`return(nogood)`, however you spell it, has no explanatory power either.

What's a good word for "Just get it from the backend, OK?" ?

uncondfetch ?

fetch_without_conditionals ?

dridi:
I suggested `return (retry-fetch)` to Walid, since it's a bit of both but not quite either.

As long as we have a return action, that's an anchor to elaborate in the reference manual.

On the "nogood" theme, you could also have `return (unusable)` (calling back to "uncacheable") that doesn't give more hints than "nogood" but at least looks familiar.

dridi:
Bugwash: why not `return (fetch(unconditional))`? It has the same drawback as `return (retry(fetch))` making the argument available anywhere, and here even worse, on the client side. But there isn't a reason against deciding early to skip revalidation, so it can technically work from anywhere.

The new drawback here is that adding `return (fetch(unconditional)` to `vcl_backend_refresh` means that a plain `return (fetch)` is also syntactically correct.

nigoroll:
With all this back end forth I wonder if we really should introduce a new return code at all. After all, the gains are marginal compared with just `return(retry)` and builtin VCL can have the default code to remove the `If-*` headers.

walid-git:
> With all this back end forth I wonder if we really should introduce a new return code at all. After all, the gains are marginal compared with just `return(retry)` and builtin VCL can have the default code to remove the `If-*` headers.

The main point about not using `return(retry)` is to avoid going through `vcl_backend_fetch` a second time and send the exact same backend request without `if-*` headers

dridi:
> * for `return(beresp)`, from `vcl_backend_refresh`, ensure the response has no body (if the 304 wrongly has a body, it gets discarded)
> 
> * `return(beresp)` should not touch the stale obj

I think that we are deviating from the original intent of the change.

Right now we unilaterally merge headers from `obj_stale` and `beresp` in core code. The goal is to give a finer-grained control over this behavior in VCL. This only matters when there is a revalidation targeting the cache, leaving aside private fetches (`return(pass)` from `vcl_recv`) that will go straight to `vcl_backend_response`, even today.

Right now the process is:

- divert 304 handling
  - transition to 200 (and raise `beresp.was_304` flag)
  - merge headers
  - copy body
- call `vcl_backend_response`

With this pull request, the new process for the happy path should become:

- divert 304 handling
  - transition to 200 (and raise `beresp.was_304` flag)
  - call `vcl_backend_refresh`
    - `merge` => merge headers
    - `obj_stale` => collect headers from `stale_oc`
    - `beresp` => keep beresp headers
  - copy body
- call `vcl_backend_response`

If we really want to skip the body copy, we can allow `unset beresp.body;` from `vcl_backend_refresh`.

I'm otherwise fine with lowering the scope with regards to the retry logic, since we failed to establish satisfying semantics. But when it comes to revalidation, we shouldn't lose sight of the core of the change: give control to VCL over the handling of beresp headers during a refresh (with the possibility to add and remove headers in `vcl_backend_refresh`) and the ability to give core code instructions on how to or not to merge.

nigoroll:
during bugwash, @walid-git has pointed out that [my summary written at the VDD was maybe not clear](https://github.com/varnishcache/varnish-cache/pull/3994#issuecomment-2909969081):

> for `return(beresp)`, from `vcl_backend_refresh`, ensure the response has no body (if the 304 wrongly has a body, it gets discarded)

IIRC this was just a clarification wrt the implementation, to consider the error case that the 304 could have a body.

> return(beresp) should not touch the stale obj

I am not exactly sure where this came from, but I read it as "should not remove it or anything", it would still copy it.

Apologies for the imprecise notes (and next time, you are all welcome to also take notes and point out if mine are unclear :) )

nigoroll:
In other words, I think that Dridis summary _still_ is what we all want.

dridi:
> I am not exactly sure where this came from, but I read it as "should not remove it or anything", it would still copy it.

But like any `stale_oc`, it is killed at the end of a successful fetch.

> > for `return(beresp)`, from `vcl_backend_refresh`, ensure the response has no body (if the 304 wrongly has a body, it gets discarded)
> 
> IIRC this was just a clarification wrt the implementation, to consider the error case that the 304 could have a body.

I don't think discarding it is reasonable, it should be a hard error.

What do we even do today?

> In other words, I think that Dridis summary _still_ is what we all want.

I hope turning this into before/after bullet points helped clarify where this change is happening.

bin/varnishd/cache/cache_fetch.c
```diff
@@ -491,14 +478,65 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
...
+		VCL_backend_refresh_method(bo->vcl, wrk, NULL, bo, NULL);
+		bo->was_304 = 1;
+		switch (wrk->vpi->handling) {
+		case VCL_RET_MERGE:
+			vbf_304_logic(bo);
+			break;
+		case VCL_RET_BERESP:
+			http_SetStatus(bo->beresp, 200, NULL);
+			http_ForceHeader(bo->beresp, H_Content_Length,
+			    HTTP_GetHdrPack(wrk, bo->stale_oc,
```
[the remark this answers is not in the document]
nigoroll: Relevant in this context:

https://github.com/varnishcache/varnish-cache/blob/51a117587524cbdd59e43567e6cbd5a76e6a39ff/bin/varnishd/cache/cache_fetch.c#L844-L863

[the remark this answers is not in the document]
nigoroll: Bugwash: move the "wait for unbusy" earlier, leave a note that we want to have a VCL controlled timeout

doc/sphinx/reference/vcl_step.rst
```diff
@@ -331,6 +331,48 @@ hashing the request method and/or request body.
...
+  |
+  | ``merge``
+  |  Merge the headers we got from the backend response with
+  |  the ones we had in the cached object.
+  |
+  | ``obj_stale``
+  |  Return the stale object's headers without performing the RFC merge logic.
+  |
+  | ``beresp``
+  |  Return the beresp headers as they were at the end of this subroutine
```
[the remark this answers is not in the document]
walid-git: https://github.com/varnishcache/varnish-cache/pull/3994/commits/63e07687ffcedda22a7f8b47c9d036d8deef628a

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -530,28 +494,73 @@ VRT_l_beresp_storage_hint(VRT_CTX, const char *str, VCL_STRANDS s)
...
-
-VCL_BOOL
-VRT_r_obj_can_esi(VRT_CTX)
-{
-	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
-	CHECK_OBJ_NOTNULL(ctx->req, REQ_MAGIC);
-	CHECK_OBJ_NOTNULL(ctx->req->objcore, OBJCORE_MAGIC);
-	return (ObjHasAttr(ctx->req->wrk, ctx->req->objcore, OA_ESIDATA));
-}
+VRT_OC_VAR_R(obj, req, REQ_MAGIC, objcore);
```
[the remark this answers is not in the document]
walid-git: https://github.com/varnishcache/varnish-cache/pull/3994/commits/87094395468f7f53f1e604581d0613825680ae40

Issue #3996, opened 2023-10-11
Handling of CVE-2023-44487 / HTTP2 Rapid Reset


lkarsten:
Hi. Brought this up on IRC, but I think there should be some public record of it as well.

**How well does Varnish handle the Rapid Reset attack?**

Consensus seems to be that some new rate limiting should be added to Varnish.

Links:
* https://ubuntu.com/security/CVE-2023-44487
* https://blog.qualys.com/vulnerabilities-threat-research/2023/10/10/cve-2023-44487-http-2-rapid-reset-attack

Related fixes:
* https://github.com/h2o/h2o/commit/28fe15117b909588bf14269a0e1c6ec4548579fe

bsdphk:
Or maybe we should simply violate this part of RFC7540 ?

  SETTINGS_MAX_CONCURRENT_STREAMS (0x3):  Indicates the maximum number
      of concurrent streams that the sender will allow.  This limit is
      directional: it applies to the number of streams that the sender
      permits the receiver to create.  Initially, there is no limit to
      this value.  [...]

I'm pretty sure that a limit of 10 or 20 would cater for all non-hostile uses.

kjetilho:
> I'm pretty sure that a limit of 10 or 20 would cater for all non-hostile uses.

I think this was what [Cloudflare had to back down on](https://blog.cloudflare.com/technical-breakdown-http2-rapid-reset-ddos-attack/#499-errors-and-the-challenges-for-http2-stream-concurrency)?

> While responding to DoS attacks enabled by [CVE-2023-44487](https://www.cve.org/CVERecord?id=CVE-2023-44487), Cloudflare reduced maximum stream concurrency to 64. Before making this change, we were unaware that clients don't wait for SETTINGS and instead assume a concurrency of 100. Some web pages, such as an image gallery, do indeed cause a browser to send 100 requests immediately at the start of a connection. Unfortunately, the 36 streams above our limit all needed to be reset, which triggered our counting mitigations. This meant that we closed connections on legitimate clients, leading to a complete page load failure. As soon as we realized this interoperability issue, we changed the maximum stream concurrency to 100.

jailbird777:
Looks like ATS went for rate limiting also: https://github.com/apache/trafficserver/commit/83161881b172f51f4b2669cb47af24ddaf6c9cdc

daghf:
> I'm pretty sure that a limit of 10 or 20 would cater for all non-hostile uses.

[...]

PR #3997, opened 2023-10-12
h2: Add a rate limit facility for h/2 RST handling ("Rapid reset" mitigation)

daghf:
This adds parameters ``h2_rst_allowance`` and ``h2_rst_allowance_period``, which govern the rate of which we allow clients to reset h/2 streams.

If the limit is exceeded the connection is closed.

The functionality is currently disabled by default (``h2_rst_allowance = 0``). This is the kind of mitigation that will always need refinement, so it might be hard to come up with a one-size-fits-all kind of default here. It needs to be generous enough to not impact regular traffic, while at the same time low enough to meaningfully mitigate malicious clients. We may come up with some sort of generous number here, and give a recommendation for users to tune this down if under attack.

Mitigates: #3996

### DIFF
--- a/bin/varnishd/http2/cache_http2.h
+++ b/bin/varnishd/http2/cache_http2.h
@@ -193,6 +193,8 @@ struct h2_sess {
 	VTAILQ_HEAD(,h2_req)		txqueue;
 
 	h2_error			error;
+	double				rst_budget;
+	vtim_real			last_rst;
 };
 
 #define ASSERT_RXTHR(h2) do {assert(h2->rxthr == pthread_self());} while(0)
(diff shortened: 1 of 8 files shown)


TomasKorbar:
Hi @daghf, i think default 3600 resets per minute is pretty generous while keeping varnish relatively safe. I tested that this limit keeps memory consumption at acceptable values during rapid reset attack. As you said, if the server is under attack, admins can restrict this to a lower value.

Dridi:
@TomasKorbar thank you for the early testing. Could you please test it in conjunction with #3998 ?

TomasKorbar:
@Dridi sure. Give me some time please.

PR #3998, opened 2023-10-13
vcl_vrt: Skip VCL execution if the client is gone

Dridi:
This is the main commit message:

> Once a client is reportedly gone, processing its VCL task(s) is just a
> waste of resources. The execution of client-facing VCL is intercepted
> and an artificial return(fail) is returned in that scenario.
> 
> Thanks to the introduction of the universal return(fail) proper error
> handling and resource tear down is already in place, which makes this
> change safe modulus unknown bugs. This adds a circuit breaker anywhere
> in the client state machine where there is VCL execution.
> 
> A new Reset time stamp is logged to convey when a task does not complete
> because the client is gone. This is a good complement to the walk away
> feature and its original circuit breaker for the waiting list, but this
> has not been integrated yet.
> 
> While the request is technically failed, it won't increase the vcl_fail
> counter, and a new req_reset counter is incremented.
> 
> Refs #3835
> Refs 61a15cbffe1141c13b87e30d48ce1402f84433bf
> Refs e5efc2c8dc0d003e5f0fa1a30b598f5949112897
> Refs ba54dc919076b1ddb85434d886ce82a6553d926b
> Refs 6f50a00f80c7f74b2a8b18bb80593f58b74816fd
> Refs b8816994cfab58261d8000ea8e6941cb5de640fa

This change is limited to h2 reset in a broader sense than just receiving an `RST_STREAM` frame, and it is conceivable to achieve a similar attack for HTTP/1 sessions with a rapid TCP reset, but this would also be less effective. It should also be easier to detect and mitigate with firewall rules.

However implementing client polling for HTTP/1 clients is not as straightforward as h2, so an HTTP/1 client can effectively trigger a workload and not stick around, yet the VCL will be processed until completion. We could probably use the waiter facility, but this needs some research.

### DIFF
--- a/bin/varnishd/http2/cache_http2_session.c
+++ b/bin/varnishd/http2/cache_http2_session.c
@@ -438,6 +438,16 @@ h2_new_session(struct worker *wrk, void *arg)
 	wrk->vsl = NULL;
 }
 
+static int v_matchproto_(vtr_poll_f)
+h2_poll(struct req *req)
+{
+	struct h2_req *r2;
+
+	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	CAST_OBJ_NOTNULL(r2, req->transport_priv, H2_REQ_MAGIC);
+	return (r2->error ? -1 : 1);
+}
+
 struct transport HTTP2_transport = {
 	.name =			"HTTP/2",
 	.magic =		TRANSPORT_MAGIC,
@@ -447,4 +457,5 @@ struct transport HTTP2_transport = {
 	.req_body =		h2_req_body,
 	.req_fail =		h2_req_fail,
 	.sess_panic =		h2_sess_panic,
+	.poll =			h2_poll,
 };
--- a/bin/varnishd/cache/cache_transport.h
+++ b/bin/varnishd/cache/cache_transport.h
@@ -44,6 +44,7 @@ typedef void vtr_sess_panic_f (struct vsb *, const struct sess *);
 typedef void vtr_req_panic_f (struct vsb *, const struct req *);
 typedef void vtr_req_fail_f (struct req *, stream_close_t);
 typedef void vtr_reembark_f (struct worker *, struct req *);
+typedef int vtr_poll_f (struct req *);
 typedef int vtr_minimal_response_f (struct req *, uint16_t status);
 
 struct transport {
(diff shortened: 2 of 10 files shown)


Dridi:
I intend to submit a port to the 6.0 branch after getting this change approved, if approved.

bin/varnishd/http2/cache_http2_session.c
```diff
@@ -438,6 +438,23 @@ h2_new_session(struct worker *wrk, void *arg)
...
+static int v_matchproto_(vtr_poll_f)
+h2_poll(struct req *req)
+{
+	struct h2_req *r2;
+	int r = 1;
+
+	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	CAST_OBJ_NOTNULL(r2, req->transport_priv, H2_REQ_MAGIC);
+
+	Lck_Lock(&r2->h2sess->sess->mtx);
```
nigoroll, line 450: I am not convinced that we should take the lock here. Yes, if we don't, we might return a false positive, but only intermittently and only for bad requests. While the lock overhead hits everyone.

nigoroll (approved):

TomasKorbar:
@Dridi memory situation improved even further after applying this change, CPU load improved significantly. I am trying to backport this to 6.6.2 but unfortunately the 02014 test is failing for me.

***  c3    rx: stream: 0, type: WINDOW_UPDATE (8), flags: 0x00, size: 4
***  c3.0  winup->size: 1048576
***  c3    rx: stream: 1, type: WINDOW_UPDATE (8), flags: 0x00, size: 4
***  c3.1  winup->size: 1048576
---- c3.1  Wrong frame type WINDOW_UPDATE (8) wanted RST_STREAM
**   c3    Waiting for stream 0

Seems the test is expecting RST_STREAM but for some reason is not getting it. @Dridi got any idea what could be wrong?

bin/varnishd/http2/cache_http2_session.c
Dridi: bugwash: +1

Dridi:
> @Dridi memory situation improved even further after applying this change, CPU load improved significantly.

Thank you very much for your testing. Your observations match my expectations, and this change was approved.

> Seems the test is expecting RST_STREAM but for some reason is not getting it. @Dridi got any idea what could be wrong?

I had similar problems while working on this patch series, changing the behavior sometimes breaks test cases without making them irrelevant and `t02014.vtc` took the blunt of it.

I suspect you are witnessing the lack of h2 rxbuf for request bodies (#3661) and I don't think it will be easy to make this test case stable. Just ignore it and you should be fine.

TomasKorbar:
Okay, will do. Thanks a lot for the suggestion.

PR #3999, opened 2023-10-16
Add vmod_h2 to control http rapid reset rate limits per session

nigoroll:
(edit: reduced to just two commits now that #3997 has been merged)

During bugwash on 2023-10-16, I suggested to support overrides of the rate limits from VCL, because nowadays many environments can provide friend/foe metrics like "has a valid session" or "has passed a Turing test".

Using tight rate limits is important to keep adversaries in check, but for (likely) known good clients, one might want to lift limits.

This patch adds such per-session controls in the form of vmod `h2`.

The implementation via a custom vmod was preferred during bugwash over adding session variables like `sess.h2_rapid_reset` because they are transport specific. During the last VDD, we also agreed to move protocols into extensions with additional controls provided by vmods. This patch is a first step in this direction.

Ref #3996


nigoroll:
Agreed on irc

PR #4000, opened 2023-10-17
mgt_main: Preserve VSM after startup panics

Dridi:
When a panic occurs early during the cache process startup, before it starts listening to its CLI, we make sure to bypass MGT exit handlers to preserve the VSM.

This restores our ability to perform post-mortem inspections.


Dridi:
Turning into a draft, there is more to preserve and I have a working PoC, but no time to work on an actual solution.

nigoroll:
👍🏽 on the idea

PR #4004, opened 2023-10-18
h2: Integrate h2_rxbuf_storage (6.0)

Dridi:
This is a port of #3661 preceded by cherry-picks of varnishtest changes it relies on (plus a cherry pick of a VTC stabilization made afterwards).

This is in preparation for a port of #3998 to the 6.0 branch.

Issue #4002, opened 2023-10-17
vtc: We need to split test cases


dridi:
It looks like circleci has lowered the CPU budget we get, and jobs running with ASAN are taking 1.5 the time they used to, and some tests are just plainly timing out on a regular basis:

- [ ] m00043.vtc
- [ ] r00962.vtc
- [ ] u00008.vtc (also it seems to leak to varnishtest's stdout)
- [ ] v00018.vtc
- [ ] v00020.vtc
- [ ] v00051.vtc
- [ ] blob_b00008.vtc
- [ ] r02262.vtc (this one looks different, the nested varnishtest times out)

nigoroll:
SAAS 🤯

dridi:
The problem occurs much less often, this may have been a transient problem on cci's infrastructure. Regarding slow sanitizer jobs, we have #4118 now.

PR #4006, opened 2023-10-18
 vcl_vrt: Skip VCL execution if the client is gone (6.0)

Dridi:
This is a port of #3998 with b4c503026e5d331e75926b32f17c7b2eda96814a folded in for the 6.0 branch.


Dridi:
Upon inspection, most of the failing tests were already registered in #4002 and a couple more very slow tests were added.

One failing test is related to the change set and needs further inspection.

Dridi:
The failing test that got me worried is the result of a missing back-port in #4004 of a patch that came after #3661. So I ended up troubleshooting the same bug twice with the same diagnostic, but I didn't even remember that I had already fixed it...

I pushed the missing back-port here and this should be ready to merge once CI completes.

PR #4007, opened 2023-10-18
varnishd: add backend age & reuses to Open & Close

asadsa92:
This extra information comes handy when trubleshooting idle timeout from the backend.


include/tbl/vsl_tags.h
```diff
@@ -107,27 +107,31 @@ SLTM(SessClose, 0, "Client connection closed",
...
-	"\t|  +--------- Backend display name\n"
-	"\t+------------ Connection file descriptor\n"
+	"\t%d %s %s %.6f %ld [ %s ]\n"
+	"\t|  |  |  |    |     |\n"
+	"\t|  |  |  |    |     +- Optional reason\n"
+	"\t|  |  |  |    +------- Backend reuses\n"
+	"\t|  |  |  +------------ Backend age in seconds\n"
+	"\t|  |  +--------------- \"close\" or \"recycle\"\n"
+	"\t|  +------------------ Backend display name\n"
+	"\t+--------------------- Connection file descriptor\n"
```
Dridi, line 134: This is a breaking change, and since we are adding the same fields to two tags, maybe we should limit this to `BackendOpen`? We can deduce the values for `BackendClose` anyway based on whether we created a brand new connection or reused one from the pool.

include/tbl/vsl_tags.h
```diff
@@ -107,27 +107,31 @@ SLTM(SessClose, 0, "Client connection closed",
...
+	"\t|  |  |  |  |  |  |  |    |\n"
+	"\t|  |  |  |  |  |  |  |    +- Backend reuses\n"
+	"\t|  |  |  |  |  |  |  +------ Backend age in seconds\n"
+	"\t|  |  |  |  |  |  +--------- \"connect\" or \"reuse\"\n"
+	"\t|  |  |  |  |  +------------ Local port\n"
+	"\t|  |  |  |  +--------------- Local address\n"
+	"\t|  |  |  +------------------ Remote port\n"
+	"\t|  |  +--------------------- Remote address\n"
+	"\t|  +------------------------ Backend display name\n"
+	"\t+--------------------------- Connection file descriptor\n"
```
Dridi, line 120: I think I would rather refer to `Connection uses` and `Connection age` (everything is in seconds since 4.0 anyway).

Also, I would probably make these fields optional, they aren't really helpful in the "connect" case.

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -127,6 +130,20 @@ PFD_Fd(struct pfd *p)
 	return (&(p->fd));
 }
 
+vtim_real
+PFD_Age(const struct pfd *p)
+{
+	CHECK_OBJ_NOTNULL(p, PFD_MAGIC);
+	return (VTIM_real() - p->created);
+}
```
Dridi, line 138: We probably have a `now` at the call site we can pass here.

Dridi (changes requested):

asadsa92:
Sorry for the long RTT, I must have forgotten about it.
All of your comments should have been addressed now, if not let me know.

gquintard:
will it conlict in goals, or supersede https://github.com/varnish/varnish-cache-plus/pull/1716 ?

nigoroll (approved):
OK, as I have given up arguing for use of `VTIM_mono`...

include/tbl/vsl_tags.h
```diff
@@ -138,15 +138,17 @@ SLTM(SessClose, 0, "Client connection closed",
...
-	"\t|  |  |  |  |  +---- Local port\n"
-	"\t|  |  |  |  +------- Local address\n"
-	"\t|  |  |  +---------- Remote port\n"
-	"\t|  |  +------------- Remote address\n"
-	"\t|  +---------------- Backend display name\n"
-	"\t+------------------- Connection file descriptor\n"
+	"\t%d %s %s %s %s %s %s [ %.6f %ld ]\n"
+	"\t|  |  |  |  |  |  |    |    |\n"
+	"\t|  |  |  |  |  |  |    |    +- Connection uses\n"
+	"\t|  |  |  |  |  |  |    +------ Connection age\n"
```
dridi, line 144: Apologies for flip-flopping on this one but looking at the reused field it should probably say "Connection reuses" here.

bin/varnishd/cache/cache_backend.c
```diff
@@ -325,9 +325,15 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
...
-	    *fdp, VRT_BACKEND_string(dir), abuf2, pbuf2, abuf1, pbuf1,
-	    PFD_State(pfd) == PFD_STATE_STOLEN ? "reuse" : "connect");
+	if (PFD_State(pfd) != PFD_STATE_STOLEN) {
+		VSLb(bo->vsl, SLT_BackendOpen, "%d %s %s %s %s %s connect",
+		    *fdp, VRT_BACKEND_string(dir), abuf2, pbuf2, abuf1, pbuf1);
+	} else {
+		VSLb(bo->vsl, SLT_BackendOpen,
+		    "%d %s %s %s %s %s reuse %.6f %ju", *fdp,
+		    VRT_BACKEND_string(dir), abuf2, pbuf2, abuf1, pbuf1,
+		    PFD_Age(pfd, VTIM_real()), PFD_Reused(pfd));
```
dridi, line 335: There's no point having a `now` argument in `PFD_Age()` if we end up passing `VTIM_real()`, we just create ceremony. My expectation was that we should have a `ctx->now`, or one `ctx->wrk->lastused` freshly computed from the last timestamp update.

If we are not reusing a real timestamp available in this scope then there is no point having a `now` argument or using real time at all.

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -63,6 +63,9 @@ struct pfd {
 	struct conn_pool	*conn_pool;
 
 	pthread_cond_t		*cond;
+
+	vtim_real		created;
+	uint64_t		reused;
```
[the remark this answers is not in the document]
asadsa92: Sorry @nigoroll , I intended to ack on it but forgot.
I have made the switch over to `vtim_mono`.

include/tbl/vsl_tags.h
asadsa92: Thanks, addressed.

bin/varnishd/cache/cache_backend.c
asadsa92: I have for now dropped the `now` argument. In the future we might add it back as `vtim_mono`, but let's not delve into this right now.

include/tbl/vsl_tags.h
```diff
@@ -138,15 +138,17 @@ SLTM(SessClose, 0, "Client connection closed",
...
-	"\t%d %s %s %s %s %s %s\n"
-	"\t|  |  |  |  |  |  |\n"
-	"\t|  |  |  |  |  |  +- \"connect\" or \"reuse\"\n"
-	"\t|  |  |  |  |  +---- Local port\n"
-	"\t|  |  |  |  +------- Local address\n"
-	"\t|  |  |  +---------- Remote port\n"
-	"\t|  |  +------------- Remote address\n"
-	"\t|  +---------------- Backend display name\n"
-	"\t+------------------- Connection file descriptor\n"
+	"\t%d %s %s %s %s %s %s [ %.6f %ld ]\n"
```
dridi, line 141: I just realized we don't specify the precision in the manual, we simply go for `%f` (probably from a log consumer point of view).

bin/varnishd/cache/cache_conn_pool.h
```diff
@@ -45,6 +45,8 @@ struct pfd;
 
 unsigned PFD_State(const struct pfd *);
 int *PFD_Fd(struct pfd *);
+vtim_real PFD_Age(const struct pfd *);
```
dridi, line 48: vtim_dur

dridi (changes requested):

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -131,6 +134,25 @@ PFD_Fd(struct pfd *p)
...
+vtim_dur
+PFD_Age(const struct pfd *p)
+{
+	vtim_mono now;
+
+	CHECK_OBJ_NOTNULL(p, PFD_MAGIC);
+	now = VTIM_mono();
+	assert(now >= p->created);
+
+	return (now - p->created);
```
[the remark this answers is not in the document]
asadsa92: Taken form `man 2 clock_gettime`:
```
 All CLOCK_MONOTONIC
              variants guarantee that the time returned by  consecutive  calls
              will not go backwards, but successive calls may—depending on the
              architecture—return identical (not-increased) time values.
```

include/tbl/vsl_tags.h
asadsa92: Yeah, good point. Did not realize we have not done this in the past.
I am leaning toward consistency, but having the precision allows you to select the correct data type to store this value.
Are we fine making such promises? or should I simply use `%f` here?

bin/varnishd/cache/cache_conn_pool.c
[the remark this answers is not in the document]
nigoroll: In general, if you switched to vtim_mono now (thank you), I would prefer to not call variables holding it `now`, but rather `t_mono` or somthing.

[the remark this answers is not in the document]
asadsa92: I did actually grep before naming the variable: 
```
$ git grep -rne "now = VTIM_mono" | wc -l
bin/varnishd/mgt/mgt_vcl.c:676:         now = VTIM_mono();
bin/varnishd/mgt/mgt_vcl.c:1089:        now = VTIM_mono();
bin/varnishstat/varnishstat_curses.c:1181:              now = VTIM_mono();
lib/libvarnishapi/vsl_dispatch.c:942:   now = VTIM_mono();
lib/libvarnishapi/vsl_dispatch.c:1453:  now = VTIM_mono();
lib/libvarnishapi/vut.c:401:    now = VTIM_mono();
```

I can rename the one in this PR here, or we can handle it as one commit tree-wide after this PR has gone in.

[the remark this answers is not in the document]
nigoroll: site note: we got no type safety of the three `vim_*` types. Years ago I attempted struct-wrapping them, which achieved the goal, but was messy. So for now, we need to be careful to not confuse them.

include/tbl/vsl_tags.h
dridi: We should put `%f` here to be consistent with the rest of the tags.

PR #4013, opened 2023-10-31
Fix and optimize 304 object copying

nigoroll:
(edit: Added a second commit which depends on the first)

TL;DR: Fix trimstore use in 304 and copy objects efficiently ("defragmented")

### Commit 1: Call trimstore only once when copying the body after a 304

It is my understanding that the objtrimstore stevedore API function is only to be called once when the object is complete, which I believe is also in line with the comment on `ObjExtend()` that _The final flag must be set on the last call_.

If this understanding of the API is correct, we did not adhere to it in the fetch code when we made a copy of an existing stale object after a 304 response: There, we iterated over the stale object and did not set the final flag just once when the object was complete, but rather after each storage segment was copied.

This commit fixes this, adds some pedentry to the simple storage and extends b00062.vtc to test this behavior specifically. On top, g6.vtc also triggered without the fix but the duplicate trim detection in place.

This issue has originally surfaced in the SLASH/fellow storage where the trimstore function implicitly asserted to only be called once.

Ref 115742b07c8bad6d465f1c981ee264f934a4492b
Ref https://gitlab.com/uplex/varnish/slash/-/issues/33

### Commit 2: Optimize 304 template object copying

When copying a stale object after a 304 revalidation, we iterated over it and allocated new storage for each storage segment.
    
So, at best, we kept the fragmentation of the existing object, or made it even worse. This is particularly relevant when the existing object was created from a chunked response, in which case the original segments might have been particularly small and, consequently, many in number.
    
Because we wait for the stale object to complete, we know the total length (`OA_LEN`) upfront, and can ask the storage serving the copy for exactly the right length. This is much more efficient, as fragmentation is lowered and the storage engine might make a much better allocation decision when it knows the full length rather than getting piecemeal requests.
    
We implement the improved allocation scheme through additional state kept for the duration of the template object copy: `struct vbf_objiter_priv` holds the pointer to the newly created busy object, the total yet unallocated length (initialized to `OA_LEN` of the existing object) and pointer/length to the unused portion of the currently open segment, if any.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -760,28 +760,56 @@ vbf_stp_fetchend(struct worker *wrk, struct busyobj *bo)
 /*--------------------------------------------------------------------
  */
 
+struct vbf_objiter_priv {
+	unsigned		magic;
+#define VBF_OBITER_PRIV_MAGIC	0x3c272a17
+	struct busyobj		*bo;
+	// not yet allocated
+	ssize_t		l;
+	// current allocation
+	uint8_t		*p;
+	ssize_t		pl;
+};
+
 static int v_matchproto_(objiterate_f)
 vbf_objiterate(void *priv, unsigned flush, const void *ptr, ssize_t len)
 {
-	struct busyobj *bo;
+	struct vbf_objiter_priv *vop;
 	ssize_t l;
 	const uint8_t *ps = ptr;
-	uint8_t *pd;
 
-	CAST_OBJ_NOTNULL(bo, priv, BUSYOBJ_MAGIC);
+	CAST_OBJ_NOTNULL(vop, priv, VBF_OBITER_PRIV_MAGIC);
+	CHECK_OBJ_NOTNULL(vop->bo, BUSYOBJ_MAGIC);
 
 	flush &= OBJ_ITER_END;
 
 	while (len > 0) {
-		l = len;
-		if (VFP_GetStorage(bo->vfc, &l, &pd) != VFP_OK)
-			return (1);
-		l = vmin(l, len);
-		memcpy(pd, ps, l);
-		VFP_Extend(bo->vfc, l, flush && l == len ? VFP_END : VFP_OK);
+		if (vop->pl == 0) {
+			vop->p = NULL;
+			AN(vop->l);
+			vop->pl = vop->l;
+			if (VFP_GetStorage(vop->bo->vfc, &vop->pl, &vop->p)
+			    != VFP_OK)
+				return (1);
+			if (vop->pl < vop->l)
+				vop->l -= vop->pl;
+			else
+				vop->l = 0;
+		}
+		AN(vop->pl);
+		AN(vop->p);
+
+		l = vmin(vop->pl, len);
+		memcpy(vop->p, ps, l);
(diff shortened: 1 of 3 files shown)


mbgrydeland (commented):
Isn't there a problem with this patch set wrt to revalidation of objects that are still streaming?

That is, due to a TTL that is shorter than the time it takes to fetch the object, a 304 revalidation is started which then needs to stream the data from the still fetching object while creating the revalidated object.

I wonder if this will then be using the wrong length (OA_LEN not yet set, which happens in vbf_stp_fetchend), causing an assert it looks like. It would I think regardless mess up both the allocation sizes and the logic to set VFP_END once and only once.

mbgrydeland:
> Isn't there a problem with this patch set wrt to revalidation of objects that are still streaming?

Hmm, seems I was confused about the behaviour here. We do an explicit wait for BOS_FINISHED before creating the revalidated object, so that should be OK.

mbgrydeland:
There is a problem with the patch when revalidating zero length objects. Varnish won't attempt to do a revalidation for these objects (when the length is zero), but a missbehaving backend answering 304 anyways will cause a panic. See the test case.

```
varnishtest ""

server s1 {
	rxreq
	txresp -nolen -hdr "Content-Length: 0" -hdr {Etag: "foo"}

	rxreq
	txresp -status 304 -nolen -hdr "Content-Length: 0" -hdr {Etag: "foo"}
} -start

varnish v1 -vcl+backend {
	sub vcl_backend_response {
		set beresp.ttl = 1s;
	}
} -start

client c1 {
	txreq
	rxresp

	txreq
	rxresp
} -run

delay 1

client c2 {
	txreq
	rxresp
} -run

delay 0.1

client c3 {
	txreq
	rxresp
} -run

```

nigoroll:
bugwash: waiting for @mbgrydeland's ok

bin/varnishd/cache/cache_fetch.c
```diff
@@ -770,15 +770,16 @@ vbf_objiterate(void *priv, unsigned flush, const void *ptr, ssize_t len)
 	uint8_t *pd;
 
 	CAST_OBJ_NOTNULL(bo, priv, BUSYOBJ_MAGIC);
-	(void)flush;
+
+	flush &= OBJ_ITER_FLUSH;
```
mbgrydeland, line 785: Isn't it `OBJ_ITER_END` that is the flag to test for here? I guess the logic holds for 304 revalidations because there wouldn't be any VDPs in the mix. But if there was e.g. a VDP gunzip in the mix there, the `OBJ_ITER_FLUSH` would be set multiple times during iteration. This looks to me like it opens future gotchas.

mbgrydeland (commented):
Did you end up dropping the panic test case from the PR? I can't see it as part of the commits currently.

nigoroll:
> Did you end up dropping the panic test case from the PR? I can't see it as part of the commits currently.

I added it directly to trunk: https://github.com/varnishcache/varnish-cache/commit/3fc690dfc6c81ac2d9e96768bbe0516e8c67b75f https://github.com/varnishcache/varnish-cache/commit/9a4653bbca9284d91dc96b9f454b78743c2ca72b

@mbgrydeland sorry for accidentally re-requesting your review too early.

mbgrydeland (approved):

PR #4016, opened 2023-11-07
param: Enable absolute bits parameters

Dridi:
With the introduction of `param.set -j` performing the regular `param.set` operation but outputting `param.show -j` we made it possible for "operators" to atomically set a parameter to a specific value, and know how it would be rendered. This enables safe periodic consistency checks to undo potential manual tweaks.

This change however, left a blind spot for bits parameters, since we either set them to their default value via the "none" or "default" special values, and otherwise with relative values in the form of a series of `-xxx` or `+yyy` tokens.

This patch series adds a new "all" value and allows "all" and "none" to appear in the bits list:

```
param.set foo none,+bar,+baz
```

It also changes the relative default values of some bits parameters to make them all absolute.

The "default" special value is marked as deprecated in favor of `param.reset`.

The `param.reset` command grew a `-j` option to behave like `param.set`, for consistency.

Last but not least, it fixes the handling of "none" and "default" that could spuriously leave up to 7 most significant bits raised.


nigoroll (approved):
Makes a lot of sense, I looked through the commits, discussed two details with Dridi, but did not test otherwise. Should be ok (flw).

Dridi:
I reworded the first commit message, hence the force-push.

PR #4021, opened 2023-11-15
vav: fix issues and improve tests

walid-git:
This PR  improves `vav_test` and fixes two issues in `VAV_ParseTxt`:

- `vav_test` was always returning 0, which makes it always pass make check. Failed tests as expected were not printed in the final report. And the test suite was aborting as soon as a single test would fail.
- `VAV_ParseTxt` was not correctly handling `"foo" "bar"` with no flags, and `"foo", "bar"` with the ARGV_COMMA flag.

### DIFF
--- a/lib/libvarnish/vav.c
+++ b/lib/libvarnish/vav.c
@@ -441,9 +441,6 @@ test_run(const struct test_case *tc, int *ret)
 		return (argv);
 	}
 
-	if (tc->argv[0] != NULL)
-		return (argv);
-
 	for (i = 1; i < argc && tc->argv[i] != NULL && argv[i] != NULL; i++) {
 		if (!strcmp(tc->argv[i], argv[i]))
 			continue;
@@ -455,7 +452,7 @@ test_run(const struct test_case *tc, int *ret)
 		return (argv);
 	}
 
-	if (tc->argv[i] != NULL || argv[i] != NULL) {
+	if (tc->argv[0] == NULL && (tc->argv[i] != NULL || argv[i] != NULL)) {
 		act = i < argc ? "less" : "more";
 		printf(
 		    "ERROR: Parsing string <%s> with flags %x, "
--- a/lib/libvarnish/vav.c
+++ b/lib/libvarnish/vav.c
@@ -192,6 +192,12 @@ VAV_ParseTxt(const char *b, const char *e, int *argc, int flag)
 			b++;
 			continue;
 		}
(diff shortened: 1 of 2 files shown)


lib/libvarnish/vav.c
```diff
@@ -482,6 +482,8 @@ main(int argc, char **argv)
 	for (tc = tests; ret == 0 && *tc != NULL; tc++) {
 		argv = test_run(*tc, &ret);
 		VAV_Free(argv);
+		if(ret)
+			return (1);
 	}
 
 	return (0);
```
Dridi:
```diff
-if(ret)
+if (ret)
```

lib/libvarnish/vav.c
```diff
@@ -474,18 +474,17 @@ int
...
 
-	for (tc = tests; ret == 0 && *tc != NULL; tc++) {
-		argv = test_run(*tc, &ret);
+	for (tc = tests; *tc != NULL; res = 0, tc++) {
+		argv = test_run(*tc, &res);
 		VAV_Free(argv);
-		if(ret)
-			return (1);
+		if(res)
+			ret++;
```
Dridi: Leave `ret` as it is to avoid noise, and add a `fail` variable instead. Assign it 1 instead of incrementing it, we may other wise hit "magic" exit statuses from the test toolchain.

lib/libvarnish/vav.c
```diff
@@ -455,7 +452,7 @@ test_run(const struct test_case *tc, int *ret)
 		return (argv);
 	}
 
-	if (tc->argv[i] != NULL || argv[i] != NULL) {
+	if (tc->argv[0] == NULL && (tc->argv[i] != NULL || argv[i] != NULL)) {
```
Dridi, line 463: It would have been nice to see a before/after example in the commit message.

Dridi (changes requested):
Good job!

lib/libvarnish/vav.c
walid-git: fixed

lib/libvarnish/vav.c
walid-git: fixed

lib/libvarnish/vav.c
walid-git: updated the commit log

Dridi (approved):

walid-git:
Bugwash: ok

PR #4029, opened 2023-12-11
Rename (struct req).filter_list to vdp_filter_list

nigoroll:
for consistency with (struct busyobj).vfp_filter_list

I would not normally PR this, but I recently got pushback on a VRT version bump...


bsdphk:
I'm OK with this.

PR #4030, opened 2023-12-14
Backend connection queue

walid-git:
This patch allows a task to be queued when a backend reaches its max_connections. The task will queue on the backend and wait for a connection to become available, rather than immediately failing. This capability is off by default and must be enabled with new parameters.

The following parameters have been added:
`backend_wait_timeout`: the amount of time a task will wait (default 0.0). 
`backend_wait_limit`: the maximum number of tasks that can wait (default 0).

The two parameters can also be overriden for individual backends in vcl:
```
    backend foo {
        .host = "bar.com";
        .wait_timeout = 3s;
        .wait_limit = 10;
    }
```

Authored by: @drodden (with minor rearrangements)

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -444,9 +444,6 @@ struct busyobj {
 	const char		*err_reason;
 
 	const char		*client_identity;
-
-	VTAILQ_ENTRY(busyobj)	cw_list;
-	pthread_cond_t		cw_cond;
 };
 
 #define BUSYOBJ_TMO(bo, pfx, tmo)					\
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -444,6 +444,9 @@ struct busyobj {
 	const char		*err_reason;
 
 	const char		*client_identity;
+
+	VTAILQ_ENTRY(busyobj)	cw_list;
+	pthread_cond_t		cw_cond;
 };
 
 #define BUSYOBJ_TMO(bo, pfx, tmo)					\
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -165,6 +165,15 @@ vbe_connwait_dequeue_locked(struct backend *bp, struct connwait *cw)
 	cw->cw_state = CW_DEQUEUED;
 }
 
+static void
+vbe_connwait_fini(struct connwait *cw)
+{
+	CHECK_OBJ_NOTNULL(cw, CONNWAIT_MAGIC);
+	assert(cw->cw_state != CW_QUEUED);
+	PTOK(pthread_cond_destroy(&cw->cw_cond));
+	FINI_OBJ(cw);
+}
+
 /*--------------------------------------------------------------------
  * Get a connection to the backend
  *
@@ -234,7 +243,7 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 		     "backend %s: busy", VRT_BACKEND_string(dir));
 		bp->vsc->busy++;
 		VSC_C_main->backend_busy++;
-		PTOK(pthread_cond_destroy(&cw->cw_cond));
+		vbe_connwait_fini(cw);
 		return (NULL);
 	}
 
@@ -249,7 +258,7 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 			vbe_connwait_dequeue_locked(bp, cw);
 			Lck_Unlock(bp->director->mtx);
 		}
-		PTOK(pthread_cond_destroy(&cw->cw_cond));
+		vbe_connwait_fini(cw);
 		return (NULL);
 	}
 	bo->htc->doclose = SC_NULL;
@@ -271,7 +280,7 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 			vbe_connwait_dequeue_locked(bp, cw);
 			Lck_Unlock(bp->director->mtx);
 		}
-		PTOK(pthread_cond_destroy(&cw->cw_cond));
+		vbe_connwait_fini(cw);
 		return (NULL);
 	}
 
@@ -310,7 +319,7 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 		bp->vsc->req--;
 		vbe_connwait_signal_locked(bp);
 		Lck_Unlock(bp->director->mtx);
-		PTOK(pthread_cond_destroy(&cw->cw_cond));
+		vbe_connwait_fini(cw);
 		return (NULL);
 	}
 	bo->acct.bereq_hdrbytes += err;
@@ -329,7 +338,7 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 	    bo->htc->first_byte_timeout, bo, bp);
 	FIND_TMO(between_bytes_timeout,
 	    bo->htc->between_bytes_timeout, bo, bp);
-	PTOK(pthread_cond_destroy(&cw->cw_cond));
+	vbe_connwait_fini(cw);
 	return (pfd);
 }
 
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -54,6 +54,12 @@
 
 /*--------------------------------------------------------------------*/
 
+enum connwait_e {
+	CW_DO_CONNECT = 1,
+	CW_QUEUED,
+	CW_BE_BUSY,
+};
+
 static const char * const vbe_proto_ident = "HTTP Backend";
 
 static struct lock backends_mtx;
@@ -105,6 +111,38 @@ VBE_Connect_Error(struct VSC_vbe *vsc, int err)
 			dst = cache_param->tmx;				\
 	} while (0)
 
+#define BE_BUSY(be)	\
+	(be->max_connections > 0 && be->n_conn >= be->max_connections)
+
+/*--------------------------------------------------------------------*/
+
+static void
+vbe_connwait_signal_locked(struct backend *bp)
+{
+	struct busyobj *bo;
+
+	Lck_AssertHeld(bp->director->mtx);
+
+	if (bp->n_conn < bp->max_connections) {
+		bo = VTAILQ_FIRST(&bp->cw_head);
+		if (bo != NULL) {
+			CHECK_OBJ(bo, BUSYOBJ_MAGIC);
+			PTOK(pthread_cond_signal(&bo->cw_cond));
+		}
+	}
+}
+
+static void
+vbe_connwait_dequeue(struct backend *bp, struct busyobj *bo, int lock_it)
+{
+	if (lock_it)
+		Lck_Lock(bp->director->mtx);
+	VTAILQ_REMOVE(&bp->cw_head, bo, cw_list);
+	vbe_connwait_signal_locked(bp);
+	if (lock_it)
+		Lck_Unlock(bp->director->mtx);
+}
+
 /*--------------------------------------------------------------------
  * Get a connection to the backend
  *
@@ -118,9 +156,12 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 	struct busyobj *bo;
 	struct pfd *pfd;
 	int *fdp, err;
+	enum connwait_e cw_state;
 	vtim_dur tmod;
 	char abuf1[VTCP_ADDRBUFSIZE], abuf2[VTCP_ADDRBUFSIZE];
 	char pbuf1[VTCP_PORTBUFSIZE], pbuf2[VTCP_PORTBUFSIZE];
+	vtim_dur wait_tmod;
+	vtim_dur wait_end;
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
 	CHECK_OBJ_NOTNULL(ctx->bo, BUSYOBJ_MAGIC);
@@ -136,7 +177,32 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 		return (NULL);
 	}
 
-	if (bp->max_connections > 0 && bp->n_conn >= bp->max_connections) {
+	/* temporary: need new param */
+	FIND_TMO(connect_timeout, wait_tmod, bo, bp);
+
+	Lck_Lock(bp->director->mtx);
+	cw_state = CW_DO_CONNECT;
+	if (!VTAILQ_EMPTY(&bp->cw_head) || BE_BUSY(bp)) {
+		cw_state = CW_BE_BUSY;
+	}
+	if (cw_state == CW_BE_BUSY && wait_tmod > 0.0) {
+		VTAILQ_INSERT_TAIL(&bp->cw_head, bo, cw_list);
+		bp->cw_count++;
+		cw_state = CW_QUEUED;
+		wait_end = VTIM_real() + wait_tmod;
+		do {
+			err = Lck_CondWaitUntil(&bo->cw_cond, bp->director->mtx,
+			    wait_end);
+		} while (err == EINTR);
+		bp->cw_count--;
+		if (err != 0) {
+			VTAILQ_REMOVE(&bp->cw_head, bo, cw_list);
+			cw_state = CW_BE_BUSY;
+		}
+	}
+	Lck_Unlock(bp->director->mtx);
+
+	if (cw_state == CW_BE_BUSY) {
 		VSLb(bo->vsl, SLT_FetchError,
 		     "backend %s: busy", VRT_BACKEND_string(dir));
 		bp->vsc->busy++;
@@ -149,6 +215,8 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 	if (bo->htc == NULL) {
 		VSLb(bo->vsl, SLT_FetchError, "out of workspace");
 		/* XXX: counter ? */
+		if (cw_state == CW_QUEUED)
+			vbe_connwait_dequeue(bp, bo, 1);
 		return (NULL);
 	}
 	bo->htc->doclose = SC_NULL;
@@ -165,6 +233,8 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 		     VRT_BACKEND_string(dir), err, VAS_errtxt(err));
 		VSC_C_main->backend_fail++;
 		bo->htc = NULL;
+		if (cw_state == CW_QUEUED)
+			vbe_connwait_dequeue(bp, bo, 1);
 		return (NULL);
 	}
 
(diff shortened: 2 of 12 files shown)


walid-git:
With this change, it seems that the default value for `thread_pool_stack` on 32 bit systems is no more sufficient (reason why `e00029.vtc` is failing on ubuntu_bionic).

bsdphk:
As an initial matter I would prefer ``queue_length`` and ``queue_timeout`` which I think are almost self-explanatory.

I'm not sure I think it is a good idea however, but default-off mitigates that.

Dridi:
FWIW this is a partial implementation of what we agreed on for VIP 31.

https://github.com/varnishcache/varnish-cache/wiki/VIP31%3A-Backend-connection-queue

There are two loose points not covered, disembarking and health status. Disembarking fetch tasks is a large project, one we can disregard or move to its own VIP. The saturation of both `.max_connections` and `.wait_limit`/`backend_wait_limit` is probably reasonable to address before closing VIP 31.

bin/varnishd/cache/cache.h
```diff
@@ -443,6 +443,9 @@ struct busyobj {
 	const char		*err_reason;
 
 	const char		*client_identity;
+
+	VTAILQ_ENTRY(busyobj)	cw_list;
+	pthread_cond_t		cw_cond;
```
nigoroll, line 448: A `pthread_cond_t` is pretty heavy, can't we re-use one we already have, like `(struct boc).cond` or `(struct worker).cond`?

FTR, I have not checked the details and in particular I have not looked at the impact on code clarity, but we should we sure the footprint is worth it.

bin/varnishd/cache/cache_backend.c
```diff
@@ -105,6 +111,38 @@ VBE_Connect_Error(struct VSC_vbe *vsc, int err)
...
+			CHECK_OBJ(bo, BUSYOBJ_MAGIC);
+			AZ(pthread_cond_signal(&bo->cw_cond));
+		}
+	}
+}
+
+static void
+vbe_connwait_dequeue(struct backend *bp, struct busyobj *bo, int lock_it)
+{
+	if (lock_it)
```
nigoroll, line 146: um no to `lock_it`.
Can we please have a `_locked` function, or just do the `VTAILQ_REMOVE()` wherever we want the locked variant?

bin/varnishd/cache/cache.h
nigoroll: OR, even better, couldn't the two `cw_` members live on the stack, because the wait only exists within `vbe_dir_getfd()` ?

include/vrt.h
```diff
@@ -58,6 +58,8 @@
  * binary/load-time compatible, increment MAJOR version
  *
  * NEXT (2024-03-15)
+ *	struct vrt_backend.backend_wait_timeout added
```
nigoroll, line 61: The commit does not actually bump the API version, but adds VRT changes.

nigoroll (changes requested):
In general, I am 👍🏽

include/vrt.h
Dridi: Because we normally bump as part of the release process.

nigoroll: Exactly. I was referring to the commit message [Add vrt ABI bump](https://github.com/varnishcache/varnish-cache/pull/4030/commits/88f83c44a4e5af6d790bddf0feabf15efd9ad7ef)

Dridi: Oh, I believe it is because a bump actually happened in the downstream commit that was cherry-picked here.

@walid-git can you please change the commit message to something like this instead?

> vrt: Register VBE connection queue changes

bin/varnishd/cache/cache_backend.c
drodden: Originally the dequeue was performed at each location, doing what was needed at that spot.  But questions were raised about duplicating code, which led to this helper. I'd prefer it just go back inline where needed.

bin/varnishd/cache/cache.h
drodden: The original implementation used, I believe, `(struct boc).cond`.  But there were concerns raised about hijacking that `cond` for something like this. To alleviate those concerns it was given its own `cond`, which also helps code clarity (IMO).

drodden: > couldn't the two `cw_` members be packed into a struct which lives on the stack

I'm not sure I understand.  You want to put a list shared across threads on the stack?

nigoroll: Basically yes.
In terms of [queue(3)](https://man.netbsd.org/queue.3) which we use prefixed with `V`, there is a list head and entries linked with (double) pointers. As long as we make sure that we remove it from the list before the stack frame becomes invalid, we can sure put an entry on the stack.

drodden: I think it is better where it is.

Dridi: If we end up moving it to the stack, I'd prefer seeing it in a followup commit to keep this port aligned with the original change.

No opinion otherwise.

nigoroll: @drodden "I think it is better where it is" is not quite a convincing argument. My arguments are:

* The additional memory requirement for _every_ `struct busyobj` is relevant

```
(gdb) p sizeof(struct busyobj)
$1 = 440
(gdb) p sizeof(pthread_cond_t)
$2 = 48
```

* The cv is only needed in _that_ particular function, which is only called when queuing, so adding overhead to the general case (memory, initialization, destruction) is not justified if there is a sensible alternative option.

Re Dridi: sure, why not

drodden: @nigoroll, no it was not a very good response and I know better than to post anything before fully waking up and having my morning tea.

If you'd like to move it, I'm with Dridi in doing as a follow up.

walid-git: Moved the `cw_*` members to the stack

bin/varnishd/cache/cache_backend.c
walid-git: Removed `vbe_connwait_dequeue` and replaced it with inlined calls

include/vrt.h
walid-git: Done

nigoroll:
Thank you, this looks good to me overall.

I would still have some suggestions, but would also be OK to polish after merge, if you agree:

* `struct backend_cw` should have magic value which we check for every access. While I did suggest to move it to the stack, I am well aware of the risk of smashing foreign stacks, and magic checks can help raise a flag if that happens.
* the magic would be initialized with `INIT_OBJ()`, which would also initialize the list pointers for clarity. The code is still correct, IIUC, even with uninitialized list pointers, but I think it helps debugging a lot if they are zeroed.
* `cw_state` should be moved into `struct backend_cw`. This allows for additional assertions (e.g. in `vbe_connwait_signal_locked`, we can assert that the state is `CW_QUEUED`.
* when we dequeue, we should change the state to a new state, e.g. `CW_DEQUEUED`.
* We should add a `backend_cw_fini` function which asserts that the state is `!= CW_QUEUED` before destroying the condition variable.
* Regarding the inlining of the dequeue, I think I would actually prefer a `vbe_connwait_dequeue_locked` for clarity, because that would now also change the state.

nigoroll (approved):
see top level comment

bin/varnishd/cache/cache_backend.c
nigoroll: see top level comment

bin/varnishd/cache/cache_backend.c
```diff
@@ -143,6 +143,14 @@ vbe_connwait_dequeue_locked(struct backend *bp, struct backend_cw *cw)
 	cw->cw_state = CW_DEQUEUED;
 }
 
+static void
+backend_cw_fini(struct backend_cw *cw)
```
Dridi, line 147: Why not `vbe_conwait_fini()`?

bin/varnishd/cache/cache_backend.c
```diff
@@ -143,6 +143,14 @@ vbe_connwait_dequeue_locked(struct backend *bp, struct backend_cw *cw)
...
+{
+	CHECK_OBJ_NOTNULL(cw, BACKEND_CW_MAGIC);
+	AN(cw->cw_state != CW_QUEUED);
```
Dridi, line 150: Use the `assert()` macro if you pass an actual condition.

bin/varnishd/cache/cache_backend.c
```diff
@@ -143,6 +143,14 @@ vbe_connwait_dequeue_locked(struct backend *bp, struct backend_cw *cw)
...
+	PTOK(pthread_cond_destroy(&cw->cw_cond));
+}
```
Dridi, line 152: Please `ZERO_OBJ()` for good measure.

Dridi (changes requested):

bin/varnishd/cache/cache_backend.c
Dridi: It will work in this function that takes a `struct backend_cw *cw` ;-)

bin/varnishd/cache/cache_backend.h
```diff
@@ -44,34 +44,50 @@ struct vrt_ctx;
...
+	CW_BE_BUSY,
+};
+
+struct backend_cw {
+	unsigned			magic;
+#define BACKEND_CW_MAGIC		0x75c7a52b
+	enum connwait_e			cw_state;
+	VTAILQ_ENTRY(backend_cw)	cw_list;
+	pthread_cond_t			cw_cond;
+};
```
Dridi: I think you should be able to get away with just an opaque `struct connwait;` here and keep the actual type definitions contained in `cache_backend.c`.

Please note how I'm implicitly suggesting to rename the struct :grin:

Dridi (changes requested):

bin/varnishd/cache/cache_backend.h
walid-git: Moved the struct to `cache_backend.c` and renamed as suggested.

bin/varnishd/cache/cache_backend.c
walid-git: Changed to `vbe_connwait_fini`.

bin/varnishd/cache/cache_backend.c
walid-git: Fixed.

walid-git:
Rebased and addressed all review comments. Ready for a (hopefully) last review.

bin/varnishd/cache/cache_backend.c
```diff
@@ -60,6 +60,14 @@ enum connwait_e {
 	CW_BE_BUSY,
 };
 
+struct connwait {
+	unsigned			magic;
+#define CONNWAIT_MAGIC			0x75c7a52b
+	enum connwait_e			cw_state;
+	VTAILQ_ENTRY(connwait)		cw_list;
+	pthread_cond_t			cw_cond;
+};
```
Dridi, line 70: Didn't we conclude that we could use `bo->wrk->cond` for the connection queue?

walid-git: @Dridi that's not what I understood from the discussion [here](https://github.com/varnishcache/varnish-cache/pull/4030#discussion_r1430199392)
See Daryl's comment:

> The original implementation used, I believe, (struct boc).cond. But there were concerns raised about hijacking that cond for something like this. To alleviate those concerns it was given its own cond, which also helps code clarity (IMO).

Dridi: Right, but this was about the boc cond, not the worker cond, right? It looks like we never discussed the latter.

walid-git: It might have been discussed on the original implementation, not sure, maybe @drodden can provide some input ?

drodden: I vaguely remember something about not using the worker cond and found this comment on the original PR:

> [daghf] [Jan 16, 2023] 
> Indeed. I also like the idea of not repurposing wrk->cond, this rhymes better with the stated future task of task disembarking.

bin/varnishd/cache/cache_backend.c
```diff
@@ -150,14 +151,11 @@ vbe_connwait_signal_locked(struct backend *bp)
...
+vbe_connwait_dequeue_locked(struct backend *bp, struct connwait *cw)
 {
-	if (lock_it)
-		Lck_Lock(bp->director->mtx);
 	VTAILQ_REMOVE(&bp->cw_head, cw, cw_list);
 	vbe_connwait_signal_locked(bp);
-	if (lock_it)
-		Lck_Unlock(bp->director->mtx);
+	cw->cw_state = CW_DEQUEUED;
 }
```
Dridi: `Lck_AssertHeld()` ?

Dridi (approved):

nigoroll (changes requested):
I always feel bad when it looks like I was holding things up, so I would like to apologize for not having spotted some issues earlier.

bin/varnishd/cache/cache_backend.c
```diff
@@ -199,6 +311,7 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 		bp->vsc->conn--;
 		bp->vsc->req--;
```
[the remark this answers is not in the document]
walid-git: Good catch, will add it, thanks.

bin/varnishd/cache/cache_backend.c
```diff
@@ -149,6 +246,12 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 	if (bo->htc == NULL) {
 		VSLb(bo->vsl, SLT_FetchError, "out of workspace");
 		/* XXX: counter ? */
+		if (cw->cw_state == CW_QUEUED) {
+			Lck_Lock(bp->director->mtx);
+			vbe_connwait_dequeue_locked(bp, cw);
```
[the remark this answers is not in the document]
walid-git: Why would we allocate workspace before we are sure that we can get a backend connection ?

bin/varnishd/cache/cache_backend.c
```diff
@@ -121,6 +184,10 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 	vtim_dur tmod;
 	char abuf1[VTCP_ADDRBUFSIZE], abuf2[VTCP_ADDRBUFSIZE];
 	char pbuf1[VTCP_PORTBUFSIZE], pbuf2[VTCP_PORTBUFSIZE];
+	unsigned wait_limit;
```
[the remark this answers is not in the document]
walid-git: Fixed

bin/varnishd/cache/cache_backend.c
```diff
@@ -105,6 +120,54 @@ VBE_Connect_Error(struct VSC_vbe *vsc, int err)
...
+			dst = cache_param->tmx;				\
+	} while (0)
+
+#define BE_BUSY(be)	\
+	(be->max_connections > 0 && be->n_conn >= be->max_connections)
+
+/*--------------------------------------------------------------------*/
+
+static void
+vbe_connwait_signal_locked(struct backend *bp)
```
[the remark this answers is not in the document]
walid-git: Fixed

bin/varnishd/cache/cache_backend.c
```diff
@@ -135,12 +202,42 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
...
+		wait_end = VTIM_real() + wait_tmod;
+		do {
+			err = Lck_CondWaitUntil(&cw->cw_cond, bp->director->mtx,
+			    wait_end);
+		} while (err == EINTR);
+		bp->cw_count--;
+		if (err != 0) {
+			VTAILQ_REMOVE(&bp->cw_head, cw, cw_list);
+			VSC_C_main->backend_wait_fail++;
+			cw->cw_state = CW_BE_BUSY;
```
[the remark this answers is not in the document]
walid-git: I have pushed a new commit where I removed the EINTR check and added a check for `BE_BUSY(bp)` when we wake up after a timeout. Let me know if this works for you.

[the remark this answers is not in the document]
Dridi: Bugwash discussion: status quo on spurious EINTR vs cond wait.

Dridi (approved):

bin/varnishd/cache/cache.h
```diff
@@ -444,6 +444,7 @@ struct busyobj {
 	const char		*err_reason;
 
 	const char		*client_identity;
+
 };
```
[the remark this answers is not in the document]
walid-git: Fixed

bin/varnishd/cache/cache_busyobj.c
```diff
@@ -145,6 +145,7 @@ VBO_GetBusyObj(const struct worker *wrk, const struct req *req)
 
 	memcpy(bo->digest, req->digest, sizeof bo->digest);
 
+
 	return (bo);
 }
```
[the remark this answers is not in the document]
walid-git: Fixed

bin/varnishd/cache/cache_backend.c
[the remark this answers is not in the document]
walid-git: I have added an XXX comment as suggested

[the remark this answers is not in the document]
nigoroll: > On the other hand, if we are effectively not connecting and workspace would have been too tight, then we fail for the wrong reason, and we don't visit vcl_backend_error at all.

I disagree on "fail for the wrong reason". This code only runs because we do want to connect and having enough workspace is a precondition for the connect to succeed. Potentially running into the connection or wait limit does not make it a "wrong reason" to fail for insufficient workspace.

nigoroll:
bugwash:

 proposed (re)name:

global parameters:
* `backend_queue_limit`
* `backend_queue_timeout`

VCL:

```vcl
backend foo {
  .queue_limit = 42;
  .queue_timeout 1m;
}
sub vcl_backend_fetch {
  set bereq.queue_limit = 42;
  set bereq.queue_timeout = 1m;
}
```

gquintard:
hi all! I'd really like this to get into the next release, and from what I'm reading, it's only a naming exercise from now on.

As a refresher, the current PR offers `backend_wait_timeout`/`backend_wait_limit` and the bugwash proposed `backend_queue_timeout`/`backend_queue_limit`. I feel like this isn't a big enough contention point to block a merge.

Any chance to get the original names in? I hate to bring it up and I'll probably get slapped for it but: we have customers using the feature and consistency is pretty important, I don't want to have to translate parameter names depending on the platform people are running.

nigoroll:
bugwash approved.
please implement connection purge on health change to sick

dridi:
Bugwash: when a probe transitions effectively to sick, the queue should be cleared to let tasks fail immediately.

Effective transitions to sick:

- probe goes sick, not overridden by CLI (`backend.set_health healthy`)
- `backend.set_health sick` (where previously healthy from probe or CLI)
- `backend.set_health probe` (with sick probe, where previously healthy from CLI)

walid-git:
PR updated:
- [e8f35fb](https://github.com/varnishcache/varnish-cache/pull/4030/commits/e8f35fbf2af0c7bbf78c76e661136b0f75e9734c) changes wait_timeout to match with new timeout defaults.
- [871f784](https://github.com/varnishcache/varnish-cache/pull/4030/commits/871f784f174f93f6777b5e7ba66d48be239f61e4) Notify client requests in the wait queue when the backend goes sick (as agreed during bugwash)

nigoroll:
bugwash: merge but the last commit, open new PR for the flush

walid-git:
As per bugwash: squashed [e8f35fb](https://github.com/varnishcache/varnish-cache/pull/4030/commits/e8f35fbf2af0c7bbf78c76e661136b0f75e9734c) and removed [871f784](https://github.com/varnishcache/varnish-cache/pull/4030/commits/871f784f174f93f6777b5e7ba66d48be239f61e4) from this PR (to be submitted separately).
Will merge once CCI is done.

PR #4034, opened 2023-12-27
Generalize the VDP API (WIP, early comments welcome)

nigoroll:
This commit is to prepare for use of the VDP API also for the backend side to filter `bereq.body` through `bereq.filters`:

The `req` member of `struct vdp_ctx` is removed, because it should only be used for initialization and, consequently, is already nulled in `VDP_DeliverObj()`.

`vdp_init_f()` and its caller `VDP_Push()` gain arguments:

- `req` is only present if the VDP is used on the client side.

- `hd` is a pointer to the outgoing headers corresponding to the body being worked on: `(struct req).resp` on the client side and `(struct busyobj).bereq` on the backend side (tbd).

- `cl` is a pointer to the content-length. -1 indicates 'unknown'.

VDPs should aim for compatibility with both the client and backend use case by using only `hd` and `cl` instead of `req`. If they depend on `req`, they should return an error if `req` is `NULL`.

Note that none of the new arguments would be necessary, they could all (including the existing `oc` argument) be derived from the `VRT_CTX`. The reasons for having these arguments are code clarity and avoidance of code duplication, e.g. for repeatedly figuring out the correct header and length pointer on the client vs. backend side.

### DIFF
--- a/bin/varnishd/cache/cache_deliver_proc.c
+++ b/bin/varnishd/cache/cache_deliver_proc.c
@@ -62,19 +62,16 @@ VDP_Panic(struct vsb *vsb, const struct vdp_ctx *vdc)
 
 
 void
-VDP_Init(struct vdp_ctx *vdc, struct worker *wrk, struct vsl_log *vsl,
-    struct req *req)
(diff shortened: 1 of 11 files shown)


nigoroll:
Continued in #4035

PR #4041, opened 2024-01-18
expire: Force a cache MISS when req.ttl = 0s

AlveElde:
This commit changes the undefined behavior of setting req.ttl = 0s. Setting req.ttl = 0s would previously cause req.ttl to have no effect, but this breaks with setting req.grace = 0s, which does have an effect.

The new behavior of setting req.ttl = 0s is similar to setting req.hash_always_miss = true, the main difference is that req.ttl allows for request coalescing. Useful for short-lived non-private objects.

I would like to define this behavior in the documentation somewhere, but didn't find a good place to put it. 

Both req.ttl and req.grace are initialized to -1, which is why this change does not break everything. But setting req.ttl/grace to -1s in VCL ends up setting them to 0s, as these two req attributes have extra logic in their respective definitions:
https://github.com/varnishcache/varnish-cache/blob/5f67d3aee3a756beb46bae610e1a75548e03c07f/bin/varnishd/cache/cache_vrt_var.c#L566-L569

Since this commit effectively prevents VCL writers from "clearing" req.ttl by setting it to 0s, maybe we should consider allowing users to clear req.ttl/grace by setting them to -1s?


Dridi:
> Since this commit effectively prevents VCL writers from "clearing" req.ttl by setting it to 0s, maybe we should consider allowing users to clear req.ttl/grace by setting them to -1s?

Maybe this is for the best? What about this instead?

```vcl
unset req.ttl;
unset req.grace;
```

nigoroll:
My take: Integrate Dridis suggestion and update the docs, then it's a good improvement!

Dridi:
A thought occurred to me, shouldn't we use `NAN` when `req.{ttl,grace}` are unset?

Dridi:
I think it would be more accurate to say "consider all objects stale when req.ttl = 0s".

Aren't we creating a perpetual waiter if the client disregards both TTL and grace? That is, when there is a busy object, the outcome will always be to disembark the request. So until the request wins the miss race, it will compete with other "req.ttl = req.grace = 0s" requests.

This is waiting list serialization once again, except that we swapped beresp with req to trigger it. I can try combining this change with #4032 to check whether it would eventually be mitigated.

Dridi:
> Aren't we creating a perpetual waiter if the client disregards both TTL and grace?

We are not since we treat zero as the lack of TTL limit for the client. Let's revisit that specific point once `unset` is implemented and we treat `NAN` as the lack of limit instead.

hermunn:
> This is waiting list serialization once again

Yeah, you are right, my comment does not add much, and I think we should simply go with the PR as it is now, with all the timestamp confusion which exist.

hermunn:
Sorry, strike that, I still think we need a test case demonstrating how you can get a hit, even when the request set both `req.ttl` and `req.grace` to `0s` because of the waiting list and timestamps.

hermunn (approved):
I have looked at the test case again, and I think it does what it should do. I think we can merge this.

AlveElde:
If you, like me, were wondering why the test case provided does not return a grace hit, this is the reason: https://github.com/varnishcache/varnish-cache/pull/2422. The object in cache is still fresh, and since we ignore req.ttl, we get no candidates for a grace hit. Whether this is the correct behavior is a subject for future discussion.

Dridi:
I think the trade off from #2422 can be let go in favor of the trade off from #4032. You should end up with cache hits again, for a different reason.

bin/varnishd/cache/cache_vrt_var.c
```diff
@@ -561,12 +561,25 @@ VRT_r_req_##nm(VRT_CTX)							\
...
+{									\
+	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);				\
+	CHECK_OBJ_NOTNULL(ctx->req, REQ_MAGIC);				\
+	ctx->req->elem = val;						\
+	ctx->req->elem = val;						\
+}
+
 REQ_VAR_R(backend_hint, director_hint, VCL_BACKEND)
 
 REQ_VAR_L(ttl, d_ttl, VCL_DURATION, if (!(arg>0.0)) arg = 0;)
```
nigoroll, line 577: do we still want to set negative arguments to zero? Could it be a better option to fail?

nigoroll (changes requested):
Overall, I am happy, but the one question I have might be relevant.

bin/varnishd/cache/cache_vrt_var.c
AlveElde: IMO it's reasonable to expect that `req.ttl` can be set to a negative duration, we even do so in a test case:
https://github.com/varnishcache/varnish-cache/blob/e0b164f9f134c0e594df8cb543036a441836da25/bin/varnishtest/tests/v00020.vtc#L150

Failing on a negative duration has a good chance of breaking existing VCL.

nigoroll: That is a coverage test.
What are the semantics of a negative duration for `req.ttl` and `req.grace`?

AlveElde: The same as `beresp.ttl` and `beresp.grace`:

https://github.com/varnishcache/varnish-cache/blob/e0b164f9f134c0e594df8cb543036a441836da25/bin/varnishd/cache/cache_vrt_var.c#L743-L744

nigoroll: I understand that this "has always been like that", but also please remember that for quite some time, varnish had no way of properly failing at runtime (`VRT_fail()`).

The particular lines you are referring to are [from 10 years ago](https://github.com/varnishcache/varnish-cache/commit/71a357ab29bb4ec4f8e671665395e101ad885d2e) and I am still wondering what a negative ttl/grace is supposed to mean. ttl: already expired before it even went into the cache? If those were the semantics, then we would not be allowed to take the respective objects as "just refreshed". grace: I would think a negative value should move an object into grace mode _earlier_ than ttl.

So, as we come across these questions, we should answer them.

AlveElde: I totally agree that "has always been like that" is a weak argument, but I personally prefer to limit breakage if I can. It's not obvious to me what a negative TTL/grace should mean, and I think that question is bigger than the scope of this PR.

nigoroll: Let's try to cover this quickly during the next bugwash. I am also fine with continuing to ignore negative values, but I'd like to at least ask for opinions.

nigoroll:
> Unless we want to convert NAN to some number when reading `req.ttl` and `req.grace`, I suggest we stick with -1 as the default value.

I think we should print NAN

AlveElde:
Sorry for the noise, ready for review

bin/varnishd/cache/cache_vrt_var.c
dridi: I think the point here is to cap the TTL, and once we run out of TTL the object goes in grace mode if applicable, so once we run out of TTL `req.grace` applies.

This is why in my opinion `req.ttl` should only limit the ability to look up a fresh object, and that negative values that could be the result of some arithmetic expression should be bumped to zero.

doc/sphinx/reference/vcl_var.rst
```diff
@@ -246,11 +246,13 @@ req.grace
...
 	Writable from: client
 
+	Unsetable from: client
 
 	Upper limit on the object grace.
 
 	During lookup the minimum of req.grace and the object's stored
-	grace value will be used as the object's grace.
+	grace value will be used as the object's grace. Setting req.grace
+	to 0s prevents hits on stale objects.
```
dridi, line 255: Maybe now would be a good time to document that negative values are swallowed.

doc/sphinx/reference/vcl_var.rst
```diff
@@ -510,8 +512,10 @@ req.ttl
 
 	Writable from: client
 
+	Unsetable from: client
 
 	Upper limit on the object age for cache lookups to return hit.
+	Lookups always miss when req.ttl is set to 0s.
```
dridi, line 518: Maybe now would be a good time to document that negative values are swallowed.

bin/varnishd/cache/cache_expire.c
```diff
@@ -73,7 +73,7 @@ EXP_Ttl(const struct req *req, const struct objcore *oc)
 	CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
 
 	r = oc->ttl;
-	if (req != NULL && req->d_ttl >= 0. && req->d_ttl < r)
+	if (req != NULL && !isnan(req->d_ttl) && req->d_ttl < r)
 		r = req->d_ttl;
```
dridi, line 77: `assert(req->d_ttl >= 0.);` ?

bin/varnishd/cache/cache_expire.c
```diff
@@ -91,7 +91,7 @@ EXP_Ttl_grace(const struct req *req, const struct objcore *oc)
 	CHECK_OBJ_NOTNULL(oc, OBJCORE_MAGIC);
 
 	g = oc->grace;
-	if (req != NULL && req->d_grace >= 0. && req->d_grace < g)
+	if (req != NULL && !isnan(req->d_grace) && req->d_grace < g)
 		g = req->d_grace;
```
dridi, line 95: `assert(req->d_grace >= 0.);` ?

dridi (approved):
LGTM, see small suggestions too.

bin/varnishd/cache/cache_vrt.c
```diff
@@ -770,6 +770,8 @@ VRT_REAL_string(VRT_CTX, VCL_REAL num)
 {
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
+	if (isnan(num))
+		return (WS_Printf(ctx->ws, "NAN"));
```
dridi, line 774: Why allocate a static string from the workspace instead of returning a static string?

nigoroll: how could my efficiency fetish miss this ;)

bin/varnishd/cache/cache_vrt_var.c
nigoroll: @dridi maybe I do not understand your answer, but why should we not fail immediately if an attempt is made to set any of the timers to a negative value?

doc/sphinx/reference/vcl_var.rst
nigoroll: If. Or document that they are invalid and will trigger a failure.

doc/sphinx/reference/vcl_var.rst
```diff
@@ -510,8 +512,10 @@ req.ttl
 
 	Writable from: client
 
+	Unsetable from: client
```
gquintard, line 515:
```suggestion
	Unsettable from: client
```

bin/varnishd/cache/cache_vrt_var.c
dridi: If you use an expression instead of a constant to compute the acceptable TTL limit and the result is negative, I think it would be too harsh to fail the transaction.

If your "clever" `req.ttl` formula can yield something negative, then negative should be considered as allowing no TTL, just like zero.

dridi (approved):
LGTM!

I think you could squash the 4th commit in the 2nd and move the result last to reduce diff churn.

dridi:
@nigoroll the sanitizer job is failing on a legit leak in the VDP stack unrelated to this change.

doc/sphinx/reference/vcl_var.rst
```diff
@@ -536,8 +539,11 @@ req.ttl
 
 	Writable from: client
 
+	Unsettable from: client
 
 	Upper limit on the object age for cache lookups to return hit.
+	Lookups always miss when req.ttl is set to 0s. Negative values
```
hermunn, line 545: Maybe it is worth noting that setting `req.ttl` to `1ms` will keep request coalescing intact, making this a better way of making sure that a fresh object is in the cache, without the risk of inserting many. `set req.ttl = 1ms;` is also better than setting `hash_always_miss` in many (all?) cases. Of course, such notes can be made elsewhere, in a different PR. Still, are we sure that _n_ requests coming in at the same time (before the first one finishes `sub vcl_backend_response` with the subsequent _unbusy_), all getting a `set req.ttl = 0s`, will produce _n_ backend requests, potentially inserting _n_ objects?

nigoroll:
@AlveElde @dridi pointed out that I pushed a change which created a merge conflict with this PR, and I am sorry for having done that unintentionally. I see no semantic conflict, though, whatever the value of "unset" is, we can have it with or without this PR.
That said, I take review of this ticket as homework for next week.

PR #4047, opened 2024-01-29
For BackendClose, also log the short tag as with SessClose

nigoroll:
this is for consistency, to simplify parsing and to reduce the amount of data logged with VSL.

Filed as a PR because it is a breaking change to the VSL format.

Motivated by #4042


Dridi (approved):

nigoroll:
ok'ed by extra bugwash

PR #4050, opened 2024-02-01
v1p: Fix invalid poll() usage with a new VTIM_poll()

Dridi:
The expectations on FreeBSD are stricter than on Linux.

### DIFF
--- a/bin/varnishd/http1/cache_http1_pipe.c
+++ b/bin/varnishd/http1/cache_http1_pipe.c
@@ -149,13 +149,12 @@ V1P_Process(const struct req *req, int fd, struct v1p_acct *v1a,
 		fds[1].revents = 0;
 		tmo = cache_param->pipe_timeout;
 		if (tmo == 0.)
-			tmo = -1.;
+			tmo = NAN;
 		if (deadline > 0.) {
 			tmo_task = deadline - VTIM_real();
-			tmo = (tmo > 0.) ? vmin(tmo, tmo_task) : tmo_task;
-			tmo = vmax(tmo, 0.);
+			tmo = vmin(tmo, tmo_task);
 		}
-		i = poll(fds, 2, (int)(tmo * 1e3));
+		i = poll(fds, 2, VTIM_poll_tmo(tmo));
 		if (i == 0)
 			sc = SC_RX_TIMEOUT;
 		if (i < 1)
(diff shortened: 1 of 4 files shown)


asadsa92:
Could we maybe put it in `vtcp.c` instead? As we already got `VTCP_read()` there.
Also note that we got a poll() inside of `VTCP_read()`.

But, an argument can be made for neither to be in the "correct" place.

Dridi:
> But, an argument can be made for neither to be in the "correct" place.

I totally agree, I wasn't sure where it belonged, but VTCP could be a better home for this one.

Dridi:
> Also note that we got a poll() inside of `VTCP_read()`.

I was planning to review out poll usage once that is merged.

bsdphk:
Tested, works on FreeBSD, but:

It feels somewhat unsanitary to put a wrapper for poll(2) in VTIM.

A ``int VTIM_poll_timeout(vtim_dur)`` would be cleaner.

Dridi:
I actually moved it to `VTCP_poll()` and forgot to push the change to my branch.

bsdphk:
IMO ``VTCP_poll()`` has nothing to do with TCP, apart from working on TCP socket fds...

I still think we should start with a ``int VTIM_poll_timeout(vtim_dur)`` and unless it has some TCP specific handling, I dont see a need for ``VTCP_poll()`` at all ?

Dridi:
Ack, I will amend the branch and merge it once ready.

FWIW I find it a bad fit for all the places I considered: VTIM, VTCP, VUS, VFIL, VEV and Waiter.

asadsa92:
I agree with not going VTCP, but why not just go for `VTIM_poll(vtim_dur);` ?
The `timeout` is implicit, both the name `poll` and the duration parameter.

Dridi:
> I agree with not going VTCP, but why not just go for `VTIM_poll(vtim_dur);` ?

Obviously I'm biased, this is what I picked initially... But I guess we could also have a `VTIM_poll_deadline(vtim_real)`. Going right away with `VTIM_poll_timeout()` would make both explicit I suppose.

nigoroll:
bugwash: make it a vtim_dur-to-poll-argument conversion function and use sth like `poll(fd, n, VTIM_poll_tmo(duration)`. Can merge without further review.

Issue #4053, opened 2024-02-09
Parent ESI fails even when VCL sets resp.status = 200 for the included ESI


slimhazard:
### Expected Behavior

[What's New](https://varnish-cache.org/docs/trunk/whats-new/changes-7.3.html#esi-processing-changes) for Varnish 7.3 says of ESI includes:

> ... only objects with 200 and 204 status will be included and any other status code will fail the parent ESI request.
> 
> If objects with other status should be delivered, they should have their status changed to 200 in VCL, for instance in sub vcl_backend_error{}, vcl_synth{} or vcl_deliver{}.

So we expected that this snippet in `vcl_deliver{}` would prevent failure of the including parent in all cases:

```
sub vcl_deliver {
    if (req.esi_level > 0 && resp.status != 200 && resp.status != 204) {
        set resp.status = 200;
    }
} 
```

### Current Behavior

Even with the VCL snippet shown above, the parent ESI can evidently fail under certain circumstances, in fact apparently in default configuration. I'll add a VTC test case, based on Varnish's current `e00035.vtc`, to illustrate that.

The VTC suggests that `beresp.do_stream` may be related to the problem. If `beresp.do_stream` is set to false and the VCL snippet is included in `vcl_deliver`, then as expected, the parent response does not fail, and the client response body is not truncated. If the `beresp.do_stream == true`, which is the default, then the parent response fails and the body is truncated, even with the `vcl_deliver` snippet.

That would at least make sense to me, since it's familiar that when streaming is on, a client response status may be seen as 200, even when the fetch actually failed (for example when there were too few bytes for the value in `Content-Length`), and would have been set to 503 if streaming had been off. On that logic, `vcl_deliver` doesn't "see" `resp.status != 200` when that happens.

But the VTC shows a bit of strangeness, because it also seems to depend on whether or not `beresp.do_esi` is conditionally or unconditionally set to `true` in VCL. The problem happens if `beresp.do_esi` is set conditionally, as in `e00035.vtc`. If VCL just says `set beresp.do_esi = true`, then the problem doesn't happen in the VTC.

Obviously that takes some explanation. I'll go through the expectations when I add the VTC test case.

### Possible Solution

In `vcl_backend_response`, set `beresp.do_esi = true` unconditionally. It may also be necessary to set `beresp.do_stream = false` (but in that case we lose the advantages of streaming).

### Steps to Reproduce (for bugs)

I'll add a VTC test case to the issue, and describe what we expected to see.

[...]

PR #4054, opened 2024-02-12
esi: Add VCL override for esi_include_onerror

Dridi:
Inspired by the discussion from #4053.


Dridi:
Discussion from #4053:

> One thing I'm also wondering is whether to always initialize this from the parameter or do that only for req_top (technically, resp_top except that it does not exist) and then inherit the default value from req_top in sub-requests.

nigoroll:
extra bugwash: after release, add vcl control over all esi feature flags

PR #4060, opened 2024-02-22
[doc] return(retry|return) must follow a std.rollback() call

gquintard:
just a docfix to avoid user hitting vcl-triggered panics, like here: https://github.com/varnish/varnish-modules/issues/222


walid-git:
Should we also restrict `std.rollback` to subroutines where `return (re{start,try})` are permitted ?

Also, note that CI is failing due to the space before the closing `**`

gquintard:
thanks @walid-git , fixed

vmod/vmod_std.vcc
```diff
@@ -582,6 +582,10 @@ $Function VOID rollback(HTTP h)
 
 Restores the *h* HTTP headers to their original state.
 
+.. note: after calling this function, you must immediately either `return(restart)` or `return(retry)`.
```
walid-git, line 585: I think this is what you are looking for:
```suggestion
.. note:: after calling this function, you must immediately either `return(restart)` or `return(retry)`.
```

The current version is treated as a comment and is not being rendered in the generated man page:
```
man -l man/vmod_std.3 | grep -B 1 -A 6 Restores
   VOID rollback(HTTP h)
       Restores the h HTTP headers to their original state.

       Failure in doing so may result in Varnish panicking.

       Example:

          std.rollback(bereq);
```
Also, please fix the indentation of next line so that it is correctly aligned with the NOTE bloc.

nigoroll:
I agree we should _recommend_ a restart/retry shortly after the rollback, but not require it and I would consider this issue an actual bug in Varnish-Cache.
What happens here is that VCL assigns `VRT_priv_{task,top}()` to pointers in the preamble of each VGC function for a sub, so when rollback happens within the sub, these pointers become stale:

vcl:

```vcl
sub t {
	var.set("foo", "bar");
	std.rollback(req);
	var.set("foo", "bar");
}
```

vgc:

```c
VGC_function_t(VRT_CTX, enum vcl_func_call_e call,
    enum vcl_func_fail_e *failp)
{
  assert(call == VSUB_STATIC);
  assert(failp == NULL);
  assert(ctx->method & (VCL_MET_DELIVER|VCL_MET_HASH|VCL_MET_HIT|VCL_MET_MISS|VCL_MET_PASS|VCL_MET_PURGE|VCL_MET_RECV|VCL_MET_SYNTH));
  struct vmod_priv *ARG_priv_task_var = VRT_priv_task(ctx, &VGC_vmod_var);
  if (ARG_priv_task_var == NULL) {
    VRT_fail(ctx, "failed to get task priv for vmod var");
    return;
  }

  /* ... from ('/tmp/t.vcl' Line 5 Pos 7) */
  {
    {
      VPI_count(ctx, VGC_NREFS, 1);
      if (UNLIKELY(ctx->vpi->trace)) VPI_trace(ctx, 1);
      END_;
      Vmod_vmod_var_Func.f_set(ctx,
        ARG_priv_task_var,
        "foo",
        "bar"
      );
      END_;
      Vmod_vmod_std_Func.f_rollback(ctx,
        VRT_r_req(ctx)
      );
      END_;
      Vmod_vmod_var_Func.f_set(ctx,
        ARG_priv_task_var,
        "foo",
        "bar"
      );
      END_;
    }
  }
}
```

For the first `Vmod_vmod_var_Func.f_set()` call, `ARG_priv_task_var` is valid, but for the second, it is not, because after the call to `Vmod_vmod_std_Func.f_rollback()`, the workspace on which the priv_task state resides, is reset.

Note that this is only an issue for VMODs using the `PRIV_TASK` and `PRIV_TOP` arguments in the vcc file. VMODs using `VRT_priv_{task,top}()` explicitly get a fresh value for each call, and thus do not have this issue.

I see the following options to avoid this:

* move the `VRT_priv_{task,top}()` call from the preamble to before each function call
* make the preamble an explicit data structure (probably a bound array), which gets passed to `std.rollback` to invalidate

Dridi:
> * make the preamble an explicit data structure (probably a bound array), which gets passed to `std.rollback` to invalidate

Probably the other way around. A data structure that `std.rollback()` would find through the VPI.

nigoroll:
addition: we could also retire the `PRIV_TASK` and `PRIV_TOP` vmod function/method arguments and simply require vmods to call the respective VRT functions.

Re @Dridi 

> Probably the other way around. A data structure that std.rollback() would find through the VPI.

Yes, and I would like to add that this data structure does already exist and that `std.rollback()` already finds it, so this new data structure would basically be a "per sub" view on it. Not sure if it's worth it...

vmod/vmod_std.vcc
```diff
@@ -576,12 +576,16 @@ Example::
 		std.set_ip_tos(0);
 	}
 
-$Restrict client
+$Restrict vcl_deliver vcl_hit vcl_miss vcl_pass vcl_pass vcl_purge vcl_recv vcl_synth vcl_backend_error vcl_backend_response
```
Dridi, line 579: Should we support a syntax similar to bits parameters?

```
$Restrict client backend -vcl_recv -vcl_backend_fetch
```

walid-git: I think I like the idea of not having to write all the vcl subroutines but one, but it can also be confusing, what if we have: `$Restrict -vcl_recv`, does it mean we allow everything but `vcl_recv` ? Or would it be necessary to use one of `client/backend/housekeeping` with the `-` syntax ?

vmod/vmod_std.vcc
```diff
@@ -576,7 +576,7 @@ Example::
 		std.set_ip_tos(0);
 	}
 
-$Restrict client
```
walid-git, line 579: This is `$Restrict` for `set_ip_tos`, `$Restrict` always apply to the [**preceding**](https://varnish-cache.org/docs/trunk/reference/vmod.html#functions-and-methods-scope-restriction) function/method

walid-git (changes requested):

vmod/vmod_std.vcc
gquintard: Ah shoot, I'll fix, thanks

vmod/vmod_std.vcc
Dridi: No, I think we should just treat a minus prefix as clearing the relevant bit(s) and adding a "negative" note to the documentation. I don't think vmodtool cares about what we tells `$Restrict` today and will happily compile a VMOD restricted to `vcl_non_existent`.

gquintard:
maybe, just maybe, it's ok now

nigoroll:
We have merged #4066, so the doc change is no longer adequate.
Also, shouldn't the `$Restrict`ion be dropped now?

PR #4062, opened 2024-02-23
SAN fixes on Ubuntu 22

stevendore:
Fix up some issues when configuring varnish with ASAN, UBSAN and workspace emulator on Ubuntu22. I tested locally, with all tests passing, on Bookworm as well since that is what CCI uses for the SAN builds. 

1. 109543e87fbfbda0f9dc1fb5397327f6000de277- Initialize the VDP CTX in `Req_new()`. I think just using INIT_OBJ is okay here and aligns with the rest of the structs initialized in this section.
2. 6ea6c934ae1c69ef2b5bf867e4709b66bacd540f- Increase `thread_pool_stack` in  e29.vtc. Without the increase the test would segfault on Ubuntu 22.
3. 569c68a8630f38431b6394be1e738ba4391bea1b - Free value created from `create_bogo_n_arg()`. Let me know if there is a more desired way to structure the temporary value to free. 

Ubsan error from `Req_new()`:
```
In function ‘memset’,
    inlined from ‘Req_New’ at cache/cache_req.c:176:2:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:59:10: error: ‘__builtin_memset’ offset [0, 63] is out of the bounds [0, 0] [-Werror=array-bounds]
   59 |   return __builtin___memset_chk (__dest, __ch, __len,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   60 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Ubsan error from exec_file:
```
vtc.c: In function ‘exec_file’:
vtc.c:602:9: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
  602 |         macro_def(vltop, NULL, "tmpdir", "%s", tmpdir);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[4]: *** [Makefile:851: varnishtest-vtc.o] Error 1
```

### DIFF
--- a/bin/varnishd/cache/cache_req.c
+++ b/bin/varnishd/cache/cache_req.c
@@ -173,7 +173,7 @@ Req_New(struct sess *sp)
 	p = (void*)PRNDUP(p + sizeof(*req->vfc));
 
 	req->vdc = (void*)p;
-	memset(req->vdc, 0, sizeof *req->vdc);
+	ZERO_OBJ(req->vdc, sizeof *req->vdc);
 	p = (void*)PRNDUP(p + sizeof(*req->vdc));
 
 	req->htc = (void*)p;
--- a/bin/varnishtest/tests/e00029.vtc
+++ b/bin/varnishtest/tests/e00029.vtc
@@ -1,5 +1,7 @@
 varnishtest "ESI max_esi_depth"
 
+feature !sanitizer
+
 # test that the default stack size is sufficient for hitting
 # max_esi_depth
 
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -567,7 +567,7 @@ mgt_process_f_arg(struct cli *cli, unsigned C_flag, void **fap)
 	return (retval);
 }
 
-static const char *
+static char *
 create_bogo_n_arg(void)
 {
 	struct vsb *vsb;
@@ -842,8 +842,11 @@ main(int argc, char * const *argv)
 
 	assert(d_flag == 0 || F_flag == 0);
 
-	if (C_flag && n_arg == NULL)
-		n_arg = create_bogo_n_arg();
+	p = NULL;
+	if (C_flag && n_arg == NULL) {
+		p = create_bogo_n_arg();
+		n_arg = p;
+	}
 
 	if (S_arg != NULL && !strcmp(S_arg, "none")) {
 		fprintf(stderr,
@@ -860,6 +863,8 @@ main(int argc, char * const *argv)
 
 	workdir = VIN_n_Arg(n_arg);
 	AN(workdir);
+	if (p != NULL)
+		free(p);
 
 	if (i_arg == NULL || *i_arg == '\0')
 		i_arg = mgt_HostName();
--- a/bin/varnishtest/vtc.c
+++ b/bin/varnishtest/vtc.c
@@ -562,6 +562,8 @@ exec_file(const char *fn, const char *script, const char *tmpdir,
 	struct vsb *vsb;
 	const char *p;
 
+	AN(tmpdir);
+
 	(void)signal(SIGPIPE, SIG_IGN);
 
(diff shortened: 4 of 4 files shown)


bin/varnishtest/tests/e00029.vtc
```diff
@@ -16,7 +16,7 @@ server s1 {
 	chunkedlen 0
 } -start
 
-varnish v1 -cliok "param.set thread_pool_stack 80k"
+varnish v1 -cliok "param.set thread_pool_stack 180k"
```
nigoroll, line 19: this change defeats the purpose of the test.
Should we skip it for ASAN runs?

bin/varnishd/cache/cache_req.c
```diff
@@ -173,7 +173,7 @@ Req_New(struct sess *sp)
 	p = (void*)PRNDUP(p + sizeof(*req->vfc));
 
 	req->vdc = (void*)p;
-	memset(req->vdc, 0, sizeof *req->vdc);
+	INIT_OBJ(req->vdc, VDP_CTX_MAGIC);
```
nigoroll, line 176: we only add a magic value in the respective struct's initializer, in this case `VDP_Init()` or such initializer does not exist as a function and initialization is basically _just_ the `INIT_OBJ()`. This is the case for the other structs initialized here.

Can we find a better way to make sanitizers happy?

nigoroll (changes requested):

bin/varnishd/cache/cache_req.c
Dridi: > Can we find a better way to make sanitizers happy?

I understand that mempools already back our main workspaces, but my suggestion would be to make the mempools have their own "mpl" workspace. This way we could allocate struct req, struct busyobj, and other things as workspace allocations as we do today, and allocate the task workspace from whatever remains from the mempool workspace.

This way, when working with sanitizers, the workspace emulator would turn those pointer-arithmetic pseudo-allocations we have today into individual mallocs and we'd finally be able to trigger an overflow in structs such as req and busyobj.

bin/varnishtest/tests/e00029.vtc
stevendore: I'm not sure if it is possible or overly specific but would it make sense to only skip the test if there is gcc and a sanitizer since it does pass fine on clang?

bin/varnishd/cache/cache_req.c
nigoroll: helpful clarification from bugwash: What Dridi means is `req = WS_Alloc(sizeof *req)`

stevendore: Sorry I haven't had much time to get back to this. I'm not sure initially where we'd get the workspace from. The session? Since this function is the one that initializes the workspace itself.

Noted in here: https://github.com/varnishcache/varnish-cache/pull/4062#issuecomment-1971312126 I realized the issue is more from using (my version of) GCC with ubsan than my distro itself. So using `ZERO_OBJ` instead of an explicit memset passes compilation fine. Though this feels more like tricking the compiler but the compiler was not fully right in the first place.

bin/varnishtest/tests/e00029.vtc
stevendore: Getting back to this, it looks like using GCC (11.4.0) with both asan and ubsan causes the failure. When using either asan or ubsan alone the test passes. This test seems like a much more nice to have with sanitizers on. Any thoughts? Just deal with the failure by saying it should be running with clang when using sanitizers?

stevendore:
Force pushed to use `ZERO_OBJ` instead of `INIT_OBJ` to clear `req->vdc` and skip e29.vtc when using sanitizers.

PR #4051, opened 2024-02-06
DONOTMERGE: Investigating archlinux failure on circlecli

Dridi:
Not expecting much from the initial debugging attempt.

PR #4063, opened 2024-02-24
[cci] add ubuntu:noble, fix arch

gquintard:
looks like the remote docker image we were using was ancient, that seems to have broken more recent builds, like `noble` and `arch`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,7 +209,7 @@ jobs:
         description: the Linux distribution (debian|ubuntu)
         type: string
       release:
-        description: the release name (buster|bullseye|bionic|focal|jammy)
+        description: the release name (buster|bullseye|bookworm|focal|jammy|noble)
         type: string
       make_target:
         description: the make target to execute during the build
@@ -419,11 +419,6 @@ workflows:
           name: build_debian_bookworm
           dist: debian
           release: bookworm
-      # latest ubuntu uses sanitizers
-      - build:
-          name: build_ubuntu_bionic
-          dist: ubuntu
-          release: bionic
       - build:
           name: build_ubuntu_focal
           dist: ubuntu
@@ -459,7 +454,6 @@ workflows:
           matrix:
             parameters:
               platform:
-                - ubuntu:bionic
                 - ubuntu:focal
                 - ubuntu:jammy
                 - ubuntu:noble
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -223,8 +223,7 @@ jobs:
       - image: centos:7
     working_directory: /workspace
     steps:
-      - setup_remote_docker:
-          version: 20.10.11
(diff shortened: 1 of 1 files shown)


nigoroll:
With the release on the horizon, this should still go in, right?

.circleci/config.yml
```diff
@@ -434,6 +434,10 @@ workflows:
           name: build_ubuntu_jammy
           dist: ubuntu
           release: jammy
+      - build:
+          name: build_ubuntu_noble
+          dist: ubuntu
+          release: noble
```
Dridi, line 440: Should we also remove bionic and make the focal job 32bit?

Dridi (approved):

nigoroll (approved):
I am happy with whatever dridi decides.

.circleci/config.yml
gquintard: no opinion, `bionic` is still supported for a few more months, it depends on how proactive we want to be.
Just say the word and I'll update the PR. Or we can do it post-merge

Dridi: the word

Dridi: I tried to reproduce the focal 32bit CCI failure on my laptop, but for some reason I can't even get a hold of such an image:

```bash
$ podman manifest inspect ubuntu:bionic | jq .manifests[].platform.architecture | grep 386
"386"
$ podman manifest inspect ubuntu:focal | jq .manifests[].platform.architecture | grep 386
$ podman manifest inspect debian:latest | jq .manifests[].platform.architecture | grep 386
"386"
```

So maybe we should have the latest ubuntu run sanitizers and the oldest debian run 32bit?

the word

gquintard: `buster` on 32bit stack overflows the same way the old `ubuntu` did
meanwhile `alpine` is racy on the `distclean-local` job

Dridi: > `buster` on 32bit stack overflows the same way the old `ubuntu` did

Yes but unlike the old one, I can look at this one locally.

> meanwhile `alpine` is racy on the `distclean-local` job

Same as arch it would seem (#4051).

gquintard: a bit different actually, it's not a permission thing:

```
find: ./libtool: No such file or directory
```

Dridi: As predicted, I was able to observe stack overflows on my workstation. Could you please rebase your branch against current master to see how cci fares with 55fecc94d3ca59ab156f1e76746bbeefeaec07d8?

Dridi: > a bit different actually, it's not a permission thing:

I think it's the same race, but crashing in a different spot.

Dridi: Since we reached the deadline for the docker image I rebased and pushed directly to your branch to get cci running before merging.

PR #4064, opened 2024-02-26
varnishd: always shutdown both ends to a backend

asadsa92:


Dridi (approved):

nigoroll (commented):
Looks like a good catch to me, but should we not handle the `pfd` as in `vcp_handle()` here?
* call the `close` method, because otherwise we leak file descriptors
* free the `pfd`, because otherwise we leak memory

Dridi:
> Looks like a good catch to me, but should we not handle the `pfd` as in `vcp_handle()` here?

The loop below is waiting for waiters to notice the shutdown and call `vcp_handle()`.

Dridi:
Actually on this topic, this sleeping loop is something we should move as a CLI hook, waiting asynchronously for connection pools to be freed before deleting them for good.

nigoroll:
> The loop below is waiting for waiters to notice the shutdown and call `vcp_handle()`.

Ah, this makes sense, thank you.

FWIW, @asadsa92 please add such explanations to the issue description or commit message. Leaving it to reviewers to figure out for themselves is no good use of their time.

> this sleeping loop is something we should move as a CLI hook, waiting asynchronously for connection pools to be freed

yes.

nigoroll (approved):

Dridi:
bugwash: merge

PR #4065, opened 2024-02-27
esi: Restore the original onerror behavior

### LINKED ISSUE #4053: Parent ESI fails even when VCL sets resp.status = 200 for the included ESI

Dridi:
This is a partial revert of 582ded6a2d6ae1a4467b1eb500f2725b42888016 to restore the assumed onerror=continue behavior for ESI includes, unless the feature flag esi_include_onerror is raised.

The part of the change that considers all status codes besides 200 and 204 to be errors for ESI includes remains. A test case covers VCL's ability to "bless" error responses by overriding resp.status, allowing ESI delivery to continue on this criterion.

Fixes #4053

### DIFF
--- a/bin/varnishtest/tests/r03241.vtc
+++ b/bin/varnishtest/tests/r03241.vtc
@@ -1,24 +1,16 @@
 varnishtest "ESI include out of workspace"
 
-
 server s1 {
 	rxreq
 	expect req.http.esi0 == "foo"
-	txresp -body {
-		<html>
-		Before include
-		<esi:include src="/body" sr="foo"/>
-		After include
-		</html>
-	}
+	txresp -body {before <esi:include src="/body" sr="foo"/> after}
 	rxreq
 	expect req.url == "/body1"
 	expect req.http.esi0 != "foo"
-	txresp -body {
-		Included file
-	}
+	txresp -body "include"
 } -start
 
+varnish v1 -cliok "param.set feature +esi_disable_xml_check"
 varnish v1 -vcl+backend {
 	import vtc;
 
@@ -48,9 +40,8 @@ logexpect l1 -v v1 -g raw {
 client c1 {
 	txreq -hdr "Host: foo"
 	rxresp
-	# XXX this is actually wrong (missed include)
-	expect resp.bodylen == 57
 	expect resp.status == 200
+	expect resp.body == "before  after"
 }  -run
 
 logexpect l1 -wait
--- a/bin/varnishtest/tests/r03241.vtc
+++ b/bin/varnishtest/tests/r03241.vtc
@@ -4,8 +4,12 @@ varnishtest "ESI include out of workspace"
 server s1 {
 	rxreq
 	expect req.http.esi0 == "foo"
-	txresp -body {<html>Before include<esi:include
-		src="/body" sr="foo"/>After include</html>
+	txresp -body {
+		<html>
+		Before include
+		<esi:include src="/body" sr="foo"/>
+		After include
+		</html>
 	}
 	rxreq
 	expect req.url == "/body1"
@@ -43,11 +47,10 @@ logexpect l1 -v v1 -g raw {
 
 client c1 {
 	txreq -hdr "Host: foo"
-	rxresphdrs
+	rxresp
+	# XXX this is actually wrong (missed include)
+	expect resp.bodylen == 57
 	expect resp.status == 200
-	rxchunk
-	expect_close
-	expect resp.body == {<html>Before include}
 }  -run
 
 logexpect l1 -wait
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -881,8 +881,9 @@ ved_deliver(struct req *req, struct boc *boc, int wantbody)
 
 	status = req->resp->status % 1000;
 
-	if (ecx->abrt && status != 200 && status != 204) {
-		ved_close(req, boc, 1);
+	if (FEATURE(FEATURE_ESI_INCLUDE_ONERROR) &&
+	    status != 200 && status != 204) {
+		ved_close(req, boc, ecx->abrt);
 		return;
 	}
 
--- a/bin/varnishd/cache/cache_esi_deliver.c
+++ b/bin/varnishd/cache/cache_esi_deliver.c
@@ -65,7 +65,7 @@ struct ecx {
 	ssize_t		l;
 	int		isgzip;
 	int		woken;
-	int		incl_cont;
+	int		abrt;
 
 	struct req	*preq;
 	struct ecx	*pecx;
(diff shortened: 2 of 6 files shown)


nigoroll:
Note: if we merge this, we should also have #4054

nigoroll:
> The part of the change that considers all status codes besides 200 and 204 to be errors for ESI includes remains.

But also only if `esi_include_onerror` is enabled, right?

Dridi:
> > The part of the change that considers all status codes besides 200 and 204 to be errors for ESI includes remains.
> 
> But also only if `esi_include_onerror` is enabled, right?

Correct, it adds (or leaves alone) one more error condition eligible to abort. But the feature flag must be raised to stop defaulting to "continue", just like when said feature flag was introduced.

Dridi:
> I think the decision to be made here is if we want more explicit control as suggested in [#4053 (comment)](https://github.com/varnishcache/varnish-cache/issues/4053#issuecomment-1941693695) (tri-state `esi_include_onerror = attribute | abort | continue`) or if we are happy with just "taming" the feature as suggested here.
> 
> My personal opinion is that this PR _together with #4054_ is the minimum we should do, but Varnish-Cache has its strength in the flexibility of its configuration and I think that it might very well be possible to get to a point were we want VCL to override to `abort`.

My opinion is that this PR is the minimum we should do, that is, restoring the previous behavior while retaining the "new" status code effect.

If we want more than a flag, then we should consider promoting it to its own parameter. Regarding other ESI feature flags, we could also consider a new `feature_esi` parameter, similar to what we did with `feature_vcc` (except that in this case we couldn't leave compatibility aliases behind).

nigoroll:
Interesting additional complication uncovered by @xcir [here](https://gitlab.com/uplex/varnish/libvdp-pesi/-/issues/16#note_1795415262): What happens when an included object is neither a 200 nor a 204, but has an empty body?

slimhazard:
@nigoroll isn't that entirely covered by the existing requirements?

I'd say: if the `esi_include_onerror` feature is off, then the top-level response is aborted. If it's on, and the esi tag has `onerror="continue"` or `resp.status` is set to 200 or 204, then include the empty response body. Since it's empty, it looks like nothing in the top-level rendered response body, which is presumably the same thing that happens for an included response with status 204.

All of which is to say "do what Varnish does now" (or is intended to do, assuming a fix for #4053). How is it a complication?

Dridi:
> Interesting additional complication uncovered by @xcir [here](https://gitlab.com/uplex/varnish/libvdp-pesi/-/issues/16#note_1795415262): What is supposed to happen when an included object is neither a 200 nor a 204, but has an empty body?

For the sake of the argument, I'll say the same behavior since Varnish 3.0, which is my oldest reference.

The default behavior is an implicit onerror="continue" attribute, so the erroneous status code of an ESI fragment does not influence the req_top delivery, whether the ESI fragment has a body or not.

nigoroll:
> if the `esi_include_onerror` feature is off, then the top-level response is aborted.

I tend to agree, but this is not [what we currently do](https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_esi_deliver.c#L871-L872).

slimhazard:
@nigoroll hmm ... I've worked with ESIs that return status 204 and an empty body, since Varnish 3.0.x (same site as the one I've been talking about). Those have always rendered 0 bytes into the including response. An empty included response does something different now, if status != 20x? That would be very strange (and again I'm puzzled as to why).

nigoroll:
extra-bugwash: merge

bin/varnishtest/tests/r03241.vtc
```diff
@@ -43,11 +47,10 @@ logexpect l1 -v v1 -g raw {
 
 client c1 {
 	txreq -hdr "Host: foo"
-	rxresphdrs
+	rxresp
+	# XXX this is actually wrong (missed include)
+	expect resp.bodylen == 57
```
Dridi, line 52: This comment should be changed:

> 15:24 < phk> so this comment should really say "if you want this to be correct, set esi_include_onerror ?

It will be done in a separate commit to keep the revert minimal.

Dridi (commented):
bugwash: merge

nigoroll:
brought up by phk: we need to adjust docs

Dridi:
Rebased against master to give cci a headstart while working on docs.

Dridi:
Rebased against current master, most notably to incorporate #4071.

nigoroll (approved):
thank you!

PR #4066, opened 2024-02-28
Avoid stale PRIV_{TASK,TOP} pointers by calling VRT_priv_{task,top} for each such argument

nigoroll:
Description mostly identical to that of the second commit:

Before this change, we would call `VRT_priv_*` only once per subroutine, which can be *) a nice performance optimization, but leaves us with stale pointers after a rollback.

Rather than adding complications for the rollback case just to keep the option of the "per subroutine pointer cache", just retrieve a fresh priv pointer every time.

The other use of the per subroutine initialization was error handling, which needs additional code outside the function arguments, simply because a `return` statement is not possible within function arguments. We remove the requirement for error handling by making sure that `VRT_priv_{task,top}` always return a valid pointer: They now fall back to a heap allocation if the workspace is full, which is unlikely and should not impact the happy path. In other words, the fallback only exists to avoid the need for error handling.

Fixes https://github.com/varnish/varnish-modules/issues/222

Alternative implementation to #4060

----

*) It is not an optimization in all cases, for example the priv pointers were intialized unconditionally, even if code using them was not reached - but then again, this is something C compilers might optimize...

### DIFF
--- a/bin/varnishd/cache/cache_vrt_priv.c
+++ b/bin/varnishd/cache/cache_vrt_priv.c
@@ -40,9 +40,15 @@
 #include "vcl.h"
 #include "vcc_interface.h"
 
+enum vrt_priv_storage_e {
+	VRT_PRIV_ST_WS = 1,
+	VRT_PRIV_ST_HEAP
+};
+
 struct vrt_priv {
 	unsigned			magic;
 #define VRT_PRIV_MAGIC			0x24157a52
+	enum vrt_priv_storage_e		storage;
 	VRBT_ENTRY(vrt_priv)		entry;
 	struct vmod_priv		priv[1];
 	uintptr_t			vmod_id;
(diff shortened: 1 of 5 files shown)


Dridi (approved):
I spent some time looking at this yesterday, and a little more today (in particular the generated code) and couldn't spot a problem.

For example, `sub priv_task` from the first VCL loaded in `m00000.vtc`:

```c
void v_matchproto_(vcl_func_f)
VGC_function_priv_5f_task(VRT_CTX, enum vcl_func_call_e call,
    enum vcl_func_fail_e *failp)
{
  assert(call == VSUB_STATIC);
  assert(failp == NULL);
  assert(ctx->method & (VCL_MET_BACKEND_ERROR|VCL_MET_BACKEND_FETCH|VCL_MET_BACKEND_RESPONSE|VCL_MET_DELIVER|VCL_MET_FINI|VCL_MET_HASH|VCL_MET_HIT|VCL_MET_INIT|VCL_MET_MISS|VCL_MET_PASS|VCL_MET_PIPE|VCL_MET_PURGE|VCL_MET_RECV|VCL_MET_SYNTH));
  struct vmod_priv *ARG_priv_task_debug = VRT_priv_task(ctx, &VGC_vmod_debug);
  if (ARG_priv_task_debug == NULL) {
    VRT_fail(ctx, "failed to get task priv for vmod debug");
    return;
  }

  /* ... from ('<vcl.inline>' Line 24 Pos 23) */
  {
    {
      VPI_count(ctx, VGC_NREFS, 3);
      if (UNLIKELY(ctx->vpi->trace)) VPI_trace(ctx, 3);
      END_;
      Vmod_xyzzy_debug_Func.f_test_priv_task_get(ctx
      );
      END_;
      Vmod_xyzzy_debug_Func.f_test_priv_task(ctx,
        ARG_priv_task_debug,
        "foo"
      );
      END_;
    }
  }
}
```

For reference:

```
$Function VOID test_priv_task_get() 
$Function STRING test_priv_task(PRIV_TASK, STRING s="")
```

After this change:

```c
void v_matchproto_(vcl_func_f)
VGC_function_priv_5f_task(VRT_CTX, enum vcl_func_call_e call,
    enum vcl_func_fail_e *failp)
{
  assert(call == VSUB_STATIC);
  assert(failp == NULL);
  assert(ctx->method & (VCL_MET_BACKEND_ERROR|VCL_MET_BACKEND_FETCH|VCL_MET_BACKEND_RESPONSE|VCL_MET_DELIVER|VCL_MET_FINI|VCL_MET_HASH|VCL_MET_HIT|VCL_MET_INIT|VCL_MET_MISS|VCL_MET_PASS|VCL_MET_PIPE|VCL_MET_PURGE|VCL_MET_RECV|VCL_MET_SYNTH));

  /* ... from ('<vcl.inline>' Line 24 Pos 23) */
  {
    {
      VPI_count(ctx, VGC_NREFS, 3);
      if (UNLIKELY(ctx->vpi->trace)) VPI_trace(ctx, 3);
      END_;
      Vmod_xyzzy_debug_Func.f_test_priv_task_get(ctx
      );
      END_;
      Vmod_xyzzy_debug_Func.f_test_priv_task(ctx,
        VRT_priv_task(ctx, &VGC_vmod_debug),
        "foo"
      );
      END_;
    }
  }
}

```

And to better see what's happening:

```diff
@@ -5,11 +5,6 @@
   assert(call == VSUB_STATIC);
   assert(failp == NULL);
   assert(ctx->method & (VCL_MET_BACKEND_ERROR|VCL_MET_BACKEND_FETCH|VCL_MET_BACKEND_RESPONSE|VCL_MET_DELIVER|VCL_MET_FINI|VCL_MET_HASH|VCL_MET_HIT|VCL_MET_INIT|VCL_MET_MISS|VCL_MET_PASS|VCL_MET_PIPE|VCL_MET_PURGE|VCL_MET_RECV|VCL_MET_SYNTH));
-  struct vmod_priv *ARG_priv_task_debug = VRT_priv_task(ctx, &VGC_vmod_debug);
-  if (ARG_priv_task_debug == NULL) {
-    VRT_fail(ctx, "failed to get task priv for vmod debug");
-    return;
-  }
 
   /* ... from ('<vcl.inline>' Line 24 Pos 23) */
   {
@@ -21,7 +16,7 @@
       );
       END_;
       Vmod_xyzzy_debug_Func.f_test_priv_task(ctx,
-        ARG_priv_task_debug,
+        VRT_priv_task(ctx, &VGC_vmod_debug),
         "foo"
       );
       END_;
```

The only arguable regression is the resulting panic in place of a VCL failure when a task runs out of workspace and fails to allocate the priv from the heap. I'm willing to dismiss this scenario in which varnishd would likely choke on a failing infallible allocation somewhere else.

Issue #4067, opened 2024-02-28
VDP closing is still incomplete (and wrong since the last change)


nigoroll:
This issue will be followed up by a PR.

### Expected Behavior

VDPs should be closed for all possible code paths and closing VDPs should be a function of the transport alone.

I shall correct my own mistakes.

### Current Behavior

VDPs get [initialized in cnt_transmit()](https://github.com/varnishcache/varnish-cache/blob/4741f409119fa8e7d11e8ad88cf9f5adcf41acd3/bin/varnishd/cache/cache_req_fsm.c#L463-L469) and [ever since VDP_close() got introduced](https://github.com/varnishcache/varnish-cache/commit/d37800783d93d7314d0f370cee51a94bb96d40f6), it was up to the transport layer to close them.
Some time ago I noticed that we would miss to close VDPs in some cases and added 6d423aa5c60940ebde76c91f0a482767346df405 thinking that this might be a simple way to fix the missing close calls from a central place, avoiding repetition. But that was incomplete and wrong, I overlooked more cases and also [created a race condition for pESI](https://gitlab.com/uplex/varnish/libvdp-pesi/-/issues/19).
[cnt_transmit() calls the transport's delivery function whenever setting up the VDPs was successful](https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_req_fsm.c#L496), so all delivery functions should close the VDPs under all circumstances, error or not.

Places where we miss a `VDP_Close()` are:
* in `ved_deliver()` [for](https://github.com/varnishcache/varnish-cache/blob/4741f409119fa8e7d11e8ad88cf9f5adcf41acd3/bin/varnishd/cache/cache_esi_deliver.c#L872) [all](https://github.com/varnishcache/varnish-cache/blob/4741f409119fa8e7d11e8ad88cf9f5adcf41acd3/bin/varnishd/cache/cache_esi_deliver.c#L879) [explicit](https://github.com/varnishcache/varnish-cache/blob/4741f409119fa8e7d11e8ad88cf9f5adcf41acd3/bin/varnishd/cache/cache_esi_deliver.c#L883) [returns](https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishd/cache/cache_esi_deliver.c#L903)
[...]

PR #4068, opened 2024-02-28
Call VDP_Close() for all cases and revert regression

nigoroll:
This PR fixes the issues described in #4067 : 

We add missing `VDP_Close()` calls to the deliver functions of esi and http1, revert a wrong addition of a `VDP_Close()` call, add one at the right place and add a `VDP_Fini()` function with an assertion to ensure all cases are taken care of.


nigoroll:
merged as of bb3d027bf6647185be548050243e1cb9d36741a7 but it seems there is an issue with gh noticing the merge

PR #4069, opened 2024-02-29
Specify how to disable timeouts

### LINKED ISSUE #3045: Setting backend timeouts to zero does not wait forever
According to the docs, setting certain timeouts like `first_byte_timeout` and `between_byte_timeout` to `0` causes them to disable and wait forever, but this is not the case.

https://github.com/varnishcache/varnish-cache/blob/f4dcf8f02a2fb059dacba4c7bedefccbe9897cd2/include/tbl/params.h#L248-L249

These timeouts zero out, but then get unzeroed here:

https://github.com/varnishcache/varnish-cache/blob/f4dcf8f02a2fb059dacba4c7bedefccbe9897cd2/bin/varnishd/cache/cache_session.c#L333-L334

VTC:

```vcl
varnishtest "Check first_byte_timeout=0 works"

server s1 {
	rxreq
	delay 2
	txresp
} -start

varnish v1 -arg "-p first_byte_timeout=0" -vcl+backend {
} -start

client c1 {
	txreq
	rxresp
	expect resp.status == 200
} -run
```

Dridi:
Following consensus, this patch series formalizes several things:

- `NAN` means that a timeout is not set (only possible in VCL, parameters must have a default value)
- it is possible to `unset` timeout variables in VCL
  - they are unset by default (it was already the case)
  - they fall back to the relevant parameter when read (it was already the case for `sess.*timeout*`)
- `INFINITY` disables a timeout
  - it translates to `VRT_DECIMAL_MAX` in VCL
  - in the `poll(2)` case it translates to -1 (aka `INFTIM`) 
  - in the `setsockopt(2)` case it translates to a zero timeval
  - a zero timeout for `setsockopt(2)` translates to a 1us timeout (should it be 1ms?)
- a new duration tweak is added for parameters (inconsequential for designated parameters)
- the existing timeout tweak can  now take a special value "never"
- `bereq` timeouts have two levels of fallback
  - backend timeouts defined in VGC can't use `INFINITY` (no `math.h`)
  - backend timeouts default to a negative value (the lack of zero override was also reported by @walid-git)
  - this is a VRT API breakage! (not the ABI)

I tried to be as thorough as possible, following the trail of relevant timeout parameters.

Fixes #3045


Dridi:
Of course I had to forget something, and it was `Lck_CondWaitTimeout()`, which I already added to the patch series. I decided to be prudent and keep support for zero meaning never timing out, but looking closely the only case where we use zero is actually in the `Lck_CondWait()` function, and `Lck_CondWaitTimeout()` when we forward a zero to `Lck_CondWaitUntil()`.

So I think we could safely make zero no longer special, and in that case have a 1ms timed wait, or less.

PR #4072, opened 2024-03-04
DONOTMERGE: Investigating sanitizer failure on circlecli

Dridi:
It was #4051 until #4063 happened.


Dridi:
The change succeeded here and has succeeded thrice in a row with no failure after pushing the change to the master branch.

PR #4078, opened 2024-03-07
Add VCL_Shutdown to wait for VCL references to vanish

nigoroll:
[A panic with SLASH/fellow](https://gitlab.com/uplex/varnish/slash/-/issues/61) exposed a problem in Varnish-Cache: We shut down stevedores unconditionally, no matter if still used or not. While an argument could be made that stevedores should wait for all object references to be returned before actually shutting down (which SLASH/fellow does, but limited to a fixed number of retries), this would not have helped in this particular case, because a `.free_space` variable was queried which, by definition, requires no reference on stevedore objects whatsoever.

So this PR adds `VCL_Shutdown()` to wait for all VCL references to vanish before closing stevedores.

### DIFF
--- a/bin/varnishd/cache/cache_main.c
+++ b/bin/varnishd/cache/cache_main.c
@@ -560,6 +560,7 @@ child_main(int sigmagic, size_t altstksz)
 	VCA_Shutdown();
 	BAN_Shutdown();
 	EXP_Shutdown();
+	STV_warn();
 	VCL_Shutdown();
 	STV_close();
 
--- a/bin/varnishd/cache/cache_main.c
+++ b/bin/varnishd/cache/cache_main.c
@@ -388,6 +388,20 @@ cli_quit(int sig)
 }
 
 /*=====================================================================
+ * XXX Generalize?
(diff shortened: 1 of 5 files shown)


nigoroll:
bugwash
* @mbgrydeland your feedback on anything like this would be greatly appreciated.
* it has been pointed out that whatever wait might be indefinite, depending on the use case, it it remains undecided if this should be something to address within varnish or from the outside (read: `SIGKILL`)
* using the warn call should allow the stevedore to replace its function pointers with some which always fail, making the problem at least less likely to be hit
*

idl0r:
This looks somewhat "trivial". Is there anything blocking this change/feature?

nigoroll:
When I last worked in this feature, the CCI checks were timing, so I suspected that something might be fundamentally wrong. I am re-checking now and have rebased to begin with.

idl0r:
Ah, ok. Thanks!

idl0r:
@mbgrydeland Any chance for you to take a look at this?

PR #4079, opened 2024-03-08
vcl_int.h: Fix VSL_VER macro

walid-git:
`VSL_VERMASK` is used in `VSL_VER()` macro, which currently always returns 0


Dridi (approved):

nigoroll:
Do I understand correctly that `VSL_VER()` is not currently used anywhere and thus we do not have a problem?

nigoroll (approved):

walid-git:
> Do I understand correctly that VSL_VER() is not currently used anywhere and thus we do not have a problem?

yes this is correct

PR #4080, opened 2024-03-11
fetch: Log bgfetch_no_thread

AlveElde:
The log emitted for a bgfetch task when there are no available threads in the thread pools currently looks like this:
```
1004 Begin           b bereq 1003 bgfetch
1004 VCL_use         b vcl1
1004 BereqAcct       b 0 0 0 0 0 0
1004 End             b 
```
This PR makes it easier to understand what happened:
```
1004 Begin           b bereq 1003 bgfetch
1004 VCL_use         b vcl1
1004 FetchError      b No thread available for bgfetch
1004 BereqAcct       b 0 0 0 0 0 0
1004 End             b 
```


Dridi (approved):
I think this one is safe for the March release.

PR #4091, opened 2024-03-27
range: New http_GetContentRange() semantics

### LINKED ISSUE #4089: 503 Invalid content-range header on partial response with start-end/*
### Expected Behavior

Context:
- A client makes a partial request against Varnish: `Range: bytes=10-19`.
- Varnish is configured to `pass` the request
- The server responds with `Content-Range: bytes 10-19/*"` (and `Transfer-Encoding: chunked`, but not sure if that's necessary to reproduce)

Expected behavior:
- Varnish passes the partial response through

Current behavior:
- Varnish responds with a 503; varnishlog shows `Invalid content-range header`

After delving deeper, this happens in `cache_http.c` at `http_GetContentRange`: `lo` gets set to 10, `hi` to 19, `cl` to -1, as expected; but then `if (*lo >= cl || *hi >= cl) return (-2);` returns -2 and fails the request, because `10 >= -1`.

Not sure I fully understand the function, but should the condition be closer to `if (*hi - *lo < cl) return (-2);` ?

Here's a MWE VTC. Feel free to reuse.

```
varnishtest "Content-Range failure reproducer"

server s1 {
	rxreq
	txresp -nolen -hdr "Transfer-Encoding: chunked" -hdr "content-range: bytes 10-19/*"
	chunkedlen 10
} -start

varnish v1 -vcl+backend {
	sub vcl_recv {
		return (pass);
	}
} -start

client c1 {
	txreq -hdr "range: bytes=10-19"
	rxresp
	expect resp.status == 206
} -run
```

### Current Behavior

TBD

### Possible Solution

_No response_

_No response_

### Context

TBD

### Varnish Cache version

_No response_

### Operating system

_No response_

_No response_

Dridi:
This function returns the content length and puts range indices in lo and hi arguments. A content length of -1 used to be interpreted as something to ignore, but there is a case where the content length may be unknown.

Since we can't represent a zero-length range, because both bounds are inclusive, zero now denotes the lack of content-range header.

Unknown range units are treated as errors as they wouldn't pass the busyobj check for the range header, even for pass transactions. The only way to use "extension" range units is to turn `http_range_support` off.

The calling convention for `http_GetContentRange()` remains otherwise the same, and for good measure `http_GetContentLength()` received a similar description.

Fixes #4089

### DIFF
--- a/bin/varnishd/cache/cache_http.c
+++ b/bin/varnishd/cache/cache_http.c
@@ -873,7 +873,14 @@ http_GetHdrField(const struct http *hp, hdr_t hdr,
 	return (i);
 }
 
-/*--------------------------------------------------------------------*/
+/*--------------------------------------------------------------------
+ * Parse content-length header
+ *
+ * Return:
+ *	-2	invalid
+ *	-1	unknown length
+ *	>= 0	body length
+ */
 
 ssize_t
 http_GetContentLength(const struct http *hp)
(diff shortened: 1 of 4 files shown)


nigoroll (approved):
lgtm

dridi:
I haven't had a chance to look at your suggestion from https://github.com/varnishcache/varnish-cache/issues/4089#issuecomment-2129395473.

PR #4094, opened 2024-04-09
hash: Close an expiry loophole

Dridi:
The opposite of 'EXP_Ttl(req, oc) > req->t_req' should not have been 'EXP_Ttl(NULL, oc) < req->t_req'. If we somehow enter the lookup when the two operands are equal, the objcore suffers a phenomenon known as Schrödinger's expiry.

The chances of running into this scenario range from epsilon to 100%.

Because 't_req' is stable across restarts, a soft purge will reliably trigger this case.

Test case by @AlveElde who first demonstrated the problem.


Dridi:
bugwash: ok

PR #4096, opened 2024-04-25
h2: Ensure highest_stream is consistent also on Upgrade: h2c

daghf:
`highest_stream` gets used both in transmitting goaway frames, and also for internal protocol consistency checks.

For the `Upgrade: h2c` handling we fail to bump this when kicking off the stream for the request that got the upgrade, so any subsequent frame transmitted on that stream will be treated with an inconsistent view on `highest_stream`. In particular, this leads to an erroneous `PROTOCOL_ERROR` on reception of a `WINDOW_UPDATE`.


Dridi (approved):

bsdphk:
Looks good to me.

PR #4097, opened 2024-04-29
POC: New beresp.error VCL variable

Dridi:
Following several VDD discussions, this patch series shows how getting a backend error could look like in VCL, and how it could be articulated in the core code.

This should check the boxes from the last consensus:

- a plain text error message
- a new VCL variable to get the error (`beresp.error`)
- VCL constants for safe comparisons (`error.*`)

The first two commits are unrelated, I found a bug in libvcc while I was working on this. I will submit a pull request once I'm done with that bug (only partly fixed here) so the relevant commits to review are the ones starting with "POC".

If we still have consensus on this approach, I will find someone™ to submit something comprehensive.

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -266,7 +266,13 @@ sub vcl_builtin_backend_error {
   </head>
   <body>
     <h1>Error "} + beresp.status + " " + beresp.reason + {"</h1>
-    <p>"} + beresp.reason + {"</p>
+    <p>"};
+	if (beresp.error) {
+		set beresp.body += beresp.error;
+	} else {
+		set beresp.body += beresp.reason;
+	}
+	set beresp.body += {"</p>
     <h3>Guru Meditation:</h3>
     <p>XID: "} + bereq.xid + {"</p>
     <hr>
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -442,6 +442,7 @@ struct busyobj {
 
 	uint16_t		err_code;
 	const char		*err_reason;
+	const char		*err_resp;
 
 	const char		*client_identity;
 };
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -49,6 +49,7 @@
 #include "cache_vcl.h"
 #include "http1/cache_http1.h"
 #include "proxy/cache_proxy.h"
+#include "vrt_obj.h"
 
 #include "VSC_vbe.h"
 
@@ -60,7 +61,7 @@ static struct lock backends_mtx;
 
 /*--------------------------------------------------------------------*/
 
-void
+const char *
 VBE_Connect_Error(struct VSC_vbe *vsc, int err)
 {
 
@@ -81,7 +82,7 @@ VBE_Connect_Error(struct VSC_vbe *vsc, int err)
 		break;
 	case ECONNREFUSED:
 		vsc->fail_econnrefused++;
-		break;
+		return (VRT_c_error_connection_refused());
 	case ENETUNREACH:
 		vsc->fail_enetunreach++;
 		break;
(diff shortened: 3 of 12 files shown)


Dridi:
Bugwash summary:

* Turn errors into a `struct vrt_error { const char *type, *msg };`
  * @nigoroll wanted to categorize errors (example: "backend timeout")
  * I suggested an `(ERROR).type` VCL attribute
  * It coud also be a list of tags (not brought up during bugwash)
* Turn `beresp.error` into a proper `ERROR`
* Teach `$Error` to vmods (add vmod_debug coverage)
* Error messages from the documentation should be logged (I meant to initially and forgot)

Then we reevaluate the error reporting.

dridi:
I wanted to revisit this ticket this morning and noticed that I forgot about the libvcc bug. I pushed 0c8448e0ef...b77b709510 to fix it.

nigoroll:
bugwash label removed because this ticket needs an update

dridi:
New proposal:

```c
struct vrt_error {
        const char *msg;
        const char *tags; /* comma-separated list of VCL identifiers */
};
```

Can be produced by a VMOD and put in context, to be consumed from `vcl_backend_error` (and probably `vcl_synth` too). The error can be null, like a header.

VCL defines new variables:

```
beresp.error

	Type: ERROR

	Readable from: vcl_backend_error

	TODO: description (has a tostring yielding `msg` or `NULL`).

beresp.error.*

	Type: BOOL

	Readable from: vcl_backend_error

	TODO: description, tells whether the error has an arbitrary tag.
```

We gradually start defining ERROR constants in VCL, as we increase our `beresp.error` coverage for at least faults happening in core code:

```
error.connection_refused

	Type: ERROR

	Error message: Connection refused by the backend

	Tags: backend, connect

	TODO: description (maybe mention ECONNREFUSED etc)
```

The `error.*` symbols are implicitly read-only and unrestricted in scope, in other words global constants.

In VCL you could do something like this:

```vcl
sub vcl_backend_error {
        if (!beresp.error.backend || beresp.error == error.connection_refused) {
                return (retry);
        }
}
```

Likewise, VMODs learn to define errors in their VCC file:

```
$Error "msg" tag, tags... # XXX: implicit tag with the name of the VMOD?
```

And you can use them also:

```vcl
import foo;

sub vcl_backend_error {
        if (beresp.error.foo && beresp.error != foo.bar) {
                return (abandon);
        }
}
```

nigoroll:
> New proposal:

That's nice! Can you please illustrate the use of tags?

dridi:
> That's nice! Can you please illustrate the use of tags?

The idea of tags is to classify errors and allow operating on classes of errors instead of individual ones. We should be conservative about tags, as we can easily add them, but removing is a breaking change.

> In VCL you could do something like this:
> 
> ```varnish-vcl
> sub vcl_backend_error {
>         if (!bereq.error.backend || bereq.error == error.connection_refused) {
>                 return (retry);
>         }
> }
> ```

You can read this example as "I'm retrying the backend fetch if the error was not related to a backend, unless it was specifically a connection refused".

I suppose it would be more accurate like this for the `return (error)` case:

```vcl
sub vcl_backend_error {
        if (bereq.error && (!bereq.error.backend ||
            bereq.error == error.connection_refused)) {
                return (retry);
        }
}
```

In summary, multiple errors may share one or more tags.

Tag ideas:

- backend
- timeout
- proto
- workspace

Something we can figure out as we start populating errors coming from core code.

nigoroll:
@dridi I still do not get what is the tag in the vcl example. Can you please illustrate?

nigoroll:
VDD:
* General agreement that we should improve:
  * clarity of the log
  * visibility from vcl
* categories (is "tag" a good name?) should be available from vcl and also be reflected in the log
* should counters reflect the same errors (might be too many)? going back to VSL for details should be ok
* add error codes with cross-reference to documentation?

please think about the details and make a concrete proposal
* which categories do we want? (with operational experience)
* How will it look like in vcl and vsl? Keep strings or move to IDs?
  * VSL: gate behind a feature for backwards compatibility?

Issue #4098, opened 2024-04-29
Intermittent timeout in r02310.vtc


stevendore:
### Expected Behavior

Test to pass consistently.

### Current Behavior

Test fails after about 30 consecutive runs.

<details><summary>VTC output</summary>
<p>
**** dT    0.000
*    top   TEST tests/r02310.vtc starting
**** top   extmacro def pkg_version=trunk
**** top   extmacro def pkg_branch=trunk
**** top   extmacro def pwd=/home/swojcik/code/varnish-cache/bin/varnishtest
**** top   extmacro def date(...)
**** top   extmacro def string(...)
**** top   extmacro def localhost=127.0.0.1
**** top   extmacro def bad_backend=127.0.0.1:34979
**** top   extmacro def listen_addr=127.0.0.1:0
**** top   extmacro def bad_ip=192.0.2.255
**** top   extmacro def topbuild=/home/swojcik/code/varnish-cache
**** top   extmacro def topsrc=/home/swojcik/code/varnish-cache
**** top   macro def testdir=/home/swojcik/code/varnish-cache/bin/varnishtest/tests
**** top   macro def tmpdir=/tmp/vtc.1323156.6e021e02
**** top   macro def vtcid=vtc.1323156.6e021e02
**   top   === varnishtest "#2310: Panic on premature hangup after Upgrade:...
*    top   VTEST #2310: Panic on premature hangup after Upgrade: h2c
**   top   === barrier b1 cond 2
**   top   === server s1 {
**   s1    Starting server
**** s1    macro def s1_addr=127.0.0.1
**** s1    macro def s1_port=40005
**** s1    macro def s1_sock=127.0.0.1:40005
*    s1    Listen on 127.0.0.1:40005
**   top   === varnish v1 -vcl+backend {} -start
**** dT    0.001
**   s1    Started on 127.0.0.1:40005 (1 iterations)
**** dT    0.005
**   v1    Launch
[...]

PR #4099, opened 2024-04-30
vtc: Stabilize r02310.vtc

walid-git:
> The test is racy, since #3998 we no longer perform the backend fetch if the client disconnects before it is started. Thus, the test will be stuck at [this](https://github.com/varnishcache/varnish-cache/blob/5e98181a6180a50df0836539b2bc12b2264fb8fe/bin/varnishtest/tests/r02310.vtc#L37) line until it times out since the [other](https://github.com/varnishcache/varnish-cache/blob/5e98181a6180a50df0836539b2bc12b2264fb8fe/bin/varnishtest/tests/r02310.vtc#L7C2-L7C9) `barrier b1 sync` will never be reached.

Fixes: #4098

### DIFF
--- a/bin/varnishtest/tests/r02310.vtc
+++ b/bin/varnishtest/tests/r02310.vtc
@@ -11,6 +11,7 @@ server s1 {
 varnish v1 -vcl+backend {} -start
 
 varnish v1 -cliok "param.set feature +http2"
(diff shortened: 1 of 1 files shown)


Dridi (approved):
Excellent.

PR #4105, opened 2024-05-12
Argument defaults from environment variables?

nigoroll:
After 15 years of tirelessly repeating `-n <foo>` for <s>each and every</s> varnish admin command, I wonder: is it time to set defaults from environment variables?

This PR is not remotely complete. It is merely to ask about opinions.

### DIFF
--- a/bin/varnishadm/varnishadm.c
+++ b/bin/varnishadm/varnishadm.c
@@ -467,6 +467,7 @@ main(int argc, char * const *argv)
 		printf(OPTARG "\n");
 		exit(0);
 	}
+	n_arg = getenv("VARNISH_DEFAULT_N");
 	/*
 	 * By default linux::getopt(3) mangles the argv order, such that
 	 *	varnishadm -n bla param.set foo -bar
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -614,7 +614,7 @@ main(int argc, char * const *argv)
 	const char *i_arg = NULL;
 	const char *j_arg = NULL;
 	const char *h_arg = "critbit";
-	const char *n_arg = NULL;
+	const char *n_arg = getenv("VARNISH_DEFAULT_N");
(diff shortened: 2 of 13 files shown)


gquintard:
Yes please, that would make containers way easier to use

bin/varnishadm/varnishadm.c
```diff
@@ -467,6 +467,7 @@ main(int argc, char * const *argv)
 		printf(OPTARG "\n");
 		exit(0);
 	}
+	n_arg = getenv("VARNISH_WORKING_DIR");
```
Dridi, line 470: Look like something we would want in libvarnishapi instead.

Mandatory bikeshedding: `VARNISH_WORKDIR`

nigoroll: > Look like something we would want in libvarnishapi instead.

Yes, just `varnishadm` does not use `VUT`.

> Mandatory bikeshedding: `VARNISH_WORKDIR`

please take a look:

```
$ git grep -Ei 'work.?dir' 
$ git grep -Ei 'working.?dir' 
```

If I did not overlook anything, there is one error message where it is "workdir", otherwise "workdir" is only used in code and all other messages and documentation are for "working directory". I abbreviated to "WORKING_DIR", but probably it should be `VARNISH_WORKING_DIRECTORY`?

Dridi: > Yes, just `varnishadm` does not use `VUT`.

I was referring to `VSM_Attach()`.

> all other messages and documentation are for "working directory". I abbreviated to "WORKING_DIR", but probably it should be `VARNISH_WORKING_DIRECTORY`?

If anything, I'd rather keep environment variables short and have the documentation use the whole "working directory" phrase.

nigoroll:
bugwash: the bikeshed color is going to be `VARNISH_DEFAULT_N` and we are keeping it there for now, as phk is skeptical about other defaults. But at least the name scheme is generic.

Issue #44, opened 2016-03-04
DOC input:  ReqEnd shmem record (from Trac)


bsdphk:
Old ticket imported from Trac
See archived copy here: https://varnish-cache.org/trac/ticket/44

PR #4109, opened 2024-05-21
Add ObjWaitExtend2() to provide a consistent boc state

nigoroll:
For context see [SLASH/fellow issue #44](https://gitlab.com/uplex/varnish/slash/-/issues/44).

Clients to the Object API need to know not only the current extension (new length) of streaming objects, but also the streaming state - in particular `BOS_FINISHED` and `BOS_FAILED`. The latter for obvious reasons, and the former to call the delivery function with `OBJ_ITER_END`, which then likely results in `VDP_END` sent down the delivery pipeline.

Background:

It is important for efficient delivery to not receive an additional `VDP_END` with a null buffer, but rather combined with the last chunk of data, so, consequently, it is important to reliably send `OBJ_INTER_END` also with the last chunk of data.

Consequent to all of this, `ObjWaitExtend()` callers need to know when `BOS_FINISHED` has been reached for some extension.

The current API, however, does not provide a consistent view of the streaming state, which is only available from within the critical region of `ObjWaitExtend()`.

Thus, we add the streaming state as an optional return value.

With this commit, we also remove a superfluous line to set `rv` again: Because `boc->fetched_so_far` must only be updated while holding the `boc` mutex, reading the value again provides no benefit.

### DIFF
--- a/bin/varnishd/cache/cache_obj.c
+++ b/bin/varnishd/cache/cache_obj.c
@@ -265,8 +265,10 @@ ObjExtend(struct worker *wrk, struct objcore *oc, ssize_t l, int final)
  */
 
 uint64_t
-ObjWaitExtend(const struct worker *wrk, const struct objcore *oc, uint64_t l)
+ObjWaitExtend(const struct worker *wrk, const struct objcore *oc, uint64_t l,
+    enum boc_state_e *statep)
 {
+	enum boc_state_e state;
 	uint64_t rv;
 
 	CHECK_OBJ_NOTNULL(wrk, WORKER_MAGIC);
@@ -282,15 +284,16 @@ ObjWaitExtend(const struct worker *wrk, const struct objcore *oc, uint64_t l)
 			oc->boc->delivered_so_far = l;
 			PTOK(pthread_cond_signal(&oc->boc->cond));
 		}
-		if (rv > l || oc->boc->state >= BOS_FINISHED)
+		state = oc->boc->state;
+		if (rv > l || state >= BOS_FINISHED)
 			break;
 		(void)Lck_CondWait(&oc->boc->cond, &oc->boc->mtx);
 	}
-	rv = oc->boc->fetched_so_far;
 	Lck_Unlock(&oc->boc->mtx);
+	if (statep != NULL)
+		*statep = state;
 	return (rv);
 }
-
 /*====================================================================
  */
 
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -48,7 +48,7 @@
 
 #define VRT_MAJOR_VERSION	19U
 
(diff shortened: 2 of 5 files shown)


bin/varnishd/cache/cache_obj.c
```diff
@@ -265,8 +265,10 @@ ObjExtend(struct worker *wrk, struct objcore *oc, ssize_t l, int final)
  */
 
 uint64_t
-ObjWaitExtend(const struct worker *wrk, const struct objcore *oc, uint64_t l)
+ObjWaitExtend2(const struct worker *wrk, const struct objcore *oc, uint64_t l,
+    enum boc_state_e *statep)
```
dridi, line 269: I think it would be much simpler to change `ObjWaitExtend()` that has a single call site in tree.

nigoroll: Last time I suggested a breaking changes in #4035 I received concerns, so this time I suggested the other way. But actually I agree that changing `ObjWaitExtend()` is the better option.

nigoroll: I have changed the PR accordingly, squashed commits, improved commit messages, updated changes.rst and vrt.h and updated the PR description.

include/vrt.h
```diff
@@ -58,6 +58,8 @@
  * binary/load-time compatible, increment MAJOR version
  *
  * NEXT (2024-09-15)
+ * 19.1 (xxx)
+ *	ObjWaitExtend() gained statep argument
```
dridi, line 62: Not a VRT change, shouldn't trigger a VRT bump.

dridi (approved):

include/vrt.h
nigoroll: We have a long tradition of recording many more changes in `vrt.h` than changes to functions prefixed `VRT_`, and we have not even defined what VRT really is (#3555).
Using the VRT version to mark changes like this one GREATLY simplifies use in vmods/vexts, in particular in light of the de-autotoolification at zero + epsilon cost. So what is the gain of NOT using vrt logs and versions?

dridi: We've already been there, I find it confusing to make a VRT bump for something other than a VRT change. And that includes the systematic "better safe than sorry" systematic major (!) bump for biannual releases.

This particular function is in `cache_varnishd.h`, which should settle the VRT-or-not question in this case.

Note that I approved the pull request and didn't even try to "request changes". I'm merely mentioning that this should say "NEXT" instead of "19.1" but I'm not vetoing it.

dridi: I agree that VEXTs are a different beast of their own.

nigoroll:
bugwash: phk agrees

PR #4110, opened 2024-05-24
Create a directory owned by the worker process for use as a TMPDIR

nigoroll:
Processes created by the worker might require a `TMPDIR` for temporary files (think: filters). Create a dedicated directory for this purpose, which is otherwise hard to achieve in a universal manner.

### DIFF
--- a/bin/varnishd/mgt/mgt_main.c
+++ b/bin/varnishd/mgt/mgt_main.c
@@ -319,6 +319,7 @@ mgt_Cflag_atexit(void)
 		VJ_rmdir("vext_cache");
 	}
 	VJ_rmdir("vmod_cache");
+	VJ_rmdir("worker_tmpdir");
 	(void)chdir("/");
 	VJ_rmdir(workdir);
 }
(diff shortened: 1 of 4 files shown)


bin/varnishd/mgt/mgt_main.c
```diff
@@ -881,6 +882,13 @@ main(int argc, char * const *argv)
 		    workdir, VAS_errtxt(errno));
 	}
 
+	if (VJ_make_subdir("worker_tmpdir",
+	    "TMPDIR for the worker process", NULL)) {
+		ARGV_ERR(
+		    "Cannot create vmod directory (%s/worker_tmpdir): %s\n",
```
dridi, line 888: Is this limited to VMODs? Are you assuming enough privileges in the first place for VEXTs?

asadsa92: Should Varnish provide helper function to create the temp files to avoid collision between the consumer of this temp directory?

dridi: We should at least provide a macro or a `const char *` (for actual linking) with the directory name and let VMODs be responsible for their own stuff. Not sure I'd want to reinvent `mktemp` for VMODs... :shrug:

nigoroll: The proposed directory is for any code executing within the worker process, hence the name.
I am 🤷🏽 on the macro/string also, because it would just be `worker_tmpdir` relative to cwd.

dridi: Likewise, I approved it as it is.

nigoroll:
backporters: fec25970c8addc28fb90d4dcd8d4fea592a03b13

PR #4111, opened 2024-05-29
Generalize the VDP API: struct vrt_init_ctx

nigoroll:
This commit is to prepare use of the VDP API also for the backend side to filter bereq.body through bereq.filters by adding a struct vrt_init_ctx argument to the VDP init function.

For background, see #4035


dridi:
As mentioned during bugwash, I prefer the #4112 approach.

PR #4114, opened 2024-06-03
vbt: Move backtrace code to libvarnish

dridi:
This is an old patch series from more than a year ago that I mentioned briefly to @bsdphk back then. The idea was to move the back-trace rendering code to libvarnish and make the panic code use it.

We needed to print back-traces outside of `varnishd` to track down an alleged unknown bug in our `getaddrinfo()` usage, that would only happen under an unknown set of circumstances, but it turned out to be a silly overflow not reported as such by glibc.

In other words, it took us a back-trace to realize we were facing another incarnation of #3765 because 247eb50f47d4dae5eeacc54375eb3f2adde086a1 had not been back-ported to the 6.0 branch. Once the problem was solved the VBT code was forgotten until last Friday.

Apologies for the delay.

### DIFF
--- a/bin/varnishd/cache/cache_panic.c
+++ b/bin/varnishd/cache/cache_panic.c
@@ -31,12 +31,6 @@
 
 #include "config.h"
 
-#ifdef WITH_UNWIND
-#  include <libunwind.h>
-#else
-#  include <execinfo.h>
-#endif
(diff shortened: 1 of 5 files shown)


bsdphk:
OK with me.

PR #4117, opened 2024-06-06
vsc: Revamp vsc.c and VSC_Iter()

dridi:
This is a change implemented mostly by @mbgrydeland in the first commit:

> This patch polishes and fixes up the VSC parts of libvarnishapi. Main
> points are:
> 
> 1) The merging of the VSM list and the VSC segment list is done in a
> separate loop from where the callbacks are executed. This prevents a
> VSM_Status() being executed in a callback from interfering with the update
> of the VSC segment list to match the VSM list.
> 
> 2) Failures to map counters does not remove the counter from the VSC seg
> list. Previously we would remove VSC seg list elements on map failures,
> but doing so would then cause issues when attempting to consolidate the
> VSM list and the VSC seg list on the next call to VSC_Iter(). With this
> patch, map failures are recorded as a state without removing the VSC seg
> list, and map failures will be retried on subsequenct VSC_Iter() calls.
> 
> 3) Handle failures to map the corresponding DOC seg during mapping of a
> counter seg gracefully. Previously this would cause an assert. Now it will
> cause a map failure of the counter seg.

It is a followup to ba7bf7ff6dba835e2cba9ad0a07131819272fb8e that was fixing one of two symptoms of failed mappings, at "unmap" time. The other symptom was failing to find the VSC_DOC segment of a VSC segment, and we eventually identified a scenario that could lead to this inconsistency, and this scenario could also trigger the first symptom.

The problem with the single VSC iteration loop (coordinating a loop over two same-ordered lists of VSM segments and their VSC counterparts) is that we could fail to map a doc segment (on Linux `vm.max_map_count` is our suspect in the `sysctl` department), go on to the next iteration, clean up stale segments (reducing the number of maps) and successfully mapping a VSC segment lacking documentation.

The changes are strictly internal and otherwise maintain the current observable behavior.

### DIFF
--- a/lib/libvarnishapi/vsc.c
+++ b/lib/libvarnishapi/vsc.c
@@ -89,6 +89,7 @@ struct vsc_seg {
 #define VSC_SEG_MAGIC		0x801177d4
 	enum vsc_seg_type	type;
 	VTAILQ_ENTRY(vsc_seg)	list;
+	VTAILQ_ENTRY(vsc_seg)	doc_list;
 	struct vsm_fantom	fantom[1];
 	const struct vsc_head	*head;
 	const char		*body;
@@ -110,6 +111,7 @@ struct vsc {
 	unsigned		raw;
 	struct vsc_sf_head	sf_list;
 	struct vsc_seg_head	segs;
+	struct vsc_seg_head	docs;
 
 	VSC_new_f		*fnew;
 	VSC_destroy_f		*fdestroy;
@@ -144,6 +146,7 @@ VSC_New(void)
 		return (vsc);
 	VTAILQ_INIT(&vsc->sf_list);
 	VTAILQ_INIT(&vsc->segs);
+	VTAILQ_INIT(&vsc->docs);
 	return (vsc);
 }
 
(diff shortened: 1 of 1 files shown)


mbgrydeland (approved):
This looks to be the complete backport of our changes to fix these issues.

lib/libvarnishapi/vsc.c
```diff
@@ -307,91 +319,148 @@ vsc_fill_point(const struct vsc *vsc, const struct vsc_seg *seg,
...
+	 * isn't ready, fail the mapping. The transitions inside
+	 * varnishd that we are waiting for are just some memcpy()
+	 * operations, so there is no reason to allow a long retry
+	 * time. */
+	for (retry = 10; retry > 0 && head->ready == 0; retry--)
+		usleep(10000);
+
+	if (head->ready == 0) {
+		VSM_Unmap(vsm, sp->fantom);
+		return (-1);
```
dridi, line 410: Missing check for `VSM_Unmap()`.

dridi (commented):
bugwash: `vmb.h` no longer needed, and missing assertion.

In addition, we should revisit the VSM machinery and see whether growing and shrinking mmaps works better on Linux these days to limit our maps consumption.

dridi:
> In addition, we should revisit the VSM machinery and see whether growing and shrinking mmaps works better on Linux these days to limit our maps consumption.

I went ahead and created a VIP for this:

https://github.com/varnishcache/varnish-cache/wiki/VIP34:-Conservative-map-usage-in-shared-memory

PR #4118, opened 2024-06-06
circleci: Run sanitizers on a larger system

dridi:
This is an attempt at speeding up this job that usually takes three time as long as other distcheck jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,7 +127,7 @@ jobs:
         description: the Linux distribution, with release, e.g. debian:buster, centos:7
         type: string
       rclass:
-        description: the resource class to use, usuall arm.medium or medium
+        description: the resource class to use, usually arm.medium or medium
         type: string
     machine:
       image: ubuntu-2004:202111-02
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -433,8 +438,9 @@ workflows:
           name: build_ubuntu_noble
           dist: ubuntu
           release: noble
+          rclass: large
           extra_conf: --enable-asan --enable-ubsan --enable-workspace-emulator
-          make_target: check -j2
+          make_target: check
```
asadsa92, line 443: Is this a typo? The large instance have 4 vCPU (instead of 2 for medium), should we not use `-j4` here?

dridi: `-j4` is the default, so we no longer need a `-j2` override.

asadsa92 (approved):
LGTM, we reduced the runtime of ASAN by 50% but I think we can do better (see my comment).

I inspected at our usage and we have only been using maximum 44% of the credits we have been given by CCI for open-source. So, it should be fine in that area,

dridi:
Approximately 15m (29m before this change) for the sanitizer job vs 11m for other jobs.

The large instance will consume more cci credits, but we should also save credits on time spent running jobs.

.circleci/config.yml
asadsa92: Aha, nvm then.

PR #4120, opened 2024-06-12
VTC output race

scance:
This aims at collecting varnishd's output when something went wrong in a test.


dridi (approved):
If I remember correctly, the problem occurred very early during the varnish startup when we ended up with no output explaining why it failed.

scance:
Yes exactly, the issue is `varnishtest` exits before collecting the output meaning the log does not contain why `varnishd` failed, for example due to permission errors or a missing file etc.

bsdphk:
Looks good.

PR #4122, opened 2024-06-14
Allow VEV_Stop() called from within a VEV callback

daghf:
The routine in VEV looping to report on all of the fd events gets confused if an event is removed while in another callback event.

This will typically cause the mgt event loop to just spin, making the manager unresponsive.

Original patch and commit message by Martin Blix Grydeland.


dridi (approved):
LGTM

bsdphk:
Looks OK to me.

Issue #4121, opened 2024-06-13
Child killed as it was unresponsive - stuck in deadlock with __GI___lll_lock_wait during vsl_get?


ctheune:
### Expected Behavior

Children should not get killed ;)

### Current Behavior

Children get killed, likely due to a deadlock, potentially with the master process?

### Steps to Reproduce (for bugs)

Unfortunately, this has been a one off AFAICT and we can't reproduce at this moment.

### Context

We've seen random stuck children every now and then but only received a decipherable core dump now.

Interestingly we had to restart the master process as well to get this unstuck.

```
Jun 05 09:57:11 xxx varnishd[679713]: Error: Unexpected reply from ping: 400 CLI communication error (hdr)
Jun 05 09:57:11 xxx varnishd[679713]: Child (2168345) not responding to CLI, killed it.
Jun 05 09:57:11 xxx varnishd[679713]: Unexpected reply from ping: 400 CLI communication error (hdr)
Jun 05 09:57:57 xxx systemd-coredump[4051844]: Could not parse number of program headers from core file: invalid `Elf' handle
Jun 05 09:57:57 xxx systemd-coredump[4051594]: [🡕] Process 2168345 (cache-main) of user 75 dumped core.
                                                    
                                                    Module /var/spool/varnish/xxx/vmod_cache/_vmod_std.974d8b4da76817f9997109be9ede98ec6f04539caf6eee61b9c40b2529081113 without build-id.
                                                    Module /var/spool/varnish/xxx/vcl_boot.1707858351.045632/vgc.so without build-id.
                                                    Module /var/spool/varnish/xxx/vcl_boot.1707858351.045632/vgc.so
                                                    Module libgcc_s.so.1 without build-id.
                                                    Module libstdc++.so.6 without build-id.
                                                    Module libjemalloc.so.2 without build-id.
                                                    Module libpcre2-8.so.0 without build-id.
                                                    Module .varnishd-wrapped without build-id.
                                                    Stack trace of thread 2168345:
                                                    #0  0x00007fbc8ef52d0f __GI___lll_lock_wait (libc.so.6 + 0x82d0f)
                                                    #1  0x00007fbc8ef59122 __pthread_mutex_lock@GLIBC_2.2.5 (libc.so.6 + 0x89122)
                                                    #2  0x0000000000457990 vsl_get (.varnishd-wrapped + 0x57990)
                                                    #3  0x0000000000457d67 vslr (.varnishd-wrapped + 0x57d67)
[...]

PR #4123, opened 2024-06-14
varnishd: Warn when workdir is not  on tmpfs

walid-git:
Adds a warning when workdir is not on a tmpfs mounted partition.
Refs #4121

### DIFF
--- a/bin/varnishd/mgt/mgt_jail.c
+++ b/bin/varnishd/mgt/mgt_jail.c
@@ -42,6 +42,11 @@
 //lint -efile(766, sys/statvfs.h)
 #include <sys/statvfs.h>
 
+#if defined(HAVE_SYS_VFS_H) && defined(HAVE_LINUX_MAGIC_H)
+#  include <sys/vfs.h>
+#  include <linux/magic.h>
+#endif
+
 #include "mgt/mgt.h"
 #include "common/heritage.h"
 #include "vav.h"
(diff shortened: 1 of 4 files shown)


bin/varnishd/mgt/mgt_jail.c
```diff
@@ -177,7 +182,12 @@ VJ_make_workdir(const char *dname)
 		    "can not reside on a file system mounted noexec\n", dname);
 	}
 #endif
+#if defined(HAVE_FSTATFS) && defined(TMPFS_MAGIC)
+	struct statfs sfs[1];
 
+	if (!fstatfs(i, sfs) && sfs->f_type != TMPFS_MAGIC)
+			MGT_Complain(C_WARN, "Workdir is not mounted on tmpfs\n");
```
dridi, line 189: Double indent.

dridi (approved):

dridi:
bugwash: make a new linux jail to isolate linuxisms like this one and make it reuse the bits from the unix jail.

nigoroll:
I can see the motivation of the bugwash decision, but I feel like it is complicating things for no relevant gain. Today, the linux jail would be identical to the unix jail except for ~2 lines added, and we could still (easily) start the linux jail once more specifics arise.
I think this addition has its merits already.

bsdphk:
The linux jail can just call into the standard unix-jail for everything which does not differ.

The reason I want this kind of platform-magic segregated is that other wise we end up with a nightmare of #ifdefs all over the place.

dridi:
> The linux jail can just call into the standard unix-jail for everything which does not differ.

This is how we are approaching this. It should be "that simple" initially, and we should actually be able to grow new linux-specific containment features. While I sympathize with @nigoroll's goal to get this in now, and take care of the linux jail later, I also sympathize with @bsdphk's view that if let this go in, we lose an incentive to actually materialize the linux jail and centralize linux-specific checks.

Luckily for us, @cartoush is already working on it. There's also another linux-specific piece of code in the unix jail, and it will find a new home in the process.

I think we can close it for good.

PR #4127, opened 2024-06-19
 circleci: Retire EOL platforms (7.4)

dridi:
https://github.com/varnishcache/varnish-cache/pull/4119 and other commit dependencies applied to the 7.4 branch.

PR #4128, opened 2024-06-19
 circleci: Retire EOL platforms (6.0)

dridi:
A forced update of the circleci configuration to 7.4 (from a 7.2 equivalent) followed by the commits from #4127.


dridi:
It turns out we needed a lot more back-ports than I initially thought to catch up with the circleci configuration from trunk.

One thing I had to give up for this round was sanitizer coverage (see 006c0e05c56d6aa85e3178f99aad0e4d8512def5) that is currently not exercised in 6.0 and not actionable because of leaks that should be investigated.

It will be revisited.

Likewise, there was no 32bit coverage before this pull request, and a few test cases were missing back-ports.

There is one thing that I don't plan to revisit. I had to disable one test case on Alpine Linux (see 69de4eb5ec082092b5660ace3a1f1037266f3ef9) that suddenly started to fail. I was not able to reproduce the problem in a local alpine-latest container.

I will need to check packaging jobs before considering this work complete.

PR #4129, opened 2024-06-19
vtc_logexp: Exclude SLT_Witness by default

dridi:
They have not posed a problem for a while, but one change in the locking department can be enough to break a test case.

---

Initially pushed to #4128 (still work in progress) since too many test cases broke in the presence of lock Witness logs.

This should future-proof our logexpect usage in trunk.


dridi:
Never mind, I didn't realized that we handled this already in f8817b8b6cfc504912436ffc74b1b7d1e851d35a.

PR #4130, opened 2024-07-03
ws: Prep work for trailers support

dridi:
Originally from #4125, submitted independently as per bugwash to reduce the size of #4125.

### DIFF
--- a/bin/varnishd/cache/cache_session.c
+++ b/bin/varnishd/cache/cache_session.c
@@ -258,12 +258,20 @@ HTC_Status(enum htc_status_e e, const char **name, const char **desc)
 void
 HTC_RxInit(struct http_conn *htc, struct ws *ws)
 {
-	unsigned l;
+	unsigned rollback;
+	int l;
 
 	CHECK_OBJ_NOTNULL(htc, HTTP_CONN_MAGIC);
 	htc->ws = ws;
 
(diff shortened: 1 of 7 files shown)


nigoroll:
The second (hpack) commit seems to make sense to me, but I do not understand the first: What are your plans with the pipeline logic? #4125 does not seem to make use of the conditional rollback and I do not understand what the meaning of a "pipeline at the end of the workspace" would be?

dridi:
> What are your plans with the pipeline logic? #4125 does not seem to make use of the conditional rollback

The new condition for the rollback ended up in this pull request, in `HTC_RxInit()`.

> I do not understand what the meaning of a "pipeline at the end of the workspace" would be?

I can't find a reference to the text you quoted.

dridi:
> What are your plans with the pipeline logic?

I forgot to answer this specific question.

Fetching trailers is done with the same building blocks that fetch headers. And they feed from pipelined bytes before fetching more. Your observation is on point, the pull request currently doesn't leave any room for pipelining, because chunked framing is still fetched one byte at a time.

Moving away from this suboptimal (euphemism) mode of operations is definitely a goal. I opted for status quo, because that reduced the amount of changes for the first step.

dridi:
It prevents `HTC_RxInit()` from rolling back a req workspace for trailers.

bugwash: merge and rebase #4125 onto trunk.

PR #4132, opened 2024-07-08
Add a parameter for the panic buffer size

nigoroll:
64k ought to be enough for everyone. /s

### DIFF
--- a/include/tbl/params.h
+++ b/include/tbl/params.h
@@ -733,6 +733,22 @@ PARAM_SIMPLE(
 )
 
 PARAM_SIMPLE(
+	/* name */	panic_buffer,
+	/* type */	bytes_u,
+	/* min */	"4k",
+	/* max */	"10m",
+	/* def */	"64k",
+	/* units */	"bytes",
+	/* descr */
+	"Size of the panic message buffer.\n"
+	"The panic buffer is allocated in the working directory as memory "
+	"shared between the management and worker process, so sufficient "
+	"working directory space should be accounted for if this value is "
+	"adjusted. Panic messages are truncated to the configured size.",
+	/* flags */	MUST_RESTART
+)
+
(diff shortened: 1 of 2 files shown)


dridi (approved):
I would change the maximum from 1GB to 1MB, because 1GB seems excessive.

include/tbl/params.h
```diff
@@ -732,6 +732,18 @@ PARAM_SIMPLE(
...
 
+PARAM_SIMPLE(
+	/* name */	panic_buffer,
+	/* type */	bytes_u,
+	/* min */	"4k",
+	/* max */	"1g",
+	/* def */	"64k",
+	/* units */	"bytes",
+	/* descr */
+	"Size of the panic message buffer.",
```
asadsa92, line 743: Maybe add a note here that enough space where shared memory is stored is required.
I am not sure about the wording, but a small note here would help.

asadsa92 (approved):
LGTM, minor comment below

nigoroll:
@dridi 

> I would change the maximum from 1GB to 1MB, because 1GB seems excessive.

You mean, 1MB ought to be enough for everyone?

I would prefer a limit which is waaaay out. At first I thought it could be 1TB, but we _really_ should have a _lot_ of headroom.

@asadsa92 thank you for your suggestions, do you see them reflected in the SQUASHME commit?

asadsa92 (approved):
Thanks, the docs looks good now.

bsdphk:
Ok with me.

dridi:
> You mean, 1MB ought to be enough for everyone?

YES!

I could be convinced to go one order of magnitude higher, but 10MB sure ought to really be enough for everyone. You can fit a couple holy books in 10MB, and I don't expect anyone to read through a book-worth of panic output.

PR #3977, opened 2023-09-07
varnishd: add lock duration

asadsa92:
Our instrumented locks lacks a way to tell how long a lock was held before released. Just looking at the number of time the lock was taken is not enough. Some locks are more expensive than others (held for longer).

This PR adds duration to each instrumented lock given in microsecond resolution.

Example output:
```
LCK.ban.creat                                 1         0.01 Created locks
LCK.ban.destroy                               0         0.00 Destroyed locks
LCK.ban.locks                                12         0.12 Lock Operations
LCK.ban.dur_us                               14         0.14 Lock duration
LCK.ban.dbg_busy                              0         0.00 Contended lock operations
LCK.ban.dbg_try_fail                          0         0.00 Contended trylock operations
LCK.busyobj.creat                             0         0.00 Created locks
LCK.busyobj.destroy                           0         0.00 Destroyed locks
LCK.busyobj.locks                             0         0.00 Lock Operations
LCK.busyobj.dur_us                            0         0.00 Lock duration
LCK.busyobj.dbg_busy                          0         0.00 Contended lock operations
LCK.busyobj.dbg_try_fail                      0         0.00 Contended trylock operations
LCK.cli.creat                                 1         0.01 Created locks
LCK.cli.destroy                               0         0.00 Destroyed locks
LCK.cli.locks                                46         0.46 Lock Operations
LCK.cli.dur_us                             1393        13.79 Lock duration
LCK.cli.dbg_busy                              0         0.00 Contended lock operations
LCK.cli.dbg_try_fail                          0         0.00 Contended trylock operations
LCK.director.creat                            1         0.01 Created locks
LCK.director.destroy                          0         0.00 Destroyed locks
LCK.director.locks                            0         0.00 Lock Operations
LCK.director.dur_us                           0         0.00 Lock duration
LCK.director.dbg_busy                         0         0.00 Contended lock operations
LCK.director.dbg_try_fail                     0         0.00 Contended trylock operations
LCK.exp.creat                                 1         0.01 Created locks
LCK.exp.destroy                               0         0.00 Destroyed locks
LCK.exp.locks                                34         0.34 Lock Operations
LCK.exp.dur_us                              282         2.79 Lock duration
LCK.exp.dbg_busy                              0         0.00 Contended lock operations
LCK.exp.dbg_try_fail                          0         0.00 Contended trylock operations
LCK.hcb.creat                                 1         0.01 Created locks
LCK.hcb.destroy                               0         0.00 Destroyed locks
[...]

PR #4133, opened 2024-07-08
Vsync proposal

scance:
This proposes an API to abstract pthread synchronization devices (mutex & conditions) as part of libvarnish making it usable outside of varnishd while offering the ability to leverage varnishd VSC instrumentation.

### DIFF
--- a/lib/libvsc/VSC_cond.vsc
+++ b/lib/libvsc/VSC_cond.vsc
@@ -1,0 +1,47 @@
+..
+	This is *NOT* a RST file but the syntax has been chosen so
+	that it may become an RST file at some later date.
+
+.. varnish_vsc_begin::	cond
+	:oneliner:	Condition Counters
+	:order:		70
+
+	Counters which track the activity in the different classes
+	of condition variables.
+
+.. varnish_vsc:: init
+	:type:	counter
+	:level:	debug
+	:oneliner:	Initialized condition variables.
+
+.. varnish_vsc:: fini
+	:type:	counter
+	:level:	debug
+	:oneliner:	Uninitialized condition variables.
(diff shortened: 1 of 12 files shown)


lib/libvsc/VSC_cond.vsc
```diff
@@ -0,0 +1,47 @@
...
+
+.. varnish_vsc:: waiting
+    :type:	gauge
+	:level:	debug
+	:oneliner:	Number of currently waiting thread on condition variables.
+
+.. varnish_vsc:: wait_duration_ns
+	:type:	counter
+	:level:	debug
+	:oneliner:	Amount of nanoseconds spent waiting of condition variables.
```
asadsa92, line 45: fyi, an attempt at this was rejected last time around: https://github.com/varnishcache/varnish-cache/pull/3977
Not sure if adding this to non-core locks is OK.

scance: This amount of time is an amount of time spent waiting on a condition, it is not related to locks per se. I would therefore argue that it is not the same. Isn't the change you mention measures the amount of time in a critical section ?

asadsa92: It was both, time spent waiting for the lock or condvar.
As I said, it could be OK outside of the core. I just mentioned it as it was closely related.

scance:
The test failure seems to come from a flaky test (I am not reproducing the issue nor do I see how it could relate to this PR).

lib/libvsc/VSC_cond.vsc
scance: alright :+1:

PR #4134, opened 2024-07-15
Backend: Wakeup clients on waitqueue when BE goes sick

walid-git:
This is a followup for #4030, it adds an optimization to purge the backend wait queue when the backend becomes sick.

Three possible transitions trigger this:
- Admin health is set to sick.
- Probe health goes sick when admin health is auto.
- Admin health is set to auto and current probe state is sick.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -139,6 +139,37 @@ VBE_Connect_Error(struct VSC_vbe *vsc, int err)
 
 /*--------------------------------------------------------------------*/
 
+int
+VBE_is_ah_auto (const struct backend *bp)
+{
+	CHECK_OBJ_NOTNULL(bp, BACKEND_MAGIC);
+	return (bp->director->vdir->admin_health == VDI_AH_AUTO);
+}
+
+void
+VBE_connwait_signal_all(const struct backend *bp)
+{
+	struct connwait *cw;
+	unsigned wait_limit;
+	vtim_dur wait_tmod;
+
+	CHECK_OBJ_NOTNULL(bp, BACKEND_MAGIC);
+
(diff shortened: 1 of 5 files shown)


walid-git:
Thanks for the review. I pushed a new version without the header pollution. I also added phk's suggestion of only taking the lock if backend queueing is enabled for that backend.

nigoroll (approved):
looks ok

bin/varnishd/cache/cache_backend.c
```diff
@@ -139,6 +139,37 @@ VBE_Connect_Error(struct VSC_vbe *vsc, int err)
...
+	unsigned wait_limit;
+	vtim_dur wait_tmod;
+
+	CHECK_OBJ_NOTNULL(bp, BACKEND_MAGIC);
+
+	FIND_BE_PARAM(backend_wait_limit, wait_limit, bp);
+	FIND_BE_TMO(backend_wait_timeout, wait_tmod, bp);
+
+	if (wait_limit <= 0 || wait_tmod <= 0)
+		return;
```
dridi, line 162: Do we really need to bail out? What if the global parameters were used and reset while clients entered the queue?

dridi: For reference: https://github.com/varnishcache/varnish-cache/pull/4182

dridi: For reference: 57e660bd46d141ffcf064f6162daac6405fbb515

nigoroll: FTR: Rightly changed in 57e660bd46d141ffcf064f6162daac6405fbb515

We could add an unlocked check for the queue being empty if we accepted the chance of dequeue events being missed occasionally.

PR #4056, opened 2024-02-14
Move decision on fetch_chunksize to the storage engine

nigoroll:
For chunked encoding, we do not know how big the object is ultimately going to be, so `VFP_GetStorage()` called `ObjGetSpace()` with the `fetch_chunksize` parameter in this case.

Yet which size is best might differ for different storage engines, and having the information that the caller does not know the final size might be relevant. Storage engines could guess that if a request came in for `fetch_chunksize` that this _might_ be the "chunked" case, but that heuristic would be wrong for Objects of just that size advertised via `Content-Length`.

So this patch takes the guesswork out of the game by just passing the magic 0 value down to the storage engine to mean "give me some good chunk of bytes, I do not know how much I am going to need".

---

For context: Over the past two years, this particular aspect has given me a hard time again and again, it was the main driver for an [unnecessarily complicated heuristic](https://gitlab.com/uplex/varnish/slash/-/blob/c57d7879a4ec3509edadd967411757abede9fbf3/src/fellow_cache.c#L5116-5134) and is related to [another issue](https://gitlab.com/uplex/varnish/slash/-/issues/60) which I just fixed.

PR #4135, opened 2024-07-17
Move implementation of transit_buffer to the storage engine

nigoroll:
By the same argument as #4056, storage engines might be able to make a better decision on the layout of object bodies if they know the expected length upfront, thus the transit buffer limit should be implemented at the storage layer.
This is also in line with the "other end" of the transit buffer, the "free behind" from the client side, being implemented in the object iterator.


nigoroll:
FTR: During bugwash, @bsdphk also noticed that the previous code for transit buffer did not limit `fetch_chunksize

PR #4136, opened 2024-07-19
-a port ranges

AlveElde:
This PR extends the syntax of the varnishd `-a` argument to support port ranges. The range is inclusive, and failing to bind to any port in the range will terminate startup. This change should not affect any existing setup.

### DIFF
--- a/lib/libvarnish/vss.c
+++ b/lib/libvarnish/vss.c
@@ -106,34 +106,31 @@ vss_parse_range(char *str, char **addr, char **port, unsigned long *lo,
     unsigned long *hi)
 {
 	const char *errp;
-	char *hyphen, *end;
+	char *end;
 	unsigned long l, h;
 
 	errp = vss_parse(str, addr, port);
-	if (*port == NULL)
+	if (errp != NULL)
 		return (errp);
-
-	hyphen = strchr(*port, '-');
-	if (hyphen == NULL)
+	if (*port == NULL || **port == '-')
 		return (NULL);
-	if (**port == '-')
-		return ("Range start missing");
 
 	l = strtoul(*port, &end, 10);
-	if (end != hyphen)
-		return ("Range start invalid");
-	if (strchr(hyphen + 1, '-') != NULL)
-		return ("Only one hyphen allowed");
+	if (end[0] != '-' || end[1] == '\0')
+		return (NULL);
+	if (strchr(end + 1, '-') != NULL)
+		return (NULL);
+	h = strtoul(end + 1, &end, 10);
+	if (end[0] != '\0')
+		return (NULL);
+
+	/* Port range of the form 80-81 */
 	if (l == 0)
 		return ("Range start cannot be 0");
-	if (hyphen[1] == '\0')
-		return ("Range end missing");
-
-	h = strtoul(hyphen + 1, &end, 10);
-	if (*end != '\0' || h > 65535)
-		return ("Range end invalid");
 	if (h < l)
 		return ("Range start higher than range end");
+	if (h > 65535)
+		return ("Range end higher than 65535");
 
 	*lo = l;
 	*hi = h;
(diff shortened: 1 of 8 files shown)


lib/libvarnish/vss.c
```diff
@@ -195,6 +235,44 @@ VSS_resolver(const char *addr, const char *def_port, vss_resolved_f *func,
...
+		return (VSS_resolver(addr, def_port, func, priv, errp));
+	}
+
+	for (i = lo; i <= hi && !error; i++) {
+		/* pp points to the first character of the range definition.
+		 * The range definition includes the biggest port number, so the
+		 * buffer must be big enough to fit each number individually.
+		 */
+		sprintf(pp, "%lu", i);
+		error = VSS_resolver(hp, pp, func, priv, errp);
```
walid-git, line 270: `hp` here might be `NULL` when `addr` is for example `:5555-5559`, which results in a panic in `VSS_resolver_socktype` (called by `VSS_resolver`) that expects `addr` to never be `NULL`

walid-git (changes requested):
I don't know if we want to accept port ranges on wildcard addresses, I would say yes, but if this is not the case, the error should still be handled more gracefully.

lib/libvarnish/vss.c
```diff
@@ -100,6 +101,45 @@ vss_parse(char *str, char **addr, char **port)
...
+
+	errp = vss_parse(str, addr, port);
+	if (*port == NULL)
+		return (errp);
+
+	hyphen = strchr(*port, '-');
+	if (hyphen == NULL)
+		return (NULL);
+	if (**port == '-')
+		return ("Range start missing");
```
walid-git, line 120: While looking at this PR I found another issue (unrelated to this PR): We never output the error messages, we always get the same generic `"Got no socket(s) for .."` message for any type of failures. We shoud probably output `err` here: https://github.com/varnishcache/varnish-cache/blob/32850ee59cdfd6e35b6ca64367c5f6e05fbac8d2/bin/varnishd/mgt/mgt_acceptor.c#L377

lib/libvarnish/vss.c
AlveElde: Good catch,  I added a fix and test coverage.

I think it makes sense to support port ranges without addresses

lib/libvarnish/vss.c
AlveElde: Added a commit to include the error reason

walid-git (approved):
LGTM !

nigoroll:
bugwash: please limit the parser to numericals for x-y

AlveElde:
The parser now considers a range to be exactly two integers separated by a single dash, anything else falls back to  `VSS_resolver()`.

lib/libvarnish/vss.c
```diff
@@ -100,6 +101,42 @@ vss_parse(char *str, char **addr, char **port)
...
+	errp = vss_parse(str, addr, port);
+	if (errp != NULL)
+		return (errp);
+	if (*port == NULL || **port == '-')
+		return (NULL);
+
+	l = strtoul(*port, &end, 10);
+	if (end[0] != '-' || end[1] == '\0')
+		return (NULL);
+	if (strchr(end + 1, '-') != NULL)
```
nigoroll, line 121: why do we special case a second hyphen check here? Should the `end[0] != '\0'` not suffice?

nigoroll (approved):

lib/libvarnish/vss.c
AlveElde: That is because `strtoul()` "helpfully" interprets `-81` as a negative number, so a port range defined as `80--81` would result in a `Range end higher than 65535` error message. This line should guard against any attempt at making a negative numbers, even if there is whitespace inbetween (`strtoul()` also ignores leading whitespace).

nigoroll: that is a very good reason, thank you!

gquintard:
@AlveElde, ok to merge?

AlveElde:
Ok with me!

PR #4137, opened 2024-07-23
builtin.vcl: Delete Content-Range header where it has no meaning

nigoroll:
Since 4ab110047130e3a89936104d74f4729b650676f9 (see also #3246), we fail backend responses with an unexpected `Content-Range`.

Yet [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-14.4) states:

> The Content-Range header field has no meaning for status codes that do not explicitly describe its semantic.

The semantic is described for status codes 206 and 416, so the obvious implementation change would be for core code to only consider `Content-Range` for these status codes.

But there might be scenarios where a stricter-than-RFC check is intended, so we keep that in core code and change `builtin.vcl` to remove the header where it has no semantic.

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -210,6 +210,7 @@ sub vcl_backend_response {
 }
 
 sub vcl_builtin_backend_response {
+	call vcl_beresp_range;
 	if (bereq.uncacheable) {
 		return (deliver);
 	}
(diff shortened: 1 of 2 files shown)


bin/varnishd/builtin.vcl
```diff
@@ -245,6 +246,12 @@ sub vcl_beresp_vary {
 	}
 }
 
+sub vcl_beresp_range {
+	if (beresp.status != 206 && beresp.status != 416) {
```
nigoroll, line 250: bugwash: comment ponting to rfc here

PR #4138, opened 2024-07-23
waiter: report WAITER_REMCLOSE correctly

asadsa92:
There seems to be a discrepancy between the epoll and kqueue
implementation. One implementation is not reporting WAITER_REMCLOSE,
while the other implementation is reporting WAITER_REMCLOSE too soon.

epoll:
We sometimes incorrectly report WAITER_ACTION when it is supposed to be
WAITER_REMCLOSE. We can check EPOLLRDHUP before reading 1 byte with MSG_PEEK to
detect WAITER_REMCLOSE. This saves us from a pontential session detour from the
Waiter to a worker thread.

kqueue:
We incorrectly report WAITER_REMCLOSE when there could be more data to be
read from the socket. Change this so that we perform read of 1 byte with
MSG_PEEK to detect closure when EV_EOF is set.

DESCRIPTION:EVFILT_READ:Sockets:

> It is possible for EOF to be  returned (indicating the connection is gone)
> while there is still data pending in the socket buffer.

- [kqueue (2)](https://man.freebsd.org/cgi/man.cgi?kqueue)

### DIFF
--- a/bin/varnishd/cache/cache_session.c
+++ b/bin/varnishd/cache/cache_session.c
@@ -712,7 +712,8 @@ SES_NewPool(struct pool *pp, unsigned pool_no)
 	pp->mpl_sess = MPL_New(nb, &cache_param->pool_sess,
 	    &cache_param->workspace_session);
 
-	pp->waiter = Waiter_New();
+	bprintf(nb, "pool%u", pool_no);
+	pp->waiter = Waiter_New(nb);
 }
 
 void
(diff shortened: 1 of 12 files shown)


nigoroll (approved):

asadsa92:
intwash: Ok to merge. But, we should keep an eye on it w.r.t regression.

asadsa92:
Rebased to solve a filename conflict for one of the test case.
Should be ready to merge after the CCI passes.

PR #4141, opened 2024-07-25
Introduce a Linux jail

cartoush:
The idea for this PR originates [here](https://github.com/varnishcache/varnish-cache/pull/4123)
It introduces a Linux jail which overloads the UNIX one
It currently only re enables core dumps after playing with uid/gid and outputs some logs when the working directory is not on a tmpfs partition

### DIFF
--- a/bin/varnishd/mgt/mgt_jail_linux.c
+++ b/bin/varnishd/mgt/mgt_jail_linux.c
@@ -35,6 +35,7 @@
 
 #include <fcntl.h>
 #include <grp.h>
+#include <linux/magic.h>
 #include <pwd.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -42,6 +43,7 @@
 #include <unistd.h>
 #include <sys/prctl.h>
 #include <sys/stat.h>
+#include <sys/vfs.h>
 
 #include "mgt/mgt.h"
 #include "common/heritage.h"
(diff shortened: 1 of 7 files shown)


bin/varnishd/mgt/mgt_jail_unix.c
```diff
@@ -229,17 +229,6 @@ vju_subproc(enum jail_subproc_e jse)
 	} else {
 		AZ(setuid(vju_uid));
 	}
-
-#ifdef __linux__
```
walid-git, line 233: Can you please remove this one too: https://github.com/varnishcache/varnish-cache/pull/4141/commits/31ef342958ebc8d62845818b2903d037d9edee07#diff-c66495bf983bac2ecc16d0b634508370d82a479bff35ab7ba642ea8fd18bfa66R47-R49 I don't think it is needed anymore.

bin/varnishd/mgt/mgt_jail_linux.c
```diff
@@ -71,7 +73,26 @@ static int vjl_make_subdir(const char *dname, const char *what, struct vsb *vsb)
...
+			VSB_printf(vsb, "Could not stat working directory '%s': %s (%d)\n", dname, VAS_errtxt(errno), errno);
+		else
+			MGT_Complain(C_ERR, "Could not stat working directory '%s': %s (%d)", dname, VAS_errtxt(errno), errno);
+		return (1);
+	}
+	if (info.f_type != TMPFS_MAGIC) {
+		if (vsb != NULL)
+			VSB_printf(vsb, "Working directory not mounted on tmpfs partition\n");
+		else
+			MGT_Complain(C_INFO, "Working directory not mounted on tmpfs partition");
```
walid-git, line 92: Should we add a "warning" log level to `MGT_Complain` for this, instead of using info ? (3f877ea925c71c563345fc8ac9940c18c98f892e)

cartoush: I think it could be nice indeed, I was surprised there was no warning log level

nigoroll:
bugwash: OK, should also warn about failing vsm mlock

bsdphk:
I'm OK with this.

Thanks for taking the ball and running with it.

bin/varnishd/mgt/mgt_jail_linux.c
walid-git: Bugwash: keep C_INFO

Issue #4140, opened 2024-07-25
deadlock (resulting in big quit) from vcl_BackendEvent() racing vcldir_retire()


nigoroll:
### Expected Behavior

.

### Current Behavior

seen while trying to reproduce https://github.com/nigoroll/libvmod-dynamic/issues/117:

This is similar to [the problem](https://github.com/nigoroll/libvmod-dynamic/issues/110) fixed by #4048, but here the situation is not to add a director, but rather to remove it:

A director thread (updating backends) is racing a CLI cold event which has a `pthread_join()` for the director thread. It never completes because the thread is waiting for the `vcl_mtx`, which the CLI event holds because it is returning the last director reference.

```
Thread 218 (Thread 0x7f84bb7ff6c0 (LWP 462388)):
#0  futex_wait (private=0, expected=2, futex_word=0x7f84cf422b88) at ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait (futex=futex@entry=0x7f84cf422b88, private=0) at ./nptl/lowlevellock.c:49
#2  0x00007f84cfcab2d2 in lll_mutex_lock_optimized (mutex=0x7f84cf422b88) at ./nptl/pthread_mutex_lock.c:48
#3  ___pthread_mutex_lock (mutex=0x7f84cf422b88) at ./nptl/pthread_mutex_lock.c:93
#4  0x0000563f69cb8f64 in Lck__Lock (lck=0x563f69e2f908 <vcl_mtx>, p=0x563f69dbf89a "vcldir_retire", l=279) at cache/cache_lck.c:124
#5  0x0000563f69cfa71a in vcldir_retire (vdir=0x7f84bdc53f80) at cache/cache_vrt_vcl.c:279
#6  0x0000563f69cf5df3 in vcldir_deref (vdir=0x7f84bdc53f80) at cache/cache_vrt_vcl.c:305
#7  0x0000563f69cf5fbe in VRT_Assign_Backend (dst=0x7f84bdc14af0, src=0x0) at cache/cache_vrt_vcl.c:343
#8  0x00007f84c2fa81f0 in ref_del (ctx=<optimized out>, r=0x7f84bdc14ad0) at vmod_dynamic.c:428
#9  0x00007f84c2faa176 in dom_update (now=1721895825.5200291, priv=<optimized out>, res=0x7f84c2fbaba0 <res_getdns>, dom=<optimized out>) at vmod_dynamic.c:74
2
[...]

PR #4142, opened 2024-07-26
Add a facility for unlocked director iteration

nigoroll:
Pondering #4140 made me realize that our central per-vcl director list constitutes a classic case of iterating a mutable list. That besides the already known fact that running potentially expensive iterator functions while holding the vcl_mtx is a very bad idea.

So this patch is a suggestion on how to get out of this situation. It does not go all the way (it still leaves unsolved a similar problem of iterating over all backends of _all vcls_), but shows an attempt on how to tackle the "iterate over one VCL's backends".

We add a fittingly named 'vdire' facility which manages the director list and, in particular, director retirement. The main change is that, while iterators are active on the director list, vcl_mtx is _not_ held and any request of a director to retire only ends up in a resignation, which manifests by this director being put on a second list.

When all iterators have completed, the resignation list is worked and the actual retirement carried out.

<s>NOTE: I am particularly _not_ happy about the vdire_(start|end)_iter() interface being exposed. What I would like to see is a FOREACH macro which does both the start and end. Having to use an iterator with a callback is clumsy, and having to directly call start/end with a VTAILQ_FOREACH feels like crossing interfaces.</s> See https://github.com/varnishcache/varnish-cache/pull/4142#issuecomment-2262299121

### DIFF
--- a/bin/varnishd/cache/cache_director.h
+++ b/bin/varnishd/cache/cache_director.h
@@ -43,7 +43,8 @@ struct vcldir {
 	struct director			*dir;
 	struct vcl			*vcl;
 	const struct vdi_methods	*methods;
-	VTAILQ_ENTRY(vcldir)		list;
+	VTAILQ_ENTRY(vcldir)		directors_list;
+	VTAILQ_ENTRY(vcldir)		resigning_list;
 	const struct vdi_ahealth	*admin_health;
 	vtim_real			health_changed;
 	char				*cli_name;
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -387,6 +387,7 @@ vdire_new(struct lock *mtx, const struct vcltemp **tempp)
 	VTAILQ_INIT(&vdire->directors);
 	VTAILQ_INIT(&vdire->resigning);
 	vdire->mtx = mtx;
+	PTOK(pthread_cond_init(&vdire->cond, NULL));
 	vdire->tempp = tempp;
 	return (vdire);
 }
@@ -402,6 +403,8 @@ vdire_start_iter(struct vdire *vdire)
 	ASSERT_CLI();
 
 	Lck_Lock(vdire->mtx);
+	while (! VTAILQ_EMPTY(&vdire->resigning))
+		(void)Lck_CondWait(&vdire->cond, vdire->mtx);
 	vdire->iterating++;
 	Lck_Unlock(vdire->mtx);
 }
@@ -425,6 +428,7 @@ vdire_end_iter(struct vdire *vdire)
 		VTAILQ_FOREACH(vdir, &resigning, resigning_list)
 			VTAILQ_REMOVE(&vdire->directors, vdir, directors_list);
 		temp = *vdire->tempp;
+		PTOK(pthread_cond_broadcast(&vdire->cond));
 	}
 	Lck_Unlock(vdire->mtx);
 
@@ -582,6 +586,14 @@ vcl_KillBackends(const struct vcl *vcl)
 	CHECK_OBJ_NOTNULL(vdire, VDIRE_MAGIC);
 
 	/*
+	 * ensure all retirement has finished (vdire_start_iter waits for it)
+	 */
+	vdire_start_iter(vdire);
+	vdire_end_iter(vdire);
+	AZ(vdire->iterating);
+	assert(VTAILQ_EMPTY(&vdire->resigning));
+
+	/*
 	 * Unlocked and sidelining vdire because no further directors can be added, and the
 	 * remaining ones need to be able to remove themselves.
 	 */
@@ -658,6 +670,7 @@ VCL_Close(struct vcl **vclp)
 	TAKE_OBJ_NOTNULL(vcl, vclp, VCL_MAGIC);
 	assert(VTAILQ_EMPTY(&vcl->filters));
 	AZ(dlclose(vcl->dlh));
+	PTOK(pthread_cond_destroy(&vcl->vdire->cond));
 	FREE_OBJ(vcl->vdire);
 	FREE_OBJ(vcl);
 }
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -368,8 +368,126 @@ VCL_Update(struct vcl **vcc, struct vcl *vcl)
 	assert((*vcc)->temp->is_warm);
 }
 
+/*--------------------------------------------------------------------
+ * vdire: Vcl DIrector REsignation Management (born out of a dire situation)
+ * iterators over the director list need to register.
+ * while iterating, directors can not retire immediately,
+ * they get put on a list of resigning directors. The
+ * last iterator executes the retirement.
+ */
+
+static struct vdire *
+vdire_new(struct lock *mtx, const struct vcltemp **tempp)
+{
+	struct vdire *vdire;
+
+	ALLOC_OBJ(vdire, VDIRE_MAGIC);
+	AN(vdire);
+	AN(mtx);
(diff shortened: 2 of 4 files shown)


nigoroll:
After rebasing on the newly understood 4e2c50da9b4b66a39191dcc7ea11bc4e3d11ab6c, I have successfully run my stress test on this patch without any issues.

nigoroll:
>  What I would like to see is a FOREACH macro which does both the start and end

Here's a macro which does that:

```c
/*
 * this macro might look a bit daunting, but it is really just two things:
 *
 * - the outer for loop is only to declare the local "next" variable
 *   and execute the inner for loop once
 *
 * - the inner for loop is VTAILQ_FOREACH_SAFE with calls to
 *   vdire_{start,end}_iter() added at the right places:
 *   start is called at initialization and end when var becomes NULL
 *   and the loop ends
 *
 * problem: none of break, goto or return must be used from the loop body
*/

#define _VDIRE_FOREACH(var, vdire, next)				\
	for (struct vcldir *next = (void *)&next;			\
	     next == (void *)&next;					\
	     next = NULL)						\
		for (vdire_start_iter(vdire),				\
		     (var) = VTAILQ_FIRST(&(vdire)->directors);		\
		     ((var) || (vdire_end_iter(vdire), 0)) &&		\
		       ((next) = VTAILQ_NEXT(var, directors_list), 1);	\
		     (var) = (next))

#define VDIRE_FOREACH(var, vdire)				\
	_VDIRE_FOREACH(var, vdire, VUNIQ_NAME(_vdirenext))

```

As stated in the comment, none of `break`, `goto` or `return` must be used from within the loop body. So I think I prefer the explicit `vdire_{start,end}_iter()`

nigoroll:
Putting this out of draft status despite not having worked on it any more, because it has received no feedback and I think it's actually important - now having seen one production case where this is at least a potential candidate.

nigoroll:
bugwash: good idea by phk: New iterators encountering retired backends should wait until they are cleared, otherwise a constant stream of iterators could prevent retirement. I do not think this is a realistic scenario at this point, but we might as well do this right from the start.

another feedback question is if we can simplify this patch with an r/w lock. I am sure that I did look at this for quite some time and came to the conclusion that it does not do much to help us, but I will revisit this question.

nigoroll:
> New iterators encountering retired backends should wait until they are cleared

done: 41c42a05cd364003f8ad7f4ef78f53c5ea9a35d2

> another feedback question is if we can simplify this patch with an r/w lock.

I do not see the simplification. With an rwlock, `vdire_resign()` would try to wrlock and, if this fails, would need to start a thread which would then wrlock and do the remove.

bsdphk:
I'm OK with this, as long as this iterator API is only for use by the CLI thread, and this should be documented in the code with an ``ASSERT_CLI()``  because allowing multiple thread to independently and concurrently iterate at the same time will require further  analysis.  (For instance, should we allow two different iterators to have their ``func`` called on the same director at the same time ?)

We should (probably) hand the resigned_list to a worker thread, rather than bog down the CLI thread with potentially many cleanup calls.

I think the iterator should skip directors which have already been retired, no point in the CLI processing them if they are already (potentially long since) gone.

Apart from that: OK with me.

nigoroll:
The next force-push has these changes (besides the rebase and typo fix):

### skip resigned backends

we can identify these by a zero refcount. [It is is already illegal to assign a director with a zero backend](https://github.com/varnishcache/varnish-cache/blob/c412418261f2a64e48b36c93536358bc53f00ef7/bin/varnishd/cache/cache_vrt_vcl.c#L349).

```diff
diff --git a/bin/varnishd/cache/cache_vcl.c b/bin/varnishd/cache/cache_vcl.c
index 12ed136a7..918537e50 100644
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -469,6 +469,8 @@ vdire_iter(struct cli *cli, const char *pat, const struct vcl *vcl,
 
        vdire_start_iter(vdire);
        VTAILQ_FOREACH(vdir, &vdire->directors, directors_list) {
+               if (vdir->refcnt == 0)
+                       continue;
                if (pat != NULL && fnmatch(pat, vdir->cli_name, 0))
                        continue;
                found++;
```

this is done only for `vdire_iter()` and not `vcl_BackendEvent()`, because `vcldir_retire()` relies on prior temperature events being delivered.

### `ASSERT_CLI()`

... is added. I expect this to be revisited at some point.

### Deferred call to `vcldir_retire()`:

I have actually written the code and reflected on it: I am _not_ going to commit this change because it will open another can of racing worms. In particular, when we cool a VCL and unload a vmod, we need to be sure that all objects using code from that vmod are gone. Delaying the actual destruction adds further complications, and the goal here is to not introduce additional complexity, but stabilize the code by handling the complications already introduced with reference counting.

On the contrary, pondering this question, I have added this change out of prudence to ensure that before the final iteration over the directors list, we have finished all retirements:

```diff
diff --git a/bin/varnishd/cache/cache_vcl.c b/bin/varnishd/cache/cache_vcl.c
index aca2b5cab..2140bc664 100644
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -585,6 +585,14 @@ vcl_KillBackends(const struct vcl *vcl)
        vdire = vcl->vdire;
        CHECK_OBJ_NOTNULL(vdire, VDIRE_MAGIC);
 
+       /*
+        * ensure all retirement has finished (vdire_start_iter waits for it)
+        */
+       vdire_start_iter(vdire);
+       vdire_end_iter(vdire);
+       AZ(vdire->iterating);
+       assert(VTAILQ_EMPTY(&vdire->resigning));
+
        /*
         * Unlocked and sidelining vdire because no further directors can be added, and the
         * remaining ones need to be able to remove themselves.
```

walid-git (approved):
Looks good to me.

dridi:
Not sure how much attention it got, but after this was merged we had a complaint from Coverity Scan:

```
*** CID 1641631:    (USE_AFTER_FREE)
/bin/varnishd/cache/cache_vcl.c: 435 in vdire_end_iter()
429                             VTAILQ_REMOVE(&vdire->directors, vdir, directors_list);
430                     temp = *vdire->tempp;
431                     PTOK(pthread_cond_broadcast(&vdire->cond));
432             }
433             Lck_Unlock(vdire->mtx);
434     
>>>     CID 1641631:    (USE_AFTER_FREE)
>>>     Dereferencing freed pointer "vdir".
435             VTAILQ_FOREACH(vdir, &resigning, resigning_list)
436                     vcldir_retire(vdir, temp);
437     }
438     
439     // if there are no iterators, remove from directors and retire
440     // otherwise put on resigning list to work when iterators end
```

walid-git:
> Not sure how much attention it got, but after this was merged we had a complaint from Coverity Scan:
> 
> ```
> *** CID 1641631:    (USE_AFTER_FREE)
> /bin/varnishd/cache/cache_vcl.c: 435 in vdire_end_iter()
> 429                             VTAILQ_REMOVE(&vdire->directors, vdir, directors_list);
> 430                     temp = *vdire->tempp;
> 431                     PTOK(pthread_cond_broadcast(&vdire->cond));
> 432             }
> 433             Lck_Unlock(vdire->mtx);
> 434     
> >>>     CID 1641631:    (USE_AFTER_FREE)
> >>>     Dereferencing freed pointer "vdir".
> 435             VTAILQ_FOREACH(vdir, &resigning, resigning_list)
> 436                     vcldir_retire(vdir, temp);
> 437     }
> 438     
> 439     // if there are no iterators, remove from directors and retire
> 440     // otherwise put on resigning list to work when iterators end
> ```

Probably needs to use `VTAILQ_FOREACH_SAFE` instead ?

nigoroll:
-> c63bd508841d3781dc99dc3109b331fffc240021 thank you

PR #4143, opened 2024-07-26
Embed the vmod version string in the binary to make the version detectable

LaurentGoderre:


nigoroll:
dup #4077

nigoroll:
bugwash: this is a strict subset of #4077, we will continue there.

PR #4144, opened 2024-07-29
vcp: use RB-tree to keep track of connection pools

asadsa92:
The cost to find an existing connection in the pool can become expensive when the linked-list is long. Heavy use of dynamic backends could exhibit such behavior.


nigoroll (approved):
👍🏽 to merge and then
* add stats for how many calls to VCP_Ref vs. how many where the pool existed
* revisit whether or not we should pre-alloc the vcp

asadsa92:
intwash: Ok to merge.

PR #4145, opened 2024-07-31
Don't go the wrong way with request bodies

nigoroll:
This PR is to revert commits which I think are going in the wrong direction. The issues raised here have become apparent when rebasing #4035 where we actually do things with request bodies and, for the first time, add some more interesting test coverage.

So what is wrong with these changes?

`BS_TAKEN` and `BS_CACHED` are exactly _not_ properties of the request, but of the request body. It can be processed on the client side, or backend side, or both, so the backend side needs to know the status, and going through `bo->req` is a bad idea long term, because we might want to decouple client and backend request more. Needless to say this also adds to complexity, because besides the request body status, also the request flags would need to be checked.

`BS_TAKEN` is important, because a request body can be consumed entirely on the client side, or backend side, and will then not be available for either a retry or a restart.

`BS_CACHED` is important, because it means that, with #4035, the body already has gone through filters.

These changes are partly justified with a new `BS_TRAILERS` to be introduced, for which I see absolutely no reason.

Regarding not caching an empty body, I see no reason why to special case this.

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -141,11 +141,10 @@ Bereq_Rollback(VRT_CTX)
 	bo = ctx->bo;
 	CHECK_OBJ_NOTNULL(bo, BUSYOBJ_MAGIC);
 
-	if (bo->htc != NULL) {
-		assert(bo->htc->body_status != BS_TAKEN);
-		if (bo->htc->body_status != BS_NONE)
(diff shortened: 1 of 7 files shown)


nigoroll:
bugwash: I was asked if we should bring back the "body_avail" status in a different way, but I think going back and re-thinking is the better option. I saw too many bad regressions in the past 1-2 years and really think we should restart here.

nigoroll:
bugwash: revert, then discuss what's needed for the trailer PR

PR #4148, opened 2024-08-01
libvcc: Fix STRING/STRANDS comparison without STRING folding

### LINKED ISSUE #4146: Regression: all STRANDS are STRINGS again now
### Expected Behavior

STRANDS have been introduced to save unnecessary string copy operations

### Current Behavior

STRANDS are converted to STRING and back to STRANDS whenever a STRANDS return value is used as a STRANDS argument

Demonstration VCL:

```vcl
vcl 4.1;

import debug;

backend proforma none;

sub vcl_recv {
	set req.http.t = debug.return_strands(debug.return_strands("foo"));
}
```

since b77b709510f1e1ff4792ac9bc9c3836d1a5249ec the relevant piece of code compiles to

```vcl
      VRT_SetHdr(ctx, &VGC_HDR_REQ_t,
        0,
        TOSTRAND(VRT_STRANDS_string(ctx,
          Vmod_xyzzy_debug_Func.f_return_strands(ctx,
            TOSTRAND(VRT_STRANDS_string(ctx,
              Vmod_xyzzy_debug_Func.f_return_strands(ctx,
                TOSTRAND("foo")
              )
            ))
          )
        ))
      );
```

notice the `TOSTRAND(VRT_STRANDS_string(...))` pattern

With 64b233606905039acecc4be58b28a51f5713202c (the previous commit), the C code is as it should be:

```vcl
      VRT_SetHdr(ctx, &VGC_HDR_REQ_t,
        0,
        Vmod_xyzzy_debug_Func.f_return_strands(ctx,
          Vmod_xyzzy_debug_Func.f_return_strands(ctx,
            TOSTRAND("foo")
          )
        )
      );
```

### Possible Solution

Implement b77b709510f1e1ff4792ac9bc9c3836d1a5249ec differently

_No response_

### Context

Noticed in [vmod-j](https://gitlab.com/uplex/varnish/libvmod-j/-/commit/4cb841477452203ae81d567d3bbc0f3cdad98b0b)

### Varnish Cache version

_No response_

### Operating system

_No response_

_No response_

nigoroll:
I do not consider myself enough of an expert on VCC and would appreciate reviews by those who do.
Side note: STRING vs STRINGS vs. STRANDS is confusing

Fixes #4146

### DIFF
--- a/bin/varnishtest/tests/v00019.vtc
+++ b/bin/varnishtest/tests/v00019.vtc
@@ -72,9 +72,9 @@ varnish v1 -errvcl {Comparison of different types: STRING '==' INT} {
 varnish v1 -vcl {
 	import debug;
 	backend be none;
-	sub vcl_recv {
-		if ("string" == "string") {
-			return (fail("should compile"));
+	sub vcl_init {
+		if ("string" != "string") {
+			return (fail("should load"));
 		}
 	}
 } -start
@@ -82,9 +82,9 @@ varnish v1 -vcl {
 varnish v1 -vcl {
 	import debug;
 	backend be none;
-	sub vcl_recv {
-		if ("string" == debug.return_strands("string")) {
-			return (fail("should compile"));
+	sub vcl_init {
+		if ("string" != debug.return_strands("string")) {
+			return (fail("should load"));
 		}
 	}
 }
@@ -92,9 +92,9 @@ varnish v1 -vcl {
 varnish v1 -vcl {
 	import debug;
 	backend be none;
-	sub vcl_recv {
-		if (debug.return_strands("string") == "string") {
-			return (fail("should compile"));
+	sub vcl_init {
+		if (debug.return_strands("string") != "string") {
+			return (fail("should load"));
 		}
 	}
 }
@@ -102,9 +102,9 @@ varnish v1 -vcl {
 varnish v1 -vcl {
 	import debug;
 	backend be none;
-	sub vcl_recv {
-		if (debug.return_strands("string") == debug.return_strands("string")) {
-			return (fail("should compile"));
+	sub vcl_init {
+		if (debug.return_strands("string") != debug.return_strands("string")) {
+			return (fail("should load"));
 		}
 	}
 }
--- a/bin/varnishtest/tests/v00019.vtc
+++ b/bin/varnishtest/tests/v00019.vtc
@@ -109,6 +109,17 @@ varnish v1 -vcl {
 	}
 }
 
+# regression test #4146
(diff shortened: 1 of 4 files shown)


walid-git (changes requested):
Even though the changes generate syntactically correct C code and that the STRING/STRANDS comparison error is gone, there is still something wrong here:
With these changes, the following VCL code:
```
 if ("string" == debug.return_strands("string")) {
                    ...
        }
```
Will generate this C code:
```
 if (
        (0 == VRT_CompareStrands(TOSTRAND("string"), vrt_null_strands))
  ) {
  ...
```
The STRANDS part of `debug.return_strands("string")` is being translated to `vrt_null_strands` which is wrong. I need to look closer to find which part of the PR is causing this, but that might be related to the first commit. We should also improve test coverage to not only test the syntactic part but also the semantic.

nigoroll:
@walid-git thank you very much for catching this significant regression. I have force-pushed an update, the delta is

```diff
diff --git a/lib/libvcc/vcc_expr.c b/lib/libvcc/vcc_expr.c
index a348e9b83..cd4b03c2f 100644
--- a/lib/libvcc/vcc_expr.c
+++ b/lib/libvcc/vcc_expr.c
@@ -159,7 +159,11 @@ static void
 vcc_strands_edit(const struct expr *e1, const struct expr *e2)
 {
 
-       if (e2->nstr == 0)
+       assert(e2->fmt == STRANDS || e2->fmt == STRINGS);
+
+       if (e2->fmt == STRANDS)
+               VSB_cat(e1->vsb, VSB_data(e2->vsb));
+       else if (e2->nstr == 0)
                VSB_printf(e1->vsb, "vrt_null_strands");
        else if (e2->nstr == 1)
                VSB_printf(e1->vsb, "TOSTRAND(%s)", VSB_data(e2->vsb));
```

bin/varnishtest/tests/v00019.vtc
```diff
@@ -109,6 +109,17 @@ varnish v1 -vcl {
 	}
 }
 
+# regression test #4146
+varnish v1 -vcl {
+	import debug;
+	backend be none;
+	sub vcl_init {
```
walid-git, line 116: Now that we actually run `vcl_init` in these tests, can you please edit the existing tests where vcl compilation is expected to succeed (replace `vcl_recv` by `vcl_init` and adapt the conditions accordingly) ? This will ensure that the generated code is correct and produces expected results.

walid-git (approved):
Thanks @nigoroll. This looks good to me now. I left a comment for test coverage improvement.

PR #4155, opened 2024-08-12
Emit a warning if mlock() of VSM fails

nigoroll:
Trivial follow-up to #4141 : We should notify users if we can not `mlock()` VSM.

Related to #4121


nigoroll:
> Should we keep track of the amount of locked memory

Because VMODs/VEXTs might lock memory, I would expect this reporting to be inaccurate and thus useless. And wrapping all system calls which could potentially lock memory seems like a hopeless endeavor.

PR #4166, opened 2024-08-20
build: add --enable-optimization-warnings

asadsa92:
Start off by adding -Wpadded and -Wno-error=padded . This is useful when investigating holes in the struct layout. Alternatively, the pahole utility can be used for this.

Dridi suggested that I add this to a new class of warnings.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -708,6 +708,11 @@ if test $? -ne 0 ; then
   AC_MSG_ERROR([wflags.py failure])
 fi
 
+OPTIMIZATION_CFLAGS=" \
+	-Wpadded \
+	-Wno-error=padded \
+"
+
 # zlib-specific flags
 AC_SUBST(VGZ_CFLAGS)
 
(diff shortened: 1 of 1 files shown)


bsdphk:
I am not entirely on board with this, because there are valid reasons not want structs tightly packed, for instance cacheline considerations.

It also bothers me that the patch calls them "OPTIMIZATION_FLAGS", they're not.  At best they're "DEVELOPER_INFORMATION_FLAGS" or something of that sort.

asadsa92:
> I am not entirely on board with this, because there are valid reasons not want structs tightly packed, for instance cacheline considerations.

Yes, we can always ignore false-positives. The intend was not to enable this by default.
We could also place a reserved variable there to fill the hole and make it more explicit.
Then, in the future, we can use this space for something else rather than expanding the struct size.

> It also bothers me that the patch calls them "OPTIMIZATION_FLAGS", they're not. At best they're "DEVELOPER_INFORMATION_FLAGS" or something of that sort.

Sure, we can rename it.
I do not think it make sense to enable this through the default `./autogen.des` invocation, at least not before we have reviewed what we got first.

PR #4167, opened 2024-08-20
Add a WRONGF() macro for dynamic explanations of wrong situations

nigoroll:
`F` for format as in `printf()`. The output will be truncated to the fixed size buffer of 255 single byte characters.

The single example use case has been chosen arbitrarily for illustration only.


nigoroll:
Good feedback from bugwash: Compilers are free to move the stack allocation to the function scope, so this change could have the effect of enlarging every call's stack frame. In this light, a global buffer is the better idea - and we might also want to move the format string option to `VAS_Fail()` in the form of a `VAS_Failf()` function.

Issue #4169, opened 2024-09-05
varnishd command line syntax for pluggable acceptors?


nigoroll:
### Expected Behavior

Working on a continuation of #3976 the question arose which command like syntax we want for pluggable acceptors. I see two basic options:

* backwards compatible `-a`

```
  -a [name=][kind,][listen_address[,PROTO|,option=value,...]]
```

  here, _kind_ would be the acceptor kind, so a valid use case would be `-a foo=tcp,127.0.0.1:4444`

* new, additional argument, for example `-A`, where _kind_ is mandatory

```
  -a [name=]kind[,listen_address[,PROTO|,option=value,...]]
```

I like to avoid adding new options, but the first option comes with some ambiguity:

> 	foo,bar

could mean

>	kind=foo, listen_address=bar

or

>	listen_address=foo, PROTO=bar

and, because of backwards compatibility, the error message might be confusing:

```
$ /tmp/sbin/varnishd -a foo=xxx,@foo -b @A -n /tmp/t
Error: Unknown protocol '@foo'
```

This can be tried out here: https://github.com/nigoroll/varnish-cache/commit/b40fc7648ba4cafde0358cf4c46d486753421efd

### Current Behavior

.

### Context

.

nigoroll:
bugwash: phk is for not adding `-a` processing for VEXTs. He suggests a `varnish.cf` file which would allow for per-extension acceptors (which I appreciate!) and we would probably also need some `-Eext.so,-a=...,-a=...` or similar.

I still think that using -E to load extensions and then use a consistent `-a` and `-s` across all implementations is cleaner and would still prefer that.

nigoroll:
I've settled on suggesting

```
[name=][%<kind>,][listen_address[,PROTO|,option=value,...]]
```

now. Reasons:

* `%` is never the first character of a valid hostname for _listen_address_
* `%` is highly unlikely as the first character of a unix domain socket
* and if it is really intended as such, `%uds,%path` can be used

I did look at other alternatives, such as:
* `[name=][kind=<kind>,][listen_address[,PROTO|,option=value,...]]`  is ambiguous again
* `[name=][listen_address[,PROTO|,option=value,...]][,kind=<kind>]` is bad because the option part should belong to the implementation

The implementation has been added to #4172 because plugging an acceptor is the basis for the test of this option.

PR #4172, opened 2024-09-06
Load VEXTs in manager

nigoroll:
We foresaw this when we added VEXTs: while storage VEXTs have been happy with loading in the worker process, to support pluggable acceptors, VEXTs need to be present in the manager for opening privileged ports etc.

An implementation of a pluggable excceptor as a VMOD can be found in https://github.com/nigoroll/varnish-cache/commits/plugable_acceptors_rebase/, which is based upon #3976


simonvik:
I'm very much in favor for this change, this also allows us to define transports (like proxy) as VEXTs (without hacking too much)

bsdphk:
», VEXTs need to be present in the manager for opening privileged ports etc.«

This is simply not true.

VEXT's get a chance to bite before privileges are dropped right after the fork().

nigoroll:
bugwash: I apologized for my mistake on the privileges, I probably had too much in mind what the solaris jail does, and need to look again at the privilege separation stuff.
Other than that, if vexts can't be loaded before `-a` processing, they need a different way to configure

nigoroll:
> », VEXTs need to be present in the manager for opening privileged ports etc.«
> 
> This is simply not true.
> 
> VEXT's get a chance to bite before privileges are dropped right after the fork().

I had another look. The problem is not that `vext_load()` happens in the child, the problem is that the initial VCA configuration including `bind()` happens in mgt, and the natural approach (at least in my mind) is that a VCA VEXT should be configured like any other VCA.

To allow VEXT VCAs but no VEXTs in mgt, we would need to move the `VCA_*()` calls from `mgt` `main()` to the child. To move `VCA_Arg()`, mgt would need to pass the respective argvs to the child, the child would parse arguments and fatal error scenarios (basically `bind()` failures) would result in a `Child failed to launch` error.

It is my understanding that we have so far been operating under the principle of detecting configuration problems before even starting the child, in particular by test-loading the VCL.

Do we want to (at least partly) give up this principle?

One other example I see where loading VEXTs in mgt makes sense is a keyserver sidecar process. I think it should have a similar lifetime as mgt and be controlled by mgt. It needs to start with elevated privileges to load a wrapping key and then drop them.

What would be reasons to for keeping VEXTs in the child only?

nigoroll:
I have added a change to this PR which is somehow unrelated, but needs the ability to plug an acceptor. See https://github.com/varnishcache/varnish-cache/issues/4169#issuecomment-2453476501

nigoroll:
I am going to continue working with the code in this ticket due to lack of alternatives, but bugwash agreed that this PR will not go in because we want to keep VEXTs out of the mgt process.
We need an alternative concept, which @bsdphk wants to write up. Thank you!

nigoroll:
FTR: In light of the existence of `LD_PRELOAD`, does it really make sense to try to "prevent" loading code into mgt?

nigoroll:
I am going to cherry-pick two commits from here which are unrelated to the overall topic

PR #4173, opened 2024-09-12
WIP Refcatorings to address Coverity findings

nigoroll:
collecting WIP here, will improve description when done
need PR number for coverity triage


lib/libvarnishapi/vsl_cursor.c
```diff
@@ -335,18 +335,20 @@ vslc_file_delete(const struct VSL_cursor *cursor)
...
+		l = read(fd, buf, n);
 		if (l <= 0)
 			return (l);
-		t += l;
+		assert(l <= n);
+		buf += l;
+		n -= l;
 	}
-	return (t);
+	return (n);
```
dridi, line 351: Isn't this no longer returning the number of bytes read?

lib/libvarnishapi/vsl_cursor.c
```diff
@@ -335,18 +335,20 @@ vslc_file_delete(const struct VSL_cursor *cursor)
...
 	ssize_t l;
 
-	while (t < n) {
-		l = read(fd, (char *)buf + t, n - t);
+	while (n > 0) {
+		l = read(fd, buf, n);
 		if (l <= 0)
 			return (l);
-		t += l;
+		assert(l <= n);
```
dridi, line 347: This one was not introduced here, but shouldn't we treat this as a short read if `t` is greater than zero?

lib/libvarnishapi/vsl_cursor.c
nigoroll: I pushed too early and you were too quick (but I marked the PR as Draft). Yes, I noticed myself...

lib/libvarnishapi/vsl_cursor.c
nigoroll: IIRC we can have short reads with some file systems

nigoroll: You mean `return (pdiff(...))` also for the `l <= 0` case? I thought about it, but did not want to change this behavior in case it mattered.

nigoroll:
bugwash: mark false positive and move on

PR #4174, opened 2024-09-12
Disable Transparent Hugepage

cartoush:


dridi:
Varnish used to crash left and right when Transparent Hugepage first landed, and we sometime still see inexplicable problems go away once we disable the feature.

I'm fine with the addition of a jail option to control this, would you agree to disabled by default?

bsdphk:
As a matter of principle, I'll be +0 on how linux_jail works and what it does.

bin/varnishd/mgt/mgt_jail_linux.c
```diff
@@ -49,10 +49,60 @@
...
+static int
+vjl_init(char **args)
+{
+	char **uargs = NULL;
+	size_t i = 0;
+	int ret = 0;
+
+	if (args != NULL) {
+		for (; args[i] != NULL; i++);
+		uargs = malloc(sizeof(char*) * (i + 1));
```
cartoush, line 84: I'd be open for another solution in order to separate Linux jail's parameters from the Unix one, this one does not look pretty imho but I havent found another (simple and elegant) way to do it

dridi: You can pass them one by one with a 2-length array living on the stack to avoid keeping track of all the non-linux arguments.

bin/varnishd/mgt/mgt_jail_linux.c
```diff
@@ -49,10 +49,60 @@
...
+			    VAS_errtxt(errno), errno);
+		}
+	} else if (!strcmp(arg, "disable") ||
+		    !strcmp(arg, "default")) {
+		if (prctl(PR_SET_THP_DISABLE, 1, 0, 0, 0) != 0) {
+			MGT_Complain(C_INFO,
+			    "Could not disable Transparent Hugepages: %s (%d)",
+			    VAS_errtxt(errno), errno);
+		}
+	} else {
```
asadsa92, line 67: Please add a option to do nothing a.k.a NOP. I guess the "default" option could be that.
Make sure to print which selection has been selected, maybe also retrieve what option the kernel use?
You could use `PR_GET_THP_DISABLE` for that.

dridi: It's actually "ignore" that will do nothing. The "default" one was the result of a misunderstanding and it will go away.

bin/varnishd/mgt/mgt_jail_linux.c
```diff
@@ -48,11 +48,61 @@
...
+	}
 
-	return jail_tech_unix.init(args);
+	for (i = 0; args[i] != NULL; i++);
+	unix_args = calloc(i + 1, sizeof *unix_args);
+	AN(unix_args);
+
+	for (i = 0; *args != NULL && ret == 0; args++) {
+		if (!strncmp(*args, "transparent_hugepage=", 22)) {
+			ret = vjl_set_thp((*args) + 22);
```
dridi, line 95: The offset is 21, because that is half 42.

Also, this is a mouthful, maybe we should call the option `huge` (like the mount option) instead of the verbose `transparent_hugepage`.

Opinions?

doc/sphinx/reference/varnishd.rst
```diff
@@ -461,6 +461,7 @@ specific options. Available jails are:
 
   - It warns when *workdir* is not on a ``tmpfs``.
   - It tries to keep the process dumpable after dropping privileges.
+  - It tries to disable Transparent Hugepage in ``varnishd``
```
dridi, line 464: Instead of just adding a bullet point here, please also document the `transparent_hugepage` parameter below, like the UNIX jail documentation. In particular, document that failing to enable or disable THP is not a fatal error.

Also, please apply this change a few lines above:

```diff
--j <linux[,user=`user`][,ccgroup=`group`][,workuser=`user`]>
+-j <linux[,transparent_hugepage=`action`][,<unix jail option>...]>
```

dridi (changes requested):
LGTM modulus a few nit picks.

bin/varnishd/mgt/mgt_jail_linux.c
dridi: Actually no, you can't, the UNIX jail init must happen in a single call.

bin/varnishd/mgt/mgt_jail_linux.c
cartoush: I feel like having a more precise name like currently makes it less confusing than using simply `huge`

bin/varnishd/mgt/mgt_jail_linux.c
```diff
@@ -48,11 +48,65 @@
...
+		    arg);
+	}
+	if (prctl(PR_SET_THP_DISABLE, val, 0, 0, 0) != 0) {
+		MGT_Complain(C_ERR,
+		    "Could not %s Transparent Hugepage: %s (%d)",
+		    arg, VAS_errtxt(errno), errno);
+	} else {
+		MGT_Complain(C_INFO, "linux jail: Transparent Hugepage now %s",
+		    prctl(PR_GET_THP_DISABLE, 0, 0, 0, 0) ?
+		    "disabled" : "enabled");
```
dridi, line 74: Not sure we need to say anything upon success. If we really want to, then this format is probably enough:

```c
"linux jail: Transparent Hugepage %sd", arg
```

We tend to not log much when something succeeds (there are cases where we do) so I'm interested in @nigoroll's opinion on this one.

doc/sphinx/reference/varnishd.rst
```diff
@@ -454,13 +454,23 @@ specific options. Available jails are:
...
+
+  The optional `transparent_hugepage` argument specifies whether Transparent
+  Hugepage should be enabled or not, it accepts the following values:
+
+  - ``enable`` to enable Transparent Hugepage
+  - ``disable`` to disable Transparent Hugepage
+  - ``ignore`` to leave Transparent Hugepage as-is
+
+  Failure to set or disable THP is not considered a fatal error.
```
dridi, line 473: I think it's a little bit too much, this is probably enough:

> The optional `transparent_hugepage` argument takes one action among `enable`, `disable` (default) and `ignore`. Failure to configure Transparent Hugepage is not a fatal error.

doc/sphinx/reference/varnishd.rst
dridi: Maybe add " by default." at the end of the sentence. At the very least, please add a dot to be consistent with @nigoroll's style change for the bullet points above.

dridi (commented):
At this point, I'm only left with docs nitpicks.

bin/varnishd/mgt/mgt_jail_linux.c
cartoush: This was done following @asadsa92 's comment https://github.com/varnishcache/varnish-cache/pull/4174#discussion_r1767002337 I guess we'll need his opinion too then :)

asadsa92: I think there is a misunderstanding, I agree we do not need to print anything for the success case.
But, `PR_GET_THP_DISABLE` was meant to retrieve whatever the current value of THP is from the kernel.
If we set it, then I agree we do not need to print the value. The kernel already reported a success there.
However, the "ignore" case should retrieve and print the current value.

dridi: If you print the value when the configuration is `ignore` but fail to retrieve the value then you also have to prepare an error message for that.

I'm not seeing this being worth the trouble and I'd rather actually ignore THP entirely, and not care about its current state. We leave it as an exercise for the operator, but we let them know when they picked a state and it couldn't be applied.

bin/varnishd/mgt/mgt_jail_linux.c
```diff
@@ -48,11 +48,64 @@
...
+		return (jail_tech_unix.init(NULL));
+	}
 
-	return jail_tech_unix.init(args);
+	for (i = 0; args[i] != NULL; i++);
+	unix_args = calloc(i + 1, sizeof *unix_args);
+	AN(unix_args);
+
+	for (i = 0; *args != NULL && ret == 0; args++) {
+		if (!strncmp(*args, "transparent_hugepage=", 21)) {
```
nigoroll, line 97: nit: can we please at least replace the magic 21 with something like
```c
	const char const *thpa = "transparent_hugepage=";
	const size_t l = strlen(thpa);
```

doc/sphinx/installation/platformnotes.rst
```diff
@@ -11,21 +11,29 @@ On some platforms it is necessary to adjust the operating system before running
...
 
-Transparent hugepages on Redhat Enterprise Linux 6
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Transparent Hugepage on Linux
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-On RHEL6 Transparent Hugepage kernel support is enabled by default.
-This is known to cause sporadic crashes of Varnish.
+On certain Linux distributions Transparent Hugepage kernel support is enabled
+by default. This is known to cause instability of Varnish.
```
nigoroll, line 18: editorial: instabilities

nigoroll (commented):
I only have nitpicks, too

bin/varnishd/mgt/mgt_jail_linux.c
dridi: FYI this was drawing inspiration from the UNIX jail.

nigoroll: https://github.com/varnishcache/varnish-cache/commit/4001ea251bac6924c2e86bac3d0f2c8d93c96bd8
FTR, in the example above I got one `const` wrong.

cartoush: maybe `sizeof("transparent_hugepage=") - 1` would be appropriate and avoid calling `strlen` just for that ?

dridi: Or `sizeof("transparent_hugepage")` to swallow the equal sign :p

A call to `strlen()` for a static string will likely be optimized out at compile time, but if anything I'd rather work with a macro here and in the UNIX jail:

```c
if ((val = ARG_MATCH(*args, transparent_hugepage)) != NULL)
        ret = vjl_set_thp(val);
```

Less ceremony, and the macro is in charge of dealing with the equal sign and value offset.

cartoush: I've added that as a separate commit and applied it to all jails, I hope its fine like that with you

dridi: That was fast, but I only meant to make a suggestion here. I'm interested in @nigoroll's opinion on my suggestion, which would do more than the macro you pushed.

nigoroll: @dridi 

> I'd rather work with a macro here and in the UNIX jail:

can you show how the macro would look like in `vju_init`? I thought about it but could not see a nice way cause of the "error if vju_get* fails, else continue"

dridi: https://github.com/varnishcache/varnish-cache/pull/4202

nigoroll:
FTR: I ended up editing this change quite heavily. See the commit message for details.

PR #4179, opened 2024-09-13
Minor r'str' fix to vmodtool.py

nyurik:
Python does not like having `\` characters followed by non-special characters, e.g. `\.` - but regex requires that. For this, python has "raw strings", e.g. `r'....'` - this way no `\` have any special meaning to python, only to the regex itself.  At least on my 3.12 Python it refused to work without it.  In general, it is a good practice to always use r-strings with regex.


dridi (approved):
Not sure why it didn't show up when 60bb4e4ed661d21de0fcb0132ade509520ebb932 happened.

nigoroll (commented):
seems to make sense

Issue #4183, opened 2024-09-16
Panic on varnishadm backend.set_health of dynamic backend


delthas:
### Expected Behavior

No panic on the VTC.

### Current Behavior

Panic:
- `Assert error in VBE_connwait_signal_all(), cache/cache_backend.c line 156:`
- `Condition((bp)->magic == 0x64c4c7c6) not true.`
- `cls_exec -> mgt_launch_child -> child_main -> CLI_Run -> VCLS_Poll -> cls_exec -> cli_backend_set_health -> VCL_IterDirector -> vcl_iterdir -> do_set_health -> VBE_connwait_signal_all -> VAS_Fail -> pan_ic`

```vtc
varnishtest "panic"

server s1 {
} -start

varnish v1 -vcl {
vcl 4.1;
import std;
import directors;
import dynamic;

backend foo none;

sub vcl_init {
    new cluster_default = directors.fallback();
    new main = dynamic.director(port = {"${s1_port}"});
    cluster_default.add_backend(main.backend(host={"${s1_addr}"}, port={"${s1_port}"}));
}
} -start

# Panic:
# Assert error in VBE_connwait_signal_all(), cache/cache_backend.c line 156:
#  Condition((bp)->magic == 0x64c4c7c6) not true.
# cls_exec -> mgt_launch_child -> child_main -> CLI_Run -> VCLS_Poll -> cls_exec -> cli_backend_set_health -> VCL_IterDirector -> vcl_iterdir -> do_set_health -> VBE_connwait_signal_all -> VAS_Fail -> pan_ic
varnish v1 -cliok "backend.set_health main(*) sick"
```

Even though my VTC involes libvmod-dynamic, this is likely an issue in Varnish, because this broke between Varnish 7.5.0 and Varnish 7.6.0

I ran a git bisect and the first failing commit is: https://github.com/varnishcache/varnish-cache/commit/e46f97278f1036f4e648dd64d3f34ada8d29f64a ("unsurprisingly" this is the commit adding the function which has the assert).

### Context

I'm using dynamic backends inside directors.

nigoroll:
Thank you for the test case, this is reproducible immediately.

The problem here is that `VBE_connwait_signal_all()` gets called on a dynamic domain director. 

looking... 👀 

```c
[...]

PR #4184, opened 2024-09-16
WIP: fix connwait signalling vs. VDI/VBE layering

nigoroll:
the idea here is to use the backend's healthy callback to signal to it a health change from outside.

this fixes the regression from #4183, but makes v74.vtc fail. And also I am not happy about the locking involved and need to spend more time...


nigoroll:
No, this is a bad idea. `VRT_Healthy()` rightly shortcuts calling the `healthy` callback.

nigoroll:
No, this is a bad idea. `VRT_Healthy()` rightly shortcuts calling the `healthy` callback.

PR #4186, opened 2024-09-17
Add director notfiy method and use it for admin health changes

nigoroll:
(still need to polish commit messages)


dridi:
Your branch didn't start on the tip of master and conflicts with the change I both suggested and pushed after verifying my suspicion in a test case.

The "problem" with this change is the ABI breakage it introduces (the real problem being of course the bug you are trying to fix).

nigoroll:
> Your branch didn't start on the tip of master and conflicts with the change I both suggested and pushed after verifying my suspicion in a test case.

Yes, I noticed, will adjust.

> The "problem" with this change is the ABI breakage it introduces (the real problem being of course the bug you are trying to fix).
> 
> edit: it's probably fine to include as part of the 7.6 series with VRT 20.0, despite the 7.6.0 release

I think this change would call for a `VRT_MINOR_VERSION` bump only.

dridi:
> I think this change would call for a `VRT_MINOR_VERSION` bump only.

Absolutely not, this is changing the layout of `struct vdi_methods` and requires a rebuild!

nigoroll:
> Absolutely not, this is changing the layout of `struct vdi_methods` and requires a rebuild!

You are right. More specifically, a member is added, so `sizeof()` changes and without a rebuild, we would cause an out-of-bounds access. Thank you!

delthas:
This causes a panic on a simple VTC, see #4189

nigoroll:
> This causes a panic on a simple VTC, see #4189

fixed with https://github.com/varnishcache/varnish-cache/pull/4186/commits/57e2a6432a6230a6ce7151594154afe36c721337

nigoroll:
OKed during bugwash

bin/varnishd/cache/cache_backend.c
```diff
@@ -139,16 +139,8 @@ VBE_Connect_Error(struct VSC_vbe *vsc, int err)
...
-{
-
-	CHECK_OBJ_NOTNULL(bp, BACKEND_MAGIC);
-	return (bp->director->vdir->admin_health == VDI_AH_AUTO);
-}
-
-void
-VBE_connwait_signal_all(const struct backend *bp)
+static void
+vbe_connwait_signal_all(const struct backend *bp)
```
dridi, line 143: Suggestion: `vbe_connwait_broadcast()`

bin/varnishd/cache/cache_backend.c
```diff
@@ -690,7 +697,8 @@ static const struct vdi_methods vbe_methods[1] = {{
 	.destroy =		vbe_destroy,
 	.panic =		vbe_panic,
 	.list =			vbe_list,
-	.healthy =		vbe_healthy
+	.healthy =		vbe_healthy,
+	.notify =		vbe_notify
```
dridi, line 701: I was bothered by the notify method because this is a breaking change as we established earlier and I wasn't satisfied with such a resolution on the 7.6 branch where we'd rather not break VRT if we can help it.

It then occurred to me that in spirit the new notify method is very close to the event method and I was wondering whether we'd agree to the creation of a new going-sick event even though that was not the original scope of the event enum.

I don't personally see a strong reason against widening the scope beyond the existing set of events.

nigoroll: Is that yes, no or something else, @dridi ?

nigoroll: Ah, phk also brought up using the `.event` method, is this what you mean? I can look at it, but it did not seem natural to me.

dridi: I was suggesting adding a new enum entry for the event callback, one that would be dedicated to backend/directors, instead of adding a new notify callback.

nigoroll: FTR: I did consider using the .`event()` method, but it seemed confusing to me because it is tied to the VCL temperature / lifetime. But as both you and phk do not share this concern, I will implement this to have something to look at.

dridi: Thanks!

nigoroll:
The alternative implementation with a `VDI_EVENT` is in #4196

nigoroll:
obsoleted by #4196

PR #4188, opened 2024-09-18
tests get larger thread pool, necessary for ppc32

ingvarha:
The former threadpool size of 80k fails on 32bit ppc (at least on my Debian/ppc32)

**** v1    CLI TX|param.set thread_pool_stack 80k
**** dT    0.197
***  v1    CLI RX  106
**** v1    CLI RX|Must be at least 128k

Upping to 128k makes the test suite run.

### DIFF
--- a/bin/varnishtest/tests/e00016.vtc
+++ b/bin/varnishtest/tests/e00016.vtc
@@ -27,7 +27,7 @@ server s1 {
 } -start
 
 # give enough stack to 32bit systems
-varnish v1 -cliok "param.set thread_pool_stack 80k"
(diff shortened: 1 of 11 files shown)


dridi:
The alternative to an increase in stack size could be `-cli` instead of `-cliok` to ignore failures when we set a stack size lower than the platform's minimum.

Could you please try that?

bsdphk:
I'm OK with s/80k/128k/

The 80k was set in the previous commit (55fecc94d3ca59) and seems like an arbitrary choice of no importance for the actual tests ?

PR #4190, opened 2024-09-19
vcc: Internal type for the reserved 'default' symbol

### LINKED ISSUE #4177: Default backend cannot contain a .via backend
### Expected Behavior

`backend default` to be allowed to use `via` backends.

### Current Behavior

Panic on VCL compilation:
```
Message from VCC-compiler:
Assert error in vcc_ParseHostDef(), vcc_backend.c line 565:
  Condition((sym) != 0) not true.
Running VCC-compiler failed, signal 6, core dumped
VCL compilation failed  
```

### Possible Solution

_No response_

Using c42.vtc as an example, changing the name of s2 to default results in a VCC panic.
https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishtest/tests/c00042.vtc#L51 

```vcl
varnish v1 -vcl {
	backend v2 { .host = "${v2_addr}"; .port = "${v2_port}"; }
	backend s1 { .via = v2; .host = "${s1_addr}"; .port = "${s1_port}"; }
	backend default { .via = v2; .host = "${s2_addr}"; .port = "${s2_port}"; }

	sub vcl_recv {
		if (req.url ~ "^/s1/") {
			set req.backend_hint = s1;
		} else if (req.url ~ "^/s2/") {
			set req.backend_hint = default;
		} else {
			return (synth(400));
		}
	}
} -start
```

### Context

Using a proxy on backend with only one backend.

### Varnish Cache version

varnish trunk, varnish 7.5

### Operating system

_No response_

_No response_

dridi:
A new DEFAULT type identifies symbols called 'default' and types that can have a default symbol carry a pseudo symbol. This approach can reconcile the overloaded default symbols that all share the generic type DEFAULT but carry their respective kinds.

This helps remove some of the special casing for default probes and backends whilst offering a sentinel DEFAULT type to safely deal with the special casing where it is actually needed, simplifying the code a wee bit in those areas.

Fixes #4177

### DIFF
--- a/bin/varnishtest/tests/c00042.vtc
+++ b/bin/varnishtest/tests/c00042.vtc
@@ -176,3 +176,9 @@ varnish v1 -errvcl "Cannot stack .via backends" {
 		set bereq.backend = c;
 	}
 }
+
+# issue #4177: backend named default with .via property
+varnish v1 -vcl {
+	backend via { .host = "${localhost}"; }
+	backend default { .via = via; .host = "${localhost}"; }
+}
--- a/lib/libvcc/vcc_backend.c
+++ b/lib/libvcc/vcc_backend.c
@@ -40,19 +40,6 @@
 #include "vcc_compile.h"
 #include "vus.h"
 
-const char *
-vcc_default_probe(struct vcc *tl)
-{
-
-	if (tl->default_probe != NULL)
-		return (tl->default_probe);
-	VSB_cat(tl->sb, "No default probe defined\n");
-	vcc_ErrToken(tl, tl->t);
-	VSB_cat(tl->sb, " at\n");
-	vcc_ErrWhere(tl, tl->t);
-	return ("");
-}
-
 /*--------------------------------------------------------------------
  * Struct sockaddr is not really designed to be a compile time
  * initialized data structure, so we encode it as a byte-string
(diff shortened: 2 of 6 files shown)


nigoroll:
I appreciate the fix, but as a vcc-noob I would be like to learn why a `DEFAULT` type is the better option to having multiple symbols named `default` with different types.
Do we not create a mess with type methods etc?

dridi:
> I appreciate the fix, but as a vcc-noob I would be like to learn why a `DEFAULT` type is the better option to having multiple symbols named `default` with different types.

When I started using Varnish it was perfectly legal to have multiple symbols of the same type, see for example 996299655c9ed5d94853a495ddaa2ae29094a9e9 that introduced coverage for a corner case. The test case contains 3 distinct `debug` symbols.

It was a deliberate choice to only allow one symbol for a given name in bd30219ea12cb558ff156815d8d8c455dea5fe67 and as you can see from the explanation in the commit message that is the turning point where symbols became unique. The `default` symbol and its special meaning turned into a pretend BACKEND symbol that was either half wrong or half correct.

The introduction of a DEFAULT type here makes it completely wrong or completely correct.

As a side note on symbol uniqueness modulus `default`, another change happened afterwards where symbols became unique per namespace in #3344. A perfect segue into the next point you raise.

> Do we not create a mess with type methods etc?

Yes but not really. Because `default.method()` becomes ambiguous depending on the underlying type, I'm more than willing to disallow calling methods or attributes directly from the `default` symbol.

bsdphk:
I think I'm OK with this handling of it.

nigoroll:
> I'm more than willing to disallow calling methods or attributes directly from the default symbol.

Not sure if this is a good idea, but under all circumstances something like
```vcl
set bereq.backend = default;
set bereq.http.healthy = bereq.backend.healthy()
```
should continue to work.

dridi:
This is fine because `bereq.backend` is a different symbol of the actual `BACKEND` type. This change does not affect other symbols of the types `BACKEND` and `PROBE`, but only the actual `default` pseudo-symbol.

nigoroll (approved):
I think this can be merged, phk agreed already

dridi:
As long as there were questions or remarks I would still consider the review to be in progress. Between here and #4177 you raised interesting points.

nigoroll:
late correction:

> have multiple symbols of the same type

I think you meant "multiple symbols of the name _name_ (and different types)" and also read this, even though this is not was was written.

PR #4194, opened 2024-09-23
vdef: Add TFOREACH() macro for txt

dridi:
Random idea of the day.


include/vdef.h
```diff
@@ -266,6 +266,8 @@ typedef struct {
 #define Tstr(s)		(/*lint -e(446)*/ (txt){(s), (s) + strlen(s)})
 #define Tstrcmp(t, s)	(strncmp((t).b, (s), Tlen(t)))
 
+#define TFOREACH(c, t) for ((c) = (t).b; (c) < (t).e; (c)++)
```
asadsa92, line 269: Nice find.
Could we instead go for `#define Tforeach(c, t)` ? or `#define TXT_FOREACH(c, t)` ?
It was not obvious to me what the macro was meant to iterate over.

dridi: T'was not obvious to me either so I went with T like in other txt-related macros and FOREACH like in other iterator macros.

`TXT_FOREACH()` was my alternative.

The pun was of course intended.

nigoroll: My useless contribution on bikeshed colors: `TXT_FOREACH_CHAR`
Please not `TFOREACH`, for consistency I'd otherwise propose `Tforeach` or `Titer`

nigoroll (approved):
You can merge this, phk agreed already

dridi:
I was waiting for CI to complete after making the change.

include/vdef.h
nigoroll: Never too late to rename for the better ;)

PR #4196, opened 2024-09-23
Add VDI_EVENT_SICK and use it for admin health changes

nigoroll:
Alternative implementation to #4186 as suggested by dridi and phk:

This patch adds an event to the existing event facility, which so far was only used for VCL-related events. We now add `VDI_EVENT_SICK` as a director-specific event, which is cheap and also helps us maintain clean layering.
The new event is posted when the admin health changes and used by VBE to cancel all waiting connection requests.

### DIFF
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -496,7 +496,9 @@ struct vmod_data {
 };
 
 /***********************************************************************
- * VCL events sent to VMODs and directors
+ * Events sent to VMODs and directors:
+ * - VCL: VCL temperature events
+ * - VDI: director events
  */
 
 enum vcl_event_e {
@@ -504,7 +506,7 @@ enum vcl_event_e {
 	VCL_EVENT_WARM,
 	VCL_EVENT_COLD,
 	VCL_EVENT_DISCARD,
-	// Only for directors
+
 	VDI_EVENT_SICK,
 };
 
--- a/include/vrt.h
+++ b/include/vrt.h
@@ -48,7 +48,7 @@
 
 #define VRT_MAJOR_VERSION	20U
 
-#define VRT_MINOR_VERSION	0U
+#define VRT_MINOR_VERSION	1U
 
 /***********************************************************************
  * Major and minor VRT API versions.
@@ -58,6 +58,8 @@
  * binary/load-time compatible, increment MAJOR version
  *
  * NEXT (2024-03-15)
+ * 20.1
+ *	VDI_EVENT_SICK added to enum vcl_event_e
  * 20.0 (2024-09-13)
  *	struct vrt_backend.backend_wait_timeout added
  *	struct vrt_backend.backend_wait_limit  added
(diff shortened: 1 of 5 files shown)


gquintard:
one note here: because the events were only sent for VCL lifetime steps, I think we were guaranteed to be running without in-flight requests (for that VCL). This changes the contract a bit.

For context, in the `rust` layer, we allow the PRIV struct to be mutated because we know that nobody has access to it, but that change would force the function to keep the mutability without the guarantee that there's no in-flight request. That's quite a big regression

nigoroll:
> one note here: because the events were only sent for VCL lifetime steps, I think we were guaranteed to be running without in-flight requests (for that VCL). This changes the contract a bit.

There is a relevant change that we extend the facility for director events, but the way I read your understanding of what was guaranteed, I think it might be incomplete: VDI temperature events were already generated outside VCL lifetime steps for dynamic backends [here](https://github.com/varnishcache/varnish-cache/blob/45faed23247471b392359dc66f26e131510fc15b/bin/varnishd/cache/cache_vrt_vcl.c#L245) and [here](https://github.com/varnishcache/varnish-cache/blob/45faed23247471b392359dc66f26e131510fc15b/bin/varnishd/cache/cache_vrt_vcl.c#L285). At least the cold event happens outside the `vcl_mtx`.

> For context, in the `rust` layer, we allow the PRIV struct to be mutated because we know that nobody has access to it, but that change would force the function to keep the mutability without the guarantee that there's no in-flight request. That's quite a big regression

I am not sure which PRIV struct you are referring to, but it sounds unrelated?

dridi:
That would be the PRIV_VCL for VMODs subscribing to events. However this event is not directed (pun intended) towards the VMOD itself so it does not violate the mutability assumption you made in the VMOD event callback.

gquintard:
Ah, thank you both, I did confuse the two events

nigoroll:
> That would be the PRIV_VCL for VMODs subscribing to events.

Yes, the director `.event` function is something different than the vmod event function.

include/vrt.h
```diff
@@ -494,14 +496,16 @@ struct vmod_data {
...
  */
 
 enum vcl_event_e {
 	VCL_EVENT_LOAD,
 	VCL_EVENT_WARM,
 	VCL_EVENT_COLD,
 	VCL_EVENT_DISCARD,
+	// Only for directors
+	VDI_EVENT_SICK,
 };
```
dridi, line 509: I would drop the "Only for directors" comment because in order to preserve the ABI we should not group events by theme and rather append them to the enum. We can instead list types of events in the comment block above:

```c
/***********************************************************************
 * VCL events sent to VMODs and directors:
 * - VCL: VCL temperature events
 * - VDI: director health events
 */
```

Or maybe something a bit more elaborated.

dridi (approved):

include/vrt.h
nigoroll: The comment could still be changed later, it does not break the ABI ;) But yea, good point!

nigoroll:
bugwash: ok

PR #4197, opened 2024-09-24
backend.list -j now outputs IPs/port information

gquintard:
since DNS resolution happens only once for static backends and can be a bit obscure for dynamic backend, it's valuable to be able to dump the IPs that are used.

In order to not break existing scripts which may `varnishadm backend.list`, only update the JSON ouput

side note: it's a bit confusing that `vbe_list` needs to produce a JSON value ending with a coma if `pflag` is true, but not if it's false. Do we want to change that? And also maybe make printing the `probe_details` and `probe_message` a responsibilty of the backend implementation, rather than splitting the key-value pair across two functions?

### DIFF
--- a/bin/varnishtest/tests/b00091.vtc
+++ b/bin/varnishtest/tests/b00091.vtc
@@ -1,0 +1,41 @@
+varnishtest "IPs/port in varnishadm backend.list -j"
+
+feature cmd "jq -V"
+
+server s0 {
+} -start
+
+
+varnish v1 -vcl+backend {
+	backend s1 {
+		.host = "127.0.0.1";
+		.port = "2222";
+		.probe = { .url = "/"; }
+	}
+	backend s2 {
+		.host = "::1:2:3:4";
+		.port = "3333";
+	}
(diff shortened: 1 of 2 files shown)


bin/varnishtest/tests/b00091.vtc
```diff
@@ -0,0 +1,41 @@
...
+		set req.backend_hint = s2;
+	}
+} -start
+
+shell {
+	set -e
+
+varnishadm -n ${v1_name} backend.list -j
+	varnishadm -n ${v1_name} backend.list -jp > result.json
+	jq '
```
nigoroll, line 32: why not pipe to jq?

dridi: I originally voted against piping before this was submitted here for several reasons, mostly because it leaves something to inspect with `varnishtest -l/-L` if the test case ever fails.

nigoroll: makes sense, thx

PR #4198, opened 2024-09-24
cache: Poll dying connection pools asynchronously

cartoush:
This PR moves the blocking sleep loop to a CLI hook.

See https://github.com/varnishcache/varnish-cache/pull/4064


nigoroll:
@cartoush IIRC your commit messages and pull requests so far all were extremely terse. Can you please elaborate such that reviewers can easily understand *which* problem you are trying to solve, *why* you chose a particular approach and, if necessary, *how* you implemented it?

I recently noticed a guide which I would be tempted to propose to adopt for our project: https://github.com/axboe/liburing/blob/master/CONTRIBUTING.md

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +259,39 @@ VCP_Rel(struct conn_pool **cpp)
...
 		cp->n_kill++;
 	}
-	while (cp->n_kill) {
-		Lck_Unlock(&cp->mtx);
-		(void)usleep(20000);
+
+	ALLOC_OBJ(dcp, DYING_CP_MAGIC);
+	dcp->cp = cp;
+	VTAILQ_INSERT_TAIL(&dying_cps, dcp, list);
+	Lck_Unlock(&cp->mtx);
```
dridi, line 266: You shouldn't need an extra data structure for this, `cp` should carry the `list` field. It could be made more explicit, for example `kill_list` (a new `VTAILQ_ENTRY` in `struct conn_pool`).

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +259,39 @@ VCP_Rel(struct conn_pool **cpp)
 		(void)shutdown(pfd->fd, SHUT_RDWR);
 		cp->n_kill++;
 	}
-	while (cp->n_kill) {
-		Lck_Unlock(&cp->mtx);
-		(void)usleep(20000);
+
+	ALLOC_OBJ(dcp, DYING_CP_MAGIC);
```
asadsa92, line 263: `AN(dcp);` ? Do we really need to allocate it? Why not reuse `cp`  ? Currently it does not belong to any tree here, either use that, or add a list member to `cp` which can be used here instead.

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +259,39 @@ VCP_Rel(struct conn_pool **cpp)
...
+	VTAILQ_FOREACH_SAFE(dcp, &dying_cps, list, dcp2) {
+		cp = dcp->cp;
+		CHECK_OBJ_NOTNULL(cp, CONN_POOL_MAGIC);
+
 		Lck_Lock(&cp->mtx);
+		while (cp->n_kill) {
+			Lck_Unlock(&cp->mtx);
+			(void)usleep(20000);
+			Lck_Lock(&cp->mtx);
+		}
```
dridi, line 284: We shouldn't block here either:

```c
Lck_Lock(&cp->mtx);
if (cp->n_kill > 0) {
        Lck_Unlock(&cp->mtx);
        continue;
}

File: bin/varnishd/cache/cache_conn_pool.c [hunk 4]
```diff
@@ -249,17 +259,39 @@ VCP_Rel(struct conn_pool **cpp)
...
+	struct conn_pool *cp;
+
+	VTAILQ_FOREACH_SAFE(dcp, &dying_cps, list, dcp2) {
+		cp = dcp->cp;
+		CHECK_OBJ_NOTNULL(cp, CONN_POOL_MAGIC);
+
 		Lck_Lock(&cp->mtx);
+		while (cp->n_kill) {
+			Lck_Unlock(&cp->mtx);
+			(void)usleep(20000);
```
asadsa92, line 282: why not `return` instead? This is called from `cli_cb_before` hook, you can not afford to block here or you make it synchronous again. :smile:

bin/varnishd/cache/cache_cli.c
```diff
@@ -78,6 +78,7 @@ cli_cb_before(const struct cli *cli)
 	VSL(SLT_CLI, NO_VXID, "Rd %s", VSB_data(cli->cmd));
 	Lck_Lock(&cli_mtx);
 	VCL_Poll();
+	VCP_RelPoll();
```
dridi, line 81: Out of curiosity, why not just `VCP_Poll()`?

dridi (changes requested):
There is room for improvement but this is a good start!

bin/varnishd/cache/cache_conn_pool.c
asadsa92: I see we `rushed` our replies.
I think he could reuse the VBRT entry instead. No need to add a new member.

dridi: That is another option, indeed.

bin/varnishd/cache/cache_conn_pool.c
asadsa92: Marking this as duplicate. See below.

bin/varnishd/cache/cache_conn_pool.c
asadsa92: Marking this as duplicate. See below.

bin/varnishd/cache/cache_cli.c
cartoush: Maybe I'm wrong but I figured that at some point it may be possible that we poll VCPs for something else than releasing them ^^

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -222,6 +224,8 @@ VCP_AddRef(struct conn_pool *cp)
  * Release Conn pool, destroy if last reference.
  */
 
+static struct vrb dying_cps = VRBT_INITIALIZER(&dying_cps);
```
dridi, line 227: Please declare it just below `conn_pools`.

For this, you should declare the tree head separately:

```diff
-static VRBT_HEAD(vrb, conn_pool) conn_pools = VRBT_INITIALIZER(&conn_pools);
+VRBT_HEAD(vrb, conn_pool);
 VRBT_GENERATE_REMOVE_COLOR(vrb, conn_pool, entry, static)
@@ ... all the VRBT_GENERATE calls ... @@
 VRBT_GENERATE_INSERT(vrb, conn_pool, entry, vcp_cmp, static)
+VRBT_GENERATE_NEXT(vrb, conn_pool, entry, static);
+VRBT_GENERATE_MINMAX(vrb, conn_pool, entry, static)
+
+static struct vrb conn_pools = VRBT_INITIALIZER(&conn_pools);
+static struct vrb dead_pools = VRBT_INITIALIZER(&dead_pools);
```

Note that I changed the head name for consistency.

Maybe as a followup commit we could rename `struct vrb` to `struct vrb_vcp`.

dridi (changes requested):
LGTM, with nitpicks.

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +254,32 @@ VCP_Rel(struct conn_pool **cpp)
...
+	VRBT_INSERT(vrb, &dead_pools, cp);
+	Lck_Unlock(&cp->mtx);
+}
+
+void
+VCP_RelPoll(void)
+{
+	struct conn_pool *cp, *cp2;
+
+	VRBT_FOREACH_SAFE(cp, vrb, &dead_pools, cp2) {
```
dridi, line 267: We should add `ASSERT_CLI();` before the loop.

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +254,32 @@ VCP_Rel(struct conn_pool **cpp)
 		(void)shutdown(pfd->fd, SHUT_RDWR);
 		cp->n_kill++;
 	}
-	while (cp->n_kill) {
-		Lck_Unlock(&cp->mtx);
-		(void)usleep(20000);
+
+	VRBT_INSERT(vrb, &dead_pools, cp);
```
dridi, line 258: I just realized that this insertion may not happen in the CLI thread, so we need to guard it. Would the existing `conn_pools_mtx` be enough or do we need a new mutex for this tree?

This implies that we also need to guard the loop in `VCP_RelPoll()` with the same lock. The entire loop or just the VRBT operations?

dridi (changes requested):
I just realized splitting this operation brings it in two different contexts, see comments. I have not thought much about how to proceed so my questions are open questions.

bin/varnishd/cache/cache_conn_pool.c
cartoush: From what I've tested reusing `conn_pools_mtx` works fine
Since the loop is non-blocking I guess we can safely guard the whole loop

dridi: I think we should lock before the loop, unlock after removing an entry, lock again at the end of the loop, and unlock after the loop.

Regarding `conn_pools_mtx` reuse one quick way to assess whether we can safely do that is running `make witness` after implementing the change. It creates a relationship between this lock and per-pool locks, if there isn't already one. You need to check that whenever those locks are taken, it's always in the same order.

nigoroll (changes requested):
As pointed out by others, locking is not correct yet.
I think we should just have a single lock for `dead_pools`, and insert each dead cp under that lock.
VCP_RelPoll (if we keep it like this) can just swap the list head under that mtx, work all entries unlocked, remembering those still active (`n_kill > 0`) on a new list, and finally appending the new list back to `dead_pools` under the mtx.
Alternatively, why do we not just hand dead pool handling to a worker thread?

dridi:
> As pointed out by others, locking is not correct yet. I think we should just have a single lock for `dead_pools`, and insert each dead cp under that lock. VCP_RelPoll (if we keep it like this) can just swap the list head under that mtx, work all entries unlocked, remembering those still active (`n_kill > 0`) on a new list, and finally appending the new list back to `dead_pools` under the mtx.

I guess we could work with a mutex dedicated to the dead pools and not even care about the per-vcp lock. Since we care about observing `n_kill == 0` we don't need to formally synchronize with waiters, we can only see `n_kill` after the waiter woke up for all connections and since this happens in a CLI hook, it will eventually be seen.

> Alternatively, why do we not just hand dead pool handling to a worker thread?

Isn't a CLI hook good-enough for this kind of cleanup sweep?

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +254,39 @@ VCP_Rel(struct conn_pool **cpp)
...
+	Lck_Lock(&conn_pools_mtx);
+	VRBT_FOREACH_SAFE(cp, vrb, &dead_pools, cp2) {
+		VRBT_REMOVE(vrb, &dead_pools, cp);
+		Lck_Unlock(&conn_pools_mtx);
+		CHECK_OBJ_NOTNULL(cp, CONN_POOL_MAGIC);
+
 		Lck_Lock(&cp->mtx);
+		if (cp->n_kill > 0) {
+			Lck_Unlock(&cp->mtx);
+			continue;
```
asadsa92, line 279: You can not do this.
You end up removing the connection pool from the tree, but if the kill count is positive we do not add it back.
(I am not saying the correct thing is to add it back, but the problem is that you end up leaking connection pools which might also have other impact such as races).

cartoush: right, my bad, I guess I was not awoken enough when I made this change ^^

nigoroll:
> > Alternatively, why do we not just hand dead pool handling to a worker thread?
> 
> Isn't a CLI hook good-enough for this kind of cleanup sweep?

It probably is, but I do not know, this was just an option to maybe consider.

I would really like to see swapping the tree head and working the cleanup outside any connection pool mtx.

cartoush:
Added a `dead_pools_mtx` instead of using `conn_pools_mtx`

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +255,35 @@ VCP_Rel(struct conn_pool **cpp)
...
-	while (cp->n_kill) {
-		Lck_Unlock(&cp->mtx);
-		(void)usleep(20000);
-		Lck_Lock(&cp->mtx);
-	}
+
+	Lck_Lock(&dead_pools_mtx);
+	VRBT_INSERT(vrb, &dead_pools, cp);
+	Lck_Unlock(&dead_pools_mtx);
 	Lck_Unlock(&cp->mtx);
```
dridi, line 262: You should be able to unlock `cp->mtx` at the end of the foreach loop and keep `dead_pools_mtx` independent. At this point only waiters may have a reference of the pool and they are not concerned with connection pools cleanup, only individual connections.

We could even consider checking `n_kill` here to avoid a trip to the CLI hook:

```c
{
        /* ... */

        if (cp->n_kill == 0) {
                vcp_destroy(&cp);
                return;
        }

        Lck_Lock(&dead_pools_mtx);
        VRBT_INSERT(vrb, &dead_pools, cp);
        Lck_Unlock(&dead_pools_mtx);
}
```

Where `vcp_destroy()` would also be called from `VCP_RelPoll()`.

dridi (changes requested):

bin/varnishd/cache/cache_conn_pool.c
cartoush: Done :)

dridi (approved):
I think I'm OK with this now, but I'm running out of brain cycles.

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -218,6 +224,24 @@ VCP_AddRef(struct conn_pool *cp)
 	Lck_Unlock(&conn_pools_mtx);
 }
 
+/*--------------------------------------------------------------------
+ */
+
+static void
+vcp_free(struct conn_pool **cpp)
```
dridi, line 231: I pointed out outside of github that `vcp_destroy()` was acting like a regular "free" function, but it was renamed to `vcp_free()` and is now behaving like a "destroy" function.

Apologies for the confusion. :sweat_smile:

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -218,6 +224,24 @@ VCP_AddRef(struct conn_pool *cp)
...
+{
+	struct conn_pool *cp;
+
+	AN(cpp);
+	cp = *cpp;
+	*cpp = NULL;
```
dridi, line 237: Use the `TAKE_OBJ_NOTNULL()` for this.

dridi (approved):

bin/varnishd/cache/cache_conn_pool.c
dridi: You did rename the function but forgot the call sites.

cartoush:
calls to `vcp_free` had not been turned into calls to `vcp_destroy`, its fixed now

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +271,33 @@ VCP_Rel(struct conn_pool **cpp)
...
+}
+
+void
+VCP_RelPoll(void)
+{
+	struct conn_pool *cp, *cp2;
+
+	ASSERT_CLI();
+	Lck_Lock(&dead_pools_mtx);
+	VRBT_FOREACH_SAFE(cp, vrb, &dead_pools, cp2) {
```
nigoroll, line 291: I still think we should take the dead_pools tree into a private tree here, try to destroy all pools on it outside the mtx and then put back any left overs.

nigoroll (changes requested):

bin/varnishd/cache/cache_conn_pool.c
cartoush: After discussing that with @dridi we think that this improvement could take place in a later PR, is that fine with you ?

nigoroll: a `VRBT_HEAD()` is just a pointer to the root node, so swapping it is trivial. Inserting all remaining elements is a little less trivial, but no big deal either.

It is obvious how this PR implements a good improvement, but as we are already doing it, I do no see why we should have `VCP_Rel()` wait for `dead_pools_mtx`.

So I do not understand why we would need incremental commits rather than just getting this done.

nigoroll: Ah, I see now that you added another unlock/lock around the destroy, so my argument about the lock is less relevant. Bus still this is unnecessary if we just swap the tree pointer.

cartoush: Alright I'll handle that

cartoush: Should be fine now

dridi (commented):
Reconciling the desire to loop without holding the lock with the reuse of the `VRBT_ENTRY()` field already present in `struct conn_pool` is complicating the change above my acceptable threshold.

If we put the cursor on the lock, then we should probably have a tail queue of dead connections as suggested before. We keep constant-time operations, including the ability to easily steal and reinject dead pools before and after the cleanup loop.

@nigoroll thoughts?

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +271,45 @@ VCP_Rel(struct conn_pool **cpp)
...
+			continue;
+		VRBT_REMOVE(vrb, &dead, cp);
+		vcp_destroy(&cp);
+	}
+
+	if (!VRBT_EMPTY(&dead)) {
+		Lck_Lock(&dead_pools_mtx);
+		VRBT_INSERT(vrb, &dead_pools, dead.rbh_root);
+		Lck_Unlock(&dead_pools_mtx);
+	}
```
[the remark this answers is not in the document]
nigoroll: this is actually pretty smart, IMHO, and I did not think of it: I originally thought we would iterate the leftovers and insert one by one, but as we do not care about ordering in the `dead_pools` tree, this _should_ work.

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +271,45 @@ VCP_Rel(struct conn_pool **cpp)
...
+	struct conn_pool *cp, *cp2;
+
+	ASSERT_CLI();
+
+	Lck_Lock(&dead_pools_mtx);
+	if (VRBT_EMPTY(&dead_pools)) {
+		Lck_Unlock(&dead_pools_mtx);
+		return;
+	}
+	TAKE_OBJ_NOTNULL(dead.rbh_root, &dead_pools.rbh_root, CONN_POOL_MAGIC);
```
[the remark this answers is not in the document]
nigoroll: Yeah, the `TAKE_OBJ` is a smart idea, but using the tree macros as intended would be my preference also.

nigoroll:
Except for the one or two nitpicks, I like the structure of the patch now. I do not understand which complication @dridi is seeing, IMHO, this is a _simplification_.

bin/varnishd/cache/cache_conn_pool.c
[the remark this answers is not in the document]
nigoroll: I checked the tree macros and indeed, the comparison function is only used for insert and lookup (as expected). So yeah, I think this is still nicer than using a union in place of the tree pointers.

dridi:
> I do not understand which complication @dridi is seeing, IMHO, this is a _simplification_.

That was in the optic of not misusing the VRBT API on purpose. I'll comment in that thread.

bin/varnishd/cache/cache_conn_pool.c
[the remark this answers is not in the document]
dridi: I wasn't suggesting a union, but if this is safe but "incorrect", we should add red tape to remind our future selves why this is good enough.

cartoush:
Fixed comments and added comments to the `VRBT_INSERT` hack

nigoroll (approved):

bin/varnishd/cache/cache_conn_pool.c
```diff
@@ -249,17 +271,56 @@ VCP_Rel(struct conn_pool **cpp)
...
+		Lck_Lock(&dead_pools_mtx);
+		/*
+		 * The following insertion will most likely result in an
+		 * unordered tree, but in this case it does not matter
+		 * as we just want to iterate over all the elements
+		 * in the tree in order to delete them.
+		 */
+		VRBT_INSERT(vrb, &dead_pools, dead.rbh_root);
+		Lck_Unlock(&dead_pools_mtx);
+	}
```
dridi, line 323: My preference is:

```
if (empty)
        return;

/* more work */
```

dridi (approved):
Good job!

nigoroll:
@bsdphk two people think this one is ripe

PR #4200, opened 2024-09-26
sml: Optimize freebehind (transient) memory usage

nigoroll:
Previously, `sml_iterator()` could not free the last segment while iterating, because it could have turned out to be over-allocated at any time and be removed.

We now delay freeing any over-allocated segments, identical to how we delay any reallocated segments, which enables us to "free behind" all sent segments.

This is relevant because, due to of the race between the fetch and iterator side of a transient operation, the "is last segment" condition can happen repeatedly while an object is busy.


nigoroll:
OKed by @mbgrydeland via email

PR #4205, opened 2024-09-30
cache_vcl: Do the temperature dance for backend creation races

### LINKED ISSUE #4199: Panic on vcl.load: Wrong turn at cache/cache_backend_probe.c:710: running
### Expected Behavior

No panic on the following VTC.

**Important reproduction notes**

This is against Varnish master + libvmod-dynamic master. I can only reproduce this when intentionally "slowing down" vmod_event with a sleep.

In my case in production, I originally had the panic when using libvmod-redis, and without adding its `import` in my VCL, the panic would not be reproduced. I suspected some kind of race condition that only happens when waiting in vmod_event somehow (because that would be the only side effect of loading libvmod-redis).

So, in order to reproduce the panic in the VCL below, add at the start of `libvmod-dynamic/src/vmod_dynamic.c` `vmod_event`:
```c
#include <unistd.h>
usleep(100000);
```
(Or run the following from `libvmod-dynamic`: `sed -z -i -E 's|(vmod_event\(VRT_CTX, struct vmod_priv \*priv, enum vcl_event_e e\)\n\{\n)|\1\#include \<unistd.h\>\nusleep\(100000\)\;\n|g' src/vmod_dynamic.c`)

This will simulate a slow down in vmod_event by sleeping for 100ms.

### Current Behavior

```vtc
varnishtest "panic"

varnish v1 -vcl {
vcl 4.1;
backend local none;
} -start

varnish v1 -vcl {
        vcl 4.1;
        import std;
        import dynamic;
        import directors;

        backend local none;

        probe p1 {}
        probe p2 {}

        sub vcl_init {
                new cluster_default = directors.fallback();
                new r1 = dynamic.resolver();
                new i1 = dynamic.director(port = {"80"}, ttl = 10s, probe = p1, resolver = r1.use(), ttl_from = dns);
                cluster_default.add_backend(i1.backend(host={"example.com"}, port={"8000"}) );
                new r2 = dynamic.resolver();
                new i2 = dynamic.director(port = {"80"}, ttl = 10s, probe = p2, resolver = r2.use(), ttl_from = dns);
                cluster_default.add_backend(i2.backend(host={"google.com"}, port={"8001"}) );
        }
}
```

The panic is quite random, so might require several tries:
```shell
while true; do varnishtest ./panic.vtc; done
```

The panic is:
```
Wrong turn at cache/cache_backend_probe.c:712:
scheduled
Backtrace:
  ip=0x63557d52e435 sp=0x7fffaf88dde0 <VBT_format+0x35>
  ip=0x63557d459773 sp=0x7fffaf88de00 <pan_backtrace+0x33>
  ip=0x63557d459473 sp=0x7fffaf88de20 <pan_ic+0x3e3>
  ip=0x63557d52d5f5 sp=0x7fffaf88dfa0 <VAS_Fail+0x55>
  ip=0x63557d4180b1 sp=0x7fffaf88dff0 <VBP_Control+0x231>
  ip=0x63557d41514f sp=0x7fffaf88e020 <vbe_dir_event+0x15f>
  ip=0x63557d42ab8d sp=0x7fffaf88e050 <VDI_Event+0xbd>
  ip=0x63557d474b1c sp=0x7fffaf88e070 <vcl_BackendEvent+0x16c>
  ip=0x63557d473f6c sp=0x7fffaf88e0a0 <vcl_set_state+0x58c>
  ip=0x63557d476cdf sp=0x7fffaf88e0f0 <vcl_load+0x3bf>
  ip=0x63557d474f4d sp=0x7fffaf88e140 <vcl_cli_load+0xcd>
  ip=0x63557d5327d2 sp=0x7fffaf88e170 <cls_dispatch+0x392>
  ip=0x63557d5322bd sp=0x7fffaf88e1b0 <cls_exec+0x2fd>
  ip=0x63557d53101f sp=0x7fffaf88e200 <cls_feed+0x58f>
  ip=0x63557d5309ff sp=0x7fffaf88e270 <VCLS_Poll+0x3bf>
  ip=0x63557d423b02 sp=0x7fffaf8902d0 <CLI_Run+0xf2>
  ip=0x63557d4534f5 sp=0x7fffaf890300 <child_main+0x485>
  ip=0x63557d4c81da sp=0x7fffaf890370 <mgt_launch_child+0x63a>
  ip=0x63557d4c924e sp=0x7fffaf8903f0 <mch_cli_server_start+0x3e>
  ip=0x63557d5327d2 sp=0x7fffaf890420 <cls_dispatch+0x392>
  ip=0x63557d5322bd sp=0x7fffaf890460 <cls_exec+0x2fd>
  ip=0x63557d53101f sp=0x7fffaf8904b0 <cls_feed+0x58f>
  ip=0x63557d5309ff sp=0x7fffaf890520 <VCLS_Poll+0x3bf>
  ip=0x63557d4ca335 sp=0x7fffaf892580 <mgt_cli_callback2+0x25>
  ip=0x63557d534bbe sp=0x7fffaf8925a0 <VEV_Once+0x53e>
  ip=0x63557d53465d sp=0x7fffaf8925f0 <VEV_Loop+0xed>
  ip=0x63557d4d02f3 sp=0x7fffaf892620 <main+0x1ed3>
  ip=0x7b288cde1e08 sp=0x7fffaf892870
  ip=0x7b288cde1ecc sp=0x7fffaf892910 <__libc_start_main+0x8c>
  ip=0x63557d40df35 sp=0x7fffaf892970 <_start+0x25>
```

But sometimes the VCL state is `running` instead of `scheduled`.

### Possible Solution

_No response_

_No response_

### Context

This was particularly hard to track down to a MWE VTC :stuck_out_tongue: 

### Varnish Cache version

varnishd master; libvmod-dynamic master + usleep patch from ticket

### Operating system

_No response_

_No response_

nigoroll:
Dynamic backends may get created while we transition a VCL to the WARM temperature: Because we first set the new temperature and then post the event, backends might get created when the temperature is already WARM, but before the `VCL_EVENT_WARM` gets posted, which leads to double warming. This can be demonstrated with an appropriate delay during `vcl_send_event()`.

The solution to this problem is simple: Only post WARM events to backends from before the start of the transition.

In code, it looks a little more involved, but it is not complicated:

Under the vcl mutex, we take all directors onto a separate "cold" list and change the temperature, and send the warm event only to the "cold" directors. When all are warm, we concatenate.

To undo, we basically do the inverse: Save the warm directors, put the cold ones back, send a cold event to all warm ones and concatenate again.

Fixes #4199

### DIFF
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -436,7 +436,8 @@ VCL_IterDirector(struct cli *cli, const char *pat,
 }
 
 static void
-vcl_BackendEvent(const struct vcl *vcl, enum vcl_event_e e)
+vcl_BackendEvent(const struct vcl *vcl, struct director_head *head,
+    enum vcl_event_e e)
 {
 	struct vcldir *vdir;
 
@@ -444,8 +445,9 @@ vcl_BackendEvent(const struct vcl *vcl, enum vcl_event_e e)
 	CHECK_OBJ_NOTNULL(vcl, VCL_MAGIC);
 	AZ(vcl->busy);
 
+	// XXX should we go unlocked? #4199 #4142
 	Lck_Lock(&vcl_mtx);
-	VTAILQ_FOREACH(vdir, &vcl->director_list, list)
+	VTAILQ_FOREACH(vdir, head, list)
 		VDI_Event(vdir->dir, e);
 	Lck_Unlock(&vcl_mtx);
 }
@@ -585,6 +587,7 @@ vcl_print_refs(const struct vcl *vcl)
 static int
 vcl_set_state(struct vcl *vcl, const char *state, struct vsb **msg)
 {
+	struct director_head colddirs, warmdirs;
 	struct vsb *nomsg = NULL;
 	int i = 0;
 
@@ -606,7 +609,7 @@ vcl_set_state(struct vcl *vcl, const char *state, struct vsb **msg)
 			vcl->temp = VTAILQ_EMPTY(&vcl->ref_list) ?
 			    VCL_TEMP_COLD : VCL_TEMP_COOLING;
 			Lck_Unlock(&vcl_mtx);
-			vcl_BackendEvent(vcl, VCL_EVENT_COLD);
+			vcl_BackendEvent(vcl, &vcl->director_list, VCL_EVENT_COLD);
 			AZ(vcl_send_event(vcl, VCL_EVENT_COLD, msg));
 			AZ(*msg);
 		}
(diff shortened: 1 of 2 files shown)


nigoroll:
Sorry, I anyone was real quick they might have caught a little glitch. I force-pushed a correction.

bin/varnishd/cache/cache_vcl.h
```diff
@@ -37,7 +37,7 @@ struct vfilter;
 struct vcltemp;
 
 VTAILQ_HEAD(vfilter_head, vfilter);
-
+VTAILQ_HEAD(director_head,vcldir);
```
dridi, line 40: If you are motivated, missing space after the comma.

dridi (approved):
It would be nice if @AlveElde could also have a look since he spent time on the temperature state machine.

bin/varnishd/cache/cache_vcl.h
nigoroll: ✅

nigoroll:
bugwash: leave more elaborate explanation on the race here

bin/varnishd/cache/cache_vcl.c
```diff
@@ -629,17 +632,44 @@ vcl_set_state(struct vcl *vcl, const char *state, struct vsb **msg)
...
+			 * the WARM event. This assumes no deletes, see #4142
+			 */
+
+			VTAILQ_INIT(&colddirs);
+			VTAILQ_INIT(&warmdirs);
+
 			Lck_Lock(&vcl_mtx);
+			VTAILQ_SWAP(&vcl->director_list, &colddirs, vcldir, list);
 			vcl->temp = VCL_TEMP_WARM;
 			Lck_Unlock(&vcl_mtx);
```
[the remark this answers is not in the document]
nigoroll: Thank you @AlveElde, you are right that this is brittle.

Because `VTAILQ_REMOVE(head, elm, field)` only touches `head` if `elm` is last in the list, this would only affect the case where the last director added before the swap is removed after the concat, but yes, it is another race. Less relevant, but still a race.

This is not the only place where removing directors is a problem, which is why I suggested a solution in #4142.

Maybe it's best to bring that one in first and then update this PR accordingly.

nigoroll:
Set to draft because #4142 should be done first.

nigoroll:
With #4142 in place, I think there is a better way, so I will close this PR and add a replacement.

PR #4206, opened 2024-10-02
VCL: Restrict scope of req.is_hit{miss,pass}

walid-git:
Reading the values elsewhere doesn't make much sense.


dridi:
Should we also include `vcl_synth` for both? That's a possible transition after `vcl_miss` and `vcl_pass`.

walid-git:
> Should we also include `vcl_synth` for both? That's a possible transition after `vcl_miss` and `vcl_pass`.

I was hesitant about that one and thought I would wait for opinions, there is also the possibility to go to `pass` from `miss`, so we may also want to allow `is_hitmiss` in `vcl_pass`.

asadsa92:
We are not worried this change would break setups, right?
We are going to use the same excuse this time around that it is better it breaks setups when the customer is using it wrong?

AlveElde:
I would add `vcl_synth`, as it is fairly common to share logic between  `vcl_synth` and `vcl_deliver` with a subroutine called from both places.

dridi:
You will also end up in `vcl_synth` if your fetch failed.

walid-git:
Added `vcl_pass` for `is_hitmiss` and `vcl_synth` for both.

dridi (approved):

nigoroll (approved):

walid-git:
bugwash: OK after adding to release notes.

Issue #2735, opened 2018-07-22
RX_OVERFLOW returns TCP reset instead of HTTP 414


sgran:
Hi there,

I'm uncertain about whether I am seeing correct behaviour in varnish or not.  I am looking at how varnish handles RX_OVERFLOW, and it seems like there is a better behaviour available than closing the socket  - HTTP 414 seems designed for this case.

As I say, I'm not certain whether you consider this a bug.  I quite understand that closing the connection is the simplest way to deal with an overlong request header.  I have a use case where it would be better for my clients if we sent them a 4xx response code for these - request parameters that we see are typically generated programmatically from the output of other queries, so can have quite large sets of parameters to iterate over.  It's not the end of the world if each client has to learn that "connection reset" ends up meaning the same thing as "HTTP 414", but it would make it simpler to explain to them what's going on if they were getting a 4xx error code.

The behaviour appears to be by design and not an accident of compilation, but we have seen this behaviour on various versions of varnish from 4.1 to 6.0, on Debian and Ubuntu.

slimhazard:
I agree with the gist of this, except that "414 URI Too Long" doesn't sound right, unless it was just the URI that lead to the overflow (which would really be something).

"413 Payload Too Large" appears to be the right response for this:
https://tools.ietf.org/html/rfc7231#section-6.5.11

BTW @sgran +1 for the avatar.

sgran:
Hi,

Yeah, I am seeing this for the request URI (which is amazing), but it's because the query params are being programmatically generated from other request/reply events - our clients need to learn how to ask for pages instead of all things all at once.

I can see that differentiating between URI and other headers in the return code might be a good thing, so perhaps they do want different responses.

slimhazard:
We talked about some possible approaches at bugwash today, but @bsdphk was not around and we want to get his input.

A point that was mentioned is that closing the connection causes the least amount of processing for abusive clients -- avoids Varnish getting DOSsed by a flood of oversized requests.

But that doesn't mean that we aren't open to different solutions (just not ready to decide today).

bsdphk:
Originally we decided to just close on abusive input, to improve DoS resistance.

That still sounds sensible to me, but if there is a need for more civilized behaviour, we could make a feature to control it (default open to discussion)

bsdphk:
We bugwashed it, and my previous comment stands.

If we want to modulate behaviour, it has to be a featureflag, we cannot ask VCL since we cannot represent the request for inspection.

[...]

PR #4207, opened 2024-10-03
Return HTTP response to RX_OVERFLOW

cartoush:
Resolves https://github.com/varnishcache/varnish-cache/issues/2735

Before this PR, Varnish would silently close the connection to a client that triggered an RX_OVERFLOW (too long URI), this PR allows configuring Varnish to respond to RX_OVERFLOW with a minimal HTTP response that will contain a chosen HTTP code instead.

Its done by adding a new feature flag in order to enable the feature and a new parameter in order to set the HTTP code that will be sent in the response.

### DIFF
--- a/bin/varnishd/common/common_param.h
+++ b/bin/varnishd/common/common_param.h
@@ -118,6 +118,7 @@ struct params {
 #define ptyp_thread_pool_min	unsigned
 #define ptyp_timeout		vtim_dur
 #define ptyp_uint		unsigned
+#define ptyp_uint_orzero	unsigned
 #define ptyp_vcc_feature	vcc_feature_t
 #define ptyp_vsl_buffer		unsigned
 #define ptyp_vsl_mask		vsl_mask_t
--- a/bin/varnishd/http1/cache_http1_fsm.c
+++ b/bin/varnishd/http1/cache_http1_fsm.c
@@ -327,6 +327,10 @@ HTTP1_Session(struct worker *wrk, struct req *req)
 			    cache_param->http_req_size);
 			assert(!WS_IsReserved(req->htc->ws));
 			if (hs < HTC_S_EMPTY) {
+				if (hs == HTC_S_OVERFLOW && cache_param->http_req_overflow_status != 0) {
+					(void)req->transport->minimal_response(req,
+					    cache_param->http_req_overflow_status);
+				}
 				req->acct.req_hdrbytes +=
 				    req->htc->rxbuf_e - req->htc->rxbuf_b;
 				Req_AcctLogCharge(wrk->stats, req);
--- a/bin/varnishd/mgt/mgt_param.h
+++ b/bin/varnishd/mgt/mgt_param.h
@@ -86,6 +86,7 @@ tweak_t tweak_thread_pool_min;
 tweak_t tweak_thread_pool_max;
 tweak_t tweak_timeout;
 tweak_t tweak_uint;
+tweak_t tweak_uint_orzero;
 tweak_t tweak_vcc_feature;
 tweak_t tweak_vsl_buffer;
 tweak_t tweak_vsl_mask;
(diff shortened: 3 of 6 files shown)


bin/varnishtest/tests/b00092.vtc
```diff
@@ -0,0 +1,28 @@
...
...hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world&hello=world"
```
cartoush, line 10: I'll be glad if someone comes up with a clean way to write this kind of test :)

simonvik: We have `${string,repeat,4096,x}`but dunno if that works.

dridi: Replace `x` with `&hello=world` and it should :)

walid-git: maybe reduce `http_req_size` to the minimum can help too

cartoush: Alright I'll try all that

cartoush: Thx for the tips guys !

nigoroll (commented):
Thank you for writing a better PR description and better commit messages ;)

I am generally 👍🏽 , and only have this nit-pick: Instead of having a feature flag and a parameter, can't we just just an invalid status code like 0 with `req_overflow_status` to mean "disable" (= close connection)?

nigoroll:
bugwash: we would appreciate some more research regarding which scenarios would benefit from a similar mechanism. Also, the argument has been made that, once we send a response at all, we might as well do so through `vcl_synth`, but then the question arises what to present to VCL as the request (url, proto, headers).

@cartoush volunteered to spend some time on researching these questions, thank you!

cartoush:
> bugwash: we would appreciate some more research regarding which scenarios would benefit from a similar mechanism. Also, the argument has been made that, once we send a response at all, we might as well do so through `vcl_synth`, but then the question arises what to present to VCL as the request (url, proto, headers).
> 
> @cartoush volunteered to spend some time on researching these questions, thank you!

Sorry for the delay, I did some thinking on that and I think one could argue all cases where an error causes a connection to close silently could benefit from that.
But generalizing that using `vcl_synth` might not be appropriate, as you can do restart for example which be quite dangerous, something better could be creating a `vcl_invalid` subroutine which only gives access to what's relevant to handle the error.
In this case the default vcl behavior would be:
```
sub vcl_invalid {
        return (close);
}
```
and in order to handle the case here we would have something like:
```
sub vcl_invalid {
        if (req.error.reason == REQ_HTTP_OVERFLOW) {
                resp.status == 413;
                return (deliver);
        }
        return (close);
}
```
I think the the whole `req` should not be usable in `vcl_invalid` but `req.error` should be populated with all necessary information to do the processing. 
But then this would have to take place in another PR as this requires quite a bit of planning in order to decide what can be done in `vcl_invalid`.

nigoroll:
This is not as trivial as I thought to support the status range 400..599 _or_ 0.

nigoroll:
has been discussed again, but the decision remains the same...

nigoroll:
I renamed the parameter to `http_req_overflow_status` for consisteny with `http_req_size`.

PR #4208, opened 2024-10-07
vmod_debug: add debug.chksha256 / debug.chkcrc32 VDPs

nigoroll:
Usually we do not require PRs for vmod_debug additions, but in this case I feel more comfortable asking if anyone has better ideas:

Working on storage engines and delivery changes, I would like to have a simple way to ensure for longer tests (for example using [wrk](https://github.com/wg/wrk)) that delivered body data is correct. While implementing checks in the http client would ensure proper end-to-end validation, the task of handling errors looks tedious, because, for hunting bugs efficiently, getting a core dump of varnishd for failing tests is basically a requirement.

Panicking varnishd from a VDP at the right place is easy and independent of protocols. So this is a suggestion to add body checksum check VDPs.

The code has some duplications, but with just SHA256 and CRC32 I consider the savings of generalizing not worthy.


dridi:
Finally, it pays off in CI :]

```
FAIL: tools/magic_check.sh
==========================

Duplicate magic number:
#define VDP_CHKSHA256_MAGIC		0x6856e913
#define VDP_CHKCRC32_MAGIC		0x6856e913

FAIL tools/magic_check.sh (exit status: 1)
```

We should maybe move vmod_debug filter coverage to a dedicated `vmod_debug_filters.c` because `vmod_debug.c` is a bit crowded. Maybe that's good as it is.

Not a review, but since you are seeking an opinion, no objection on my end. I'm not convinced by the mechanics for panicking at first glance, but don't have time to look closer. If you want to use this outside of `varnishtest` and out of tree, maybe vmod_vtc would be a better place, or a dedicated `vmod_chk` where we could bundle various checks.

nigoroll:
bugwash: ok, split source files

PR #4210, opened 2024-10-09
Include debug.h in libvmod_debug_la_SOURCES

simonvik:


dridi (commented):
This file is generated, there should be something else to touch.

dridi:
Oh, it looks like this is how it got regenerated.

nigoroll:
A bit late, but: thank you

PR #4212, opened 2024-10-11
Generic Copyright

nigoroll:
This commit mirrors https://github.com/varnishcache/homepage/commit/0f25620719c3568e166b9beff5c2c176236bf29a

as of today, each page redendered from the sphinx source contains the line

	© Copyright 2010-2014, Varnish Software AS.

While this is true for some of the content, it is not for all of it and because we, several people from UPLEX, intend to contribute significant amounts of documentation in the near future, we would like to change this into something more generic, without the need to state specific copyright and authorship on each page.

### DIFF
--- a/doc/sphinx/conf.py.in
+++ b/doc/sphinx/conf.py.in
@@ -38,7 +38,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'Varnish Cache'
-copyright = u'2010-2014, Varnish Software AS'
+copyright = u'2010-2024, The Varnish Cache Contributors'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the


nigoroll:
bugwash: 1 week to raise any concerns, then it we apply the change

huayra:
@nigoroll the copyright notice on the varnish.org website cannot be compared to the documentation as such just because it is presented as a website in this case.

Our rationale has been that copyright of the documentation follows the [code license](https://github.com/varnishcache/varnish-cache/blob/master/LICENSE), as it is distributed as a unit. The premises for this haven't changed.

perbu:
I suspect the reason it is like it is, is because I wrote/rewrote all docs in that period so it was more or less true.

I would move the copyright statements into the RST source as comments and disregard it in the rendered pages. It really isn't that important, imo. Perhaps this would work for @nigoroll as well?

"For copyright information see the source of the document" or something like that.

perbu:
> @perbu yes, thank you for your great intial work, but we have come a long way since then *)

thank God.

> I do not see much of an advance of "For copyright information see the source of the document" over just stating that the copyright is with the respective contributors? Or, from the other end, what is the issue with a generic statement like this?

it would provide the reader with a way of figuring out what who the respective contributors are without having 1-7 copyright statements on each rendered page. copyright is implicitly owned by the contributors, though, so stating it doesn't add much. might as well drop the copyright statement.

I no longer have a horse in this race. I was just asked to clarify what I did way back when.

nigoroll:
@perbu thank you for the clarification, it is much appreciated.

I would think that the wording you are proposing works similarly, except that it makes another indirection to the source. I would like to imply the meaning of "look up the contributors to determine the copyright holders". Ultimately, that could mean to state "Copyright is established through git commits". Do we want something like this?

dridi:
> Ultimately, that could mean to state "Copyright is established through git commits". Do we want something like this?

This is at least the purpose of the author/committer dichotomy in git. One could argue that having a single author is an unfortunate limitation, but at least when used correctly we keep track of who did what regardless of who merges contributions.

nigoroll:
bugwash: give @huayra

bsdphk:
Just for completion of narative:

The agreement between the project and V-S is that V-S would defensively hold the "collection" copyright on behalf of the project, which did/does not have a legal persona which can do so.

This is clearly stated in the first two lines of the LICENSE file:

    The compilation of software known as "Varnish Cache" is distributed under the following terms:

huayra:
@nigoroll yes, there's no question that contributors own their copyright. There's no copyright transfer unless otherwise explicitly agreed.

Harmonizing the copyright note with what we already have on varnish.org makes sense for the docs. And it's more precise than yet another reference to the source as contributors likely know who they are and feel likely represented by your proposal.

I just wanted to emphasize the historical context of this regarding V-S contributions and how that line specifically was a function of that and the copyright as shown in the license. I even asked @perbu to weight in just to clarify the issue and have some perspective. I'm happy we managed to discuss different aspects of this and that we are in agreement. And I apologize for making you wait,. I thought we were kind of done :)

nigoroll:
Thank you @huayra your clarification is much appreciated.

PR #4213, opened 2024-10-11
Varnisncsa: Change matching rules to reflect reality

### LINKED ISSUE #3528: varnishncsa logs header values even if they have in fact been unset
(Maybe somewhat related to #3527 ?)

In VCL :
```
set resp.http.X-My-Header = "not-actually-sent";
unset resp.http.X-My-Header;
```

Expected behavior :
varnishncsa -f "%{X-My-Header}o" --> nothing

Current behavior :
varnishncsa -f "%{X-My-Header}o" --> "not-actually-sent"

We use varnishncsa to collect request logs in ELK, it is impossible to distinguish requests where the header was present from those where it was actually unset before sending the response

walid-git:
This PR introduces the changes discussed in https://github.com/varnishcache/varnish-cache/issues/3528

The relevant parts of [BE]requests/responses are now shown as they were received/sent from/to the peer, without taking irrelevant VCL changes into account. This PR also introduces the handling of Unset headers in varnishncsa.

Note: this breaks the default behavior of varnishncsa.

Fixes #3528.

### DIFF
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -160,6 +160,7 @@ static struct ctx {
 	const char		*handling;
 	const char		*side;
 	int64_t			vxid;
+	int			recv_compl;
 } CTX;
 
 static void parse_format(const char *format);
@@ -971,6 +972,7 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 		} else
 			continue;
 
+		CTX.recv_compl = 0;
 		CTX.hitmiss = "-";
 		CTX.handling = "-";
 		CTX.vxid = t->vxid;
@@ -1073,6 +1075,7 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 				break;
 			case SLT_VCL_call:
 				if (!strcasecmp(b, "recv")) {
+					CTX.recv_compl = 1;
 					CTX.hitmiss = "-";
 					CTX.handling = "-";
 				} else if (!strcasecmp(b, "hit")) {
@@ -1090,6 +1093,8 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 					   wrong */
 					CTX.hitmiss = "miss";
 					CTX.handling = "synth";
+				} else if (!strcasecmp(b, "backend_response")) {
+					CTX.recv_compl = 1;
 				}
 				break;
 			case SLT_VCL_return:
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -957,7 +957,7 @@ frag_line(enum format_policy fp, const char *b, const char *e,
 
 static void
 process_hdr(enum format_policy fp, const struct watch_head *head, const char *b,
-    const char *e)
+    const char *e, int unset)
 {
 	struct watch *w;
 	const char *p;
@@ -966,7 +966,12 @@ process_hdr(enum format_policy fp, const struct watch_head *head, const char *b,
 		CHECK_OBJ_NOTNULL(w, WATCH_MAGIC);
 		if (!isprefix(w->key, w->keylen, b, e, &p))
 			continue;
-		frag_line(fp, p, e, &w->frag);
+		if (unset) {
+			frag_line(fp, CTX.missing_string,
+			    NULL,
+			    &w->frag);
+		} else
+			frag_line(fp, p, e, &w->frag);
 	}
 }
 
@@ -1101,7 +1106,7 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 				break;
 			case SLT_BereqHeader:
 			case SLT_ReqHeader:
-				process_hdr(FMTPOL_REQ, &CTX.watch_reqhdr, b, e);
+				process_hdr(FMTPOL_REQ, &CTX.watch_reqhdr, b, e, 0);
 				if (ISPREFIX("Authorization:", b, e, &p) &&
 				    ISPREFIX("basic ", p, e, &p))
 					frag_line(FMTPOL_REQ, p, e,
@@ -1113,7 +1118,15 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 				break;
 			case SLT_BerespHeader:
 			case SLT_RespHeader:
-				process_hdr(FMTPOL_RESP, &CTX.watch_resphdr, b, e);
+				process_hdr(FMTPOL_RESP, &CTX.watch_resphdr, b, e, 0);
+				break;
+			case SLT_BereqUnset:
+			case SLT_ReqUnset:
+				process_hdr(FMTPOL_REQ, &CTX.watch_reqhdr, b, e, 1);
+				break;
+			case SLT_BerespUnset:
+			case SLT_RespUnset:
+				process_hdr(FMTPOL_RESP, &CTX.watch_resphdr, b, e, 1);
 				break;
 			case SLT_VCL_call:
 				if (!strcasecmp(b, "recv")) {
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -163,7 +163,44 @@ static struct ctx {
 	int			recv_compl;
 } CTX;
 
+enum format_policy {
+        FMTPOL_INTERNAL,
+        FMTPOL_REQ,
+        FMTPOL_RESP,
+};
+
 static void parse_format(const char *format);
+
+static unsigned
+frag_needed(const struct fragment *frag, enum format_policy fp)
+{
+	unsigned is_first, want_first, want_frag;
+
+        is_first = CTX.gen != frag->gen;
+
+        switch (fp) {
+        case FMTPOL_INTERNAL:
+                want_first = 1;
+                want_frag = 1;
+                break;
+        case FMTPOL_REQ:
+                want_first = *CTX.side == 'c';
+                want_frag = !CTX.recv_compl;
+                break;
+        case FMTPOL_RESP:
+                want_first = *CTX.side == 'b';
+                want_frag = (*CTX.side == 'c') || !CTX.recv_compl;
+                break;
+        default:
+                WRONG("Invalid format policy");
+        }
+
+        if (!want_frag)
+                return (0);
+        if (want_first && !is_first)
+                return (0);
+        return (1);
+}
 
 static void
 openout(int append)
@@ -851,7 +888,7 @@ isprefix(const char *prefix, size_t len, const char *b,
 }
 
 static void
-frag_fields(int force, const char *b, const char *e, ...)
+frag_fields(enum format_policy fp, const char *b, const char *e, ...)
 {
 	va_list ap;
 	const char *p, *q;
@@ -883,7 +920,7 @@ frag_fields(int force, const char *b, const char *e, ...)
 		assert(p != NULL && q != NULL);
 		if (p >= e || q <= p)
 			continue;
-		if (frag->gen != CTX.gen || force) {
+		if (frag_needed(frag, fp)) {
 			/* We only grab the same matching field once */
 			frag->gen = CTX.gen;
 			frag->b = p;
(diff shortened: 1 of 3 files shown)


bin/varnishncsa/varnishncsa.c
```diff
@@ -160,8 +160,26 @@ static struct ctx {
...
+
+#define UPDATE_REQ_C(CTX) (CTX.vcl_recv_seen ? UM_NEVER : UM_ALWAYS)
+#define UPDATE_REQ_B(CTX) (CTX.vcl_br_seen ? UM_NEVER : UM_ALWAYS)
+
+#define UPDATE_REQ(CTX) (*CTX.side == 'b' ? UPDATE_REQ_B(CTX) : UPDATE_REQ_C(CTX))
+
+#define UPDATE_RESP_C(CTX) (UM_ALWAYS)
+#define UPDATE_RESP_B(CTX) (CTX.vcl_br_seen ? UM_NEVER : UM_ALWAYS)
+
+#define UPDATE_RESP(CTX) (*CTX.side == 'b' ? UPDATE_RESP_B(CTX) : UPDATE_RESP_C(CTX))
```
dridi, line 181: I think I would prefer something like this:

```c
enum format_policy {
        FMTPOL_FIRST,
        FMTPOL_REQ,
        FMTPOL_RESP,
        FMTPOL_BEREQ,
        FMTPOL_BERESP,
};
```

Remember that you always have access to `CTX` to apply the policy, so we shouldn't need macros. A single function could tell you what to do:

```c
static unsigned
process_format(enum format_policy fp)
{

        /* switch (fp) => check CTX */
}
```

Then you can use that in functions like `frag_fields()` to bail out early:

```c
if (!process_format(fp))
        return;
```

It might prevent cascading enum arguments, I could only have a quick look at the patch series.

dridi (commented):
I think this could be better organized.

bin/varnishncsa/varnishncsa.c
```diff
@@ -160,8 +160,26 @@ static struct ctx {
 	const char		*handling;
 	const char		*side;
 	int64_t			vxid;
+	int			vcl_recv_seen;
+	int			vcl_br_seen;
```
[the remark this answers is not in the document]
walid-git: Wouldn't it be better to add some new vsl tag that indicates the end of reception from a peer so that we can rely on it in varnishncsa and avoid taking headers added by the core code before vcl_recv/vcl_backend_response into account ?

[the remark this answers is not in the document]
dridi: We should focus on the new model for NCSA formats here.

bin/varnishncsa/varnishncsa.c
walid-git: I have force pushed an updated version that includes your suggested changes.

bin/varnishncsa/varnishncsa.c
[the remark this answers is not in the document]
walid-git: updated to `recd_req` and `recd_beresp`

bin/varnishncsa/varnishncsa.c
```diff
@@ -916,7 +916,8 @@ frag_line(int force, const char *b, const char *e, struct fragment *f)
 }
 
 static void
-process_hdr(const struct watch_head *head, const char *b, const char *e)
+process_hdr(int force, const struct watch_head *head, const char *b,
+    const char *e)
```
dridi: The commit introducing the force flag is immediately undone by the next commit introducing the format policy, you can remove it entirely from the patch series.

bin/varnishncsa/varnishncsa.c
```diff
@@ -164,8 +164,26 @@ static struct ctx {
 	int			recd_beresp;
 } CTX;
 
+enum format_policy {
+        FMTPOL_FIRST = 0,
+        FMTPOL_ALWAYS
+};
```
dridi: Let's populate the enum entirely right off the bat with all policies, even if we initially honor a subset of the rules.

bin/varnishncsa/varnishncsa.c
```diff
@@ -884,7 +902,7 @@ frag_fields(int force, const char *b, const char *e, ...)
 		assert(p != NULL && q != NULL);
 		if (p >= e || q <= p)
 			continue;
-		if (frag->gen != CTX.gen || force) {
+		if (process_format_policy(fp, frag->gen != CTX.gen)) {
```
dridi, line 929: It looks like so far [1] that the "first" criterion is always a generation check. Maybe pass `frag` to `process_format_policy()` and derive first inside the function.

[1] commit "varnishncsa: Introduce process_format_policy()"

bin/varnishncsa/varnishncsa.c
```diff
@@ -166,15 +166,39 @@ static struct ctx {
...
         FMTPOL_FIRST = 0,
+        FMTPOL_REQ,
+        FMTPOL_RESP,
+        FMTPOL_BEREQ,
+        FMTPOL_BERESP,
         FMTPOL_ALWAYS
 };
 
+#define FMTPOL_REQ_SIDE() *CTX.side == 'c' ? FMTPOL_REQ : FMTPOL_BEREQ
+#define FMTPOL_RESP_SIDE() *CTX.side == 'c' ? FMTPOL_RESP : FMTPOL_BERESP
```
dridi: The enum looks a lot like what I suggested in a rushed review, so I'd like to apologize upfront for the alternative I'm going to suggest.

In the proposal from https://github.com/varnishcache/varnish-cache/issues/3528#issuecomment-2017621182 4 categories emerged:

- what we receive
- what we send
- internal events
- other

We should probably map the enum to the actual policies:

```c
enum format_policy {
        FMTPOL_SENT,
        FMTPOL_RECD,
        FMTPOL_INTERNAL,
};
```

Composing `CTX.side` with the format policy should be enough to circle back to which message was sent or received in `process_format_policy()`.

dridi, line 177: Side note since my new suggestion would obsolete the macros: make sure to guard complex expressions with parentheses in macro definitions, to be future-proof.

bin/varnishncsa/varnishncsa.c
```diff
@@ -1017,44 +1041,44 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
...
-				frag_fields(0, b, e,
+				frag_fields(FMTPOL_FIRST, b, e,
 				    1, &CTX.frag[F_h],
 				    0, NULL);
 				break;
 			case SLT_BereqMethod:
 			case SLT_ReqMethod:
-				frag_line(0, b, e, &CTX.frag[F_m]);
+				frag_line(FMTPOL_REQ_SIDE(), b, e, &CTX.frag[F_m]);
 				break;
```
dridi, line 1067: My suggestion implies that we could no longer share the same blocks for Req/Bereq switch cases, respectively Resp/Beresp.

bin/varnishncsa/varnishncsa.c
```diff
@@ -993,44 +1046,44 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
...
 				    3, &CTX.frag[F_I],
 				    4, &CTX.frag[F_O],
 				    0, NULL);
 				break;
 			case SLT_BackendOpen:
-				frag_fields(1, b, e,
+				frag_fields(FMTPOL_ALWAYS, b, e,
 				    3, &CTX.frag[F_h],
 				    0, NULL);
 				break;
```
dridi, line 1058: I think FMTPOL_ALWAYS should go away, %h is listed in the first-match "internal" category. This is the only place where we use it.

dridi (changes requested):
I think we can improve this change a bit more if we actually follow the proposal for new matching rules more closely.

bin/varnishncsa/varnishncsa.c
dridi: Forgot to check, so I just did and it looks like the generation is always the criterion, so please pass `frag` instead of a boolean.

walid-git:
Discussed this offline with @dridi. Force pushed a (hopefully) much cleaner version now. All review items should be addressed/obsolete.

nigoroll (approved):
I pass my vote on to @dridi . If he is happy with it now, I am too.

bin/varnishncsa/varnishncsa.c
```diff
@@ -163,8 +163,42 @@ static struct ctx {
 	int			recv_compl;
 } CTX;
 
+enum format_policy {
+        FMTPOL_FIRST = 0,
+        FMTPOL_REQ,
+        FMTPOL_RESP,
+};
```
dridi, line 170: My suggestion was to map the policy to the 3 categories "internal", "received" and "sent" and in our offline discussion I agreed to req/resp instead of sent/recd to make the change less intrusive than I suggested (and noted in https://github.com/varnishcache/varnish-cache/pull/4213#discussion_r1824229723).


`FMTPOL_FIRST` should really be `FMTPOL_INTERNAL` and there is no need to give it a specific value.

bin/varnishncsa/varnishncsa.c
```diff
@@ -163,8 +163,42 @@ static struct ctx {
...
+
 static void parse_format(const char *format);
 
+static unsigned
+process_format_policy(enum format_policy fp, unsigned first)
```
dridi, line 175: I requested that we pass the fragment instead of a boolean in https://github.com/varnishcache/varnish-cache/pull/4213#discussion_r1824213963 so the function signature should look like this:

```c
static unsigned
process_format_policy(enum format_policy fp, const struct fragment *frag);
```

And since we operate on a fragment, we can make it fit better with the rest of the code:

```c
static unsigned
frag_needed(const struct fragment *frag, enum format_policy fp);
```

This way we can completely centralize the policy, and call sites are easier to read too.

bin/varnishncsa/varnishncsa.c
```diff
@@ -163,8 +163,42 @@ static struct ctx {
...
+
+#define BACKEND_FLAG 2<<8
+	policy = fp | ((*CTX.side == 'c') ? 0 : BACKEND_FLAG);
+
+	switch (policy) {
+		case FMTPOL_REQ:
+		case FMTPOL_RESP | BACKEND_FLAG:
+			if (CTX.recv_compl)
+				return (0);
+			return (first);
```
dridi, line 190: In our offline discussion I hinted at decomposing the policy bits with boolean variables and returning a result in the end. Here we are still returning from everywhere and the new side-based macro is even more problematic.

We are nailing the semantics of our NCSA formats, it should be easy to understand the rules.

bin/varnishncsa/varnishncsa.c
```diff
@@ -163,8 +163,42 @@ static struct ctx {
...
+		case FMTPOL_REQ | BACKEND_FLAG:
+			if (CTX.recv_compl)
+				return (0);
+			return (1);
+		case FMTPOL_RESP:
+			return (1);
+		default:
+			WRONG("Illegal enum value");
+	}
+}
```
dridi, line 200: Putting it all together, this function should look like this instead:

```c
static unsigned
frag_needed(const struct fragment *frag, enum format_policy pol)
{
        unsigned is_first, want_first, want_frag;

        is_first = CTX.gen != frag->gen;

        switch (pol) {
        case FMTPOL_INTERNAL:
                want_first = 1;
                want_frag = 1;
                break;
        case FMTPOL_REQ:
                want_first = CTX.side == 'c';
                want_frag = !CTX.recv_compl;
                break;
        case FMTPOL_RESP:
                want_first = CTX.side == 'b';
                want_frag = (CTX.side == 'c') || !CTX.recv_compl;
                break;
        default:
                WRONG("Invalid format policy");
        }

        if (!want_frag)
                return (0);
        if (want_first && !is_first)
                return (0);
        return (1);
}
```

I have not tested it, but it should capture the semantics we established, and hopefully I'm not the only one thinking it's easier to read.

dridi (changes requested):
Something went terribly wrong with this new iteration. I probably didn't express myself correctly so I tried to be a lot more specific in this review.

walid-git:
Sorry for the mix-up, I have updated the PR now, let me know if it's still missing something.

bin/varnishncsa/varnishncsa.c
```diff
@@ -963,7 +963,12 @@ process_hdr(enum format_policy fp, const struct watch_head *head, const char *b,
 		CHECK_OBJ_NOTNULL(w, WATCH_MAGIC);
 		if (!isprefix(w->key, w->keylen, b, e, &p))
 			continue;
-		frag_line(fp, p, e, &w->frag);
+		if (unset) {
+			frag_line(fp, CTX.missing_string,
+			    CTX.missing_string + strlen(CTX.missing_string),
+			    &w->frag);
```
dridi, line 969: You could pass `CTX.missing_string, NULL` and make `frag_line()` find the end of the `b` string upon seeing a null `e` argument.

dridi (approved):
One polish suggestion.

walid-git:
bugwash: ok

I will update the docs and merge

PR #4215, opened 2024-10-16
Add a few more to gitignore

nyurik:
My repo ended up with these files in the root due to some build process. Not certain if others may run into the same, so adding to ignore.

```
vcc_if.c
vcc_if.c.tmp
vcc_if.c.tmp2
vcc_if.h
vcc_if.h.tmp
vmod_example.man.rst
vmod_example.man.rst.tmp
vmod_example.rst
vmod_example.rst.tmp
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,8 @@ cscope.*out
 /vmod/vcc_*_if.c
 /vmod/vcc_*_if.h
 /vmod/vmod_*.rst
+/vcc_if.*
+/vmod_*.*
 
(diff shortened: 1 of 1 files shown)


dridi:
Are you not in the wrong repository? There is no vmod_example here, but there is one in https://github.com/varnishcache/libvmod-example.

nyurik:
I am not certain how i ended up with these files. If both vmod and vcc_if are a result of some unusual action, i'll just clean them up

PR #4217, opened 2024-10-29
varnishncsa: Add hitmiss hitpass indicators to Varnish:handling

nigoroll:
The title should say it all. Simple straight forward extension of the handling strings.

@martin-uplex

### DIFF
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -1128,6 +1128,14 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 			case SLT_RespUnset:
 				process_hdr(FMTPOL_RESP, &CTX.watch_resphdr, b, e, 1);
 				break;
+			case SLT_HitMiss:
+				CTX.hitmiss = "miss";
+				CTX.handling = "hitmiss";
+				break;
+			case SLT_HitPass:
+				CTX.hitmiss = "miss";
+				CTX.handling = "hitpass";
+				break;
 			case SLT_VCL_call:
 				if (!strcasecmp(b, "recv")) {
 					CTX.recv_compl = 1;
@@ -1136,10 +1144,10 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
 				} else if (!strcasecmp(b, "hit")) {
 					CTX.hitmiss = "hit";
 					CTX.handling = "hit";
-				} else if (!strcasecmp(b, "miss")) {
+				} else if (!strcasecmp(b, "miss") && strcmp(CTX.handling, "hitmiss")) {
 					CTX.hitmiss = "miss";
 					CTX.handling = "miss";
-				} else if (!strcasecmp(b, "pass")) {
+				} else if (!strcasecmp(b, "pass") && strcmp(CTX.handling, "hitpass")) {
 					CTX.hitmiss = "miss";
 					CTX.handling = "pass";
 				} else if (!strcasecmp(b, "synth")) {
--- a/bin/varnishtest/tests/u00002.vtc
+++ b/bin/varnishtest/tests/u00002.vtc
@@ -1,0 +1,104 @@
+varnishtest "varnishncsa handling"
+
+server s1 -repeat 4 {
+	rxreq
+	txresp
+} -start
+
+varnish v1 -vcl+backend {
+	backend none none;
+
+	sub vcl_backend_fetch {
+		# XXX would like to do everything in v_b_e, but
+		#     return(pass(x)) is not supported
+		if (bereq.url == "/hitpass") {
+			set bereq.backend = s1;
+		} else {
+			set bereq.backend = none;
+		}
+	}
+
+	sub vcl_backend_response {
+		set beresp.ttl = 1y;
+		return (pass(1y));
+	}
+
+	sub vcl_backend_error {
+		set beresp.status = 200;
+		set beresp.ttl = 1y;
+		set beresp.body = bereq.url;
+		if (bereq.url ~ "^/hitmiss") {
+			set beresp.uncacheable  = true;
+		}
+		return (deliver);
+	}
+
+	sub vcl_miss {
+		if (req.url == "/hitmisspass" && req.is_hitmiss) {
+			return (pass);
+		}
+	}
+
+	sub vcl_recv {
+		if (req.url == "/pass") {
+			return (pass);
+		}
+		if (req.url == "/synth") {
+			return (synth(204));
+		}
+	}
+} -start
+
+client c1 {
+	# prime
+	txreq -url "/hitmiss"
+	rxresp
+	txreq -url "/hitmisspass"
+	rxresp
+	expect resp.status == 200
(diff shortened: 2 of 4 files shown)


bin/varnishtest/tests/u00002.vtc
```diff
@@ -0,0 +1,93 @@
...
+miss /hitpass
+miss /hit
+hitmiss /hitmiss
+hitpass /hitpass
+hit /hit
+pass /pass
+synth /synth
+EOF
+	varnishncsa -d -n ${v1_name} -F "%{Varnish:handling}x %U" >have
+	diff expect have
```
dridi, line 92: `diff -u` to make errors easier to review?

bin/varnishncsa/varnishncsa.c
```diff
@@ -1071,17 +1071,25 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
...
 					CTX.hitmiss = "hit";
 					CTX.handling = "hit";
-				} else if (!strcasecmp(b, "miss")) {
+				} else if (!strcasecmp(b, "miss") && strcmp(CTX.handling, "hitmiss")) {
 					CTX.hitmiss = "miss";
 					CTX.handling = "miss";
-				} else if (!strcasecmp(b, "pass")) {
+				} else if (!strcasecmp(b, "pass") && strcmp(CTX.handling, "hitpass")) {
 					CTX.hitmiss = "miss";
 					CTX.handling = "pass";
```
dridi, line 1094: Should we conditionally update only the hitmiss in these blocks?

In both `vcl_hit` and `vcl_miss` we can degrade to a pass, so technically hitmiss should be preserved (unless "-") while the handling should be unconditionally updated.

bin/varnishncsa/varnishncsa.c
```diff
@@ -1071,17 +1071,25 @@ dispatch_f(struct VSL_data *vsl, struct VSL_transaction * const pt[],
...
 				process_hdr(&CTX.watch_resphdr, b, e);
 				break;
+			case SLT_HitMiss:
+				CTX.hitmiss = "miss";
+				CTX.handling = "hitmiss";
+				break;
+			case SLT_HitPass:
+				CTX.hitmiss = "miss";
+				CTX.handling = "hitpass";
+				break;
```
dridi, line 1081: Shouldn't it be the other way around? At this point we set hitmiss to either "hitmiss" or "hitpass" (lookup outcome) and we leave the handling alone, it will be set by the relevant subroutine once we reach it.

dridi (commented):
I have the feeling that this change is happening backwards.

bin/varnishncsa/varnishncsa.c
nigoroll: My understanding is that `varnishncsa` users generally rely on `Varnish:hitmiss` to return either `hit` or `miss` in order to generate statistics and find misses. I would thus think changing this fundamental format would not be advisable.
`Varnish:handling`, however, already has more distinct values, providing more fine grained information on the nature of a hit or miss.

bin/varnishncsa/varnishncsa.c
nigoroll: I do not understand what you are trying to say. The intention is to preserve `hitmiss` and `hitpass` when we see a `call MISS` or `call PASS`. When `hitmiss` turns into a `pass`, the value ends up as `pass`. Is this what you are trying to say?

bin/varnishncsa/varnishncsa.c
nigoroll: FWIW, the format being called `Varnish:hitmiss` (read "hit or miss") is probably not the best choice in light of `hitmiss` as a lookup outcome.

gquintard: If I may: please break stuff

we have releases every 6 months, with full latitude to increment the major version. I don't say we should be as reckless as I'd hope us to be, but changing APIs isn't that big of a deal. Being correct and useful is more important

The current state of clients, as I see it is that `varnishncsa` is only a way to get data out of Varnish before shipping it to a log aggregator. These platform are fully equipped to then search based on regex, prefixes and whatnot. So the only true adjustment would be to update the search query from `== miss` to `== miss*` or similar.

nigoroll: @gquintard I do 💯% agree that _being correct and useful is more important_, this is, I hope, what I strive for all the time. And, fwiw, I think my concrete proposal in here does just that: It breaks `handling` by adding two more values.

However, I do not see the advantage of diversifying `hitmiss` over diversifying `handling`, which already _is_ more fine grained. If we have two values, of which one is actually redundant because it only depends on the other, then why should we spread additional information over the two?

Please give me a compelling reason.

dridi: Let's continue with the proposal in https://github.com/varnishcache/varnish-cache/pull/4217#discussion_r1822170362 that doesn't need to break anything.

bin/varnishncsa/varnishncsa.c
nigoroll: So you are proposing exactly the same plus `purge` plus `reset`, just as _yet another_ different tag, which represents a superset of %handling which represents a superset of %hitmiss?

To clarify: 🟢 exists already 🟡 proposed by this PR 🔵 proposed by Dridi in addition

- 🟢 `hit`
- 🟢 `miss`
- 🟡 `hitmiss`
- 🟢 `pass` (no effective lookup)
- 🟡 `hitpass`
- 🟢 `pipe` (no effective lookup)
- 🔵 `purge`
- 🟢 `synth` (no effective lookup)
- 🔵 `reset` (client disconnected before a lookup-or-not decision was made)

I have not checked if `reset` can be implemented properly, at this point I would guess it might be equivalent to the existing `-`. So all in all, the proposal is to also add `purge`, fine.

And should we have a new %format? @gquintard says no, @dridi says yes, I don't care. But I do not think that tag inflation helps much here, I would just keep %handling.

nigoroll: It seems like I forget to reply to some comments:

> What I had in mind was "%hitmiss/%handling" leading to for example "hitmiss/pass" after a return(pass) from vcl_miss in a hit-for-miss scenario.

Yes. How does this not work with the current patch? I added this to the test case: 143390be2

The `hitmiss` rightly gets turned into a `pass` because of a deliberate vcl return.

> What I mean by that is that a hit-for-pass transaction is handled the same as a regular pass transaction, whether you return(pass) from vcl_recv, vcl_hit, or vcl_miss. You essentially get a private fetch, this is how the request is handled. This is how I think %handling should be reported (and already is if I'm reading correctly).

😖 This patch proposes to report a difference between a regular `pass` and a `hitpass`. Before this patch, it was reported as `pass`, now as `hitpass`. That is intentional and I think it is helpful (unless we actually go for _yet another_ %format).

dridi: wfm

nigoroll:
I assume that phk is +0 as always on ncsa and I will thus go ahead here to not use up valuable bugwash time.

PR #4219, opened 2024-10-31
Refactor: Move V1L private pointer from wrk to the vdp entry

nigoroll:
The http1 line layer V1L is a bit special in that it is first used to send[^1] http headers and then, optionally, used as a delivery processor (VDP) to send the body.

Historically, V1L was contained directly in `struct worker` under the name WRW and [when it got separated](https://github.com/varnishcache/varnish-cache/commit/bc31f5c13883ce4e88a2d50603b3d7454d539a61), it retained its private pointer directly in `struct worker`.

From today's perspective and with the advent of more protocol support, having this special case seems like an anachronism, because we have everything in place to not require it.

This PR consists of two patches. The first prepares the ground by simple refactoring of `V1L_Open()` to return the `struct v1l *` rather than storing it in the `struct worker` itself. It contains some transient code to update `struct worker` outside `V1L_Open()`, this is only done to enable the split into two patches.

The actual change is in the second patch, which implements the goal of removing the `v1l` member from `struct worker`. This necessarily (but trivially) changes the V1L related function signatures, so the textual diff might look like a mouthful. But most of it is simple, really:

* Change `V1L_Close()` to take a `struct v1l **` 
* Change the `VDP_Push()` to pass the `struct v1l *` as the private pointer
* Adjust all other V1L related code to take `struct v1l *` instead of `struct worker *`

One detail which remains special has to do with the duality of V1L remaining in place: Because `V1L_Close()` remains responsible for closing the v1l, the VDP layer must not touch it. For this interaction to work, `V1L_Close()` needs to clear the private pointer in the VDP entry. The existing check in `VDP_Close()` for the private pointer to be cleared implies a check for V1L to be closed before VDP, so we can be certain that when the VDP closes, V1L is already closed.

Related note: Before this second iteration, I had also completed a patch series where V1L would be entirely moved under control of the VDP layer (that is, the VDP would need to be pushed whenever V1L was used, so also for the "no body" case), but I found this option too intrusive and implying too much overhead. So keeping V1L a bit special seemed like the better option.

[^1]: "send" is a simplification, actually the whole point about using V1L for both header and body is to _not_ write the headers to the HTTP1 connection individually, but rather batch both together into a single `writev()` call, if possible.


nigoroll:
I added two Flexelint commits with polishing found with less silencing. These are independent of the actual topic of this patch, but I wrote them on top of the other changed and would prefer to have them reviewed.

nigoroll:
bugwash: phk ok, walid needs more time

walid-git (approved):
LGTM

PR #4220, opened 2024-11-02
Transport disembark

nigoroll:
This is the next chapter of the book of AI, see #4209

To eventually get to event driven, asynchronous IO where a smaller number of threads handles a large number of requests, we also need a way to disembark a request from a worker and have another thread pick up where it left of, then reentering the state machine later to finish off.

Thanks to the already existing disembark facility used for waitinglist wakeups, implementing this is pretty straight forward: We basically only need to allow the vtr_deliver_f (transport deliver function) to tell the state engine to disembark, add a "finish" state, and everything else more or less works out of the box.

The last commit has a demonstration in a test case.

Everything else in this patch is refactoring.

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -537,6 +537,7 @@ struct req {
 	struct ws		ws[1];
 	struct objcore		*objcore;
 	struct objcore		*stale_oc;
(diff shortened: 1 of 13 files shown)


nigoroll:
force-pushed after #4224 merge

nigoroll:
bugwash: OK, need to resolve conflicts, will merge then

Issue #4195, opened 2024-09-23
Collection of issues for a possible point release


nigoroll:
Reminder: Discuss if we want a point release.

relevant:
* #4196
* #4154
* #4204 https://github.com/varnishcache/varnish-cache/commit/48a2a41cbc97498e044af7b7b3545427e3289872

cosmetic/maybe
* #4193 
* https://github.com/varnishcache/varnish-cache/commit/5dfbc6e2a0532460522da6868e11762d26a6c3ea (pid1 expectation)

(feel free to edit this note)

dridi:
I would add 5dfbc6e2a0532460522da6868e11762d26a6c3ea too.

stankolubomir:
Is there any possible ETA for this release? We have updated varnish to 7.6.0 and this issue https://github.com/varnishcache/varnish-cache/issues/4204 impacts our production environment. It would be great to have this patched as soon as possible.

nigoroll:
bugwash: @walid-git wants to roll a release by the end of next week

PR #4222, opened 2024-11-04
Backport series for 7.6.1

walid-git:
This is a set of backports in preparation for the 7.6.1 release.
Refs: #4195

### DIFF
--- a/bin/varnishd/common/common_vsmw.c
+++ b/bin/varnishd/common/common_vsmw.c
@@ -42,6 +42,7 @@
 #include <time.h>
 #include <unistd.h>
 #include <sys/mman.h>
+#include <sys/resource.h>
 #include <sys/stat.h>
 
 #include "vdef.h"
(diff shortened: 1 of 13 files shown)


doc/changes.rst
```diff
@@ -34,13 +34,17 @@ http://varnish-cache.org/docs/trunk/whats-new/index.html and via
 individual releases. These documents are updated as part of the
 release process.
 
-================================
-Varnish Cache 7.6.0 (2024-09-13)
-================================
+===============================
+Varnish Cache NEXT (2025-03-15)
```
nigoroll, line 38: I don't think we want this commit in a maintenance branch

doc/sphinx/reference/vsm.rst
```diff
@@ -121,3 +121,45 @@ varnishd are no longer relevant across namespaces.
 
 To disable liveness checks based on PIDs, the variable ``VSM_NOPID`` needs to
 be present in the environment of VSM readers.
+
```
nigoroll, line 124: There were two touchup commits to this one which I think we should also add (not critical)
f53bb248691a17849b46a4dbc897fa5b3861e000
5f650460977dd238af8e0d948a95caa85e617f2a

bin/varnishd/common/common_vsmw.c
```diff
@@ -275,6 +275,7 @@ vsmw_delseg(struct vsmw *vsmw, struct vsmwseg *seg)
 
 /*--------------------------------------------------------------------*/
 
+#ifdef RLIMIT_MEMLOCK
```
nigoroll, line 278: I had a follow up to this one, too: e87afd691d820d9fcb28f3621c4a23a3c0bbb780

doc/changes.rst
walid-git: I picked this commit to avoid the conflict in d98d7670956aa3bd3faf31b6d0e56eae0f5ea971, how do we deal with it in this case ? do we add `VDI_EVENT_SICK`  to the previously existing VRT section ?

walid-git: Thanks, I have updated the PR. I removed the commit and fixed the conflict in `vrt.h`. For `changes.rst`, that will be done in a separate PR later this week.

doc/sphinx/reference/vsm.rst
walid-git: Added both commits.

nigoroll (approved):

nigoroll:
Just so my apology has a place: I just pushed two commits to master which were meant to be just one

PR #4224, opened 2024-11-05
Facilitate simpler custom implementations of http1

nigoroll:
These two little changes help custom http1 implementations to only reimplement the transport's deliver callback without copying all of the code.


nigoroll:
bugwash ok

PR #4230, opened 2024-11-19
storage_simple: mark buffers

nigoroll:
Make sure that we do not confuse buffers with object data.

The flags field comes at no extra memory cost on 64bit. For 32bit, this change breaks binary compatibility with any out-of-tree storages using storage_simple.

### DIFF
--- a/bin/varnishd/storage/storage_simple.c
+++ b/bin/varnishd/storage/storage_simple.c
@@ -184,6 +184,7 @@ SML_AllocBuf(struct worker *wrk, const struct stevedore *stv, size_t size,
 	if (st == NULL)
 		return (NULL);
 	assert(st->space >= size);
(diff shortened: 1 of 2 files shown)


nigoroll:
bugwash: come back when the code making use of this is PRed

nigoroll:
The use case is now shown here: https://github.com/varnishcache/varnish-cache/pull/4209/commits/226d9ebfbe1a81ac0564da8f4c1b77a465f5e1c8. For the new VAI API, we want a "return a lease to storage" function with a lease representing either a segment of a storage object or a buffer. The caller should not need to be concerned with what it is, and we have different cases: When returned, temporary bufffers are to be freed always, while storage segments of objects are only to be freed if the object is transient ("a pass object").

PR #4103, opened 2024-05-07
build: Keep an autoconf check for pthread_setname_np()

dridi:
It is not portable by definition, so we should keep the guards.

PR #4231, opened 2024-11-19
Fix build on OpenBSD

catap:
Here a two trivial commit which unblock current master on OpenBSD.

### DIFF
--- a/bin/varnishd/acceptor/cache_acceptor.h
+++ b/bin/varnishd/acceptor/cache_acceptor.h
@@ -30,6 +30,8 @@
  *
  */
 
+#include <pthread.h>
+
 /* cache_acceptor.c */
 struct listen_sock;
 struct listen_arg;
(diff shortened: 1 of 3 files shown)


nigoroll:
Related #4103

nigoroll:
@catap I am not against having your suggestions on top of it. This has more to do with agreement within the team and reverting the appropriate change.

dridi:
> @nigoroll indeed, but @dridi patch doesn't include `pthread_set_name_np` which is the right name on OpenBSD. Plust it needs to resolve conflicts.

My focus was not on restoring the previous code that supported multiple non-portable pthread functions. The purpose of my patch was to bring back ifdef guards against a non-portable function just in case it wouldn't exist on some systems. I turned out to be on the right track.

catap:
@dridi thus, here a PR which is needed to be able to build varnish on OpenBSD again.

nigoroll:
@bsdphk can we leave the final verdict with you?

catap:
I had just rebased this PR into the last master.

Also, this fixes required for build Varnish on OpenBSD.

catap:
@nigoroll, @bsdphk ping?

PR #4233, opened 2024-11-21
fix: always wake the new head of the waiting queue when there is available connections

dmap:
There is a fundamental concurrency issue with the backend connection wait queue implementation.

With the current implementation, when there are items on the waiting queue, every in-progress backend request that complete signals the request at the head of the waiting queue to notify it that should resume obtaining a connection.  

However the current implementation assumes this means that if N in-progress requests complete then N waiting requests will be notified - but this is not the case.  This is because multiple in-progress requests can all complete simultaneously and all signal the same waiting request - the current head of the waiting queue.  Eventually that thread will obtain the lock and continue, however it reduces the maximum concurrency through the backend as N requests completed and only 1 resumed.

This patch we believe is a safe and effective way to fix the problem.  Each thread that awakes from the waiting queue simply checks to see if there is spare capacity in the connection pool and if so signals the next waiting thread that it too can continue.  When that next thread resumes it will also do the same check and so on down the queue, so eventually N threads will resume and peak concurrency is maintained.

We have run and verified both the behaviour of the unpatched code and the implementation with our patch under very high loads.  The unpatched code shows the expected (undesirable) behaviour when requests begin to queue - that is that the number of connections in use to the backend quickly drops below the maximum even though there are still new requests going on to the wait queue.  This in turn reduces overall throughput and degrades performance.   Eventually the queue gets long enough that the requests on the queue begin to timeout and fail - once enough requests off the queue fail the queue empties and normal throughput returns - until the next time requests begin to queue.

The patched code operates as expected, with the maximum number of connections maintained to the backend while there are requests on the wait queue.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -246,6 +246,12 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 	}
 	if (cw->cw_state != CW_BE_BUSY)
 		bp->n_conn++;
+
+	if (!VTAILQ_EMPTY(&bp->cw_head) && !BE_BUSY(bp)) {
+		/* Signal the new head of the waiting queue */
+		vbe_connwait_signal_locked(bp);
+	}
+
 	Lck_Unlock(bp->director->mtx);
 
 	if (cw->cw_state == CW_BE_BUSY) {


nigoroll:
bugwash: We will look at this and come back to it next week

nigoroll (approved):
Right. I can see that for a successful `vbe_dir_getfd()`, we need to check if we need to wake up another waiting thread, because the a number of `vbe_dir_finish()` calls could take the mutex before the woken up  `vbe_dir_getfd()` gets to run.
I agree that the patch looks clean and simple, thank you 👍🏽

dmap:
> Right. I can see that for a successful `vbe_dir_getfd()`, we need to check if we need to wake up another waiting thread, because the a number of `vbe_dir_finish()` calls could take the mutex before the woken up `vbe_dir_getfd()` gets to run. I agree that the patch looks clean and simple, thank you 👍🏽

Yes, exactly.  Thank you.

dridi:
Maybe the signal/broadcast operation should remove workers from the queue before waking them up. This fix looks as racy as the current behavior (thanks for the report by the way) so the solution should prevent this scenario altogether instead of optimistically waking up other workers.

Upon waking up we'd look at the state:

- queued => timed out, have worker remove itself from queue and fail
- dequeued => allowed to proceed, already removed from the queue
- etc

There's the broadcast case where we may want a state that tells us to give up because the backend went sick.

The semantics should probably be revisited to avoid loose ends.

dmap:
You are of course free to fix the bug however you think best but the fix we have proposed here I think has some merits.  I am not sure what appears racy about it?  All state checks and changes are done while holding the mutex, it cannot cause more threads to awake than there are available connections.  

It also has the advantage that we know it works (because we are using it) in a high traffic real world site.

nigoroll:
@dridi same question from me: Which race are you seeing?

On your suggestion:

until 4f36cd5900b451eafa8cf3729cce1a03717c0fdd we had `VTAILQ_REMOVE()` before sending the signal. But I think if we went back to something similar we would, besides creating additional complications, need to answer where to put a woken up thread back on the queue if it fails to get a connection. The current implementation with this change on top has, I think, the advantage of keeping a waiting fetch thread in the position it ought to be in until it either times out or succeeds.

bsdphk:
I'm OK with this.

dridi:
I don't have time to review this thoroughly, maybe ask @walid-git for a last call to review.

walid-git:
Although I think Dridi's suggestion is better, this one has the merit of being simpler and is actually fixing the bug.
I'm fine either way.

nigoroll:
> Although I think Dridi's suggestion is better, this one has the merit of being simpler and is actually fixing the bug. I'm fine either way.

Anyone is free to suggests improvements, I would just like to understand what the fixed issue or improvement is.

Issue #4232, opened 2024-11-19
Backend workspace overflow through lost header doesn't fail the response


stevendore:
### Expected Behavior

If a workspace overflow occurs due to a lost header, the request should fail.

### Current Behavior

If a workspace overflow occurs due to a lost header, deliver the response from server.

### Steps to Reproduce (for bugs)

The expectation would be that the second request would also return a 500 (or 503). 

```vcl
varnishtest "Workspace overflow through LostHeader"

varnish v1 -cliok "param.set http_max_hdr 32"

server s1 -repeat 2 {
	rxreq
	txresp
} -start

varnish v1 -vcl+backend {
	sub vcl_recv {
		if (req.url == "/1") {
			set req.http.make-overflow = "some value";
		}
	}
} -start

logexpect l1 -v v1 {
	expect * 1001 LostHeader 	"make-overflow: some value"
	expect * 1001 Error 		"out of workspace"
	expect * 1001 RespStatus 	"200"
	expect * 1001 VCL_return	"deliver"
	expect * 1001 Error 		"workspace_client overflow"

	expect * 1005 Error 		"out of workspace"
	expect * 1005 LostHeader 	"Accept-Encoding"
	expect * 1005 Error 		"out of workspace"
	expect * 1005 LostHeader 	"X-Varnish: 1005"
	expect * 1005 BerespStatus	"200"
} -start

# 1. Trigger client workspace overflow through LostHeader in vcl_recv
# 2. Go to vcl_backend_fetch and cause a backend workspace overflow and
#    still return a 200
# 3. Notice a workspace overflow after vcl_deliver and give client 500 response
client c1 {
	txreq -url "/1"    -hdr "h1: 1"   -hdr "h2: 2"   \
		-hdr "h3: 3"   -hdr "h4: 4"   -hdr "h5: 5"   \
		-hdr "h6: 6"   -hdr "h7: 7"   -hdr "h8: 8"   \
		-hdr "h9: 9"   -hdr "h10: 10" -hdr "h11: 11" \
		-hdr "h12: 12" -hdr "h13: 13" -hdr "h14: 14" \
		-hdr "h15: 15" -hdr "h16: 16" -hdr "h17: 17" \
		-hdr "h18: 18" -hdr "h19: 19" -hdr "h20: 20" \
		-hdr "h21: 21" -hdr "h22: 22" -hdr "h23: 23"
	rxresp
	expect resp.status == 500
} -run

[...]

PR #4235, opened 2024-11-25
WIP more WS_Overflowed() checks

nigoroll:
Re #4232 

Adding earlier overflow checks is pretty easy if we use VCL failure, but so far we had centralized the overflow handling on the client side in the `vtr_deliver_f`, sending downstream a 500 error and not a 503.
So my main question is: If we want consistency here, can we settle on simply handling workspace overflows as VCL failures? Or do we really want the separate error code still? If we do, improving the overflow detection on the backend side might become a bit more involved, because right now the backend side error just turns into a 503 on the client side.

Note that this PR is really for discussing the design questions, it does _not_ pass the test suite because of the changed error code.

### DIFF
--- a/bin/varnishtest/tests/r04232.vtc
+++ b/bin/varnishtest/tests/r04232.vtc
@@ -1,0 +1,61 @@
+varnishtest "Workspace overflow through LostHeader"
+
+server s1 {} -start
+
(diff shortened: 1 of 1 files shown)


nigoroll:
bugwash: turn it all into 500, update PR

PR #4236, opened 2024-11-26
Optimize HSH_Lookup

sobolevsv:
This PR optimizes the **HSH_Lookup** function to prevent Varnish Cache from becoming unresponsive for extended periods when handling a large number of bans and variations.

Earlier this year we encountered issues where Varnish would suddenly become unresponsive.

The issues always followed the same pattern. They occurred infrequently—about once a week or less, typically in the early morning on weekends. Before Varnish stopped responding, the **n_expired** counter would suddenly stop updating. Simultaneously, the rate of **bans_tests_tested** would increase while **bans_tested** remained unchanged. Then, the number of threads would increase until it reached the thread pool limit, at which point Varnish stopped accepting new connections.

Initially, we suspected a deadlock. We upgraded from version 5 to 6.6, but this did not resolve the issue. We also increased the thread pool size, which helped once—Varnish managed to unblock itself before the DevOps team restarted it. This indicated that the root cause was not a deadlock.

A thorough investigation revealed the following:

We used a TTL of 900 seconds for cacheable pages and 1 day for non-cacheable pages. Several non-cacheable pages were configured in the VCL file to bypass the cache. For other non-cacheable pages, hit-for-miss was configured with a TTL of 1 day.

After 1 day of uptime, Varnish's cache could contain 500,000–1.5 million objects, many of which are hit-for-miss objects (as showed the investigation).

We use lurker-friendly bans to invalidate objects in the cache. Each ban uses a regex to match objects by their URL. We add around 50k bans per hour. The lurker was unable to keep up with the rate of adding bans due to the large number of objects in the cache. After 1 day of uptime, Varnish could have as many as 300k active bans.

Additionally, we use Vary by User-Agent. While this is not ideal, it is part of our current configuration and it worked for years. As a result, a single page could be accessed by more than a thousand unique User-Agent daily, which creates a long list of variations for a single object in the cache.

Bans were added more or less at a constant rate throughout the day, but traffic significantly decreased during nighttime. Many objects were not requested overnight, meaning they needed to be checked against more bans when traffic resumed if the lurker had not yet processed them.

A rough estimate showed that the lurker could take up to 1 day to check all bans against all objects, so we can ignore its role in the investigation.

In the morning, as traffic increased, Varnish received a request for an object that had not been accessed overnight. This object needed to be checked against a large number of bans. If the requested object was non-cacheable with a TTL of 1 day, it also had to go through thousands of variations to find a match. Recently accessed objects were at the beginning of the list, so the probability of the requested object being near the end of the list was low. What is more likely is that a ban invalidating the requested object was added at the start of the night, followed by many more bans added throughout the night.

During lookup, once an **objhead** is found, Varnish needs to check all variants against all bans from the start of the ban list to the ban that invalidated the object. This process could take minutes to complete **(hundreds of thousands of bans multiplied by thousands of variations).** During this operation, the mutex **oh->mtx** is held, which blocks the expire thread. This explains why **n_expired** stopped updating. This mutex also blocks new incoming requests for the same object (page).  This explains why threads started increasing.

We fixed the configuration—reduced the TTL for hit-for-miss objects and added additional non-cacheable pages to bypass the cache. This helped to reduce the number of objects in the cache so the lurker could now keep up with newly added bans. As a result, the issue with Varnish disappeared. We will also revise the Vary header to pupolate the cache with fewer object varinats

While we had a not very good configuration that allowed our Varnish to work for years without any issues, a critical system software should not stop responding unexpectedly because of wrong or suboptimal configuration.

To prevent Varnish from becoming stuck unexpectedly for minutes, we need to swap the order of the Vary check and the Ban check. If we first find the object's variant that exactly matches the request and only then check it against all bans, the worker will return the object from the cache (or from backends) with a much smaller latency spike. The worker will hold the mutex for a much shorter time, allowing other requests for the same object to be served. Latency should be a top priority—returning a requested object to the client as quickly as possible. Let other object variants expire, or let the lurker clean them up asynchronously.


.gitignore
```diff
@@ -6,6 +6,7 @@ Makefile.in
 _*
 .deps/
 .libs/
+.vscode/
```
nigoroll, line 9: this should go in your global .gitignore

nigoroll (approved):
As you rightly explained, the root cause of your problems lies with the excessive use of variants and bans. Regarding variants, we introduced `vary_notice`, which does not avoid the problem, but at least helps awareness of it. Regarding bans, we added `ban_cutoff`, which, if used, would have avoided the problematic scenario you described.

That said, I can see that the change you propose makes sense, so :+1:

.gitignore
sobolevsv: fixed

sobolevsv:
Thank you for reviewing and approving my proposal, @nigoroll!
Regarding ban_cutoff - yes, I came across this during the investigation. I understood that it could serve as a workaround, but I wanted to identify the root cause. After pinpointing the issue—an excessive number of variations for non-cacheable objects—we addressed it by fixing the configuration, and the issue disappeared. Moving forward, we will consider both ban_cutoff and vary_notice as safety measures.

nigoroll:
@sobolevsv Yes, I agree that you did the right thing by tracking down the root cause.

Can you please squash your commits regarding `.gitignore` and force-push?

walid-git (commented):
Thanks for your contribution. I can only see one issue that could be problematic with this change: this is introducing a change of behavior in that a request matching a `req.*` ban expression would only evict the variant it matches, whereas previously all the variants would be evicted by the same request. This means that for variants that are only requested once in a while, the objects would stay for much longer, potentially until they expire (since the ban lurker thread does not act on `req.*` based bans). These objects would keep a reference to the ban, which will also take much longer to complete. 
I am not sure about the impact that this could have, but it seems like something we should think about.

nigoroll:
good point @walid-git. I agree that this could be a reason not to add this change, on the other hand we clearly discourage `req.*` bans.

asadsa92:
Maybe not relevant for a minor version bump, I agree, but would not a `req.*` bans to hit only the relevant variants be what the user expects? Maybe this needs be a doc fix? https://varnish-cache.org/docs/trunk/users-guide/purging.html
This seems to be defined for purges, but not bans.

nigoroll:
I find @asadsa92's argument compelling that this is actually a bug fix. Without checking Vary, we do not know if the ban should even apply.

dridi:
The workaround suggested here is simple and effective, but I think we should keep the current behavior since it appears to act as a better alarm than the `vary_notice`. I believe that the change is correct in the sense that we would no longer evaluate bans for objects we wouldn't consider for delivery, but it's just hiding the problem. However, in a sane environment with a low number of variants I would actually prefer a single lookup to test them all, especially for ban expressions with request criteria, to process them sooner (a better optimization than "correct behavior" in my opinion).

The problem faced here is "too many bans" times "too many variants" grinding lookups to a halt.

We already "solved" the ban accumulation with `ban_cutoff` but it does nothing by default. We should consider giving it a non-zero default value.

The variants accumulation could be solved with a `vary_limit` parameter that could default for example to 20 (twice `vary_notice`) that would turn new variants exceeding the limit into pass transactions. I was planning to submit (or ask someone to submit) such a change after #4073 where I briefly mention it in the pull request description.

walid-git:
FTR, here is a vtc that demonstrates the "bug" being fixed with this PR:

```shell
varnishtest "test ban + vary behavior"

server s0 {
	rxreq
	txresp -hdr "vary: version" -body "Variant A"
	rxreq
	txresp -hdr "vary: version" -body "Variant B"
	rxreq
	txresp -hdr "vary: version" -body "New variant A"
	rxreq
	txresp -hdr "vary: version" -body "New variant B"
} -start

varnish v1 -vcl+backend {} -start

client c1 {
	txreq -hdr "version: a"
	rxresp
	expect resp.body == "Variant A"
} -run

client c2 {
	txreq -hdr "version: b"
	rxresp
	expect resp.body == "Variant B"
} -run

varnish v1 -cliok "ban req.http.version == a"

# Should this remove a single variant from cache ?
client c3 {
	txreq -hdr "version: a"
	rxresp
	expect resp.body == "New variant A"
} -run

# It does not seem to be the case ..
client c4 {
	txreq -hdr "version: b"
	rxresp
	expect resp.body == "Variant B"
} -run
```

nigoroll:
> FTR, here is a vtc that demonstrates the "bug" being fixed with this PR:

Thank you. This is what I had in mind when I got confused during bugwash.

It should be noted that the VTC currently _fails_:

```
**** c4    c-l|New variant B
**** c4    bodylen = 13
**   c4    === expect resp.body == "Variant B"
---- c4    EXPECT resp.body (New variant B) == "Variant B" failed
```

So the question really is what the scope of a `req.*` ban is: Is it the hash or the object? We currently implement the former, but I lean towards the latter: my implicit understanding has always been that a `req.*` ban applies if a matching request would otherwise find the object, which is clearly not the case for the demonstration VTC.

walid-git:
> It should be noted that the VTC currently _fails_:
> 
> ```
> **** c4    c-l|New variant B
> **** c4    bodylen = 13
> **   c4    === expect resp.body == "Variant B"
> ---- c4    EXPECT resp.body (New variant B) == "Variant B" failed
> ```

Yes, it fails on master but passes with this PR. We could probably consider adding it to our test suite, as part of this PR.

sobolevsv:
@walid-git Thank you for the good example of what else this PR fixes. I added your test to the PR.

nigoroll:
bugwash:
because [phk's proposal](https://github.com/varnishcache/varnish-cache/pull/4236#issuecomment-2511729813) covers both cases, we want to take that route. The proposal basically is

```
(15:36:24) phk:     if (counter < threshold)
(15:36:27) phk:         ban-check
(15:36:38) phk:         counter++
(15:36:42) phk:     vary check
(15:36:53) phk:     if counter == threshold
(15:37:01) phk:         ban-check
```

for `threshold == 0`, we will get the POLA semantics at the expensive of potentially longer object lifetime. Proposed parameter names were `mandatory_ban_community_service`, `ban_variants`, `ban_vary_limit`m `max_ban_checks(_per_request)`.

@walid-git volunteered to write up a doc proposal.

walid-git:
Overtaken by #4253.

PR #4239, opened 2024-11-28
Add VARRAY_FOREACH and VPTRS_ITER macros to iterate over arrays and pointer arrays

nigoroll:
These patches add two macros, semantic patches to start using them and the respective changes:

The `VARRAY_FOREACH` macro is intended to simplify loops over elements of an array using pointer arithmetic instead of an array index.

The usual pattern of iterating over the array index and then accessing the i-th element of the array is replaced with calculating the end pointer and looping over the array elements.

The macro basically implements

```c
for (pointer = array, end = array + length; pointer < end; pointer++)
```

This pattern is more efficient but less clear to read when written out. The macro is similar to the `*_FOREACH` pattern which we already use in many other places with trees, lists and queues, but declares the loop variable locally.

The second macro `VPTRS_ITER` further simplifies our "array of pointers" case as, for example, used in struct strands.

Here, the loop variable is to be declared outside the macro to support keeping found elements from within the loop.

### DIFF
--- a/bin/varnishd/cache/cache_vrt.c
+++ b/bin/varnishd/cache/cache_vrt.c
@@ -379,11 +379,9 @@ VRT_HashStrands32(VCL_STRANDS s)
 
 	AN(s);
 	VSHA256_Init(&sha_ctx);
-	VARRAY_FOREACH(elem, s->p, s->n) {
-		p = *elem;
+	VPTRS_ITER(p, s->p, s->n)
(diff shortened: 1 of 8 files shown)


nigoroll:
bugwash discussed, not currently actionable

nigoroll:
I am going ahead with the idea in new code, and some generalization will arise naturally. We can then see if we want to elevate it later.

Issue #3833, opened 2022-08-04
vmod function/method name clash


nigoroll:
As demonstrated with this patch

```diff
diff --git a/vmod/vmod_debug.vcc b/vmod/vmod_debug.vcc
index df8ba6247..310a046bb 100644
--- a/vmod/vmod_debug.vcc
+++ b/vmod/vmod_debug.vcc
@@ -69,6 +69,8 @@ $Function STRING test_priv_top(PRIV_TOP, STRING)
 
 Test function for TOP private pointers
 
+$Function VOID obj_enum(ENUM { phk, des, kristian, mithrandir, martin })
+
 $Object obj(STRING string="default", ENUM { one, two, three } number=one)
 
 Test object

```

we do not support a function called `obj.name` with a method `name` for class `obj`.

Compiler error:

```
/usr/local/bin/python3.10 ../lib/libvcc/vmodtool.py --boilerplate -o vcc_debug_if ./vmod_debug.vcc
WARNING: Not emitting .RST for $Event event_function

WARNING: Not emitting .RST for $Event event_function

  CC       libvmod_debug_la-vcc_debug_if.lo
In file included from vcc_debug_if.c:10:
./vcc_debug_if.h:94:10: error: conflicting types for 'xyzzy_obj_enum'
VCL_VOID xyzzy_obj_enum(VRT_CTX, struct VPFX(debug_obj) *,
         ^
./vcc_debug_if.h:89:10: note: previous declaration is here
VCL_VOID xyzzy_obj_enum(VRT_CTX, VCL_ENUM);
         ^
vcc_debug_if.c:128:18: error: typedef redefinition with different types ('VCL_VOID (const struct vrt_ctx *, VCL_ENUM)' (aka 'void (const struct vrt_ctx *, const char *)') vs 'VCL_VOID (const struct vrt_ctx *, struct xyzzy_debug_obj *, VCL_ENUM)' (aka 'void (const struct vrt_ctx *, struct xyzzy_debug_obj *, const char *)'))
typedef VCL_VOID td_xyzzy_debug_obj_enum(VRT_CTX, VCL_ENUM);
                 ^
vcc_debug_if.c:32:18: note: previous definition is here
typedef VCL_VOID td_xyzzy_debug_obj_enum(VRT_CTX,
                 ^
vcc_debug_if.c:201:28: error: duplicate member 'obj_enum'
        td_xyzzy_debug_obj_enum         *obj_enum;
                                         ^
[...]

PR #4240, opened 2024-12-03
vmodtool: Add option to specify c names of arguments and avoid "bool"

nigoroll:
I tried gcc version 15.0.0 20241203 and even without -std=c23, it would reserve "bool":

```
In file included from vcc_std_if.c:10:
vcc_std_if.h:78:33: error: two or more data types in declaration specifiers
   78 |         VCL_BOOL                bool;
      |                                 ^~~~
```

(and more)

So this patch adds the option to specify the c name of arguments to vmod functions/methods by separating it with a colon:

	vclname:cname

We use this facility to rename the "bool" argument to vmod_std functions to "boola".

**Implementation**
    
 cname needs to be specified by the VMOD author because it is also used in the vmod implementation.
    
Obviously, VCC needs to know that name, too, to generate argument structs, so we need to transport it from the VMOD to VCC via the JSON spec. A logical place for  is next to the name, because the other attributes following in the argument spec array are optional. So this changes the JSON spec format, and, consequently, we need to increase the version number.
    
So, ultimately, this is a breaking change with respect to vmodtool: One can not import vmods built before this change. We could add compatibility for this case, but as we have a tradition of forcing rebuilds with each release by bumping the VRT major number anyway, I did not see my time well spent on implementing backwards compatibility.

### DIFF
--- a/vmod/vmod_std_conversions.c
+++ b/vmod/vmod_std_conversions.c
@@ -149,15 +149,15 @@ vmod_integer(VRT_CTX, struct VARGS(integer) *a)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 
-	nargs = a->valid_s + a->valid_bool + a->valid_bytes +
+	nargs = a->valid_s + a->valid_boolean + a->valid_bytes +
 	    a->valid_duration + a->valid_real + a->valid_time;
 
 	if (!onearg(ctx, "integer", nargs))
 		return (0);
 
 	r = NAN;
-	if (a->valid_bool)
-		return (a->bool ? 1 : 0);
+	if (a->valid_boolean)
+		return (a->boolean ? 1 : 0);
 
 	if (a->valid_bytes)
 		return (a->bytes);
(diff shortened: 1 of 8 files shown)


nigoroll:
To draft: I overlooked VCC for the argstruct.

vmod/vmod_std_conversions.c
```diff
@@ -149,15 +149,15 @@ vmod_integer(VRT_CTX, struct VARGS(integer) *a)
...
 	    a->valid_duration + a->valid_real + a->valid_time;
 
 	if (!onearg(ctx, "integer", nargs))
 		return (0);
 
 	r = NAN;
-	if (a->valid_bool)
-		return (a->bool ? 1 : 0);
+	if (a->valid_boola)
+		return (a->boola ? 1 : 0);
```
dridi, line 160: I'm not fond of "boola" here, why not "boolean" instead?

dridi (commented):
Excellent idea. I didn't look further than the diff.

vmod/vmod_std_conversions.c
nigoroll: yeah... `walid_boolean` ;)

nigoroll:
bugwash:
see how a generic "prefix for all c names" would look like

nigoroll:
bugwash: phk wants to prototype alternatives.

nigoroll:
I pondered this again: We have a similar case in #3833, which could be solved similarly to this suggestion.

So I made up my mind and I do not think that the option to explicitly specify the cname is the right way, while a "wholesale prefix" is not, because it introduces a lot of noise, reduces legibility of the code and might not even fix all problems, like #3833.

nigoroll:
Sorry for the delay on my end. I just promised to finish this RSN.

nigoroll:
FYI, next step I have in the work is some polishing of vmodtool and improved vtcs.

Issue #2107, opened 2016-10-04
HEAD requestes changed to GET


b3k:
It looks like varnish replaces HEAD requests to GET in before ```vcl_backend_fetch```.

Not sure if this is bug or feature, but I didn't found anything about that in documentation.

VCL looks like (less important parts was cut-off):
```
sub vcl_recv {
set req.http.X-Req-Method-Entry = req.method;
return (hash);
}

sub vcl_backend_response {
set beresp.http.X-Varnish-Method-Entry = bereq.http.X-Req-Method-Entry;
set beresp.http.X-Req-Method = bereq.method;
return (deliver);
}
```

varnishlog (headers: X-Req-Method-Entry and X-Req-Method):
```
-   ReqStart       185.24.255.255 16410
-   ReqMethod      HEAD
-   ReqURL         /api/rest/health
-   ReqProtocol    HTTP/1.1
-   ReqHeader      Host: 178.216.255.255
-   ReqHeader      User-Agent: curl/7.50.3
-   ReqHeader      Accept: */*
-   ReqHeader      X-Forwarded-For: 185.24.255.255
-   VCL_call       RECV
-   ReqUnset       Host: 178.216.255.255
-   ReqHeader      Host: 178.216.255.255
-   ReqURL         /api/rest/health
-   ReqHeader      X-Varnish-URL: /api/rest/health
-   ReqHeader      Surrogate-Capability: key=ESI/1.0
-   ReqHeader      X-Req-Method-Entry: HEAD
-   VCL_return     hash
-   VCL_call       HASH
-   VCL_return     lookup
-   VCL_call       MISS
-   VCL_return     fetch
-   Link           bereq 32771 fetch
-   Timestamp      Fetch: 1475573555.869367 0.079345 0.079345
-   RespProtocol   HTTP/1.1
-   RespStatus     200
-   RespReason     OK
-   RespHeader     Server: nginx
-   RespHeader     Date: Tue, 04 Oct 2016 09:32:35 GMT
-   RespHeader     Content-Type: application/json; charset=utf-8
-   RespHeader     Content-Length: 2
-   RespHeader     X-Powered-By: React/alpha
-   RespHeader     Set-Cookie: frontend=d8ce5dc1d896a1c3ef344f7a1b; expires=Tue, 04 Oct 2016 12:32:35 GMT; max-age=10800; domain=test.net; path=/; httponly
-   RespHeader     X-Varnish-URL: /api/rest/health
[...]

PR #4247, opened 2025-01-02
cache_req_fsm: keep the cache object's Content-Length for HEAD always

### LINKED ISSUE #4245: Content-Length is always 0 for content obtained from a response to a HEAD request
### Expected Behavior

When the backend returns a  `Content-Length` while replying to a HEAD request, it should be passed to the client.

### Current Behavior

When the content was fetched in response to a HEAD request, the client receives `Content-Length: 0`.

### Possible Solution

_No response_

1. Build Varnish from 772d7385dbe2cfa8b70d98af0b5da1236727e58b
2. Use the following VCL (see #2107):
```vcl
vcl 4.1;

backend default {
    # A Debian archive mirror
    .host = "129.102.1.37";
    .port = "80";
}


sub vcl_recv {
    unset req.http.X-Fetch-Method;
    if (req.restarts == 1) {
        return (hash);
    }
}

sub vcl_hash {
    if (req.restarts == 1) {
        hash_data(req.method);
    }
}

sub vcl_miss {
    if (req.method == "HEAD") {
        if (req.restarts == 0) {
            return (restart);
        }
        set req.http.X-Fetch-Method = "HEAD";
    }
}

sub vcl_backend_fetch {
    if (bereq.http.X-Fetch-Method) {
        set bereq.method = bereq.http.X-Fetch-Method;
    }
    unset bereq.http.X-Fetch-Method;
}
```
3. Start `tshark -f "host 129.102.1.37" -t ad -V` 
4. Execute `curl http://localhost/pub/debian/README.html -I`
5. tshark show the request:
```text
Hypertext Transfer Protocol                                                                                                                  
    HEAD /pub/debian/README.html HTTP/1.1\r\n                                                                                                
```
and the reply:
```text
Hypertext Transfer Protocol                                                                                                                  
    HTTP/1.1 200 OK\r\n                                                                                                                      
        [Expert Info (Chat/Sequence): HTTP/1.1 200 OK\r\n]                                                                                   
            [HTTP/1.1 200 OK\r\n]                                                                                                            
            [Severity level: Chat]                                                                                                           
            [Group: Sequence]                                                                                                                
        Response Version: HTTP/1.1                                                                                                           
        Status Code: 200                                                                                                                     
        [Status Code Description: OK]                                                                                                        
        Response Phrase: OK                                                                                                                  
    Date: Mon, 30 Dec 2024 16:42:24 GMT\r\n                                                                                                  
    Server: Apache\r\n                                                                                                                       
    X-XSS-Protection: 1\r\n                                                                                                                  
    Content-Security-Policy: default-src 'self' *.ircam.fr\r\n                                                                               
    X-Content-Type-Options: nosniff\r\n                                                                                                      
    Last-Modified: Sat, 09 Nov 2024 09:23:03 GMT\r\n                                                                                         
    ETag: "b67-626776a22999e-gzip"\r\n                                                                                                       
    Accept-Ranges: bytes\r\n                                                                                                                 
    Vary: Accept-Encoding\r\n                                                                                                                
    Content-Encoding: gzip\r\n                                                                                                               
    Content-Length: 967\r\n                                                                                                                  
        [Content length: 967]                                                                                                                
    Content-Type: text/html\r\n                                                                                                              
    \r\n                                                                                                                                     
    [HTTP response 1/1]                                                                                                                      
    [Time since request: 0.012166528 seconds]                                                                                                
    [Request in frame: 4]                                                                                                                    
    [Request URI: http://localhost/pub/debian/README.html]                                                                                   
```
6. The client receives this:
```
HTTP/1.1 200 OK
Date: Mon, 30 Dec 2024 16:42:24 GMT
Server: Apache
X-XSS-Protection: 1
Content-Security-Policy: default-src 'self' *.ircam.fr
X-Content-Type-Options: nosniff
Last-Modified: Sat, 09 Nov 2024 09:23:03 GMT
ETag: "b67-626776a22999e-gzip"
Vary: Accept-Encoding
Content-Type: text/html
X-Varnish: 3
Age: 0
Via: 1.1 ns3082119 (Varnish/trunk)
Accept-Ranges: bytes
Content-Length: 0
Connection: keep-alive
```

### Context

Sending a Content-Length of 0 looks livalid:
> in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET. 

according to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13

It also breaks some clients such as (Go-http-client used by Packer doesn't even attempts to send a GET after the HEAD as it assumes that the file is empty).

### Varnish Cache version

varnishd (varnish-trunk revision 772d7385dbe2cfa8b70d98af0b5da1236727e58b)

### Operating system

Debian 12

N/A

nigoroll:
Previously, we would only keep the Content-Length header for HEAD requests on hit-for-miss objects, now we simply keep it always to enable "fallback" caching of HEAD requests.

The added vtc implements the basics of the logic to enable the (reasonable) use case documented in
https://github.com/varnishcache/varnish-cache/issues/2107#issuecomment-2536642262 but using Vary instead of cache key modification plus restart.

Fixes #4245


bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -484,15 +484,8 @@ cnt_transmit(struct worker *wrk, struct req *req)
...
-			/*
-			 * Don't touch C-L header (debatable)
-			 *
-			 * The only way to do it correctly would be to GET
-			 * to the backend, and discard the body once the
-			 * filters have had a chance to chew on it, but that
-			 * would negate the "pass for huge objects" use case.
-			 */
+		} else if (head) {
+			req->resp_len = 0;
```
dridi, line 488: Maybe leave a minimal "Don't touch C-L header" comment?

dridi (commented):
LGTM, but this was not a thorough review.

bin/varnishtest/tests/r04245.vtc
```diff
@@ -0,0 +1,77 @@
...
+	    set req.http.X-Fetch-Method = "HEAD";
+	} else {
+	    unset req.http.X-Fetch-Method;
+	}
+    }
+
+    sub vcl_backend_fetch {
+	if (bereq.http.X-Fetch-Method) {
+	    set bereq.method = bereq.http.X-Fetch-Method;
+	}
```
sbraz, line 27: In this example configuration, the `X-Fetch-Method` headers can't be unset here before sending the request to the backend or it breaks the `Vary` part, right? I don't mind sending an extra header to my backend but it's one thing that differs from the `restart`-based solution.

nigoroll: Correct. We need to (un)set the header before the cache lookup such that the right variant gets hit, if present. For a miss, the header gets copied to the backend request and, when it completes (after `vcl_backend_response {}` returns), the header's value gets added to the Vary specification for that cache object.

So, in short, the header needs to be present during cache lookup and at the end of `vcl_backend_response {}`. For practical reasons, it is also needed to signal the backend side to activate the `Vary` handling.

With these requirements in mind, we can change the code to not send the header by deleting it in `vcl_backend_fetch {}` and restoring it in `vcl_backend_response {}`, but we need a vmod to do so. Here's how the test case adjustment looks like with a [`taskvar.bool`](https://gitlab.com/uplex/varnish/varnish-objvar/-/blob/master/src/vmod_taskvar.rst?ref_type=heads) object as a simple marker to activate the vary handling:

```diff
diff --git a/bin/varnishtest/tests/r04245.vtc b/bin/varnishtest/tests/r04245.vtc
index 27244e053..42982b79a 100644
--- a/bin/varnishtest/tests/r04245.vtc
+++ b/bin/varnishtest/tests/r04245.vtc
@@ -13,6 +13,12 @@ server s1 {
 } -start
 
 varnish v1 -vcl+backend {
+    import taskvar;
+
+    sub vcl_init {
+       new vary_x_fetch_method = taskvar.bool();
+    }
+
     sub vcl_recv {
        if (req.method == "HEAD") {
            set req.http.X-Fetch-Method = "HEAD";
@@ -24,13 +30,17 @@ varnish v1 -vcl+backend {
     sub vcl_backend_fetch {
        if (bereq.http.X-Fetch-Method) {
            set bereq.method = bereq.http.X-Fetch-Method;
+           # use marker to avoid sending the header to the backend
+           unset bereq.http.X-Fetch-Method;
+           vary_x_fetch_method.set(true);
        }
     }
 
     sub vcl_backend_response {
        # NOTE: this use of Vary is specific to this case, it is
        # usually WRONG to only set Vary for a specific condition
-       if (bereq.http.X-Fetch-Method) {
+       if (vary_x_fetch_method.get()) {
+           set bereq.http.X-Fetch-Method = bereq.method;
            if (beresp.http.Vary) {
                set beresp.http.Vary += ", X-Fetch-Method";
            } else {
```

For the purpose within the varnish-cache tree, we only use bundled vmods, so this change can not be applied to the proposed patch.

An even simpler way would be to use `bereq.method == "HEAD"` as the marker in `vcl_backend_response {}`, which should be possible if the additional logic is only used for `HEAD`. That is, it should work exactly as in the test case, but might cause trouble in real world VCL:

```diff
diff --git a/bin/varnishtest/tests/r04245.vtc b/bin/varnishtest/tests/r04245.vtc
index 27244e053..44edbd5bc 100644
--- a/bin/varnishtest/tests/r04245.vtc
+++ b/bin/varnishtest/tests/r04245.vtc
@@ -24,13 +24,15 @@ varnish v1 -vcl+backend {
     sub vcl_backend_fetch {
        if (bereq.http.X-Fetch-Method) {
            set bereq.method = bereq.http.X-Fetch-Method;
+           unset bereq.http.X-Fetch-Method;
        }
     }
 
     sub vcl_backend_response {
        # NOTE: this use of Vary is specific to this case, it is
        # usually WRONG to only set Vary for a specific condition
-       if (bereq.http.X-Fetch-Method) {
+       if (bereq.method == "HEAD") {
+           set bereq.http.X-Fetch-Method = bereq.method;
            if (beresp.http.Vary) {
                set beresp.http.Vary += ", X-Fetch-Method";
            } else {
```

nigoroll:
notes from bugwash:
* There should be a way for VCL to stop sending `C-L`. 

my own homework:
* understand why the current code works for pass

nigoroll:
homework: why does the current code work?
```diff
diff --git a/bin/varnishd/cache/cache_req_fsm.c b/bin/varnishd/cache/cache_req_fsm.c
index bbcb3824f..91ec23780 100644
--- a/bin/varnishd/cache/cache_req_fsm.c
+++ b/bin/varnishd/cache/cache_req_fsm.c
@@ -493,6 +493,7 @@ cnt_transmit(struct worker *wrk, struct req *req)
                         * filters have had a chance to chew on it, but that
                         * would negate the "pass for huge objects" use case.
                         */
+                       VSLb(req->vsl, SLT_Debug, "HEAD with OC_F_HFM");
                } else {
                        http_Unset(req->resp, H_Content_Length);
                        if (req->resp_len >= 0)
```

```
$ ./varnishtest -iv tests/b00065.vtc | grep -C 5 'HEAD wi'
**** v1    vsl|       1004 RespHeader      c Via: 1.1 v1 (Varnish/trunk)
**** v1    vsl|       1004 VCL_call        c DELIVER
**** v1    vsl|       1004 VCL_return      c deliver
**** v1    vsl|       1004 Timestamp       c Process: 1738839586.533446 0.014269 0.000088
**** v1    vsl|       1004 Filters         c 
**** v1    vsl|       1004 Debug           c HEAD with OC_F_HFM
**** v1    vsl|       1004 RespHeader      c Connection: keep-alive
**** v1    vsl|       1004 Timestamp       c Resp: 1738839586.533676 0.014500 0.000230
**** v1    vsl|       1004 ReqAcct         c 53 0 53 165 0 165
**** v1    vsl|       1004 End             c 
**** v1    vsl|       1003 SessClose       c REM_CLOSE 0.016
```

so the answer is: Because we set `OC_F_HFM` for passes.

nigoroll:
I the following, I refer to a response which, by definition, does not have an HTTP body ([CONNECT or HEAD request and any response with a 1xx (Informational), 204 (No Content), or 304 (Not Modified) status code](https://www.rfc-editor.org/rfc/rfc9112#section-6.3)) as _without body_, and all others as _with body_, even if the body may be empty.

>     * There should be a way for VCL to stop sending `C-L`.

I have pondered the _how_ and worked on an implementation, and for now, I am unhappy with what I have: `unset resp.body` seems wrong. It would mean for responses _with body_ to send a `Content-Length: 0`, while for responses _without body_ it would mean to clear the `Content-Length: 0` header. This really is un-pola.

Coming up with a better alternative is surprisingly hard, because, for most cases, we either create or recreate the `Content-Length` header when VCL has already finished. Also, trying to find a way to give VCL control over `Content-Length` gets messy soon because of streaming...

Hence, I lean towards a very simple solution to make the change after VCL has finished:

Add a "removeCL" filter, which

* for responses _without body_ just removes `Content-Length`
* for responses _with body_ removes `Content-Length` and prevents a body to be sent, that is, for chunked encoding, it also sends an end chunk.

This implies that filters will also need to run for responses without a body, but, at least from my perspective, this is already an overdue change for consistency.

The vcl interface would be simple, for example:

```vcl
sub vcl_deliver {
    set resp.filters += " removeCL";
}
```

@dridi I guess you might have opinions, do you?

dridi:
> Also, trying to find a way to give VCL control over `Content-Length` gets messy soon because of streaming...

We specifically made content-length and transfer-encoding read-only headers because of their role in HTTP framing (especially HTTP/1.x). So it shouldn't be direct control.

> Add a "removeCL" filter

This is interesting, but I don't really understand what you are proposing.

> * for responses _without body_ just removes `Content-Length`
> 
> * for responses _with body_ removes `Content-Length` and prevents a body to be sent, that is, for chunked encoding, it also sends an end chunk.

In both cases we end up without a body delivery, so this really looks like a case for `unset resp.body` (consistent with `unset bereq.body`). It would be both a simpler interface and simpler behavior to explain:

- discard the response body if there is one
- discard framing headers (both content-length and transfer-encoding)

We may want to also discard content-encoding if there was a body.

> This implies that filters will also need to run for responses without a body, but, at least from my perspective, this is already an overdue change for consistency.

No opinion, I haven't given much thought, but I am sensitive to the consistency argument. I'm pretty sure that filters today already fiddle with headers, so having headers-only filters make sense (making the `VDP::bytes()` and `VFP::pull()` callbacks optional).

I'm not a big fan of header manipulation in core or VMOD code, but I can several cases in favor of this kind of "rendez-vous point" ability to tweak headers before/after a delivery/fetch. Essentially cases where VCL syntax is too limited.

nigoroll:
Re @dridi 

> > Also, trying to find a way to give VCL control over `Content-Length` gets messy soon because of streaming...
> 
> We specifically made content-length and transfer-encoding read-only headers because of their role in HTTP framing (especially HTTP/1.x). So it shouldn't be direct control.

Yes, I understand what we did and why, but still this might not have been the best solution to the problem.

This ticket is about the response to `HEAD` requests with respect to the `Content-Length` returned to clients. We have the following different cases to consider:

1. Stable cache object with known length: The `Content-Length` header stored with the cache object might still be wrong, and we need correct it based on the actual length of the cached body data.

2. Busy object (streaming) with `Content-Length`: The header might turn out to be wrong, but when VCL runs, it is the best approximation we have.

3. Busy object with chunked encoding: We have no `Content-Length` at all.

If we wanted to have a correct `Content-Length` for the busy cases (2) and (3), we could wait for the object to be completely received, but that would not work with `transit_buffer`.

So, for the case of the `HEAD` request, we can send a "probably correct" `Content-Length` with (1), a "maybe correct" `Content-Length` with (2), and no `Content-Length` with (3). Which is what this PR does in its current form.

Now the bugwash decision was that VCL should have a way to _prevent_ sending `Content-Length` with a response to `HEAD`, and the current question is _HOW_.

In order to not repeat myself, please re-read https://github.com/varnishcache/varnish-cache/pull/4247#issuecomment-2640309332 with the above in mind. The problem is that `unset resp.body` is, I think, just wrong for the case of "prevent sending `Content-Length` in response to `HEAD`".

> > Add a "removeCL" filter
> 
> This is interesting, but I don't really understand what you are proposing.
> 
> > * for responses _without body_ just removes `Content-Length`
> > * for responses _with body_ removes `Content-Length` and prevents a body to be sent, that is, for chunked encoding, it also sends an end chunk.
> 
> In both cases we end up without a body delivery, so this really looks like a case for `unset resp.body` (consistent with `unset bereq.body`). It would be both a simpler interface and simpler behavior to explain:

It's not the same, because for `unset resp.body` we would preferably send `Content-Length: 0` in response to a `GET`.

nigoroll:
While @dridi and me should find concensus, I think this PR could be merged. The question how to avoid sending `Content-Length` is I think separate enough.

nigoroll:
The vtc I had added was doing all kinds of things, but not actually test for the `Content-Length`. Sorry for that. Fixed now in force-push

nigoroll:
bugwash: the "delete C-L" option should exist
Would be good to finish the discussion with @dridi

nigoroll:
@dridi ?

Issue #4249, opened 2025-01-06
VCL include +glob with relative path is not working


masarlabs:
### Expected Behavior

I created a minimal default.vcl that includes an empty file
```
vcl 4.1;

backend default {
  .host = "[127.0.0.1]:3081";
}

include       "test.d/test.vcl";
include +glob "/etc/varnish/test.d/*.vcl";
include +glob "test.d/*.vcl";
```

then test the code with
```
varnishd -C -f /etc/varnish/default.vcl
```

it should go without error

### Current Behavior

the first two includes work as expected, but the last one says

```
Message from VCC-compiler:
glob pattern matched no files.
('/etc/varnish/default.vcl' Line 12 Pos 15)
include +glob "test.d/*.vcl";
--------------##############-
Running VCC-compiler failed, exited with 2
VCL compilation failed
```

### Possible Solution

I think that current directory is not set correctly because it works with full path.

### Context

Cannot use relative path in vcl

### Varnish Cache version

varnishd (varnish-7.6.1 revision c3d5882003eb87e5e93dc09fb9513ca96db3ca3c)

### Operating system

alpine 3.21.0

masarlabs:
There is a typo, sorry, this is the error message
```
Message from VCC-compiler:
glob pattern matched no files.
('/etc/varnish/default.vcl' Line 12 Pos 15)
include +glob "test.d/*.vcl";
--------------##############-
Running VCC-compiler failed, exited with 2
VCL compilation failed

```

nigoroll:
bugwash agrees that with `+glob`, relative paths should be tried relative to `vcl_path` members.

walid-git:
@masarlabs  can you give #4250 a try ?

masarlabs:
Compiled
```
git clone https://github.com/walid-git/varnish-cache.git
cd varnish-cache/
git switch origin/fix_4249
sh autogen.sh
sh configure
make
make install
```
Version
```
# varnishd -V
varnishd (varnish-trunk revision 0edd8bcfda20ae5fbf2dd360ea809a687ef07379)
Copyright (c) 2006 Verdens Gang AS
Copyright (c) 2006-2024 Varnish Software
Copyright 2010-2024 UPLEX - Nils Goroll Systemoptimierung
```
Test
```
# varnishd -C -p vcl_path=/etc/varnish -f /etc/varnish/default.vcl
...
```
No error, it seems to work.
Thanks

walid-git:
[...]

PR #4250, opened 2025-01-07
VCC: Fix +glob with relative path

walid-git:
Fixes: #4249

### DIFF
--- a/bin/varnishtest/tests/c00053.vtc
+++ b/bin/varnishtest/tests/c00053.vtc
@@ -67,3 +67,35 @@ client c1 {
 	expect resp.http.foo == foo
 	expect resp.http.bar == bar
 } -run
+
+shell {
+	rm ${tmpdir}/top.vcl
+	mkdir -p ${tmpdir}/vclpath/vcl1
+	mkdir -p ${tmpdir}/vclpath/vcl2
+	echo 'sub vcl_synth { set resp.http.foo = "vcl1"; }' > ${tmpdir}/vclpath/vcl1/sub_foo.vcl
+	echo 'sub vcl_synth { set resp.http.bar = "vcl2"; }' > ${tmpdir}/vclpath/vcl2/sub_bar.vcl
+	cat > ${tmpdir}/vclpath/top.vcl <<-EOF
+		vcl 4.1;
+
+		backend default {
+			.host = "0:0";
+		}
+
+		include +glob "vcl*/sub_*.vcl";
+
+		sub vcl_recv {
+			return (synth(200));
+		}
+	EOF
+}
+
+varnish v1 -cliok "param.set vcl_path ${tmpdir}/vclpath"
+varnish v1 -cliok "vcl.load globtest top.vcl"
+varnish v1 -cliok "vcl.use globtest"
+
+client c1 {
+	txreq
+	rxresp
+	expect resp.http.foo == vcl1
+	expect resp.http.bar == vcl2
+} -run


walid-git:
One question that came to my mind regarding how we handle `+glob` includes: Should we include all the matching files  in all `vcl_path` directories, or do we only include the matching files from the first directory in `vcl_path`  where we have a match (as we do for regular includes) ?

nigoroll:
> One question that came to my mind regarding how we handle `+glob` includes

This is a very good point. My first reaction is to include all matches within all path elements, but I am not entirely sure.

walid-git:
> Right. But would it not make more sense to add an option to `VFIL_searchpath` to run the glob, rather than adding a different interface with some code duplication?

That could be an option. I will give it a try and see how that would look.

> This is a very good point. My first reaction is to include all matches within all path elements, but I am not entirely sure.

 That is also my opinion, but I need to think more about it.

walid-git:
> One question that came to my mind regarding how we handle `+glob` includes: Should we include all the matching files in all `vcl_path` directories, or do we only include the matching files from the first directory in `vcl_path` where we have a match (as we do for regular includes) ?

Bugwash: 
- By default, make +glob only expect to find a match in one of the paths in `vcl_path` and error out otherwise.
- Add another +all flag to include everything

walid-git:
New proposal: don't allow the use of relative paths with `+glob`, here is what the docs would look like:
```diff
diff --git a/doc/sphinx/reference/vcl.rst b/doc/sphinx/reference/vcl.rst
index 295a55e797..0bf4c6d15a 100644
--- a/doc/sphinx/reference/vcl.rst
+++ b/doc/sphinx/reference/vcl.rst
@@ -249,10 +249,19 @@ To include a VCL file in another file use the ``include`` keyword::
 
     include "foo.vcl";
 
+The file can be specified as either an absolute path, a relative path
+starting with "./", in which case the included file is searched
+relatively to the including VCL file, or a relative path which is
+searched in one of the ``vcl_path`` directories.
+
 Optionally, the ``include`` keyword can take a ``+glob`` flag to include all
 files matching a glob pattern::
 
-    include +glob "example.org/*.vcl";
+    include +glob "/etc/varnish/example.org/*.vcl";
+
+Note that using +glob with a relative path that does not start with "./"
+results in an error, which means that ``+glob`` includes cannot be searched in
+``vcl_path`` directories.
 
 Import statement
 ----------------
diff --git a/include/tbl/params.h b/include/tbl/params.h
index 7c093baa95..2495e565e7 100644
--- a/include/tbl/params.h
+++ b/include/tbl/params.h
@@ -1800,7 +1800,8 @@ PARAM_STRING(
        "VCL files in both the system configuration and shared "
        "data directories to allow packages to drop their VCL "
        "files in a standard location where relative includes "
-       "would work.",
+       "would work. Includes using +glob cannot be searched "
+       "in vcl_path.",
        /* flags */     BUILD_OPTIONS,
        /* dyn_min_reason */    NULL,
        /* dyn_max_reason */    NULL,
```

bin/varnishtest/tests/c00053.vtc
```diff
@@ -67,3 +67,35 @@ client c1 {
...
+			.host = "0:0";
+		}
+
+		include +glob "vcl*/sub_*.vcl";
+
+		sub vcl_recv {
+			return (synth(200));
+		}
+	EOF
+}
```
bsdphk, line 90: This should use the "filewrite" feature of v(arnish)test instead of shell.

(But maybe filewrite should have a "make dirs as necessary" option ?)

PR #4251, opened 2025-01-07
doc: Add more guidance to CONTRIBUTING

nigoroll:
mostly taken from @axboe's https://github.com/axboe/liburing/blob/master/CONTRIBUTING.md and edited.


nigoroll:
bugwash: point to jens' guide and name differences

Issue #3352, opened 2020-06-15
7.0 release checklist


nigoroll:
please collect todo/decision items here as a reminder checklist

* retire hash #3276 ?
* retire std.type2type

Dridi:
Recycle as 8.0 checklist?

nigoroll:
removed action label for now because the next release is going to be 7.7

nigoroll:
flagged for next bugwash because we need to make decisions

nigoroll:
notes on bugwash votes:
* retire hash +3: yes, because we remain interface compatible, we just swap out the implementation under the hood
* retire deprecated cookie.format_rfc1123 +3
* vmod_cookie overhaul -> ENOTIMPL
* (re)define VRT
https://github.com/varnishcache/varnish-cache/issues/3355 -> ENOTIMPL
* remove v_deprecated_ APIs: +3
* change  ban_any_variant to 0: +3  with very explicit note in release notes
* remove return(pipe) for unknown methods: +3
  * replace `return(pipe)` with `return(405)`
* change acl +fold default to on: +3

dridi:
> * replace `return(pipe)` with `return(405)`

I would argue 501 (Not Implemented) is more appropriate here.

gquintard:
random question: can we make `-g request` the default for `varnishlog`? It's what we want 95% of the time on support (both professionally and on the open interwebz). The current default means we usually miss the backend transaction with is where 50% of the juicy stuff happens

PR #4253, opened 2025-01-15
HSH: Limit the number of non matching variants evaluated against the ban list during lookup

walid-git:
This is an implementation of the solution suggested by @bsdphk in https://github.com/varnishcache/varnish-cache/pull/4236#issuecomment-2511729813

It adds a parameter to limit the number of possibly non matching variants that we evaluate against the ban list during a lookup.

Leaving this as a draft until we agree on the parameter name.

Refs: #4236

### DIFF
--- a/bin/varnishd/cache/cache_hash.c
+++ b/bin/varnishd/cache/cache_hash.c
@@ -371,6 +371,8 @@ HSH_Lookup(struct req *req, struct objcore **ocp, struct objcore **bocp)
 	const uint8_t *vary;
 	intmax_t boc_progress;
 	unsigned xid = 0;
+	unsigned ban_checks;
+	unsigned ban_any_variant;
 	float dttl = 0.0;
 
 	AN(ocp);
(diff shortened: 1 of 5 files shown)


nigoroll:
after bugwash discussion:
* we settled on `ban_any_variant` for the name
* the default should be changed to 0 with a major release (added a reminder to #3352)
* the parameter should be copied to a local variable for consistency, otherwise both `ban_checks` related conditions or none could become true

walid-git:
> * we settled on `ban_any_variant` for the name
> * the default should be changed to 0 with a major release (added a reminder to [8.0 release checklist #3352](https://github.com/varnishcache/varnish-cache/issues/3352))
> * the parameter should be copied to a local variable for consistency, otherwise both `ban_checks` related conditions or none could become true

Also added a paragraph in _purging.rst_  about the new `ban_any_variant` parameter.

nigoroll (approved):

walid-git:
bugwash: make docs clearer

walid-git:
Added https://github.com/varnishcache/varnish-cache/pull/4253/commits/744e840e633a20bf8a48cdb52cb2aeefe3f6d2a4. @bsdphk let me know if this is clearer now.

nigoroll:
I was not addressed, but I think the document improvement is very good.

PR #4254, opened 2025-01-15
vmod_blob: Add convencience functions for urldecode and urlencode

nigoroll:
For the common cases, the generic and versatile syntax is really cumbersome.

Why exactly did we not add these earlier?


simonvik:
Could it be worth writing somewhere that blob.urldecode might be unsafe and truncate values if the url contains %00?

nigoroll:
@simonvik good point.
Are there any established alternatives to handling `%00` like throwing an error?
If yes, I think we should rather review any `IDENTITY` encoding to `STRING`.

nigoroll:
This triggered  #4255

PR #4257, opened 2025-02-03
VCC: Prohibit the use of relative paths with +glob

### LINKED ISSUE #4249: VCL include +glob with relative path is not working

walid-git:
This is a replacement for https://github.com/varnishcache/varnish-cache/pull/4250, where [it appeared](https://github.com/varnishcache/varnish-cache/pull/4250#issuecomment-2575730478) that finding an unambiguous and easy to predict behavior when using `+glob` includes with relative paths involving `vcl_path` was not an easy task.

After some discussions, the consensus was that relative paths should not be allowed when using `+glob` includes.

Refs #4250
Closes #4249

### DIFF
--- a/bin/varnishtest/tests/c00053.vtc
+++ b/bin/varnishtest/tests/c00053.vtc
@@ -67,3 +67,8 @@ client c1 {
 	expect resp.http.foo == foo
 	expect resp.http.bar == bar
 } -run
+
+varnish v1 -errvcl "+glob can only be used with absolute paths or relative paths starting with './'" {
+	include +glob "sub_*.vcl";
+	backend default none;
+}
--- a/doc/sphinx/reference/vcl.rst
+++ b/doc/sphinx/reference/vcl.rst
@@ -258,7 +258,11 @@ The included file can be specified as follows:
 Optionally, the ``include`` keyword can take a ``+glob`` flag to include all
 files matching a glob pattern::
 
-    include +glob "example.org/*.vcl";
+    include +glob "/etc/varnish/example.org/*.vcl";
+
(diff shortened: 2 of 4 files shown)


nigoroll (approved):

nigoroll:
bugwash OK

walid-git:
bugwash: ok

nigoroll:
FTR, I just came across precedent where, I think, the same decision had been taken decades ago:

https://perldoc.perl.org/perlfunc#do-EXPR

> Using do with a relative path (except for ./ and ../), like `do 'foo/stat.pl';` will search the [@INC](https://perldoc.perl.org/perlvar#%40INC) directories

PR #4259, opened 2025-02-04
vdire: plug the remaining vcl temperature change races

### LINKED ISSUE #4199: Panic on vcl.load: Wrong turn at cache/cache_backend_probe.c:710: running

nigoroll:
The context for this change is #4142, but while it builds upon it, it is still causally unrelated. This PR replaces #4205

Problem
=======

To motivate this change, let's look at `vcl_set_state()` as of before the patch, and consider some non-issues and issues:

transition to COLD
------------------

```c
                        Lck_Lock(&vcl_mtx);
                        vcl->temp = VTAILQ_EMPTY(&vcl->ref_list) ?
                            VCL_TEMP_COLD : VCL_TEMP_COOLING;
                        Lck_Unlock(&vcl_mtx);
			// *A*
                        vcl_BackendEvent(vcl, VCL_EVENT_COLD);
```

non-issues:

At point *A*, backends could get added. For `VCL_TEMP_COOLING`, this fails (see `VRT_AddDirector()`). For `VCL_TEMP_COLD`, the state is consistent, because backends get created cold and that's it.

At point *A*, backends could get removed. `vdire_resign()` from #4142 ensures that the temperature passed to `vcldir_retire()` is consistent with the events the backends have received.

transition to WARM (success)
----------------------------

```c
                        Lck_Lock(&vcl_mtx);
                        vcl->temp = VCL_TEMP_WARM;
                        Lck_Unlock(&vcl_mtx);
			// *B* ...
                        i = vcl_send_event(vcl, VCL_EVENT_WARM, msg);
                        if (i == 0) {
				// ... *B*
                                vcl_BackendEvent(vcl, VCL_EVENT_WARM);
                                break;
                        }
```

issues:

(1) In region *B*, backends could get removed. They will not have received a WARM event, but will be called with a WARM temperature, so they will receive a bogus COLD event.

(2) Also in region *B*, backends could get added. They will implicitly receive a WARM event (see `VDI_event()` in `VRT_AddDirector()`), and then another from `vcl_BackendEvent()`

transition to WARM (failed)
---------------------------

```c
                        AZ(vcl_send_event(vcl, VCL_EVENT_COLD, &nomsg));
                        AZ(nomsg);
                        vcl->temp = VCL_TEMP_COLD;
```

issues:

(3) Backends added in region *B* will have received the implicit WARM event, and thus need a COLD event for the "cancel".

Solution
========

To solve these issues, we need to do two things:

There needs to be some kind of transaction which begins with the temperature change and ends when all backends have been notified appropriately. Backends can not get deleted while the transaction is in progress.

We need a notion of "backends from before the temperature change" and "backends from after".

The first part is already delivered by #4142: The vdire facility already provides the transaction during which backends do not actually get deleted and it ensures that the right temperature gets passed when the deletion is carried out. So for this part, we only need to use vdire.

But issues (2) and (3) are not yet covered. For these, we add a checkpoint, such that we know which directors from the list are the "base" from before the temperature change and which are the "delta" added after it.

That's this patch.

`vdire_start_event()` atomically (under the `vcl_mtx`) sets the checkpoint and the new temperature.

`vdire_end_event()` just uses the existing `vdire_end_iter()` and clears the checkpoint.

`vcl_BackendEvent()` gets split into two:

`backend_event_base()` notifies backends from before the checkpoint. `backend_event_delta()` atomically sets a new temperature and notifies backends from after the checkpoint (but not from after its temperature change).

Fixes #4199


nigoroll:
One addition to the bugwash discussion: Dynamic backend's temperature events are, by design, not tied to VCL temperature events. When they are created on a warm vcl, they receive a warm event, and, conversely, a cold event when they are removed.

bin/varnishd/cache/cache_vcl.c
```diff
@@ -567,11 +602,54 @@ vcl_BackendEvent(const struct vcl *vcl, enum vcl_event_e e)
...
 	AZ(vcl->busy);
 	vdire = vcl->vdire;
+	AN(vdire->iterating);
 
-	vdire_start_iter(vdire);
-	VTAILQ_FOREACH(vdir, &vdire->directors, directors_list)
+	if (vdire->checkpoint == NULL)
+		return;
+
+	VTAILQ_FOREACH(vdir, &vdire->directors, directors_list) {
```
walid-git, line 610: Could we use `VTAILQ_FOREACH_REVERSE_FROM` here, where `from` is `vdire->checkpoint` ? That would avoid the:
```
if (vdir == vdire->checkpoint)
        break;
```

bin/varnishd/cache/cache_vcl.c
```diff
@@ -567,11 +602,54 @@ vcl_BackendEvent(const struct vcl *vcl, enum vcl_event_e e)
...
+
+	ASSERT_CLI();
+	CHECK_OBJ_NOTNULL(vcl, VCL_MAGIC);
+	AZ(vcl->busy);
+	vdire = vcl->vdire;
+	AN(temp);
+	AN(vdire->iterating);
+
+	Lck_Lock(vdire->mtx);
+	if (vdire->checkpoint == NULL)
```
walid-git, line 638: Same here, I think we can use `VTAILQ_FOREACH_FROM`, that already takes care of checking if `checkpoint == NULL` and fall back to `VTAILQ_FIRST` when it is the case.

walid-git (approved):
LGTM (after addressing previous points)

bin/varnishd/cache/cache_vcl.c
nigoroll: In this case, I do not think that using `VTAILQ_FOREACH_FROM` is helpful, because we want to start at either the first element or the element _after_ the checkpoint. So we need `vdir = VTAILQ_NEXT(vdire->checkpoint, directors_list);` anyway, and if _that_ is `NULL`, we do not want to iterate at all, so we do not want to fall back to `VTAILQ_FIRST()` as `VTAILQ_FOREACH_FROM` would. Plus we also have a custom termination element taken under the lock, which we then check for unlocked.

bin/varnishd/cache/cache_vcl.c
nigoroll: I added this change. Besides switching to `VTAILQ_FOREACH_REVERSE_FROM`, it also adds magic checks.

```diff
diff --git a/bin/varnishd/cache/cache_vcl.c b/bin/varnishd/cache/cache_vcl.c
index 0dd480be7..cc9c109d1 100644
--- a/bin/varnishd/cache/cache_vcl.c
+++ b/bin/varnishd/cache/cache_vcl.c
@@ -604,13 +604,15 @@ backend_event_base(const struct vcl *vcl, enum vcl_event_e e)
        vdire = vcl->vdire;
        AN(vdire->iterating);
 
-       if (vdire->checkpoint == NULL)
+       vdir = vdire->checkpoint;
+       if (vdir == NULL)
                return;
 
-       VTAILQ_FOREACH(vdir, &vdire->directors, directors_list) {
+       CHECK_OBJ(vdir, VCLDIR_MAGIC);
+       VTAILQ_FOREACH_REVERSE_FROM(vdir, &vdire->directors,
+           vcldir_head, directors_list) {
+               CHECK_OBJ(vdir, VCLDIR_MAGIC);
                VDI_Event(vdir->dir, e);
-               if (vdir == vdire->checkpoint)
-                       break;
        }
 }

```

Issue #3963, opened 2023-07-31
Via backends: Sending invalid SNI headers


delthas:
The SNI spec states that the SNI header should contain exactly a hostname: neither IP addresses nor ports.

Currently, `.authority` defaults to `.host_header`, then `.host`.
- Typical host headers are `example.com:1234`, `1.2.3.4`. In those cases, Varnish would by default send an invalid authority PROXY TLV (translated to an invalid SNI header).
- Another example is if there is no `.host_header`, but an IP literal in `.host`: `.host = "1.2.3.4";`. In this case, Varnish would send an IP address in the authority by default

I see two possible solutions here:
- Not fallback to `.host_header` then `.host`, just `.host` -> fixes the first issue
- Checking the `.authority` value if it was a fallback, and if it is not a plain hostname, drop it entirely
- We could also mention it in the docs (as in, "you can override the authority if it's an IP address"), but I think it would be nice to have a default behavior where no invalid SNI header is sent

PR #4264, opened 2025-02-07
vcc_backend: Forbid ip addresses as .authority with .via

nigoroll:
Second half of the fix for #3963:

So far, we deliberately allowed IP addresses in the `.authority` field of backends, which might be derived from `.host_header` or `.host`.

There is a fair argument in #3963 that this should not be allowed.

This patch adds a `getaddrinfo()` call via `VSS_ResolveOne()` to check if the value parses as an ip address and, if so, fails VCL compilation.


nigoroll:
OK from bugwash

PR #4267, opened 2025-02-09
docs: clarify that A-E header needs to be set on bereq.http for pass requests

kaiburjack:
We had a hard to track issue on one of our production systems that resulted in Varnish sending compressed content to clients that did not advertize it in Accept-Encoding:gzip (some old smart televisions) for POST requests (that were a pass in Varnish).

While I was following the advise on: https://varnish-cache.org/docs/trunk/phk/gzip.html#tuning-tweaking-and-frobbing 
> In vcl_pass{} the clients Accept-Encoding header is copied to the backend request unchanged. Even if the client does not support gzip, you can force the A-C header to “gzip” to save bandwidth between the backend and varnish, varnish will gunzip the object before delivering to the client.

in order to reduce bandwidth between the backend and Varnish, I accidentally did the following:
```
sub vcl_pass {
  set req.http.Accept-Encoding = "gzip";
}
```
Together with (minus any checks for eligible Content-Types of course):
```
sub vcl_backend_response {
  set beresp.do_gzip = true;
}
```
and the backend responding with gzip responses when Varnish asks for it with Accept-Encoding:gzip, this resulted in clients that *do not* advertize gzip support to _still_ receive gzip compressed responses.

After a discussion with @gquintard on the Varnish discord, he pointed out that the issue actually was me setting Accept-Encoding on **req.http** in **vcl_pass**, while I should have done that on the **bereq.http** object in **vcl_backend_fetch** instead.

This PR makes sure that when people follow the advice on the page https://varnish-cache.org/docs/trunk/phk/gzip.html#tuning-tweaking-and-frobbing about still making the backend respond with gzip for passes, that they also use bereq.http in vcl_backend_fetch and not req.http in vcl_pass, as the current documentation may intend (and me falling for it).

### DIFF
--- a/vmod/vmod_debug.c
+++ b/vmod/vmod_debug.c
@@ -1353,12 +1353,14 @@ xyzzy_log_strands(VRT_CTX, VCL_STRING prefix, VCL_STRANDS subject, VCL_INT nn)
 	AN(subject);
 	if (nn > INT_MAX)
 		n = INT_MAX;
+	if (nn < 0)
+		n = 0;
 	else
 		n = nn;
 
 	for (i = 0; i < subject->n; i++) {
 		const char *p = subject->p[i];
 		mylog(ctx->vsl, SLT_Debug, "%s[%d]: (%s) %p %.*s%s", prefix, i,
-		    ptr_where(ctx, p), p, n, p, strlen(p) > n ? "..." : "");
+		    ptr_where(ctx, p), p, n, p, strlen(p) > (unsigned)n ? "..." : "");
 	}
 }


nigoroll:
First of all: thank you! Then regarding GH showing this PR as closed:

GH is weird. I followed a command line workflow which I thought was valid, but somehow it resulted in GH closing the PR:

```
git remote add kaiburjack git@github.com:kaiburjack/varnish-cache.git
git fetch kaiburjack
git checkout -b kaiburjack_master kaiburjack/master
git rebase master
git push -f kaiburjack master
git checkout master
git merge kaiburjack_master
git push
```

I guess anything which works without clicking a button is now evil to Microsoft...

dridi:
Maybe this should have been handled by @bsdphk since "PHK's random outbursts" is his tribune in the documentation.

The is specific page goes way back, and maybe now is a good time to revisit all of them to adjust them to today's reality. Maybe this should be a biannual task before March/September releases. (Last time I checked "security barriers" the diagram was still pretty accurate!)

PR #4269, opened 2025-02-11
Stash objcore references until the end of the task to avoid copies

nigoroll:
This proposal was motivated by #3768, which is about avoiding to make copies of constant strings by special casing. This PR does not yet include one additional detail from #3768 [^1], but it solves the underlying root cause.

### Context

Any reference handled in VCL needs to have at least [`PRIV_TASK`](https://varnish-cache.org/docs/trunk/reference/vmod.html#private-pointers-memory-management) lifetime. We notoriously shied away from formalizing this definition, but it is a factual consequence from how workspaces work. Rollbacks reset the respective task's workspace and thus finalize all `PRIV_TASK`s.

One way of ensuring `PRIV_TASK` lifetime is by copying the referenced value (usually a string) to the task's workspace, and we do this today.

Yet this is wasteful, because static strings from VCL and pointers to memory on the heap already outlive the task lifetime.

The only objects which did not already have `PRIV_TASK` lifetime were attributes from objects, because object references got returned before restarts. [b92.vtc](https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishtest/tests/b00092.vtc) illustrates this case.

For this reason and this reason only do we currently copy all strings to the respective workspace.

### Avoid copies by giving object references task lifetime

This proposal gives object references the same lifetime as `PRIV_TASK` and removes the then unnecessary workspace copies. As a side effect, it also solves the case for #3768 [^1], because this also avoids most copies of static strings to workspace.

In the past, an argument had been made (IIRC by @mbgrydeland) that keeping object references until the end of the task would increase their lifetime by too much, but restarts in VCL really should be done within milliseconds in most cases - and if keeping references _is_ an actual problem in specific situations, those can be avoided by either not restarting or rolling back also. In general, until now we have charged all Varnish-Cache users with the cost for specific use cases, but we should rather only charge the specific use cases instead.

---

[^1]: When we set an existing header to a new header with the same name as in `set req.http.name = resp.http.name`, we currently create a new HEADER on the workspace. This could be avoided by looking left of the string if it is `<LEN>name: ` where it is safe.


bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -78,6 +80,68 @@
...
+	struct ocstash *stash;
+	unsigned u;
+
+	AN(stashp);
+	if (*stashp == NULL)
+		return;
+	TAKE_OBJ_NOTNULL(stash, stashp, OCSTASH_MAGIC);
+	assert(stash->n <= stash->l);
+	for (u = 0; u < stash->n; u++)
+		(void)HSH_DerefObjCore(wrk, &stash->ocs[u], HSH_RUSH_POLICY);
```
dridi, line 108: We can't take any objcore and assume they're all supposed to wake `rush_exponent` requests up.

bin/varnishd/cache/cache.h
```diff
@@ -117,6 +117,7 @@ struct cli;
 struct http_conn;
 struct listen_sock;
 struct mempool;
+struct ocstash;
```
dridi, line 120: Using this line as an anchor to mention that you had a text editor accident with your first patch. This line leaked from the git commit template at the end of the first commit message:

> t Please enter the commit message for your changes. Lines starting

dridi (commented):
> In the past, an argument had been made (IIRC by @mbgrydeland) that keeping object references until the end of the task would increase their lifetime by too much, but restarts in VCL really should be done within milliseconds in most cases - and if keeping references is an actual problem in specific situations, those can be avoided by either not restarting or rolling back also.

A restart occurring after `vcl_miss` can take orders of magnitude longer than milliseconds.

I'm not convinced this is a good idea. I think objcores are resources we should always try to hold onto sparingly, and workspace allocations are an acceptable trade off.

bin/varnishd/cache/cache_req_fsm.c
nigoroll: Right. The call in `cnt_fetch` which changed the rush parameter from 1 to HSH_RUSH_POLICY is actually not needed - we only need to consider the cases where VCL had a chance to look at the object.

nigoroll:
> A restart occurring after `vcl_miss` can take orders of magnitude longer than milliseconds.

For this scenario to be relevant, it would need to be a `return(restart)` from `sub vcl_deliver {}`, which then ends up being a _miss_. But yes, I am aware and I was specifically referring to this argument in the last paragraph of the initial comment.

> I think objcores are resources we should always try to hold onto sparingly

What would be the convincing argument for that?

I brought my case, so if you disagree, what is yours?

bin/varnishd/cache/cache_req_fsm.c
nigoroll: But you made me realize that I had overlooked `sub vcl_hit {}` in `cnt_lookup()`, which has access to `obj.*`. Thank you!

dridi: I should have said "after `vcl_hash`" in my review comment, because a grace hit could actually take significant time to resume.

nigoroll: A grace hit is immediate. Do you mean a waitinglist hit?

dridi: A grace hit needs to wait for the request to be released, which can take significant time for a couple reasons. First, the bgfetch task may be queued if we are out of threads. Second, if the request has a body, it can take until it's actually consumed, at which point at least one timeout is involved (again, orders of magnitude above milliseconds).

dridi (commented):
> But yes, I am aware and I was specifically referring to this argument in the last paragraph of the initial comment.

And I quoted the last paragraph, except the last sentence I don't understand.

> This proposal gives object references the same lifetime as `PRIV_TASK` and removes the then unnecessary workspace copies. As a side effect, it also solves the case for #3768, because this also avoids most copies of static strings to workspace.

It was a quick glance, but I don't see how ensuring an at-least-task lifetime to objcores solves anything for static strings. In #3768 I added a new function to the runtime that sets a header with a string guaranteed to outlive the task (user's responsibility) and taught libvcc to better keep track of constant expressions to favor the new function when possible.

I would argue that with objcores guaranteed to outlive VCL execution, we could teach libvcc to recognize `obj.stuff` as a constant expression:

```vcl
set req.http.foo = "foo"; # 3768 would avoid a workspace copy
set req.http.foo = obj.http.foo; # 3768 could avoid a workspace copy
set req.http.foo = "bar, " + obj.http.foo; # cannot avoid copy
```

Your change is only affecting core code and it would easily compose with #3768 to extend `obj.*` "constness" to VCL code (though not very common). But it certainly doesn't solve workspace overflows on constant string assignments, I have seen cases where a significant bunch of headers are set in `vcl_deliver` and `vcl_synth` (CORS and other policies).

> I brought my case, so if you disagree, what is yours?

My main concern is that object storage is a more critical resource than workspaces. If we need a lot of workspace, we can lower task concurrency. When storage is saturated and in constant churn (when you eventually reach full capacity) it becomes crucial that dying objects actually go away. Increasing latency here means that your churn throughput will increase, reducing your cache/storage efficiency.

Forget what I said about restarts. Unlike retries, I think they should generally be avoided and be the exception instead of the norm, so I don't have that much of a problem adding the objcore retention caveat to this feature.

I had a closer look and in the normal case we stash nothing and drop the objcore reference in `cnt_finish()`, right? In that case I'm more open to it, and it would be a good occasion to revisit #3768 with more "task-constant" expressions. But I really don't like how the stash is put together. Please also note that dropping the ref in `cnt_finish()` makes the normal case "not quite task-scoped" but is it good enough? If the answer is yes, then the stash should be cleared alongside.

nigoroll:
> > But yes, I am aware and I was specifically referring to this argument in the last paragraph of the initial comment.
> 
> And I quoted the last paragraph, except the last sentence I don't understand.

The sentence is:

> In general, until now we have charged all Varnish-Cache users with the cost for specific use cases, but we should rather only charge the specific use cases instead.

My point is: This patch makes the common case cheaper.

> > This proposal gives object references the same lifetime as `PRIV_TASK` and removes the then unnecessary workspace copies. As a side effect, it also solves the case for #3768, because this also avoids most copies of static strings to workspace.
> 
> It was a quick glance, but I don't see how ensuring an at-least-task lifetime to objcores solves anything for static strings.

Without this patch, static strings need to be copied always. Without this patch, the `set HEADER = HEADER` optimization is not possible in the general case.

> In #3768 I added a new function to the runtime that sets a header with a string guaranteed to outlive the task (user's responsibility) and taught libvcc to better keep track of constant expressions to favor the new function when possible.

I think I understand what you are doing, and I think we should first have a more general improvement. We should first make sure that every reference in VCL at least has the same lifetime as `PRIV_TASK`, then we can avoid copying no matter the source. Your `set HEADER = HEADER` optimization makes sense (see footnote of the initial comment), but we do not need to mark vcl statics _constant_, because they are just a special case of "at least `PRIV_TASK` lifetime".

> I would argue that with objcores guaranteed to outlive VCL execution, we could teach libvcc to recognize `obj.stuff` as a constant expression:

Now you seem confused to me. Also with #3768, `set req.http.foo = obj.http.foo` can not save a workspace copy unless we _also_ have this PR. And yes, again, a `set HEADER = HEADER` optimization can save copies for the special case that the header names match, and for statics from VCL we can make it so that they do.

> Your change is only affecting core code and it would easily compose with #3768 to extend `obj.*` "constness" to VCL code 

It is not just affecting core code, because vmods suffer from the same copying[^1]. And I have said all the time that it would compose with  the `set HEADER = HEADER` optimization from #3768. The main point is that we do not avoid copying by marking an additional special case, the main point is that we avoid it by making sure that "all things referenced" have `PRIV_TASK` equivalent lifetime.

(Oh man, I am repeating myself for the fifth time now or so.)

> My main concern is that object storage is a more critical resource than workspaces. If we need a lot of workspace, we can lower task concurrency. When storage is saturated and in constant churn (when you eventually reach full capacity) it becomes crucial that dying objects actually go away. Increasing latency here means that your churn throughput will increase, reducing your cache/storage efficiency.

I understand what an object reference implies. Also we should note that for most objects on a typical Varnish-Cache installation, the bulk of object references will be held because of body delivery.

here the main point is that _all_ the scenarios where the reference would be kept longer than before the patch are _special cases_, which can be avoided by not restarting, or rolling back in addition to the restart. None of these cases is typical, and the argument that for some special case some object would be held for somewhat longer than before the patch really seems not significant.

> I had a closer look [...]

So was the first half of your comment from "before you looked closer" and the last paragraph from after?


> in the normal case we stash nothing and drop the objcore reference in `cnt_finish()`, right?

Yes

> But I really don't like how the stash is put together. Please also note that dropping the ref in `cnt_finish()` makes the normal case "not quite task-scoped" but is it good enough? If the answer is yes, then the stash should be cleared alongside.

The stash is only allocated if needed. The call to `ocstash_fini()` is a noop if `stash_oc()` was not called. It only gets called for the "exception path", of which _restart_ is most relevant.

[^1]: Actually [vmod_objvar](https://gitlab.com/uplex/varnish/varnish-objvar) probably benefits the most, because `taskvar.string` replacing use cases of `HEADER` will then not use any copies for `STRING` inputs and could be extended to also support `STRANDS` to avoid even more copying.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -684,8 +751,7 @@ cnt_lookup(struct worker *wrk, struct req *req)
 		WRONG("Illegal return from vcl_hit{}");
 	}
 
-	/* Drop our object, we won't need it */
-	(void)HSH_DerefObjCore(wrk, &req->objcore, HSH_RUSH_POLICY);
+	stash_oc(&req->ocstash, &req->objcore, req->ws, req->max_restarts + 1);
```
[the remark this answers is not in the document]
nigoroll: No. The stash is only allocated once for the maximum capacity the task might need.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -240,7 +304,7 @@ cnt_deliver(struct worker *wrk, struct req *req)
 
 	if (wrk->vpi->handling != VCL_RET_DELIVER) {
 		HSH_Cancel(wrk, req->objcore, NULL);
-		(void)HSH_DerefObjCore(wrk, &req->objcore, HSH_RUSH_POLICY);
+		stash_oc(&req->ocstash, &req->objcore, req->ws, req->max_restarts + 1);
```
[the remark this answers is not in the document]
nigoroll: See below.

dridi:
> The sentence is:
> 
> > In general, until now we have charged all Varnish-Cache users with the cost for specific use cases, but we should rather only charge the specific use cases instead.
> 
> My point is: This patch makes the common case cheaper.

I can parse it now, got it. I generally agree with tradeoffs in favor of the common cases.

> Without this patch, static strings need to be copied always. Without this patch, the `set HEADER = HEADER` optimization is not possible in the general case.

Agreed, with the understanding that "this patch" refers to #3768. But I think you are reading too much into my patch series, see below.

> I think I understand what you are doing, [...] Your `set HEADER = HEADER` optimization makes sense

I don't think I can take credit for the `set HEADER = HEADER` optimization, or I should look at #3768 to make sure. My optimization was `set HEADER = "literal string"`. Then I suggested marking `obj` header as constants to treat them like literal strings in libvcc, and then you came up with a generalization (assuming guaranteed task lifetime of objcores).

> > I would argue that with objcores guaranteed to outlive VCL execution, we could teach libvcc to recognize `obj.stuff` as a constant expression:
> 
> Now you seem confused to me. Also with #3768, `set req.http.foo = obj.http.foo` can not save a workspace copy unless we _also_ have this PR.

Not confused, right? The idea was assuming "objcores guaranteed to outlive VCL execution", but your `set HEADER = HEADER` idea would encompass `obj` headers.

> > Your change is only affecting core code and it would easily compose with #3768 to extend `obj.*` "constness" to VCL code
> 
> It is not just affecting core code, because vmods suffer from the same copying. And I have said all the time that it would compose with the `set HEADER = HEADER` optimization from #3768.

We actually agree, this is poor wording on my end. I think both changes would complement each other well and open the door to a new optimization.

> So was the first half of your comment from "before you looked closer" and the last paragraph from after?

Hours before the beginning and the end of my review...

> The stash is only allocated if needed. The call to `ocstash_fini()` is a noop if `stash_oc()` was not called. It only gets called for the "exception path", of which _restart_ is most relevant.

Then considering the workspace gains we can expect, we should avoid the complications of a just-in-time allocation, and certainly not fall back to a heap allocation. I think we should give it the reqtop treatment and simply always make room for the stash.

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
dridi: Likewise.

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
dridi: Please don't resolve review comments on my behalf.

I crossed two streams again. By the time I left my comment I had convinced myself that the stash should be allocated upfront, and misinterpreted the meaning of "restarts" here.

Anyway, what I'm seeing here is duplication and spread of logic. Both `stash_oc()` calls are identical, both are forcing the call site to know about internal logic, they should instead look like this:

```c
Req_StashObjcore(req);
Req_StashObjcore(req, &req->objcore); // if we ever grow the need to stash another oc
```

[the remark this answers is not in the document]
nigoroll: Yes, we can change the signature. The current generic arguments come from the idea that we might want to reuse the facility somewhere else, but we do not have that use case at the moment.

nigoroll:
As the rest of the discussion looks like it might be resolved, I will only respond to the last paragraph:

> we should avoid the complications of a just-in-time allocation

Again, I wanted to keep impact on the common case minimal.

> not fall back to a heap allocation.

This is unlikely and simplifies the calling code.

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
nigoroll: Do you like 81391a5f7c2f1b0c43cab2d43cd1d2ab93ca5881 better?

dridi:
> Again, I wanted to keep impact on the common case minimal.

The impact is already a negative workspace footprint.

> This is unlikely and simplifies the calling code.

So does a small systematic allocation (56B by default) during the `struct req` setup.

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
dridi: I actually think this all belongs in `cache_req.c`, and I'm thinking that clearing the stash could fit well in `Req_Cleanup()`.

nigoroll:
> So does a small systematic allocation (56B by default) during the `struct req` setup.

It is not like I had not considered this option. My worry is that it will impact users with high `max_restarts` values substantially, and likely needlessly.

On the malloc fallback, I think it is [generally a good idea](https://github.com/varnishcache/varnish-cache/blob/4fe462a4794584c43fb26046d966be8cb7cebf13/bin/varnishd/cache/cache_vrt_priv.c#L158) to not fail in out supporting facilities.

dridi:
> But at any rate, doing a heap allocation for this exception path is, I think, offset by the gain in simplicity: Because we might already be on an exception path when we stash an oc, we save ourselves from complicating it further.

The simplest approach is a preemptive allocation of the stash that completely removes the need for just-in-time allocations and gets rid of all allocation-related branches.

Ignoring a failed workspace allocation here is just delaying the actual workspace failure somewhere in vcl_synth for the synth/fail cases (reminder, the built-in `vcl_synth` makes allocations, but we have a candidate fix for that). In the restart case, there is no point processing an entire task again with an overflowed workspace.

This is actually increasing the distance between the root cause and the symptoms.

Again, reserving a stash upfront:

- gets rid of branches
- avoids the worst case scenario a heap allocation
  - and some malloc overhead to perform, track and free the allocation
  - and a potential delayed failure caused by the workspace overflow
- does not introduce error handling at the call site

I do agree in the dynamic priv case that it brings error handling to the caller. In fact `VRT_priv_task()` used to be fallible so it would only bring error handling back. It would also bring dynamic privs back to respecting the task's memory budget.

> FTR on the other topic: The allocation of the `struct vrt_priv` is not to be confused with the user controlled `(struct vmod_priv).priv` member.

That's exactly the distinction I was making. The struct should be allocated from the relevant workspace, the member is up to the VMOD author.

Quoting myself:

> A VMOD author is still free to allocate PRIV_TASK data from the heap.

Replace my "data" with your more accurate "`(struct vmod_priv).priv` member".

:-1: for not noticing the pun :]

I think the simplest actually is:

- manage stash in `cache_req.c`
- allocate stash upfront from mempool (like reqtop)
- export `Req_StashObjcore()`
- clear stash in `Req_Cleanup()`

Pros and cons:

- less moving parts, better encapsulation
- no new error paths
- fixed tiny workspace overhead
  - paying dividends as soon as you set up a `resp` for `vcl_deliver`

And if the stash turns out to have a prohibitive cost[^1] then we can complicate the picture and refactor the allocation to make it just-in-time. The extra care you added is premature optimization at this point.

[^1]: unlike a reqtop systematically allocated upfront even in the absence of sub-requests

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
nigoroll: Right now, the code is in _fsm.c because that's where it is used.
I think otherwise it would live in cache_hash.c, because HSH_DerefObjCore lives there.
`ocstash_fini()` / `ReqFiniObjcoreStash()` is called where `VCL_TaskLeave()` is called, because the the two are very much related. `Req_Cleanup()` is concerned with the lifetime of `struct req`, which is longer than `PRIV_TASK`.

side note: #3994 will probably also need this facility

nigoroll:
I disagree that it is the better option, because it charges all users with a workspace allocation which will not be needed in most cases. The `struct reqtop` case is different, because that struct is of fixed size and we actually do need it always to not add special casing all over the place for `vcl0`, VCL `req.top.` and `PRIV_TOP`. Yet you are right that we allocate it also for ESI subrequests, which would not be needed.
I have commented on the `Req_Cleanup` and `cache_req.c` suggestions further up.

nigoroll:
> fixed tiny workspace overhead

It is not fixed. It depends on `max_restarts`.

> paying dividends as soon as you set up a resp for vcl_deliver

No. The straight path does not need it. The common cases where it is needed are `return(restart)`  and `return(synth)`  from `vcl_deliver{}` and `vcl_hit{}`.

dridi:
> It is not fixed. It depends on `max_restarts`.

Sorry, I meant fixed for the duration of the task.

> > paying dividends as soon as you set up a resp for vcl_deliver
> 
> No. The straight path does not need it. The common cases where it is needed are `return(restart)` and `return(synth)` from `vcl_deliver{}` and `vcl_hit{}`.

I meant that your change will pay dividends with the workspace savings outweighing the upfront allocation even for the common cases.

> I have commented on the `Req_Cleanup` and `cache_req.c` suggestions further up.

I submitted #4271 to illustrate what I meant about better encapsulation. What leaks out of `cache_req.c` is minimal.

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
dridi: `Req_Cleanup()` barely outlives the task's lifetime, and at least calling it from there guarantees that privs were actually freed \*before\* clearing the stash. And a cleanup doesn't prevent `req` from being reused for keep-alive, you are confusing it with `Req_Release()`.

> side note: #3994 will probably also need this facility

You are just agreeing with me that `cache_req_fsm.c` is not a good place for this. At least with what I'm suggesting we can move the stash logic from `cache_req.c` to `cache_obj.c` and keep the call site in `Req_Cleanup()`.

We can cross that bridge when `obj_stale` makes its appearance.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -1213,6 +1282,7 @@ CNT_Request(struct req *req)
 	}
 	wrk->vsl = NULL;
 	if (nxt == REQ_FSM_DONE) {
+		ReqFiniObjcoreStash(req);
 		INIT_OBJ(ctx, VRT_CTX_MAGIC);
 		VCL_Req2Ctx(ctx, req);
 		if (IS_TOPREQ(req)) {
```
dridi, line 1288: It's not safe to clear the stash before the privs free callbacks run.

bin/varnishd/cache/cache_req_fsm.c
```diff
@@ -286,6 +353,8 @@ cnt_vclfail(struct worker *wrk, struct req *req)
 	AZ(req->objcore);
 	AZ(req->stale_oc);
 
+	ReqFiniObjcoreStash(req);
+
 	INIT_OBJ(ctx, VRT_CTX_MAGIC);
 	VCL_Req2Ctx(ctx, req);
```
dridi, line 359: It's not safe to clear the stash before the privs free callbacks run.

bin/varnishd/cache/cache_req_fsm.c
nigoroll: Hm. The free callbacks _should_ not do anything with references which the priv might still hold, but still I agree that changing the order is the safer option. Thank you for your valid point and good suggestion!

bin/varnishd/cache/cache_req_fsm.c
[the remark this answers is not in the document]
nigoroll: > We can cross that bridge when obj_stale makes its appearance.

That's were we agree.

With the slight interface modifications in #4271 and in particular calling the cleanup from `Req_*.c`, I do fully agree with the move of the code. The two important questions, however, are if the new location to call fini/cleanup is better and if we really should have the upfront allocation.

bin/varnishd/cache/cache_req_fsm.c
dridi: What I have in mind is collecting something during the task, coming from a header for example, and logging it when the task ends. We replaced the `free()` callback with a `fini()` one that takes a `VRT_CTX` to allow privs to do things.

nigoroll:
@dridi On the `ReqFiniObjcoreStash()` / `ocstash_fini()` / `ocstash_clear()` call site (called "the cleanup" in what follows to avoid the bikeshed):

I agree that calling it in the middle of `Req_Rollback()` after the `VCL_TaskLeave()` is better.
I now also agree that calling it in `CNT_Request()` for the `REQ_FSM_DONE` case is wrong, because I had overlooked that filters (VDPs) can also have a look at headers from their init callback, and because headers could reference object storage, the cleanup needs to run after the vdp init.
But calling it in `Req_Cleanup()` is _also_ wrong: If we moved the cleanup there, the concern that we held onto oc references for "way too long" became valid, because `Req_Cleanup()` is called after all of the body processing is done.

dridi:
I'm not following you here, isn't `REQ_FSM_DONE` reached from `cnt_finish()` for the normal object delivery case? In that case we already delivered the body in `cnt_transmit()`.

That's why I said that `Req_Cleanup()` is called soon after leaving the state machine in a previous comment (https://github.com/varnishcache/varnish-cache/pull/4269#discussion_r1954051048).

nigoroll:
> I'm not following you here

Ah right, I was somehow wrong and right at the same time: We should actually clear the stash before sending the body, and I thought this patch did this, but it does not - unless the new `VTR_D_DISEMBARK` facility is used. 

So this needs more deliberation, still, and I shall remember to write code first, then comment...

nigoroll:
@dridi thank you for your excellent summary.
I am now going to continue here and would like to ask you to please apologize my confusion when I wrote "But calling it in `Req_Cleanup()` is _also_ wrong" [here](https://github.com/varnishcache/varnish-cache/pull/4269#issuecomment-2656077219). Yes, we do need to keep all stashed ocs until delivery is complete, because references to them might be used by filters basically anywhere (a filter could also add any task scoped data to the body).

nigoroll:
@dridi I have now taken your version of the patch in slightly modified form and added you as Co-Auther by the de-facto standard form (I should have looked this up for some other commits added recently and done the same...).

Regarding the jit vs. upfront allocation, one argument came to my mind which, I think, had not been mentioned, but which made me change my opinion: For a scenario with a substantial amount of restarts which only happen under some circumstances, users might be taken by surprise if their code "suddenly" needs more workspace. Because, to make their use case work, they need to increase the workspace size anyway, they rather notice sooner than by surprise. So, simply put, I now agree that this is the better option by the argument of "predictability before maximum efficiency".

A diff to your version can be found in 75e88cae8. I have done the following:

* made `max_restarts` fit into `uint16_t` to make room for a magic value without spending another `sizeof(void *)`.
* Added `CHECK_OBJ_NOTNULL` accordingly
* Modified `Req_New()` slightly to make Flexelint happy and avoid zeroing memory twice (while that is still done for the other struct allocated...)

dridi (approved):
The change LGTM, and apologies if the review got on your nerves. You can otherwise tell me I'm wrong as much as you want. As long as you are saying it in good faith I can't take offense.

In the end, I'm happy with the result of our discussion.

bin/varnishd/cache/cache_req.c
```diff
@@ -39,12 +39,60 @@
...
+#include "cache_objhead.h"
 #include "cache_pool.h"
 #include "cache_transport.h"
 
 #include "common/heritage.h"
 #include "vtim.h"
 
+/*--------------------------------------------------------------------
+ * Facility to keep obcore references until the end of the task across restarts
+ */
```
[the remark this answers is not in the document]
nigoroll: "discarded objcores" to me sounds like they would be removed from cache.

bin/varnishd/cache/cache_vrt.c
```diff
@@ -93,14 +93,6 @@ VRT_synth(VRT_CTX, VCL_INT code, VCL_STRING reason)
 		return;
 	}
 
-	if (reason && !WS_Allocated(ctx->ws, reason, -1)) {
-		reason = WS_Copy(ctx->ws, reason, -1);
-		if (!reason) {
-			VRT_fail(ctx, "Workspace overflow");
-			return;
-		}
-	}
```
dridi, line 102: This might not be safe actually:

```vcl
return (synth(200, some_vmod.some_string()));
```

In that case we can't assume a lifetime beyond the `VRT_synth()` call for the `reason` argument.

bin/varnishd/cache/cache_vrt.c
```diff
@@ -445,12 +437,11 @@ VRT_StrandsWS(struct ws *ws, const char *h, VCL_STRANDS s)
...
-		if (h == NULL)
-			return ("");
-		if (WS_Allocated(ws, h, -1))
-			return (h);
-	} else if (h == NULL && WS_Allocated(ws, q, -1)) {
+	if (q == NULL && h == NULL)
+		return ("");
+	if (q == NULL)
+		return (h);
+	if (h == NULL) {
```
dridi, line 444: Likewise, we can't assume a task lifetime of the arguments.

vmod/vmod_debug.c
```diff
@@ -645,8 +645,6 @@ xyzzy_concatenate(VRT_CTX, VCL_STRANDS s)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	r = VRT_StrandsWS(ctx->ws, NULL, s);
-	if (r != NULL && *r != '\0')
-		AN(WS_Allocated(ctx->ws, r, strlen(r) + 1));
```
dridi, line 649: If our concern is coverage for `VRT_StrandsWS()` then we need to keep a check that the result effectively belongs to the workspace.

vmod/vmod_debug.c
```diff
@@ -657,8 +655,6 @@ xyzzy_collect(VRT_CTX, VCL_STRANDS s)
 
 	CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);
 	r = VRT_STRANDS_string(ctx, s);
-	if (r != NULL && *r != '\0')
-		AN(WS_Allocated(ctx->ws, r, strlen(r) + 1));
```
dridi, line 661: See previous.

vmod/vmod_debug.c
```diff
@@ -690,8 +686,6 @@ xyzzy_sethdr(VRT_CTX, VCL_HEADER hdr, VCL_STRANDS s)
 			VSLbs(ctx->vsl, SLT_LostHeader,
 			    TOSTRAND(hdr->what + 1));
 		} else {
-			if (*b != '\0')
-				AN(WS_Allocated(hp->ws, b, strlen(b) + 1));
```
dridi, line 694: See previous.

dridi (changes requested):
I need to reduce my workspace saving expectations here. I somehow convinced myself that this change was preventing copies from objcore to workspace once the objcore is guaranteed to outlive VCL processing.

It turns out we already skip copies, when most notably during `resp` setup. `HTTP_Decode()` already dips directly in the objcore without a detour from workspace.

So I still think this is a good idea, but we need to deal with `set HEADER = HEADER` to actually reap benefits.

nigoroll:
> So I still think this is a good idea, but we need to deal with `set HEADER = HEADER` to actually reap benefits.

b00092.vtc shows how copying from an object is skipped when restarting:

```
**** v1    vsl|       1001 Debug           c resp.http.method[0]: (oc) 0x7ffb8a60709c O1_M...
**** v1    vsl|       1003 Debug           c resp.http.url[0]: (oc) 0x7ffb8a607199 /o2_...
**** v1    vsl|       1005 Debug           c req.method[0]: (?) 0x7ffb8a60709c O1_M...
**** v1    vsl|       1005 Debug           c req.url[0]: (?) 0x7ffb8a607199 /o2_...
```

Yes, this is no copy to `HEADER`, just from `HEADER`. I think we discussed this topic sufficiently.

But the main benefit is that we do not need to copy for all the other cases, which you questioned again. After this PR, the contract would officially become that all pointers passed around for VCL have to have at least `PRIV_TASK` lifetime. This was already the de-facto contract, with the exception of the case addressed here.

Regarding your code comments: I do not want to add 4 more comments to this, but what kind of a pointer do you think for `return (synth(200, some_vmod.some_string()));` the vmod would return?

bin/varnishd/cache/cache_vrt.c
nigoroll: see https://github.com/varnishcache/varnish-cache/pull/4269#issuecomment-2663362814 regarding this and the next four review comments.

nigoroll:
> but what kind of a pointer do you think for return (synth(200, some_vmod.some_string())); the vmod would return?

To be very clear: I think the raised concern is without substance.

nigoroll:
@dridi because you requested changes on, I think, an unfounded basis, I would appreciate an update from your end.

nigoroll:
@dridi ?

nigoroll:
*VERY* good find by @bsdphk : `vcl_miss` gets called with an optional `stale_oc`, which we need to treat equally.

PR #4270, opened 2025-02-12
cache_req: Allocate topreq only if needed

nigoroll:
This patch avoids allocating a superfluous `struct reqtop` for ESI subrequests.

From a discussion here:
https://github.com/varnishcache/varnish-cache/pull/4269#issuecomment-2654577140


dridi (approved):
LGTM

PR #4271, opened 2025-02-13
draft: req_fsm: Stash dropped objcores seen by VCL

dridi:
My personal take on #4269.


dridi:
The goal was to serve as material for the discussion in #4269. No longer needed.

PR #4202, opened 2024-09-26
draft: jail_unix: Simplify sub-argument matching

dridi:
TODO: find a name and a home for the ARG_MATCH() macro.

Refs 4001ea251bac6924c2e86bac3d0f2c8d93c96bd8
Refs #4174

PR #4272, opened 2025-02-13
Simplify key=val argument handling

nigoroll:
@dridi I would use this in place of #4202. Do you agree?

This is a Draft because I still want to look if we can replace other similar cases and probably move the helper function to a better place as a `static inline` [^1]

----

we add a simple utility function to return the pointer to the thing after "key=" if the first argument begins with "key=".

Note that this is no worse than a macro with a "reasonable good" compiler *) and default (-O2) optimization, because the function call gets inlined and the strlen() turned into a constant.

```
  0x00000000000a1bd1 <+49>:     lea    0x6ed49(%rip),%r14        # 0x110921
  ...
  0x00000000000a1c01 <+97>:    mov    $0x5,%edx
  0x00000000000a1c06 <+102>:   mov    %rbx,%rdi
  0x00000000000a1c09 <+105>:   mov    %r14,%rsi
  0x00000000000a1c0c <+108>:   call   0x2d360 <strncmp@plt>
  0x00000000000a1c11 <+113>:   lea    0x5(%rbx),%rbp
  0x00000000000a1c15 <+117>:   test   %eax,%eax

  (gdb) p (const char *)0x110921
  $2 = 0x110921 "user="
```

Replaces #4202
Polishes 4001ea251bac6924c2e86bac3d0f2c8d93c96bd8

*) tested:
   gcc version 12.2.0 (Debian 12.2.0-14)
   Debian clang version 14.0.6

[^1]: Reminder that `inline` has no other relevance than "do not complain if unused", as the compiler is free to inline no matter what

### DIFF
--- a/bin/varnishd/mgt/mgt_jail_unix.c
+++ b/bin/varnishd/mgt/mgt_jail_unix.c
@@ -117,9 +117,15 @@ vju_getccgid(const char *arg)
 /**********************************************************************
  */
 
+// avoid line breaks
+#define JERR(what, val)						\
+	ARGV_ERR("Unix jail: %s " what " not found.\n", val)
+
 static int v_matchproto_(jail_init_f)
 vju_init(char **args)
 {
+	const char *val;
+
 	if (args == NULL) {
 		/* Autoconfig */
 		if (geteuid() != 0)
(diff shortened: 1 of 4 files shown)


bin/varnishd/mgt/mgt_jail_unix.c
```diff
@@ -132,28 +143,25 @@ vju_init(char **args)
...
-			const char * const a_user = "user=";
-			const size_t l_user = strlen(a_user);
-			if (!strncmp(*args, a_user, l_user)) {
-				if (vju_getuid((*args) + l_user))
+			val = keyval(*args, "user=");
+			if (val != NULL) {
+				if (vju_getuid(val))
 					ARGV_ERR(
 					    "Unix jail: %s user not found.\n",
 					    (*args) + 5);
```
dridi, line 151: Not ready to replace #4202 until you get rid of `(*args) + ...` occurrences.

bin/varnishd/mgt/mgt_jail_unix.c
```diff
@@ -132,28 +143,25 @@ vju_init(char **args)
...
-			const char * const a_workuser = "workuser=";
-			const size_t l_workuser = strlen(a_workuser);
-			if (!strncmp(*args, a_workuser, l_workuser)) {
-				if (vju_getwrkuid((*args) + l_workuser))
+			val = keyval(*args, "workuser=");
+			if (val != NULL) {
+				if (vju_getwrkuid(val))
 					ARGV_ERR(
 					    "Unix jail: %s user not found.\n",
 					    (*args) + 9);
```
dridi, line 159: Not ready to replace #4202 until you get rid of `(*args) + ...` occurrences.

bin/varnishd/mgt/mgt_jail_unix.c
```diff
@@ -132,28 +143,25 @@ vju_init(char **args)
...
-			const char * const a_ccgroup = "ccgroup=";
-			const size_t l_ccgroup = strlen(a_ccgroup);
-			if (!strncmp(*args, "ccgroup=", l_ccgroup)) {
-				if (vju_getccgid((*args) + l_ccgroup))
+			val = keyval(*args, "ccgroup=");
+			if (val != NULL) {
+				if (vju_getccgid(val))
 					ARGV_ERR(
 					    "Unix jail: %s group not found.\n",
 					    (*args) + 8);
```
dridi, line 167: Not ready to replace #4202 until you get rid of `(*args) + ...` occurrences.

dridi (changes requested):
s/unit/unix/ in the commit message.

bin/varnishd/mgt/mgt_jail_unix.c
nigoroll: oops! thank you!

nigoroll: That I had already overlooked in 4001ea251bac6924c2e86bac3d0f2c8d93c96bd8 🤦🏽‍♂️

PR #4274, opened 2025-02-14
[varnishstat] reset running averages

gquintard:
I've been running benchmark tests these last few days, and I've been wishing to be able to reset the average calculations without having to restart `varnishncsa`, so here we go.

### DIFF
--- a/bin/varnishstat/varnishstat_curses.c
+++ b/bin/varnishstat/varnishstat_curses.c
@@ -129,6 +129,7 @@ static int current = 0;
 static int rebuild = 0;
 static int redraw = 0;
 static int sample = 0;
+static int reset_averages = 0;
 static int scale = 1;
 static double t_sample = 0.;
 static double interval = 1.;
@@ -262,6 +263,12 @@ sample_points(void)
 		pt->t_last = pt->t_cur;
 		pt->t_cur = VTIM_mono();
 
+		if (reset_averages) {
+			pt->chg = 0;
+			pt->ma_10.n = 0;
+			pt->ma_100.n = 0;
+			pt->ma_1000.n = 0;
+		}
 		if (pt->t_last)
 			pt->chg = ((int64_t)pt->cur - (int64_t)pt->last) /
 			    (pt->t_cur - pt->t_last);
@@ -305,6 +312,11 @@ sample_hitrate(void)
 		ratio = hr / (hr + mr);
 	else
 		ratio = 0;
+	if (reset_averages) {
+		hitrate.hr_10.n = 0;
+		hitrate.hr_100.n = 0;
+		hitrate.hr_1000.n = 0;
+	}
 	update_ma(&hitrate.hr_10, ratio);
 	update_ma(&hitrate.hr_100, ratio);
 	update_ma(&hitrate.hr_1000, ratio);
@@ -318,6 +330,7 @@ sample_data(void)
 	redraw = 1;
 	sample_points();
 	sample_hitrate();
+	reset_averages = 0;
 }
 
 static void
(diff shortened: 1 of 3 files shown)


nigoroll (approved):

dridi:
Please remove all mentions to `varnishncsa`, this is `varnishstat`.

gquintard:
Ah yes, of course

dridi:
Please also make it two commits, it's easier to back-port to any branch without the guaranteed conflict in the changelog.

dridi:
I accidentally submitted the previous comment, I meant to also ask for basic test coverage.

bin/varnishstat/varnishstat_curses.c
```diff
@@ -316,8 +327,9 @@ sample_data(void)
 	t_sample = VTIM_mono();
 	sample = 0;
 	redraw = 1;
-	sample_points();
-	sample_hitrate();
+	sample_points(reset_averages);
+	sample_hitrate(reset_averages);
+	reset_averages = 0;
```
dridi, line 332: Why pass `reset_averages` to the `sample_*()` functions?

bin/varnishstat/varnishstat_curses.c
```diff
@@ -305,6 +311,11 @@ sample_hitrate(void)
...
 	else
 		ratio = 0;
+	if (reset) {
+		hitrate.hr_10.n = 0;
+		hitrate.hr_100.n = 0;
+		hitrate.hr_1000.n = 0;
+	}
 	update_ma(&hitrate.hr_10, ratio);
 	update_ma(&hitrate.hr_100, ratio);
 	update_ma(&hitrate.hr_1000, ratio);
```
dridi, line 321: Here you could use `reset_averages` directly.

bin/varnishstat/varnishstat_curses.c
```diff
@@ -262,7 +263,12 @@ sample_points(void)
...
 
-		if (pt->t_last)
+		if (reset) {
+			pt->chg = 0;
+			pt->ma_10.n = 0;
+			pt->ma_100.n = 0;
+			pt->ma_1000.n = 0;
+		} else if (pt->t_last)
 			pt->chg = ((int64_t)pt->cur - (int64_t)pt->last) /
 			    (pt->t_cur - pt->t_last);
```
dridi, line 273: To be consistent with the block of code from my previous comment, shouldn't it look like this?

```c
if (reset) { // or better: if (reset_averages)
	pt->chg = 0;
	pt->ma_10.n = 0;
	pt->ma_100.n = 0;
	pt->ma_1000.n = 0;
}

if (pt->t_last) {
	pt->chg = ((int64_t)pt->cur - (int64_t)pt->last) /
	    (pt->t_cur - pt->t_last);
}
```

For the hit ratio you reset the 10/100/1000 averages and let them be updated in the same iteration. So for consistency you should give the counters a chance to be updated too.

bin/varnishstat/varnishstat_curses.c
gquintard: this was a quick 10 minutes job, and I assumed the key presses could come at any given time and wanted to avoid different counters from being handled differently, but I see the `wgetch()` call now, I can indeed use the global variable

dridi: It's single-threaded, you are fine with globals :)

nigoroll:
I like phk's idea to use the `0` key to reset the averages.
Other than that, bugwash is +1

gquintard:
I think we are good now

nigoroll (approved):

nigoroll:
Please squash the commits.

gquintard:
added the period, squashed and force-pushed

dridi:
I will push it with an accurate commit message.

dridi:
5a26bb8160c7099593f4a212b065fd01637f9629

PR #4287, opened 2025-02-25
cache_ban: Add obj.last_hit

nigoroll:
This commit adds bans by last LRU time, basically equivalent to the existing `obj.age` from b06cdbbab4099a351e188106add56b5e1eb0b927 (the code was trivial to add).

The primary use case is to remove objects from cache which have not been accessed for a long time, and, in particular, to get rid of request bans by removing all objects which have not been touched since the request ban.


nigoroll:
This needs more preparation, ticket coming up

nigoroll:
I have implemented the agreed-upon changes. My impression from bugwash was that we agreed, but I missed to ask explicitly.

bin/varnishd/cache/cache_ban.c
```diff
@@ -498,7 +498,7 @@ ban_evaluate(struct worker *wrk, const uint8_t *bsarg, struct objcore *oc,
 	int rv;
 
 	/*
-	 * for ttl and age, fix the point in time such that banning refers to
+	 * for ttl, age and lru, fix the point in time such that banning refers to
```
walid-git, line 501: shouldn't this comment now say `last_hit` instead of `lru` ?

nigoroll: thank you!

walid-git (approved):
LGTM

PR #4289, opened 2025-02-26
cache_ban: Forward compatibility for the binary ban representation

### LINKED ISSUE #4288: bans are not forward compatible
https://github.com/varnishcache/varnish-cache/blob/a62cb553d3a535c9d13a423c8f3d01366261ab1a/bin/varnishd/cache/cache_ban.c#L550

When loading a persistent storage from a later version which has a new ban format on an older version, a panic is triggered:

```
Wrong turn at cache/cache_ban.c:550:
Wrong BAN_ARG code
version = varnish-trunk revision 6d5aa36cc13f0d09211ffa74e68d7ca607d921d2, vrt api = 20.1
ident = Linux,6.1.0-29-amd64,x86_64,-jnone,-sfellow,-sdefault,-Elibvmod_slash.so,-hcritbit,epoll
now = 9508.824577 (mono), 1740567235.533489 (real)
Backtrace:
  ip=0x55e637bc3b05 sp=0x7f027ebfdd30 <VBT_format+0x35>
  ip=0x55e637afb063 sp=0x7f027ebfdd50 <pan_backtrace+0x33>
  ip=0x55e637afadaa sp=0x7f027ebfdd70 <pan_ic+0x37a>
  ip=0x55e637bc2d35 sp=0x7f027ebfdef0 <VAS_Fail+0x55>
  ip=0x55e637ac2ea4 sp=0x7f027ebfdf40 <ban_evaluate+0x2b4>
  ip=0x55e637ac7c90 sp=0x7f027ebfe000 <ban_lurker_test_ban+0x4b0>
  ip=0x55e637ac71be sp=0x7f027ebfe0a0 <ban_lurker_work+0x1be>
  ip=0x55e637ac6f22 sp=0x7f027ebfe100 <ban_lurker+0x162>
  ip=0x55e637b3783d sp=0x7f027ebfe160 <wrk_bgthread+0x13d>
  ip=0x7f02d648c1c4 sp=0x7f027ebfe460 <pthread_condattr_setpshared+0x4e4>
  ip=0x7f02d650c85c sp=0x7f027ebfe500 <__xmknodat+0x23c>
```

For this example, an `obj.lru` ban has been created with #4287 and then the storage loaded with 6d5aa36cc13f0d09211ffa74e68d7ca607d921d2, which does not have `obj.lru`.

nigoroll:
Our ban expressions (like `obj.age > 20s`) are represented in a binary format (see [top of `cache_ban.h`](https://github.com/varnishcache/varnish-cache/blob/6d5aa36cc13f0d09211ffa74e68d7ca607d921d2/bin/varnishd/cache/cache_ban.h#L31-L56)) which allows for forward compatibility, yet at the respective places we currently just trigger an assertion failure if we hit an unknown argument or operator code.

This commit brings forward compatibility such that, when bans are loaded from persistent storage into older code which does not yet support newly introduced binary codes, we no longer panic.

Ban evaluation:

For bans, evaluating an expression to `true` is always "correct" in that the cache would not deliver banned content. It might cause objects to be removed from cache, but that is at least not incorrect. So the fail safe action this code takes is to always evaluate unknown ban expressions to true.

CLI `ban.list`:

For unsupported ban expressions, the unknown argument or operator codes are formatted as `(0x%02x)` with the string `UNSUPPORTED` as the user-specified argument. For example:

```
	1740567193.765849     0 -  (0x20) > UNSUPPORTED && obj.http.foo ~ 377.266
```

(note that here the operator `>` is supported and printed as such, and the ban contains one unsupported and one supported expression)

Logging:

For each unsupported argument or operator code, an `Error` VSL is output exactly once to vxid 0.

Statistics:

Whenever unsupported argument or operator codes are encountered, the newly added counters `MAIN.bans_inval_arg1` and `MAIN.bans_inval_oper` are incremented, respectively.

Fixes #4288


nigoroll:
bugwash: phk wants to think about it

nigoroll:
bugwash:
* we have a [related issue in SLASH/fellow](https://gitlab.com/uplex/varnish/slash/-/issues/111) which points towards an issue caused by the fact that bans use regexen which are not properly (un)serialized
* my initial proposal was to add that functionality to Varnish-Cache
* but @bsdphk proposed to export BANs in text form, which might be an even better idea, which would get us even better compatibility across versions
* then we could also go further and make the runtime representation of bans more efficient for evaluation

We should go back to https://github.com/varnishcache/varnish-cache/wiki/VDD23Q1#bans-and-persistent-storages and write down a plan how bans should look like in future, especially considering loading within a cluster and moving storage.

PR #4301, opened 2025-03-21
vmodtool: New --noinst option for vmod_debug

dridi:
The problem with vmod_debug is that we don't want to distribute it, but the automake boilerplate unilaterally adds all VMODs to $(vmoddir). This results in the installation of libvmod_debug.so but not vmod_debug.vcc and the responsibility of not redistributing libvmod_debug.so to our downstream maintainers.

A new --noinst option cuts down the amount of circumventions we have for vmod_debug and ensures that vmod_debug does not accidentally land in packages.

---

Less boilerplate and overall a negative diff stat :]


nigoroll:
CCI does not look happy.
Also: Why was it again that we check in the generated boilerplates?

nigoroll (commented):
I am surprised

dridi:
Hum, I didn't try to run the test suite, and as a `noinst` library only a static `.a` archive is built.

dridi:
I found the missing piece.

> Also: Why was it again that we check in the generated boilerplates?

Because we need it at configure time, otherwise we have a chicken-egg catch 22 when we build for the first time.

PR #4303, opened 2025-03-27
introduce VRT_VSC_AllocVariadic

gquintard:
VRT_VSC_Alloc requires `va_list` which isn't available on all platforms in `rust`. To bridge this FFI gap, provide a variadic version of it.

### DIFF
--- a/bin/varnishd/common/common_vsc.c
+++ b/bin/varnishd/common/common_vsc.c
@@ -183,6 +183,20 @@ VRT_VSC_Allocv(struct vsmw_cluster *vc, struct vsc_seg **sg,
 	return (vsg->ptr);
 }
 
+void *
+VRT_VSC_Alloc(struct vsmw_cluster *vc, struct vsc_seg **sg,
+    const char *nm, size_t sd,
+    const unsigned char *jp, size_t sj, const char *fmt, ...)
+{
+	va_list ap;
+	void *retval;
+
+	va_start(ap, fmt);
+	retval = VRT_VSC_Allocv(vc, sg, nm, sd, jp, sj, fmt, ap);
+	va_end(ap);
+	return(retval);
+}
+
 void
 VRT_VSC_Destroy(const char *nm, struct vsc_seg *vsg)
 {
--- a/bin/varnishd/common/common_vsc.c
+++ b/bin/varnishd/common/common_vsc.c
@@ -135,7 +135,7 @@ VRT_VSC_Reveal(const struct vsc_seg *vsg)
 }
 
 void *
-VRT_VSC_Alloc(struct vsmw_cluster *vc, struct vsc_seg **sg,
+VRT_VSC_Allocv(struct vsmw_cluster *vc, struct vsc_seg **sg,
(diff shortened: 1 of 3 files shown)


include/vrt.h
```diff
@@ -824,6 +824,10 @@ void VRT_VSM_Cluster_Destroy(VRT_CTX, struct vsmw_cluster **);
 #ifdef va_start	// XXX: hackish
```
gquintard, line 824: side note: I'm not sure this `ifdef` is justified anymore? There's no alternative for `VRT_VSC_Alloc` if it isn't defined, and we systematically rely on that function, don't we?

bin/varnishd/common/common_vsc.c
```diff
@@ -183,6 +183,20 @@ VRT_VSC_Alloc(struct vsmw_cluster *vc, struct vsc_seg **sg,
 	return (vsg->ptr);
 }
 
+void *
+VRT_VSC_AllocVariadic(struct vsmw_cluster *vc, struct vsc_seg **sg,
```
dridi, line 187: I would have suggested `VRT_VSC_Allocv()` to match what we do in other places but in this case it's completely backwards :-/

nigoroll (approved):
WFM, but yes, `VRT_VSC_Allocv()`

bsdphk:
I agree with @dridi that the name is non-C-ish, but I'll refrain from making suggestions, what with bikesheds and whats not.

gquintard:
so, `VRT_VSC_Allocv`->`VRT_VSC_Alloc` and `VRT_VSC_AllocVariadic`->`VRT_VSC_Allocv`?

dridi:
> so, `VRT_VSC_Allocv`->`VRT_VSC_Alloc` and `VRT_VSC_AllocVariadic`->`VRT_VSC_Allocv`?

That's a sneaky breaking change, but that would be the correct move.

You'd need two extra entries in the VRT history:

- VRT_VSC_Alloc() renamed to VRT_VSC_Allocv()
- new VRT_VSC_Alloc() added

gquintard:
CI is failing on an unrelated leak, I think this PR is ready? Do you want a `doc/changelog.md` entry?

dridi:
It can't be ready, it doesn't match what we agreed to do.

gquintard:
Erf indeed, pushed the wrong stuff. putting in draft until I sort it out

dridi (approved):

nigoroll (commented):
I think this can be merged after addressing Walid's nitpick.

nigoroll (approved):

walid-git (approved):

xcir:
This change has caused problems for some VMODs.(xcounter)
I know this is basically a change to be made before release, but could you please change `VRT_(MAJOR|MINOR)_VERSION`?
It would be easier to handle the master branch if it could be decided by ifdef.

dridi:
You can check the existence of `VRT_VSC_Allocv()` at configure time to perform the ifdef ritual.

PR #4304, opened 2025-03-30
constify headers

gquintard:
one of my vmod is complaining that headers in `tbl/http_headers.h` are not immutable and technically, there aren't because of `http_init_hdr`, but as `extern` they should be const?


gquintard:
Sad. I do get the why they aren't purely const, but I was hopeful when it compiled locally that we would be able to at least expose them as such.

Oh well, closing.

gquintard:
I'm not sure the discussion in #3215 would go a different way but it would
save me some warnings, so I'm in

dridi:
New attempt with #4308.

PR #4306, opened 2025-03-31
param: Off by one maximum http_req_overflow_status

dridi:
Spotted during a discussion with @walid-git.

### DIFF
--- a/include/tbl/params.h
+++ b/include/tbl/params.h
@@ -653,7 +653,7 @@ PARAM_SIMPLE(
 	/* name */	http_req_overflow_status,
 	/* type */	uint_orzero,
 	/* min */	"400",
-	/* max */	"499",
+	/* max */	"500",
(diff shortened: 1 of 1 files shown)


nigoroll:
Hm, this was actually deliberate on my end, but sure, we can change it.

dridi:
I'm pretty sure I was initially of the same opinion until I discussed it with Walid and gave actual thought to this angle.

nigoroll (approved):
`+0`, no strong feelings either way.

dridi:
> (but I would appreciate to not use "off by one" for the title)

It was meant to be a classic joke, but if it bothers you I can change it.

nigoroll:
@dridi feel free to merge.
I just think "off by one" is not a good decision, because it sounds like "off by one error", which it is not. It's rather "support 500 status also".

PR #4307, opened 2025-03-31
gzip: Allocate gzip buffers from storage

dridi:
Arbitrary allocation from storage was introduced to solve the h2
head-of-line blocking caused by early DATA frames, allowed by the
initial control flow window of new streams. The decision could have
been made then to allocate the window buffer from the heap, but it
was tied to storage instead, offering better control over the global
memory footprint of the cache process.

The gzip buffer is tied to a task, but too large to allocate from
workspace without posing a risk. Even worse, there may be multiple
concurrent gzip operations in a single task. For example a gunzip
needed to parse ESI followed by a gzip to store a compressed body.

For a similar reason, it was not opportune to allocate the h2 stream
window buffer from workspace, despite being tied to the task.

Following the same logic, the gzip allocation can be performed from
storage to remove one wild card in our heap consumption. Another
possibility could have been the addition of a mempool, and it could
also have been an alternative for the h2 stream window, but transient
storage seemed more appropriate and it matches the on-demand malloc
behavior.

The gzip buffer logic could have been better encapsulated, but the
amount of direct access to the m_buf field would have resulted in a
lot of noise. Most of the noise here is caused by the two function
signatures changed to take a worker argument.

### DIFF
--- a/bin/varnishd/cache/cache_esi_fetch.c
+++ b/bin/varnishd/cache/cache_esi_fetch.c
@@ -146,7 +146,7 @@ vfp_esi_end(struct vfp_ctx *vc, struct vef_priv *vef,
 	if (vef->vgz != NULL) {
 		if (retval == VFP_END)
 			VGZ_UpdateObj(vc, vef->vgz, VGZ_END);
-		if (VGZ_Destroy(&vef->vgz) != VGZ_END)
+		if (VGZ_Destroy(vc->wrk, &vef->vgz) != VGZ_END)
 			retval = VFP_Error(vc,
 			    "ESI+Gzip Failed at the very end");
 	}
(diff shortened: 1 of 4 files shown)


bsdphk:
Good idea: +1

nigoroll (approved):
:+1:, just a minor point: Do we really benefit from keeping `m_buf`?

dridi:
> 👍, just a minor point: Do we really benefit from keeping `m_buf`?

The only reason why I kept it was to limit the amount of noise in the transition from plain malloc to storage buffer.

I was planning to eventually trim the excess fields in the struct and "better encapsulate" the gzip buffer logic. I didn't do it right away because some direct field accesses are not straightforward and I don't want to rush it.

nigoroll:
@dridi so then merge it please

Issue #3844, opened 2022-09-07
(NULL == "") == true ?


nigoroll:
I am pretty sure this has been discussed before, but I stumbled over it again and would like to revisit because I experienced it to be very un-pola:

We print a C `NULL` string as `""`, but a comparison with `""` yields false.

gquintard:
agreed, it's fine for C devs, but it's confusing for the rest of the population. Changing that is a major breaking change though, all the VCLs would need to at least be reviewed and possibly updated

Dridi:
As long as a BOOL evaluation of a HEADER can be used to test existence, I have no problem with `HEADER == STRING` to turn be degraded to `STRING == STRING` (and turning a non existent header into a string would result in an empty string, just like an assignment would).

gquintard:
I think I'd prefer for the HEADER to be cast into a STRING and then into a BOOL, with `""` being false-ish, and have a `std.header_exists(HEADER)` function to actually test presence.

nigoroll:
The case here was with a vmod function returning `NULL`, but I think @gquintard 's suggestion has some merit to it: `NULL` could signify "undefined/unset", but still evaluate to `""` if used as a string, then the following could hold true:
* `std.defined(req.http.inexistent-header) == false`
* `(req.http.inexistent-header == "") == true`

nigoroll:
Summary of bugwash discussion:

The current proposal basically boils down to
* for type conversions/folding, preserve `NULL`: `Typea(NULL) => Typeb(NULL)`
* for string contexts, convert `NULL` to `""`: `String(NULL) => ""`
* for boolean contexts, convert `NULL` to false: `BOOL(NULL) => false`

So far, the rules do not require an exception for comparisons, the [examples](https://github.com/varnishcache/varnish-cache/issues/3844#issuecomment-1243632961) can be implemented with these rules only (and the existing string comparison).

Some aspects relating to `String(NULL) => ""` were discussed:

* If `NULL` handling was to be transitive, `set` would need to have `unset` semantics such that `!req.http.a == !req.http.inexistent-header` holds for
```
set req.http.a = req.http.inexistent-header;
```

* But as long as we want to preserve `String(NULL) => ""`, such that in the example above, header `a` will be set to the empty string (as it is now), we will need to define `BOOL("") == true` such that `BOOL(req.http.empty-header) == true` and `BOOL(req.http.inexistent-header) == false`. This is the case with currecnt code.

[...]

Issue #3634, opened 2021-06-25
cookie.get() return value incorrect when called with nonexistent cookie


mszabo-wikia:
## Expected Behavior
cookie.get() should return an empty string if no such cookie was present in the request header, as [documented](https://varnish-cache.org/docs/trunk/reference/vmod_cookie.html#cookie-get)

## Current Behavior
On Varnish 6.6.0, cookie.get() returns something other than an empty string/empty value.

## Steps to Reproduce (for bugs)
Simple varnishtest reproducer:
```
varnishtest "cookie.get() with nonexistent cookie"

server s1 {
    rxreq
    expect req.http.X-Cookie-Present == "0"
    txresp
} -start

varnish v1 -vcl+backend {
    vcl 4.1;
    import cookie;

    sub vcl_recv {
        if (req.http.Cookie) {
            cookie.parse(req.http.Cookie);

            if (cookie.get("no_such_cookie") != "") {
                set req.http.X-Cookie-Present = "1";
            } else {
                set req.http.X-Cookie-Present = "0";
            }
        }
        return(pass);
    }
} -start

client c1 {
    txreq -hdr "Cookie: some_cookie=foo;other_cookie=bar"
    rxresp
} -run
```

The above test also fails if the condition is changed to `if (!cookie.get("no_such_cookie"))`. `cookie.isset()` however works as expected.
## Context
I wanted to use `cookie.get("no_such_cookie") != ""` to handle both the case where the cookie is not set, or where it's set but the value is empty.

## Your Environment
* Version used: 6.6.0
* Operating System and version: Debian Buster

Dridi:
That's probably because `cookie.get()` returns null and not an empty string.

The same would happen with a non-existent header:

```
varnishtest 3634

varnish v1 -vcl {
        backend be none;
        sub vcl_recv {
                return (synth(200));
        }
        sub vcl_synth {
                if (req.http.Cookie) {
                        set resp.http.X-Cookie-Present = "1";
                } else {
                        set resp.http.X-Cookie-Present = "0";
                }
                if (req.http.Cookie == "") {
                        set resp.http.X-Cookie-Empty = "1";
                } else {
                        set resp.http.X-Cookie-Empty = "0";
                }
        }
} -start

client c1 {
        txreq
        rxresp
        expect resp.http.X-Cookie-Present == 0
        expect resp.http.X-Cookie-Empty == 0
} -run
```

hermunn:
[...]

PR #4309, opened 2025-04-07
vcl: Make 'none' a reserved keyword and use it for probes

dridi:
:warning: This is technically a VCL BREAKING CHANGE but we should make it. :warning:

This patch series does several things, but the effective goal is to enable this:

```vcl
backend improbable {
        .host = "legacy.example.com";
        .probe = none;
}
```

Having backends we can't probe also prevents the use of a default probe, so my suggestion here is to allow `none` as a special value for the `.probe` field in backend definitions.

This is a breaking change, but I believe we should treat it as a bug fix and proceed without concern for broken VCL. In exchange, it should of course be one of the headlines in the release notes.

Once `none` becomes a reserved word, we can imagine solving the recurring confusion around unset values:

- https://github.com/varnishcache/varnish-cache/issues/2846
- https://github.com/varnishcache/varnish-cache/issues/3634
- https://github.com/varnishcache/varnish-cache/issues/3844

We could imagine the two if statements becoming equivalent:

```vcl
if (req.http.foo) { }
if (req.http.foo != none) { }
```

In the :warning: ACTUALLY BREAKING :warning: department, some would like to change the behavior of `set HEADER = HEADER;`:

```vcl
unset req.http.missing;
set req.http.assigned = req.http.missing;
# expect both headers to be 'none' here
```

I personally think it's an interesting question that should extend to `set HEADER = STRING;` for null strings. For comparisons in general, it could apply to any expression that can yield a null pointer. Something to consider for the next revision of VCL.

This pull requests stops after turning `none` into the keyword it should have become on day one, in hindsight, and then uses this keyword to explicitly circumvent the default probe.

### DIFF
--- a/bin/varnishtest/tests/c00136.vtc
+++ b/bin/varnishtest/tests/c00136.vtc
@@ -1,0 +1,20 @@
+varnishtest "none probe"
+
+server s0 {
+	rxreq
+	txresp -status 500
+} -dispatch
+
+varnish v1 -vcl {
+	probe default {
+		.interval = 100ms;
+	}
+	backend s0 {
+		.host = "${s0_sock}";
+		.probe = none;
+	}
+} -start
+
+delay 1
+
+varnish v1 -cliexpect healthy "backend.list -p"
--- a/bin/varnishtest/tests/m00000.vtc
+++ b/bin/varnishtest/tests/m00000.vtc
@@ -142,7 +142,7 @@ varnish v1 -errvcl {Symbol not found: 'obj'} {
 
 varnish v1 -errvcl {rot13: VFP already registered (per-vcl)} {
 	import debug;
-	backend none none;
+	backend be none;
 	sub vcl_init {
 		debug.rot104();
 	}
@@ -167,7 +167,7 @@ varnish v1 -errvcl {Failed from VCL} {
 
 varnish v1 -errvcl {Failed initialization} {
 	import debug;
-	backend none none;
+	backend be none;
 	sub vcl_init {
 		new fails = debug.obj("fail");
 	}
@@ -177,7 +177,7 @@ shell {
 	cat >${tmpdir}/f1 <<-EOF
 	vcl 4.1;
 	import debug;
-	backend none none;
+	backend be none;
 	sub vcl_init {
 		new fails = debug.obj("fail");
 	}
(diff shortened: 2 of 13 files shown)


nigoroll:
So this looks very sensible to me.

Just one question: Is it really the best option to keep `NULL` for "default probe", or should `NULL` become "no probe" and instead introduce a magic value for "default probe"?

bin/varnishtest/tests/m00000.vtc
```diff
@@ -142,7 +142,7 @@ varnish v1 -errvcl {Symbol not found: 'obj'} {
```
nigoroll, line 142: On the commit message:
As with other versioning questions, I think we should only increase the major version for _silent_ breaking changes. In this case a VCL compilation failure ensures that VCL authors become aware of the change, so we do not need a VCL version bump.

bin/varnishtest/tests/c00136.vtc
```diff
@@ -0,0 +1,20 @@
...
+	txresp -status 500
+} -dispatch
+
+varnish v1 -vcl {
+	probe default {
+		.interval = 100ms;
+	}
+	backend s0 {
+		.host = "${s0_sock}";
+		.probe = none;
```
nigoroll, line 14: Can we also make sure please that this also works for a `VCL_PROBE` vmod function/method argument?

nigoroll (changes requested):
I am fine with the VRT argument, but noticed one important aspect during a second review round

bin/varnishtest/tests/m00000.vtc
dridi: In this case I think we should consider it a bug fix, we should never have authorized `none` to persist as a usable symbol name once we gave it a special meaning.

bin/varnishtest/tests/c00136.vtc
dridi: I don't think we can pass `none` for a backends either, that would be a subsequent change (among a couple others) should this one be approved.

bin/varnishtest/tests/m00000.vtc
nigoroll: Yes

bin/varnishtest/tests/c00136.vtc
nigoroll: We can always define a none backend and then pass that, so that is not so important. For probes we don't have that option.

dridi: Right, so you'd like to align this change with backend definitions?

```vcl
probe disabled none;

backend improbable {
       .probe = none;
}

backend also_improbable {
       .probe = disabled;
}
```

That would increase overall consistency, good idea.

nigoroll:
@dridi [your last comment](https://github.com/varnishcache/varnish-cache/pull/4309#issuecomment-3019298638) aligns with my understanding.

PR #4315, opened 2025-04-14
storage: pass the caller's struct boc * to the objiterator_f

nigoroll:
Storage body iterator functions need to hold a reference to a `struct boc` while the object is being created, if only to call `ObjWaitExtend()` and/or `ObjWaitState()` in order to only deliver available data.

Before this patch, the design was to either have zero or two `boc` references in a client worker:

```c
cnt_transmit() {
	// ...
	req->boc = HSH_RefBoc(req->objcore);

// then later, called via VDP iterate:

xxx_iterator() {
	// ...
	boc = HSH_RefBoc(oc);
```

Now between the two calls, the `boc` might transition to `BOS_FINISHED` or `BOS_FAILED`, such that the second call from the iterator does not get a `boc` reference, because `HSH_RefBoc()` rightly does not return any for these states:

```c
HSH_RefBoc(const struct objcore *oc) {
	// ...
        if (boc != NULL) {
		assert(boc->refcount > 0);
		if (boc->state < BOS_FINISHED)
			boc->refcount++;
		else
			boc = NULL;
	}
}
```

This logic is important because otherwise a `boc` might exist forever.

Now a problem arises for storage engines which need different methods to iterate over busy objects vs. non-busy objects.

https://gitlab.com/uplex/varnish/slash/-/issues/104 documents such a case where the iterating thread holds a `boc` reference from `cnt_transmit()`, but none on the iterator level and thus deadlocks waiting for an unbusy event which never comes (because the thread itself holds the boc reference).

### DIFF
--- a/bin/varnishd/cache/cache.h
+++ b/bin/varnishd/cache/cache.h
@@ -754,7 +754,7 @@ typedef int objiterate_f(void *priv, unsigned flush,
 #define OBJ_ITER_FLUSH	0x01
 #define OBJ_ITER_END	0x02
 
-int ObjIterate(struct worker *, struct objcore *,
+int ObjIterate(struct worker *, struct objcore *, struct boc *,
     void *priv, objiterate_f *func, int final);
 
 vxid_t ObjGetXID(struct worker *, struct objcore *);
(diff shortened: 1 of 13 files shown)


nigoroll:
Actually... Now that I have written up the reasoning and it all seemed to make perfect sense, I pondered the topic again to come to the conclusion that it does not: As long as the iterator respects the bounds set by `ObjWaitExtend()`, it should use whatever state it finds, irrespective of the presence of a busyobj.
Closing this issue, at least for now.

PR #4317, opened 2025-04-17
vtree: Inject contextual parent checks during import

dridi:
This is an attempt at adding assertions without keeping track of a specific line of code. Tested with tree.h from the current tip of the main branch.

Refs freebsd/freebsd-src@a962800a09a45b8f702ebd1a06b8f20bc84e97b1


tools/import_vtree_from_freebsd.sh
```diff
@@ -16,8 +16,7 @@ git diff vtree.h | git apply -R > /dev/null 2>&1 || true
 GR=f6e54eb360a78856dcde930a00d9b2b3627309ab
```
dridi, line 16: The commit hash currently tracked by `$GR` does not match our current import of `vtree.h`.

nigoroll:
bugwash ok

dridi:
bugwash: ok

PR #4319, opened 2025-04-28
Seems like fedora:latest now new awk explicitly installed

simonvik:
Not tested at all but might fix issues like : 

https://app.circleci.com/pipelines/github/varnishcache/varnish-cache/6980/workflows/5a566469-b5e6-4e5d-b6a0-bba3f3b2822a/jobs/89714


simonvik:
Note that compile on fedora:latest still fail but that needs to be added to pkg-varnish-cache i think

PR #4320, opened 2025-04-29
param: Add shutdown_timeout

nigoroll:
Similar to why we added `startup_timeout`, orderly shutdown of a varnishd worker process might take considerably longer than cli_timeout. Consequently, we add `shutdown_timeout`.


nigoroll:
Good idea by @dridi : add delay to the debug stevedore for testing

nigoroll:
bugwash: @bsdphk wants to ponder this in light of new developments wrt process managemnt

PR #4322, opened 2025-04-30
param: New uncacheable_ttl for the built-in VCL

dridi:
This way the hit-for-miss TTL can be updated without requiring a VCL
reload or a dynamic vcl_beresp_hitmiss override involving a VMOD.

Why expose parameters to VCL you ask? See below the explanation from the first commit:

--->8----->8----->8----->8----->8----->8----->8----->8----->8---

The goal is to allow constructs avoiding hard-coded values when they
could be relative to existing parameters.

Let's take the first one in the list for example, backend_idle_timeout.
In a clustered deployment, with our without multiple tiers, this timeout
can be used to increase timeout_idle for a fellow Varnish server in a
way that allows it to pool its connections:

    sub vcl_recv {
        if (req.http.via ~ "varnish") {
            set sess.timeout_idle = param.backend_idle_timeout * 1.5;
        }
    }

One would obviously have a more reliable way of identifying a member of
the cluster, but assuming a similar configuration across the cluster,
this enables adaptive timeouts without resorting to VCL reloads or VMODs
to access parameters.

--->8----->8----->8----->8----->8----->8----->8----->8----->8---

See the 3rd point from https://github.com/varnishcache/varnish-cache/issues/3246#issuecomment-2765837882.


dridi:
I think I added all the missing bits and pieces, ready for review.

dridi:
I accidentally pushed to master.

asadsa92:
A comment that I had when I was first introduced to this was if parameters should be `cached` at the beginning of the task.
With the current solution, you may end up in a limbo state where your VCL is dealing with both the old value and the new value of the parameters.

dridi:
It would be rather easy to make a snapshot of `cache_param` in a dynamic PRIV_TASK on first use.

Issue #4323, opened 2025-05-02
Security docs on read-only params outdated?


eest:
### Expected Behavior

I was looking at
https://varnish-cache.org/docs/trunk/users-guide/run_security.html#parameters
and
https://varnish-cache.org/docs/trunk/reference/varnishd.html#security-options

They both mention some parameters that can be a good idea to mark as read-only. What makes this confusing is that e.g. the mentioned `vcc_allow_inline_c` is then documented as
> Deprecated alias for the vcc_feature parameter.

This in turn made me think "OK, i will just mark `vcc_feature` as read-only then.", e.g. `-r vcc_feature`

### Current Behavior

When testing this more it turned out that even if `vcc_feature` is read-only, it can still be changed if instead doing so via `param.set vcc_allow_inline_c on` if it is not also marked as read-only, e.g.:
```
varnish> param.show vcc_allow_inline_c
200
vcc_allow_inline_c
        Value is: none,+err_unref
        Alias of: vcc_feature

        Deprecated alias for the vcc_feature parameter.

varnish> param.set vcc_allow_inline_c on
200

varnish> param.show vcc_allow_inline_c
200
vcc_allow_inline_c
        Value is: none,+err_unref,+allow_inline_c
        Alias of: vcc_feature

        Deprecated alias for the vcc_feature parameter.


varnish> param.show vcc_feature
200
vcc_feature
        Value is: none,+err_unref,+allow_inline_c
        Default is: none,+err_unref,+unsafe_path

        Enable/Disable various VCC behaviors.
           default             Set default value (deprecated: use
                               param.reset)
           none                Disable all behaviors.

        Use +/- prefix to enable/disable individual behavior:
           err_unref           Unreferenced VCL objects result in
                               error.
           allow_inline_c      Allow inline C code in VCL.
           unsafe_path         Allow '/' in vmod & include paths. Allow
                               'import ... from ...'.

        NB: This parameter is protected and cannot be changed.
```

It seems either the docs need to be updated to also mention `vcc_feature`, or maybe it is unexpected that a read-only `vcc_feature` can still be modified via the deprecated aliases? The other way around also works, if `vcc_allow_inline_c` is marked read-only the `vcc_feature` can still be modified to enable that feature by including `+allow_inline_c`.

### Context

It seems easy to not properly lock down the vcc related params the way things are currently documented/working.

### Varnish Cache version

[...]

PR #4324, opened 2025-05-02
param: Move alias resolution before protected check

dridi:
param: Move alias resolution before protected check

Instead of resolving tweaks when the function is called, this is now
done in the MGT code performing the protected check. Since aliases may
be used to reset a single bit of another parameter (namely vcc_feature)
the default value is looked up before the alias resolution.

Unfortunately, that also means resolving deprecated aliases before
showing them to the user, adding a little bit of duplicated logic.

Refs #4323

### DIFF
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -155,6 +155,18 @@ mcf_addpar(struct parspec *ps)
 	VTAILQ_INSERT_TAIL(&phead, pl, list);
 }
 
+static const struct parspec *
+mcf_alias(struct parspec *alias, const struct parspec *pp)
+{
+	const struct parspec *orig;
+
+	orig = TRUST_ME(pp->priv);
+	AN(orig);
+	memcpy(alias, orig, sizeof *alias);
+	alias->priv = TRUST_ME(orig);
+	return (alias);
+}
+
 /*--------------------------------------------------------------------
  * Wrap the text nicely.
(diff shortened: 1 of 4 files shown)


dridi:
bugwash: OK, waiting for @nigoroll's review.

bin/varnishd/mgt/mgt_param.c
```diff
@@ -155,6 +155,18 @@ mcf_addpar(struct parspec *ps)
...
 }
 
+static const struct parspec *
+mcf_alias(struct parspec *alias, const struct parspec *pp)
+{
+	const struct parspec *orig;
+
+	orig = TRUST_ME(pp->priv);
+	AN(orig);
+	memcpy(alias, orig, sizeof *orig);
```
nigoroll, line 165: nitpick `sizeof *alias` as a "I know my buffer overflows" style checkbox?

nigoroll (approved):

nigoroll:
FTR, flexelint is happy with the tree as of f26f3f7f942b9f619ba241cbbe7a646773fba3be except for `arg_xyzzy_debug_argtest` which I had looked after in the meantime (unrelated).

Reminder: Be aware of SQUASHMEs

bin/varnishd/mgt/mgt_param.c
dridi: Fair enough, I will change the `sizeof` operand.

Should we have a `COPY_OBJ(dest, src, magic)` macro with actual type checking?

PR #4330, opened 2025-05-15
Constify sentinel pointers

dridi:
I made this mistake of creating a variable pointing to a sentinel value, and forgot to seal the variable itself to prevent it from pointing to a different constant value. Then I wondered whether I could find other cases and hit 3 of them.

These are technically breaking changes since they are adding new ABI constraints, but it should only break invalid (and most likely nonexistent) use cases without compromising legitimate use cases. So I'm hereby waiving ABI concerns.


nigoroll (approved):

bsdphk:
OK with me.

PR #4332, opened 2025-05-16
cache_fetch: Send Content-Length: 0 for POST and PUT

### LINKED ISSUE #4331: on removing content-length for 0-length POST requests
### Expected Behavior

`POST` requests should be exempt from `content-length` stripping, maybe.

### Current Behavior

`POST` requests with an empty body are stripped from their `content-length` header

### Possible Solution

_No response_

```
varnishtest "POST"

server s1 {
	rxreq
	txresp
	expect req.http.content-length == 0
} -start

varnish v1 -vcl+backend {

} -start

client c1 {
	txreq -method POST -hdr "content-length: 0"
	rxresp
} -run

server s1 -wait
```

### Context

this is a follow-up on #4228, I'm not sure whether what I'm seeing here is an intended consequence of that fix.

Yvan on `discord` reported that `lighttpd` replies with a 411 if there's no `content-length` on a `POST` request, and that header is now read-only in VCL, so they can't fix it there.

Question is, who's at fault here? I would argue that Varnish is, but I'm not sure that's why I'm opening this ticket.

https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2 says:

>    A user agent SHOULD send a Content-Length in a request message when
   no Transfer-Encoding is sent and the request method defines a meaning
   for an enclosed payload body

`POST` qualifies here.

>   For example, a Content-Length header
   field is normally sent in a POST request even when the value is 0
   (indicating an empty payload body).

>   A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.

A `POST` usually expects a body, so I think it's fair to send a `content-length`.

### Varnish Cache version

varnishd (varnish-7.7.1 revision 2e8180f788715e5bc44df08479d60c9435d79bdd)

### Operating system

arch, btw

_No response_

nigoroll:
otherwise keep the existing logic of sending no Content-Length header if there is no request body.

This code runs before vcl_backend_fetch, so should VCL gain back control over Content-Length as proposed in #4331, it will be possible again to override this default.

Fixes #4331 for the special case

### DIFF
--- a/bin/varnishd/cache/cache_fetch.c
+++ b/bin/varnishd/cache/cache_fetch.c
@@ -411,8 +411,15 @@ vbf_stp_startfetch(struct worker *wrk, struct busyobj *bo)
 
 	http_PrintfHeader(bo->bereq, "X-Varnish: %ju", VXID(bo->vsl->wid));
 
-	if (bo->bereq_body == NULL && bo->req == NULL)
-		http_Unset(bo->bereq, H_Content_Length);
+	if (bo->bereq_body == NULL && bo->req == NULL) {
+		const char *met = http_GetMethod(bo->bereq);
+		if (http_method_eq(met, POST) ||
+		    http_method_eq(met, PUT) ||
+		    http_method_eq(met, PATCH))
+			http_ForceHeader(bo->bereq, H_Content_Length, "0");
+		else
+			http_Unset(bo->bereq, H_Content_Length);
+	}
 
 	VCL_backend_fetch_method(bo->vcl, wrk, NULL, bo, NULL);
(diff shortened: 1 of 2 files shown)


dridi:
See https://github.com/varnishcache/varnish-cache/issues/4331#issuecomment-2886597725.

In the absence of a bereq.body, I think we should rather strip `content-length: 0` based on the method (when it is known not to anticipate data) and force the addition of `content-length: 0` for other methods.

nigoroll:
Superseded by #4340

PR #4337, opened 2025-05-22
backend: add VSCs for backend errors

asadsa92:
Abstract vbe_close_acct, this is supposed to mimic the ses_close_acct for backends.
    
We have up until now not exposed `htc->doclose` (backend errors) to any counters as we do for `req->doclose` (session errors).
    
That is unfortunate as these errors code are useful when trubleshooting fetches, so expose them as `bc_*` counters which is a subset of the `sc_*` counters.
    
Also add counters for backend closes: `backend_closed` and `backend_closed_err`.

### DIFF
--- a/bin/varnishd/cache/cache_backend.c
+++ b/bin/varnishd/cache/cache_backend.c
@@ -350,6 +350,36 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
 	return (pfd);
 }
 
+/*--------------------------------------------------------------------
+ * Update bc_ counters by reason (implementing ses_close_acct for backends)
+ *
+ * assuming that the approximation of non-atomic global counters is sufficient.
+ * if not: update to per-wrk
+ */
+
+static void
+vbe_close_acct(const struct pfd *pfd, stream_close_t reason)
+{
+
+	if (reason == SC_NULL) {
+		assert(PFD_State(pfd) == PFD_STATE_USED);
+		VSC_C_main->bc_tx_proxy++;
+		return;
+	}
+
+#define SESS_CLOSE_BACKEND
+#define SESS_CLOSE(U, l, err, desc)					\
+	if (reason == SC_ ## U) {					\
+		VSC_C_main->bc_ ## l++;					\
+		if (err)						\
+			VSC_C_main->backend_closed_err++;		\
+		return;							\
+	}
+#include "tbl/sess_close.h"
+
+	WRONG("Wrong event in vbe_close_acct");
+}
+
 static void v_matchproto_(vdi_finish_f)
 vbe_dir_finish(VRT_CTX, VCL_BACKEND d)
 {
@@ -369,11 +399,13 @@ vbe_dir_finish(VRT_CTX, VCL_BACKEND d)
 	pfd = bo->htc->priv;
 	bo->htc->priv = NULL;
 	if (bo->htc->doclose != SC_NULL || bp->proxy_header != 0) {
+		vbe_close_acct(pfd, bo->htc->doclose);
 		VSLb(bo->vsl, SLT_BackendClose, "%d %s close %s", *PFD_Fd(pfd),
 		    VRT_BACKEND_string(d), bo->htc->doclose->name);
 		VCP_Close(&pfd);
 		AZ(pfd);
 		Lck_Lock(bp->director->mtx);
+		VSC_C_main->backend_closed++;
 	} else {
 		assert (PFD_State(pfd) == PFD_STATE_USED);
 		VSLb(bo->vsl, SLT_BackendClose, "%d %s recycle", *PFD_Fd(pfd),
--- a/bin/varnishd/http1/cache_http1_fetch.c
+++ b/bin/varnishd/http1/cache_http1_fetch.c
@@ -228,9 +228,9 @@ V1F_FetchRespHdr(struct busyobj *bo)
 			VSLb(bo->vsl, SLT_FetchError, "Received junk");
 			htc->doclose = SC_RX_JUNK;
 			break;
-		case HTC_S_CLOSE:
+		case HTC_S_EOF:
 			VSLb(bo->vsl, SLT_FetchError, "backend closed");
-			htc->doclose = SC_RESP_CLOSE;
+			htc->doclose = SC_REM_CLOSE;
 			break;
 		case HTC_S_TIMEOUT:
 			VSLb(bo->vsl, SLT_FetchError, "timeout");
--- a/include/tbl/sess_close.h
+++ b/include/tbl/sess_close.h
@@ -32,8 +32,8 @@
 /*lint -save -e525 -e539 */
 
 // stream_close_t	  sc_* stat	is_err	Description
-SESS_CLOSE(REM_CLOSE,	  rem_close,	0,	"Client Closed")
-SESS_CLOSE(REQ_CLOSE,	  req_close,	0,	"Client requested close")
+SESS_CLOSE(REM_CLOSE,	  rem_close,	0,	"Peer Closed")
+SESS_CLOSE(REQ_CLOSE,	  req_close,	0,	"Peer requested close")
 SESS_CLOSE(REQ_HTTP10,	  req_http10,	1,	"Proto < HTTP/1.1")
 SESS_CLOSE(RX_BAD,	  rx_bad,	1,	"Received bad req/resp")
 SESS_CLOSE(RX_BODY,	  rx_body,	1,	"Failure receiving body")
--- a/include/tbl/sess_close.h
+++ b/include/tbl/sess_close.h
@@ -40,18 +40,22 @@ SESS_CLOSE(RX_BODY,	  rx_body,	1,	"Failure receiving body")
 SESS_CLOSE(RX_JUNK,	  rx_junk,	1,	"Received junk data")
 SESS_CLOSE(RX_OVERFLOW,   rx_overflow,	1,	"Received buffer overflow")
 SESS_CLOSE(RX_TIMEOUT,	  rx_timeout,	1,	"Receive timeout")
-SESS_CLOSE(RX_CLOSE_IDLE, rx_close_idle,0,	"timeout_idle reached")
 SESS_CLOSE(TX_PIPE,	  tx_pipe,	0,	"Piped transaction")
 SESS_CLOSE(TX_ERROR,	  tx_error,	1,	"Error transaction")
 SESS_CLOSE(TX_EOF,	  tx_eof,	0,	"EOF transmission")
 SESS_CLOSE(RESP_CLOSE,	  resp_close,	0,	"Backend/VCL requested close")
 SESS_CLOSE(OVERLOAD,	  overload,	1,	"Out of some resource")
-SESS_CLOSE(PIPE_OVERFLOW, pipe_overflow,1,	"Session pipe overflow")
-SESS_CLOSE(RANGE_SHORT,   range_short,	1,	"Insufficient data for range")
-SESS_CLOSE(REQ_HTTP20,	  req_http20,	1,	"HTTP2 not accepted")
-SESS_CLOSE(VCL_FAILURE,	  vcl_failure,	1,	"VCL failure")
-SESS_CLOSE(RAPID_RESET,	  rapid_reset,  1,      "HTTP2 rapid reset")
-SESS_CLOSE(BANKRUPT,	  bankrupt,	1,      "HTTP2 credit bankruptcy")
+#ifndef SESS_CLOSE_BACKEND
+  SESS_CLOSE(RX_CLOSE_IDLE,	rx_close_idle,	0,	"timeout_idle reached")
+  SESS_CLOSE(PIPE_OVERFLOW,	pipe_overflow,	1,	"Session pipe overflow")
+  SESS_CLOSE(RANGE_SHORT,	range_short,	1,	"Insufficient data for range")
+  SESS_CLOSE(REQ_HTTP20,	req_http20,	1,	"HTTP2 not accepted")
+  SESS_CLOSE(VCL_FAILURE,	vcl_failure,	1,	"VCL failure")
+  SESS_CLOSE(RAPID_RESET,	rapid_reset,	1,	"HTTP2 rapid reset")
+  SESS_CLOSE(BANKRUPT,		bankrupt,	1,	"HTTP2 credit bankruptcy")
+#else
+  #undef SESS_CLOSE_BACKEND
+#endif
 #undef SESS_CLOSE
 
 /*lint -restore */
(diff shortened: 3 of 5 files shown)


include/tbl/sess_close.h
```diff
@@ -40,18 +40,22 @@ SESS_CLOSE(RX_BODY,	  rx_body,	1,	"Failure receiving body")
...
 SESS_CLOSE(TX_EOF,	  tx_eof,	0,	"EOF transmission")
 SESS_CLOSE(RESP_CLOSE,	  resp_close,	0,	"Backend/VCL requested close")
 SESS_CLOSE(OVERLOAD,	  overload,	1,	"Out of some resource")
-SESS_CLOSE(PIPE_OVERFLOW, pipe_overflow,1,	"Session pipe overflow")
-SESS_CLOSE(RANGE_SHORT,   range_short,	1,	"Insufficient data for range")
-SESS_CLOSE(REQ_HTTP20,	  req_http20,	1,	"HTTP2 not accepted")
-SESS_CLOSE(VCL_FAILURE,	  vcl_failure,	1,	"VCL failure")
-SESS_CLOSE(RAPID_RESET,	  rapid_reset,  1,      "HTTP2 rapid reset")
-SESS_CLOSE(BANKRUPT,	  bankrupt,	1,      "HTTP2 credit bankruptcy")
+#ifndef SESS_CLOSE_BACKEND
```
nigoroll, line 48: I am not sure if this is the best option.

Should we maybe rather have `SESS_CLOSE_C()` for the client-only closure reasons?

bin/varnishd/cache/cache_backend.c
```diff
@@ -350,6 +350,36 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
...
 
+/*--------------------------------------------------------------------
+ * Update bc_ counters by reason (implementing ses_close_acct for backends)
+ *
+ * assuming that the approximation of non-atomic global counters is sufficient.
+ * if not: update to per-wrk
+ */
+
+static void
+vbe_close_acct(const struct pfd *pfd, stream_close_t reason)
```
nigoroll, line 361: Is this patch only half baked? This commit only adds one call to `vbe_close_acct` and for this call, incrementing `bc_tx_proxy` looks wrong, because `reason` is _not_ `SC_NULL`. Also I wonder why `bc_tx_proxy` should need special casing, why do we suddenly care about the call site, when otherwise the reason is just that?

bin/varnishd/http1/cache_http1_fetch.c
```diff
@@ -228,9 +228,9 @@ V1F_FetchRespHdr(struct busyobj *bo)
 			VSLb(bo->vsl, SLT_FetchError, "Received junk");
 			htc->doclose = SC_RX_JUNK;
 			break;
-		case HTC_S_CLOSE:
```
nigoroll, line 231: this particular change lgtm

nigoroll (changes requested):
Besides specific comments, I am not sure if adding global counters makes much sense. If users see these increase, won't they immediately ask "yeah, but which backend?". So unless there is some specific background which might need explanations, I would suggest to skip one iteration and go straight to per-backend counters.

include/tbl/sess_close.h
asadsa92: Yes, I think that would be a little bit nicer.
It would mimic something we already do in `include/tbl/obj_attr.h`.
Let me have a look.

bin/varnishd/cache/cache_backend.c
asadsa92: I made `vbe_close_acct` into a function to mimic `ses_close_acct`. It being called from one call-site was intended.

`reason` can be `SC_NULL` because this if statement is an OR, see https://github.com/varnishcache/varnish-cache/pull/4337/files#diff-fe1fb2709016ca15daea3e926b7cec272966ee397c5bafe50d6967023dd9f4a7R401

If you remove this handling, the test `tests/o00002.vtc` will run into an assert:
```
***  v2    debug|Error: Child (1249377) Panic at: Thu, 22 May 2025 09:20:05 GMT 
  14 ***  v2    debug|Wrong turn at cache/cache_backend.c:381:                       
  15 ***  v2    debug|Wrong event in vbe_close_acct                                  
  16 ***  v2    debug|version = varnish-trunk revision f15259b053765f675a10b40e0e6d91a0847ac065, vrt api = 21.0
  17 ***  v2    debug|ident = Linux,6.1.0-34-amd64,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,epoll
  18 ***  v2    debug|now = 392362.874551 (mono), 1747905603.630655 (real)           
  19 ***  v2    debug|Backtrace:                                                     
  20 ***  v2    debug|  ip=0x564002279265 sp=0x7f60b036a390 <VBT_format+0x35>        
  21 ***  v2    debug|  ip=0x5640021acdb3 sp=0x7f60b036a3b0 <pan_backtrace+0x33>     
  22 ***  v2    debug|  ip=0x5640021acaf7 sp=0x7f60b036a3d0 <pan_ic+0x377>           
  23 ***  v2    debug|  ip=0x5640022782db sp=0x7f60b036a550 <VAS_Fail+0x4b>          
  24 ***  v2    debug|  ip=0x56400216dd91 sp=0x7f60b036a590 <vbe_close_acct+0x3a1>   
  25 ***  v2    debug|  ip=0x56400216b5ab sp=0x7f60b036a5b0 <vbe_dir_finish+0x3cb>   
  26 ***  v2    debug|  ip=0x564002181419 sp=0x7f60b036a610 <VDI_Finish+0x249>       
  27 ***  v2    debug|  ip=0x56400218e318 sp=0x7f60b036a6c0 <vbf_cleanup+0x158>      
  28 ***  v2    debug|  ip=0x5640021932f2 sp=0x7f60b036a6e0 <vbf_stp_fetchend+0x1f2> 
  29 ***  v2    debug|  ip=0x56400218f364 sp=0x7f60b036a730 <vbf_fetch_thread+0x5f4> 
  30 ***  v2    debug|  ip=0x5640021edb47 sp=0x7f60b036a810 <Pool_Work_Thread+0x7c7> 
  31 ***  v2    debug|  ip=0x5640021ed243 sp=0x7f60b036a8a0 <WRK_Thread+0x333>       
  32 ***  v2    debug|  ip=0x5640021ececb sp=0x7f60b036b430 <pool_thread+0xcb>       
  33 ***  v2    debug|  ip=0x7f60b92a81f5 sp=0x7f60b036b460 <pthread_condattr_setpshared+0x515>
  34 ***  v2    debug|  ip=0x7f60b932889c sp=0x7f60b036b500 <__xmknodat+0x23c>       
```

GDB:
```gdb
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007fdbc20a9f4f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007fdbc205afb2 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fdbc2045472 in __GI_abort () at ./stdlib/abort.c:79
#4  0x0000558df4ab5d0c in pan_ic (func=0x558df4badbe0 "vbe_close_acct", file=0x558df4bad312 "cache/cache_backend.c", line=381, cond=0x558df4badbd1 "Wrong event in vbe_close_acct", kind=VAS_WRONG) at cache/cache_panic.c:791
#5  0x0000558df4b812db in VAS_Fail (func=0x558df4badbe0 "vbe_close_acct", file=0x558df4bad312 "cache/cache_backend.c", line=381, cond=0x558df4badbd1 "Wrong event in vbe_close_acct", kind=VAS_WRONG) at vas.c:107
#6  0x0000558df4a76d91 in vbe_close_acct (pfd=0x7fdbb4a00000, reason=0x558df4c25600 <SC_NULL>) at cache/cache_backend.c:381
#7  0x0000558df4a745ab in vbe_dir_finish (ctx=0x7fdbbf5a6610, d=0x7fdbc181d7b0) at cache/cache_backend.c:403
#8  0x0000558df4a8a419 in VDI_Finish (bo=0x7fdbb667d460) at cache/cache_director.c:195
#9  0x0000558df4a97318 in vbf_cleanup (bo=0x7fdbb667d460) at cache/cache_fetch.c:134
#10 0x0000558df4a9c2f2 in vbf_stp_fetchend (wrk=0x7fdbbf5a7340, bo=0x7fdbb667d460) at cache/cache_fetch.c:748
#11 0x0000558df4a98364 in vbf_fetch_thread (wrk=0x7fdbbf5a7340, priv=0x7fdbb667d460) at cache/cache_fetch.c:1123
#12 0x0000558df4af6b47 in Pool_Work_Thread (pp=0x7fdbbf5fd000, wrk=0x7fdbbf5a7340) at cache/cache_wrk.c:496
#13 0x0000558df4af6243 in WRK_Thread (qp=0x7fdbbf5fd000, stacksize=81920, thread_workspace=2048) at cache/cache_wrk.c:157
#14 0x0000558df4af5ecb in pool_thread (priv=0x7fdbbe0000e0) at cache/cache_wrk.c:529
#15 0x00007fdbc20a81f5 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#16 0x00007fdbc212889c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
(gdb) frame 6
#6  0x0000558df4a76d91 in vbe_close_acct (pfd=0x7fdbb4a00000, reason=0x558df4c25600 <SC_NULL>) at cache/cache_backend.c:381
381		WRONG("Wrong event in vbe_close_acct");
(gdb) p reason
$1 = (stream_close_t) 0x558df4c25600 <SC_NULL>
(gdb)
```

As you can see reason is `SC_NULL`.

The reason for why `SC_NULL` require special handling is because that value is not handled from the table (`include/tbl/sess_close.h`).

nigoroll: Sorry, yes, you are right, I overlooked `|| bp->proxy_header != 0`.

So what is happening here is that we do not want to recycle a connection if it was opened with a proxy header because otherwise a request could be sent "under" a wrong proxy header.

So I can follow now the argument that this is in fact a special close reason, but I would still think it should get its own entry in `sess_close.h`, be adjusted in `vbe_dir_finish()` and handled like any other in `vbe_close_acct()`.

So why is `vbe_close_acct()` not called from `vbe_dir_getfd()`?

nigoroll: > I wanted to fix it but also did not want to introduce too many changes in this PR.

We have PRs with a lot of changes. So this is more abut structure. For example, I think you could introduce the new SC to the old code, then add `vbe_close_acct()` and it would all fall in place.

> I could be convinced to change how `vbe_dir_getfd()`

I do not know what else you might have in mind, but I do think that we should account for all the cases where we close connections, including the "proxy write failure" in `vbe_dir_getfd()`. I have not thought about the details, but I do not think we can leave out a case.

asadsa92: Great, I will make an attempt at this.

asadsa92:
> Besides specific comments, I am not sure if adding global counters makes much sense. If users see these increase, won't they immediately ask "yeah, but which backend?". So unless there is some specific background which might need explanations, I would suggest to skip one iteration and go straight to per-backend counters.

Good point, this was actually mentioned to me. I forgot to take that into account, but I think that make sense.
However, I think we should keep the global counters such that further breakdown is available for the interested users.
The global counter is a quick way to tell if looking into backends make sense.

bin/varnishd/cache/cache_backend.c
asadsa92: > > I wanted to fix it but also did not want to introduce too many changes in this PR.
> 
> We have PRs with a lot of changes. So this is more abut structure. For example, I think you could introduce the new SC to the old code, then add `vbe_close_acct()` and it would all fall into place.

So, I had an attempt at this. I am starting to understand why things are laid out as they are today.
The reason why we can not use a SC for PROXY is due to a couple of reasons:
- First, `SC_NULL` is used as a sink multiple places. If `bo->htc->doclose` is `SC_NULL` we know that the fetch has not failed yet. Changing this would mean that we would always have to take PROXY into account. No big deal, but are we sure we want to go this route?
- Another reason is that we can not latch another reason on top of `bo->htc->doclose = SC_TX_PROXY;`. Example, how do we report a fetch has failed due to a timeout (`SC_RX_TIMEOUT`) when we are dealing with a PROXY transaction?

include/tbl/sess_close.h
asadsa92: See: https://github.com/varnishcache/varnish-cache/pull/4337/commits/07f58bb9b6c694fc11424c41a5a159e92b14d38a

asadsa92:
> So I _think_ this is OK, but I would like to ask @asadsa92 to make life easier for reviewers. Could you please

Thanks for your 2nd review, we are getting there :smile: 
Yes, I will come back to you with a polished PR.

bin/varnishd/cache/cache_backend.c
```diff
@@ -306,29 +385,29 @@ vbe_dir_getfd(VRT_CTX, struct worker *wrk, VCL_BACKEND dir, struct backend *bp,
...
-		bp->n_conn--;
-		bp->vsc->conn--;
-		bp->vsc->req--;
-		vbe_connwait_signal_locked(bp);
-		Lck_Unlock(bp->director->mtx);
-		vbe_connwait_fini(cw);
-		return (NULL);
+	if (bp->proxy_header != 0) {
+		err = VPX_Send_Proxy(*fdp, bp->proxy_header, bo->sp);
+		if (err < 0) {
```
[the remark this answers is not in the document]
asadsa92: I agree, I have reverted the nesting, see https://github.com/varnishcache/varnish-cache/pull/4337/commits/725e7164758a6086a0d35a423a650f87137f447e

bin/varnishd/cache/cache_backend.c
```diff
@@ -181,6 +181,79 @@ vbe_connwait_fini(struct connwait *cw)
...
 
+/*--------------------------------------------------------------------
+ * Update bc_ counters by reason (implementing ses_close_acct for backends)
+ *
+ * assuming that the approximation of non-atomic global counters is sufficient.
+ * if not: update to per-wrk
+ */
+
+static void
+vbe_close_acct(const struct pfd *pfd, stream_close_t reason)
```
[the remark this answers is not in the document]
asadsa92: Thanks, done.

bin/varnishd/cache/cache_backend.c
```diff
@@ -181,6 +181,79 @@ vbe_connwait_fini(struct connwait *cw)
...
+		if (err)						\
+			VSC_C_main->backend_closed_err++;		\
+		return;							\
+	}
+#include "tbl/sess_close.h"
+
+	WRONG("Wrong event in vbe_close_acct");
+}
+
+static void v_matchproto_(vdi_finish_f)
```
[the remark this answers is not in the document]
asadsa92: Good idea, I found that I had misplaced the opening curly for the function. See https://github.com/varnishcache/varnish-cache/pull/4337/commits/268b18286a1a3fd50800af25ad6acfb43aa9528f

Issue #4299, opened 2025-03-18
macOS build failure: no `endian.h` found


botantony:
### Expected Behavior

I expect to compile program

### Current Behavior

Build fails with this error:
```
    CC       libvarnish_la-vsb.lo
    CC       libvarnish_la-vsha256.lo
    CC       libvarnish_la-vss.lo
    CC       libvarnish_la-vsub.lo
  vsha256.c:35:12: fatal error: 'endian.h' file not found
     35 | #  include <endian.h>
        |            ^~~~~~~~~~
  1 error generated.
  make[2]: *** [libvarnish_la-vsha256.lo] Error 1
  make[2]: *** Waiting for unfinished jobs....
  make[1]: *** [install-recursive] Error 1
  make: *** [install-recursive] Error 1
```

### Possible Solution

In file [vsha256.c](https://github.com/varnishcache/varnish-cache/blob/master/lib/libvarnish/vsha256.c#L34) you use `#ifndef __DARWIN_BYTE_ORDER`, which suppose to prevent this error. `__DARWIN_BYTE_ORDER` is not a defined macro in Xcode Clang. You can try `__APPLE__` macro instead

### Steps to Reproduce (for bugs)

Build steps (from tarball):
```shell
$ ./autogen.sh # I tried w/ and w/o, got same result
$ ./configure --localstatedir=/opt/homebrew/var --disable-debug --disable-dependency-tracking --prefix=/opt/homebrew/Cellar/varnish/7.7.0 --libdir=/opt/homebrew/Cellar/varnish/7.7.0/lib
$ make install CFLAGS=-DVARNISH_VMOD_DIR='"/opt/homebrew/lib/varnish/vmods"' -DVARNISH_VCL_DIR='"/opt/homebrew/etc/varnish:/opt/homebrew/share/varnish/vcl"'
```

### Context

I contibute to Homebrew package manager, the recent update failed CI.
Related PR: Homebrew/homebrew-core#211340

### Varnish Cache version

7.7.0

### Operating system

macOS Sonoma

### Source of binary packages used (if any)

https://varnish-cache.org/_downloads/varnish-7.7.0.tgz

daeho-ro:
Even after the fix is applied, another one is coming from `7.7.1`.

```
vcc_purge_if.c:134:46: error: argument to 'section' attribute is not valid for this target: mach-o section specifier requires a segment and section separated by a comma
[...]

PR #4339, opened 2025-05-26
build: macos compatibility build fixes

chenrui333:
fixes https://github.com/varnishcache/varnish-cache/issues/4299
relates to https://github.com/Homebrew/homebrew-core/pull/223233

### DIFF
--- a/lib/libvarnish/vsha256.c
+++ b/lib/libvarnish/vsha256.c
@@ -31,7 +31,14 @@
 
 #include "config.h"
 
-#ifndef __DARWIN_BYTE_ORDER
+#ifdef __APPLE__
+#  include <machine/endian.h>
+#  include <libkern/OSByteOrder.h>
+#  define htobe32(x) OSSwapHostToBigInt32(x)
+#  define htobe64(x) OSSwapHostToBigInt64(x)
+#  define VBYTE_ORDER __DARWIN_BYTE_ORDER
+#  define VBIG_ENDIAN __DARWIN_BIG_ENDIAN
+#else
 #  include <endian.h>
 #  ifdef _BYTE_ORDER
 #    define VBYTE_ORDER	_BYTE_ORDER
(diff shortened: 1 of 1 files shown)


nigoroll:
@chenrui333 thank you for your patches. Regarding the endianness-related changes, I would like wait what phk thinks, but the `vmodtool.py` change looks simple and clean, so I cherry-picked it. Related to that, are there any MacOS-specifics on retrieving the version from a shared object? If yes, would you maybe share the respective command to be added [here](https://github.com/varnishcache/varnish-cache/blob/d2e2bf85e399d8a5882a123fed3c39d4e9b32292/doc/sphinx/reference/vmod.rst?plain=1#L1055-L1062)?

chenrui333:
> Related to that, are there any MacOS-specifics on retrieving the version from a shared object? If yes, would you maybe share the respective command to be added [here](https://github.com/varnishcache/varnish-cache/blob/d2e2bf85e399d8a5882a123fed3c39d4e9b32292/doc/sphinx/reference/vmod.rst?plain=1#L1055-L1062)?

yeah, you can use `otool` to extract the version info as below:

```
 > otool -s __TEXT vmod_vcs lib/varnish/vmods/libvmod_debug.so \
                                                                | awk '/^[0-9a-f]/ {for(i=2;i<=NF;i++) printf "%s",$i}' \
                                                                | xxd -r -p \
                                                                | strings
18e27f081788b5e5d44c480f6d9749c07d53ddb9
```

nigoroll:
> another thing is looks like there is no CI for macos, maybe we should also add one?

we would certainly welcome a contribution :) If you would add one, please have a look at https://github.com/varnishcache/varnish-cache/blob/master/tools/vtest.sh. You need to generate an ssh keypair and send your public key to phk (or me, as a backup) to have it accepted for reporting.

nigoroll:
> yeah, you can use `otool` to extract the version info as below:

Thank you, I added your recipe

chenrui333:
np at all, any concern of merging this PR? (we have already applied it on the homebrew side)

bsdphk:
I belive this has been fixed: the Mac-Mini VTester is happy.

PR #4347, opened 2025-06-12
build: fix duplicate OS_CODE definitions on macOS

sandydoo:
`TARGET_OS_MAC` and `__APPLE__` are both true on macOS, which results in duplicate definitions for `OS_CODE`.

Upstream removed the check for `TARGET_OS_MAC`, as well as the ancient compat code, in
https://github.com/madler/zlib/commit/4bd9a71f3539b5ce47f0c67ab5e01f3196dc8ef9.

This was fixed in varnish >= 7.6 as part of https://github.com/varnishcache/varnish-cache/commit/86df12b6c1ad8208899ea353fdcbea227356fcf8.

### DIFF
--- a/lib/libvgz/zutil.h
+++ b/lib/libvgz/zutil.h
@@ -137,17 +137,8 @@ extern z_const char * const z_errmsg[10]; /* indexed by 2-zlib_error */
 #  endif
 #endif
 
-#if defined(MACOS) || defined(TARGET_OS_MAC)
+#if defined(MACOS)
 #  define OS_CODE  7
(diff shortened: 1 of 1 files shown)


walid-git:
Thanks for your PR @sandydoo.  This was addressed via https://github.com/varnishcache/varnish-cache/pull/4424, could you confirm if you are now able to build with the current 6.0 branch ?

As this is not impacting a platform that we support, it does not justify a dedicated release, and will have to wait for the next one, which we have no ETA for at this point.

PR #4350, opened 2025-06-20
mgt: Add an intentional conflict with pthread.h

dridi:
If the macro check is not enough, this one hopefully catches stragglers.

Suggestion from @asadsa92.

Refs #4346


bsdphk:
I'm OK with this,but check that the CI failures are not indicative of trouble.

PR #4352, opened 2025-06-24
circleci: Sanitizers job fail on CI

dridi:
A lot of job executions fail with a handful of test logs containing stack traces looking like this:

    abort+0xdf
    _ZN11__sanitizer5AbortEv+0x30
    _ZN11__sanitizer3DieEv+0x4d
    _ZN11__sanitizer11CheckFailedEPKciS1_yy+0x8b
    _ZN11__sanitizer14ThreadRegistry12CreateThreadEmbjPv+0x45b
    _ZN6__asan10AsanThread6CreateEPKvmjPN11__sanitizer10StackTraceEb+0xa9
    __interceptor_pthread_create+0x128

In that case errno is set to EAGAIN, indicating either an actual lack of resources in the constrained cci environment, or a limit reached. After finding bug reports describing similar problems (but not this case) in which the question of the allocator comes up, this change is an attempt at improving the reliability of the cci job running with sanitizers.

---

This starts as a draft pull request, I may attempt other changes to mitigate spurious failures in CI, triggering background jobs while I work on something else.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -409,7 +409,7 @@ workflows:
           release: noble
           rclass: large
           extra_conf: --enable-asan --enable-ubsan --enable-workspace-emulator --without-jemalloc
-          make_target: check -j2
+          make_target: check
       - build:
           name: build_alpine
           dist: alpine
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -408,7 +408,7 @@ workflows:
           dist: ubuntu
           release: noble
           rclass: large
-          extra_conf: --enable-asan --enable-ubsan --enable-workspace-emulator
+          extra_conf: --enable-asan --enable-ubsan --enable-workspace-emulator --without-jemalloc
           make_target: check
       - build:
           name: build_alpine
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -377,49 +377,12 @@ workflows:
         - << pipeline.parameters.dist-url >>
     jobs:
       - build:
-          name: build_almalinux_8
-          dist: almalinux
-          release: "8"
-      - build:
-          name: build_almalinux_9
-          dist: almalinux
-          release: "9"
-      # fedora is our witness
-      - build:
-          name: build_fedora_latest
-          dist: fedora
-          release: latest
-          make_target: witness.dot
-      - build:
-          name: build_debian_bookworm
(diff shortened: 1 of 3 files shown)


dridi:
This is still happening after reducing the concurrency of the sanitizer job. I will attempt a retry policy (`make recheck`) limited to asan failures.

dridi:
The CCI failures are caused by libasan internal checks.

See this one for example:

```
AddressSanitizer: CHECK failed: sanitizer_thread_registry.cpp:161 "((tctx->status)) == ((ThreadStatusInvalid))" (0x78211cc0, 0x0) (tid=41508)
```

I was hoping to intercept such errors with `__asan_on_error()` but since this is not an actual error spotted by ASAN, it's not called (I was able witness its call on a legit buffer overflow I manually added).

In order to get a usable sanitizers reporting I was thinking of doing this:

- wrap all `varnishd` calls to `pthread_create()`
- the new `THR_Create()` function does more in ASAN builds
  - raise and clear a thread-local "creating thread" flag
- the panic handler does more in ASAN builds
  - new exit scenario when `errno == EAGAIN` while "creating thread"
- MGT identifies the cache ASAN bailout and exits accordingly
- varnishtest identifies the MGT bailout and terminates the test with the SKIP exit status

This is a rather involved chain of changes in order to get libasan working reliably in the constrained CCI environment. I think it would be worth it.

Another (long overdue?) possibility is to migrate all ad-hoc `pthread_create()` calls to `WRK_BgThread()` to end up with exactly two `pthread_create()` call sites in the same file, which is trivial to ASAN-ize. I at least like the idea of formalizing that all non-pooled threads turn into background workers.

In both cases, we need to eventually find out about libasan's internal failed check in varnishtest to avoid the spurious failure.

Thoughts?

dridi:
My proposal is to skip sanitizer steps if the system is too low on resources. Right now it only happens in CI, where we run build in constrained containers.

Right now, there's always a couple of failing tests for this very reason. We haven't had a streak of commits with green check mark in months. I'd like to reduce this noise on two premises:

- it's always less than a handful of test cases
- it's never specific test cases
- it's always during `varnishd`'s cache startup

So we know that all tests are frequently exercised, and (very) sporadically skipped at random.

We have never seen this problem on non-sanitizer builds, hence my suggestion to act on this problem this way.

PR #4354, opened 2025-07-02
vcc_acl: Add +fold(-report) sub-flag omit +fold warnings

nigoroll:
With a lot of folding going on, the warnings can easily bury more relevant CLI output.

### DIFF
--- a/doc/sphinx/reference/vcl.rst
+++ b/doc/sphinx/reference/vcl.rst
@@ -354,7 +354,7 @@ individually:
 
   Skip and fold operations on ACL entries are output as warnings
   during VCL compilation as entries from the VCL are processed in
-  order.
+  order unless the `-report` sub-flag is also given (see below).
 
   Logging under the ``VCL_acl`` tag can change with this parameter
   enabled: Matches on skipped subnet entries are now logged as matches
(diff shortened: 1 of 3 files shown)


doc/sphinx/reference/vcl.rst
```diff
@@ -365,6 +365,13 @@ individually:
 
   Negated ACL entries are never folded.
 
+* `+fold,noreport` - Fold without reporting
```
dridi: I think we should avoid introducing new syntax with a comma, we already have a precedent somewhere else:

```
param.set vsl_mask +VfpAcct,+VdpAcct
```

I think it would be simpler to introduce a new `report` flag defaulting to `+report` and therefore allow to flag an ACL with `-report` to disable ACL warnings, starting folding warnings and including future warnings.

lib/libvcc/vcc_acl.c
```diff
@@ -837,8 +889,7 @@ vcc_ParseAcl(struct vcc *tl)
 			acl->flag_log = sign;
 			vcc_NextToken(tl);
 		} else if (vcc_IdIs(tl->t, "fold")) {
-			acl->flag_fold = sign;
-			vcc_NextToken(tl);
+			vcc_parseAclFold(tl, sign);
```
walid-git: I think we should stop the loop if we failed in `vcc_parseAclFold`, otherwise we might get double error reporting. Maybe we can turn the loop into a `while (!tl->error)` ?

PR #4365, opened 2025-07-11
Add a special-purpose synth storage engine

nigoroll:
This PR includes #4358 as of 576e2a50f98bcd030372bebddce04628956a685e

Before this patch, creating a response body in vcl_synth {} involved two memcpy to heap operations: First to a vsb, then to a storage object.

The new "synth" storage engine simplifies this drastically in tandem with special casing in cnt_synth() and VRT_l_resp_body(): Constituents of the response are not copied, but rather referenced in a list of VSCARABs, which are the directly used for delivery.

Besides this body handling, the synth storage engine only supports the bare minimum object API calls.

To accomodate the "hand out VSCARAB" semantics instead of "here is a buffer to write to", ObjGetspace() is used in an incompatible, special way. We might want to consider adding a special purpose object API instead.

Also, there currently is no way for storage functions to get hold of the request workspace directly, so it is retrieved via the pthread key.

For buffers, simple malloc()/free() is used.

Performance numbers will be coming.

### DIFF
--- a/bin/varnishd/Makefile.am
+++ b/bin/varnishd/Makefile.am
@@ -109,6 +109,7 @@ varnishd_SOURCES = \
 	storage/storage_malloc.c \
 	storage/storage_debug.c \
 	storage/storage_simple.c \
+	storage/storage_synth.c \
 	storage/storage_umem.c \
 	waiter/cache_waiter.c \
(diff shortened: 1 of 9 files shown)


nigoroll:
bugwash: needs VSCs

PR #4368, opened 2025-07-18
http2: Move all socket operations to the session thread

dridi:
This is a port of a change initially made by @mbgrydeland first in Varnish Enterprise for planning reasons.

It's a major change of the architecture:

- all socket operations happen in the `h2_sess` thread
- the session's file descriptor switches to non-blocking
- coordination with `h2_req` threads is done with a file descriptor
  - based on #4348 originally submitted independently
  - this enables polling together with the socket
- control frames are sent from a fixed-length buffer living in workspace

I was the main reviewer of the change, and I personally extracted several commits (@mbgrydeland did too based on my reviews) in order to reduce the size of the main commit as much as possible:

> http2: Rework HTTP/2 to be using non-blocking sockets

After porting the change to this branch, I wish I had more aggressively extracted small changes into self-contained commits because the amount of conflicts to resolve made the port of this commit alone several days of work.

There were a lot of explicit (line) conflicts, and a fair amount of implicit commits because at this point trunk and Varnish Enterprise (based on 6.0) diverged in many small ways:

- Varnish Enterprise notably has
  - a waiting list "walk-away" feature (see #3835)
  - listen socket management (#3959)
  - task tracking with a `varnishscoreboard` utility
- and Varnish Cache has
  - a workspace emulator and hardened workspace semantics (#3644)
    - I defaulted to ws_emu+asan in my porting endeavor, it actually helped
  - a generalized VDP API (#4035)
  - a new asynchronous iterator for delivery (#4209)

Between these major changes (and probably others I'm forgetting) and lots of small core changes accumulated over the years, there was pretty much always a merge conflict for all the commits I cherry-picked. I snuck two original commits at the beginning, to reduce noise when conflicts happened during the port.

I caught two minor bugs in this port, and besides the workspace conflict, things went rather smoothly. It was just tedious to replay, so tedious that I frequently switched to other (important) tasks, which added delays. But I ran the test suite for each commit, stressing them with load, the workspace emulator paired with sanitizers.

The original change went to production, facing real world h2 traffic, that's why the patch series ends with a bunch of bug fixes. For similar workloads it's putting overall a lower load on the system, and it's more reactive. If you wish to use OpenSSL, you can't perform reads and writes concurrently, so moving all socket operations to the session thread also removes a good deal of contention for HTTPS workloads.

I'm requesting @mbgrydeland's review so he can make sure that the port is faithful to our original work. I'm also requesting @walid-git's review because he rebased his trailer implementation (Varnish Enterprise counterpart of #4125) on this architectural change (and eventually he should do the same for #4125). It was reportedly easier to implement with this new architecture.

I have one demand.

Leave structural changes aside if they can wait until after merging this. If they must really happen as part of this change, it will require new commits. I didn't apply the bug fixes in the commit introducing the bugs because keeping the same list of commits will help us track what has or has not been ported either way (among commits relevant in both projects).

I'm otherwise open to squashing whitespace improvements or general polish in the right commits, but I will be away for a couple weeks.

### DIFF
--- a/bin/varnishd/http2/cache_http2_proto.c
+++ b/bin/varnishd/http2/cache_http2_proto.c
@@ -272,11 +272,12 @@ h2_kill_req(struct worker *wrk, struct h2_sess *h2, struct h2_req **pr2,
 
 	h2_stream_setstate(r2, H2_S_CLOSED);
 
+	Lck_Lock(&h2->sess->mtx);
 	if (r2->scheduled) {
-		Lck_Lock(&h2->sess->mtx);
 		h2_reqbody_kick(r2);
 		Lck_Unlock(&h2->sess->mtx);
 	} else {
+		Lck_Unlock(&h2->sess->mtx);
 		h2_del_req(wrk, &r2);
 	}
 }


dridi:
Hello from a bus in the middle of nowhere. It was brought to my attention that someone was "not terribly happy about" me doing a "and don't do anything else, while I'm away on vacation" salute.

Apologies for giving this impression, I'll try to clarify what I wrote on a late Friday night since I was not sure I would get a reliable-enough internet access today to submit it.

> Leave structural changes aside if they can wait until after merging this.

What I mean here is not "don't do anything else" but rather, if you can help it, "do anything else" after merging. In other words, if you register review comments that could be applied afterwards, let's apply them afterwards.

> If they must really happen as part of this change, it will require new commits.

If you must insist on making changes before merging, "do anything else" as new commits at the end of the patch series.

> I didn't apply the bug fixes in the commit introducing the bugs because keeping the same list of commits will help us track what has or has not been ported either way (among commits relevant in both projects).

I could have submitted a smaller patch series, but that would make coordination harder for us, so in order to facilitate that kind of contribution (already gone to production) from us, it was easier to retain the same list of applicable patches.

Please do "anything else" while I'm on vacation, except introducing structural changes in the middle of the patch series. :saluting_face:

bin/varnishd/http2/cache_http2_proto.c
```diff
@@ -218,10 +219,16 @@ h2_del_req(struct worker *wrk, struct h2_req **pr2)
 		AZ(stvbuf);
 	}
 
-	Req_Cleanup(sp, wrk, r2->req);
+	req = r2->req;
+	CHECK_OBJ_NOTNULL(req, REQ_MAGIC);
+	r2->magic = 0;
+	req->transport_priv = NULL;
```
mbgrydeland, line 223: I believe these two lines need to be moved after `Req_Cleanup()`. This due to `Req_Cleanup()` calling `VDP_Fini()`, which I guess may for some VDP call on code expecting to see the complete structure.

mbgrydeland (approved):
I have gone through all the patches, and I believe this is a faithful port of the code. Only a couple of minor comments made.

dridi:
Thank you for your feedback. In order to move this forward, I plan to revert the latest VSV patches at the beginning of the patch series, and add the patches we originally wrote for this new architecture at the end.

But I'm not starting without a sign that this will receive attention.

bin/varnishd/http2/cache_http2_proto.c
dridi: Good catch, I forgot about the increased scope of `Req_Cleanup()` in this branch.

walid-git:
There is something confusing here:

>   err_code = 1, err_reason = (null),

`req->err_code` is used to make the difference between an opportunistic h2 upgrade and a prior knowledge one, and the value 1 corresponds to `H2_PU_MARKER` (prior knowledge). However, `req->err_code` is cleared [here](https://github.com/varnishcache/varnish-cache/blob/022025f/bin/varnishd/http2/cache_http2_session.c#L387), at the beginning of the h2 session. 
But, the panic indicates that the h2 session has been running for a while and has seen several streams:

>  open_streams = 0, highest_stream = 113, goaway_last_stream = 0,
 
 These two items seem to be contradicting each other.

PR #4373, opened 2025-08-01
[stat] use -1 is not outputting to a TTY

gquintard:
Simple quality-of-life improvement to automatically detect that the `varnishstat` output is being piped, and to use `-1` if that's the case.


walid-git:
Thanks @gquintard, the changes look good. Please proceed with the squashing and I will merge it.

gquintard:
done!

PR #4378, opened 2025-08-13
-x json parameter

gquintard:
I've been thinking that it would be pretty nice to have shell completion for `varnishd`, but the parameter part is a bit unwieldy because to discover all the parameters, one would need to parse the `varnishd -x parameter` output, which isn't fun.

So, I brute-forced a `-x paramer-json` implementation to gauge interest in implementing this.

If we do want to do this, I'm make sure we have tests before we merge, but the actual completion scripts can/should be worried about later

### DIFF
--- a/bin/varnishd/mgt/mgt_param.c
+++ b/bin/varnishd/mgt/mgt_param.c
@@ -968,3 +968,84 @@ MCF_DumpRstParam(void)
 		printf("\n\n");
 	}
 }
+
+void
+MCF_DumpJsonParam(void)
+{
+	struct plist *pl;
+	const struct parspec *pp;
+	const char *p, *q;
+	unsigned flags;
+	struct vsb *vsb = VSB_new_auto();
+
+	CHECK_OBJ_NOTNULL(vsb, VSB_MAGIC);
+
+	p = "\n";
+	VSB_cat(vsb, "{");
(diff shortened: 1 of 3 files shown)


bin/varnishd/mgt/mgt_param.c
```diff
@@ -968,3 +968,84 @@ MCF_DumpRstParam(void)
...
+	const struct parspec *pp;
+	const char *p, *q;
+	unsigned flags;
+	struct vsb *vsb = VSB_new_auto();
+
+	CHECK_OBJ_NOTNULL(vsb, VSB_MAGIC);
+
+	p = "\n";
+	VSB_cat(vsb, "{");
+	VSB_indent(vsb, 2);
```
dridi, line 985: You should use the VBOB API to build your object model and the VBOR API to turn it into JSON.

dridi (commented):
We now have an alternative to the VJSN API and hand-rolled JSON, could you please give it a try?

It's used in several places, and `vbor.h` tries to describe how the APIs work. Having someone other than @daghf (who reviewed it) giving it a try would be a good benchmark for its usability.

If the docs in `vbor.h` and examples from other places are not enough, then please complain in my general direction.

gquintard:
bump

PR #4381, opened 2025-08-18
No builtin pipe

walid-git:
It was agreed during a bugwash that we no longer want to pipe unknown request methods by default and instead return a synthetic 405 (Method Not Allowed) for the next 8.0 release.

Refs #3352


nigoroll:
bugwash:
a) add return(fail(status)) as a "lightweight synth" (without additional vcl control)
b) change core code such that we respect a `set req.http.connection = "close";` from vcl

Issue #4203, opened 2024-09-30
Unix socket backend leads to high cpu usage in some cases


jogoossens:
### Expected Behavior

No unexplained high cpu load when using unix socket on backend.

### Current Behavior

We had some longtime intermittent high cpu usage with varnish software. We spent a lot of hours troubleshooting and searching, finally we changed our backend to port 8080 and the high cpu usage stayed away.

When we (re)start varnish, initially there is no  high load at all, but after some time (one hour or a few hours) the vcache process experiences high cpu usage.

### Steps to Reproduce (for bugs)

We now switched back and forth between the following settings, and this change alone makes it "stable" versus "high cpu usage".

High cpu usage

```
backend default_backend {
    .path = "/var/lib/nginx/default.sock";
    .connect_timeout        = 5s;
    .first_byte_timeout     = 600s;
}
```

No high cpu usage

```
backend default_backend {
    .host = "127.0.0.1";
    .port = "8080";
    .connect_timeout        = 5s;
    .first_byte_timeout     = 600s;
}
```

### Context

We tried perf top and see high cpu usage in copying data, but it's not easy to see the source so far.

We have this problem for a long time (at least with varnish 7.0 and all following versions, probably 6.x  too but unsure)

### Varnish Cache version

varnishd 7.6.0

### Operating system

Debian 12

nigoroll:
We'll look at it, but it would be _really_ helpful if you could get some perf data, yes. Or a reproducer. Also could you please report which VMODs you use, if any?

nigoroll:
> We'll look at it, but it would be _really_ helpful if you could get some perf data, yes. Or a reproducer. Also could you please report which VMODs you use, if any?

ping @jogoossens ?

jogoossens:
Hi, somehow we bypassed the problem now (it was there for easily a year and earlier tries never got rid of it... ).

We disabled http2 completely with a startup parameter, not sure if that could be related. Possible we did other things too , but we tried so many things while testing this.

nigoroll:
Without any tangible information this issue is not actionable. If anyone has anything for us to work on, please reopen.

jogoossens:
I upgraded to varnish 8 as a test, but I still encounter 100% cpu usage. I assume the reason is something else unrelated to the unix sockets...

So any ways we could debug this? :)

PR #4386, opened 2025-09-01
vtcp,vus: Consistently establish blocking connections

dridi:
It was originally observed by @scance that unix-domain sockets were in a non-blocking state, which was unexpected. After some digging, he found the root cause and fixed it. It was also independently observed by @fwsGonzo.

This patch series ensures that all successful outcomes of `V{TCP,US}_connect()` calls yield a socket in blocking mode for consistency.

### DIFF
--- a/lib/libvarnish/vtcp.c
+++ b/lib/libvarnish/vtcp.c
@@ -298,6 +298,8 @@ VTCP_connect(const struct suckaddr *name, int msec)
 	AZ(setsockopt(s, IPPROTO_TCP, TCP_NODELAY, &val, sizeof val));
 
 	i = connect(s, sa, sl);
+	if (i == 0 && msec > 0)
+		VTCP_blocking(s);
 	if (i == 0)
 		return (s);
 	if (errno != EINPROGRESS) {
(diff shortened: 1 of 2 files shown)


dridi:
The root cause is that we create the socket, make it non-blocking, and upon success make it blocking again in `V{TCP,US}_connected()`, unless it was an immediate success.

dridi:
It could very well cause the problem described in #4203, creating a tight `EAGAIN` loop or some similar ill effect where we'd rather block once. I have not looked closer than the issue discussion.

dridi:
@nigoroll good point, but maybe `msec >= 0` instead for the blocking case. But it shouldn't make a big difference since the caller for `msec < 0` is not supposed to use the socket until after a successful `VTCP_connected()` call.

But with your suggestion (modulus `>=`) we would save one system call.

dridi:
It was brought to my attention that we don't bother making the socket non-blocking for the `msec == 0` case so I'll go with @nigoroll's suggestion.

It was also brought to my attention that we could make the socket non-blocking directly from the `socket(2)` call, but that looks like a linuxism so I'm not venturing in non-portable territory for a bug fix.

nigoroll:
also 👍🏽 for the improved commit message

dridi:
I didn't author the original commits, and as a reviewer myself I also complained about the commit message. I shared the bug fix as it was merged initially on our end.

Issue #4396, opened 2025-09-19
H2 workspace related panic


walid-git:
Originally posted by @desdic in https://github.com/varnishcache/varnish-cache/issues/4368#issuecomment-3291111550

The server was running the master branch and not #4368

```
Child (23116) Panic at: Sat, 13 Sep 2025 17:08:11 GMT
Assert error in h2_del_sess(), http2/cache_http2_session.c line 185:
  Condition(!WS_IsReserved(req->ws)) not true.
version = varnish-trunk revision 43fbfd6051f0ea3760d04b893ebebcb5351ad263, vrt api = 21.0
ident = Linux,6.8.0-51-generic,x86_64,-jlinux,-smalloc,-sdefault,-hcritbit,epoll
now = 11612746.444027 (mono), 1757783219.043331 (real)
Backtrace:
  ip=0x5931b56e2ab5 sp=0x734bb316a300 <VBT_format+0x65>
  ip=0x5931b56562cd sp=0x734bb316a420 <pan_ic+0x24d>
  ip=0x5931b56e1d39 sp=0x734bb316a580 <VAS_Fail+0x19>
  ip=0x5931b569e19b sp=0x734bb316a590 <h2_del_sess+0x11b>
  ip=0x5931b569ed37 sp=0x734bb316a5c0 <h2_new_session+0x8a7>
  ip=0x5931b5683122 sp=0x734bb316a870 <WRK_Thread+0x3c2>
  ip=0x5931b568370c sp=0x734bb316b420 <pool_thread+0x3c>
  ip=0x734bb329caa4 sp=0x734bb316b450 <pthread_condattr_setpshared+0x684>
  ip=0x734bb3329c3c sp=0x734bb316b500 <__clone+0x24c>
errno = 110 (Connection timed out)
argv = {
  [0] = \"/usr/sbin/varnishd\",
  [1] = \"-a\",
  [2] = \":80\",
  [3] = \"-a\",
  [4] = \"127.0.0.1:81,PROXY\",
  [5] = \"-a\",
  [6] = \"/var/run/varnish.sock,PROXY,user=vcache,group=varnish,mode=666\",
  [7] = \"-f\",
  [8] = \"/etc/varnish/onecom.vcl\",
  [9] = \"-T\",
  [10] = \":6082\",
  [11] = \"-t\",
  [12] = \"120\",
  [13] = \"-l\",
  [14] = \"512M\",
  [15] = \"-s\",
[...]

PR #4397, opened 2025-09-19
h2: Consistently release the session workspace in h2_rxframe

walid-git:
> All other code paths returning from h2_rxframe guarantee that the h2 session workspace is released as they all go through HTC_RxStuff. The only exception was this code path that is particularly hard to hit, as it requires h2->goaway to be set without having h2->error at the same time.

That code path is currently not covered by our tests ([gcov](https://varnish-cache.org/gcov/bin/varnishd/http2/cache_http2_proto.c.html)), I spent a lot of time trying to make a test case that reproduces the panic, but it is particularly hard with all the things going on concurrently in h2.

Fixes: #4396


walid-git:
bugwash: ok

PR #4400, opened 2025-09-24
4399 invalid refresh

nigoroll:
(tbd)


bin/varnishd/cache/cache_fetch.c
```diff
@@ -272,7 +272,7 @@ vbf_stp_mkbereq(struct worker *wrk, struct busyobj *bo)
 	}
 	http_ForceField(bo->bereq0, HTTP_HDR_PROTO, "HTTP/1.1");
 
-	if (bo->stale_oc != NULL &&
+	if (bo->stale_oc != NULL && (! (bo->stale_oc->flags & OC_F_DYING)) &&
 	    ObjCheckFlag(bo->wrk, bo->stale_oc, OF_IMSCAND) &&
 	    (bo->stale_oc->boc != NULL || ObjGetLen(wrk, bo->stale_oc) != 0)) {
 		AZ(bo->stale_oc->flags & (OC_F_HFM|OC_F_PRIVATE));
```
dridi, line 278: You may want to cherry-pick https://github.com/varnishcache/varnish-cache/pull/4082/commits/c461fb2779216e7e22a783508f6cba4a292835ed from #4082 that comes with test coverage.

nigoroll: done, thank you!

I edited the commit message slightly and added a comment that I did.

dridi: The wording is this way because of the preceding commit in #4082.

I'm +1 on not initiating a condfetch with a stale OC, but I don't have a strong opinion on abandoning the "object template" if the stale objcore is dying. The problem with a dying objcore at 304 time is that it could either have expired meanwhile, or have been actively invalidated.

That was at least the theoretical problem for #4082. Based on the comment of your original commit:

> It does not seem to make much sense to start a conditional request on a stale
object which VRT_r_obj_stale_is_valid() would classify it as invalid.

You may want to also pick the test case from https://github.com/varnishcache/varnish-cache/pull/4082/commits/588d103454f401f50824041f75ae5305b7238450 to have coverage for the built-in `sub vcl_refresh_valid`. You may even want to take the entire commit and bail out of revalidation, because VCL could bypass the built-in check, but then you are back to the question of passive expiry vs active invalidation. You could also decide that if VCL let it slide, then it can be reused.

I was misled to believe that #4082's scope was part of the `vcl_backend_refresh` work at the time, and didn't pay enough attention to make sure the dying stale objcore question got settled then and that's on me (but I do remember consensus against a retry straight to backend fetch return action that was also part of #4082 and that is fine).

bin/varnishd/builtin.vcl
```diff
@@ -230,7 +230,10 @@ sub vcl_refresh_conditions {
 # We currently only revalidate 200 responses
 sub vcl_refresh_status {
 	if (obj_stale.status != 200) {
-		return (error(503, "Invalid object for refresh"));
+		# Can happen if VCL adds If-Modified-Since / If-None-Match
+		# for non-200 status objects. If deliberate, this sub can
+		# be overridden with sub vcl_refresh_status { return; }
+		return (error(503, "Invalid object for refresh (status)"));
```
dridi, line 236: Again, I would change the error message to something more specific:

> Can only refresh for status code 200

And for the contents of the comment, I think they belong in either VCL design patterns in the sphinx docs, or the documentation for `vcl_refresh_status` if there's a dedicated section in a manual.

nigoroll: You are right that this should be covered there. But I also think we should keep the comments in built-in VCL, simply because this is just more helpful to users. We already have some comments in the same direction and I believe we should continue to add such hints.

PR #4402, opened 2025-10-03
fetch: Close data race between BOC state and objcore flags

dridi:
The changes from #4073 increases the likelihood of a data race between a BOC state check followed immediately by an objcore flag check:

```c
ObjWaitState(oc, BOS_STREAM);
AZ(oc->flags & OC_F_BUSY);
if (oc->boc->state == BOS_FAILED)
        AN(oc->flags & OC_F_FAILED);
```

Once we reach `BOS_STREAM`, we touch `oc->flags` to check `OC_F_BUSY`.

If the backend fetch failed immediately after reaching `BOS_STREAM`, we may then observe `BOS_FAILED` when we touch `oc->boc->state` but still see `oc->flags` from before the failure in this thread.

Because `oc->flags` and `oc->boc->state` are guarded by different locks, we can't safely rely on this construct. One option could be to move the `OC_F_BUSY` check down, and another, could be to capture the BOC state:

```c
state = ObjWaitState(oc, BOS_STREAM);
AZ(oc->flags & OC_F_BUSY);
if (state == BOS_FAILED)
        AN(oc->flags & OC_F_FAILED);
```

This pull request implements the state capture.


dridi:
I'm not sure what "here" is referring to, the link you shared is highlighting `if (state == BOS_FAILED)`.

Around this line there are two accesses to `oc->flags`:

```c
state = ObjWaitState(oc, BOS_STREAM);
AZ(oc->flags & OC_F_BUSY);
if (state == BOS_FAILED)
        AN(oc->flags & OC_F_FAILED);
```

The first one is always true so it's not a problem, and the second is safe too as long as the observed state is BOS_FAILED. We first fail the oc before failing the boc, so capturing `BOS_FAILED` is our warranty.

Whereas before, the sequence could have been:

- client resumes upon `BOS_STREAM`
- client observes no `OC_F_BUSY`
- backend fetch fails
- client observes `BOS_FAILED`
- client cannot observe `OC_F_FAILED`

nigoroll (approved):

PR #4404, opened 2025-10-06
various fixes to get master branch to compile and run on macOS

perbu:
First three commits are just trying to get things to compile without warnings. 

The last one is an overflow. No /proc on macOS. Falls back to sysconf(_SC_OPEN_MAX) returns LONG_MAX, which overflows the int and then assert(maxfd > 0) isn't true anymore.

I've built and ran make check on Linux afterwards. Seems to work.


perbu:
I suspect that failure is because of a flaky test. I can't see the relevance.

Issue #4403, opened 2025-10-05
%{VSL:...}x formats are hard to use and not well documented


rfc1036:
### Expected Behavior

varnishncsa(1) documents that the `%{VSL:tag:prefix[field]}x` format can be used to extract specific fields from the VSL.

### Current Behavior

I am unable to extract specific fields from VCL "TTL" tags. I wonder if this is caused by "TTL" not being "followed by a colon". But I am not sure of which VSL tags actually have prefixes followed by a colon, so it is not obvious to me when this functionality can be used. `%{VSL:TTL[2]}x` works, while `%{VSL:TTL:RFC[2]}x` does not.

Also, TTL tags may start with "VCL" instead of "RFC", but I never see the first ones logged.

Another related issue is that `%{VSL:TTL}x` only works in backend mode, while generally logging is enabled only for client mode. I have partially solved my use case by using `%{VSL:Hit[2]}x` instead in client mode, but it only works for cache hits.

### Steps to Reproduce (for bugs)

```
varnishncsa -b -F '%{VCL_Log:ip}x %l %u %t "%r" %s %b "%{Referer}i" "%{User-agent}i" handling=%{Varnish:handling}x ttl=%{VSL:TTL:RFC[2]}x:%{VSL:TTL}x'
```

Will output lines like:

```
 - - [05/Oct/2025:01:58:49 +0200] "GET ... HTTP/1.1" 200 27189 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36 Edg/114.0.1264.71" handling=- ttl=-:RFC 172800 10 0 1759622330 1759622330 1759622329 1759795129 172800 cacheable
```

This line shows that the TTL VSL tag is found, but it is not possible to extract a field.

### Context

I am trying to augment the standard logs with information about when responses are coming from the cache or not and what is their TTL.

### Varnish Cache version

varnishncsa (varnish-7.1.1 revision 7cee1c581bead20e88d101ab3d72afb29f14d87a)

### Operating system

Debian 12

### Source of binary packages used (if any)

http://deb.debian.org/

PR #4405, opened 2025-10-13
varnishncsa: Support space separator in %{VSL:tag:record-prefix[field]}x format

walid-git:
This makes varnishncsa also accept `record-prefix` that are followed by whitespace (in addition to colon) for the `%{VSL:tag:record-prefix}x` formatter. This can be useful to match specific log records in multiple situations.

Fixes: #4403

### DIFF
--- a/bin/varnishncsa/varnishncsa.c
+++ b/bin/varnishncsa/varnishncsa.c
@@ -647,7 +647,7 @@ addf_vsl(enum VSL_tag_e tag, long i, const char *prefix)
 	assert(i <= INT_MAX);
 	w->idx = i;
 	if (prefix != NULL) {
-		w->prefixlen = asprintf(&w->prefix, "%s:", prefix);
+		w->prefixlen = asprintf(&w->prefix, "%s", prefix);
 		assert(w->prefixlen > 0);
 	}
 	VTAILQ_INSERT_TAIL(&CTX.watch_vsl, w, list);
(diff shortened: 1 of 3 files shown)


bin/varnishncsa/varnishncsa.c
```diff
@@ -895,12 +895,19 @@ parse_format(const char *format)
...
 static int
 isprefix(const char *prefix, size_t len, const char *b,
-    const char *e, const char **next)
+    const char *e, const char **next, const char *sep)
 {
 	assert(len > 0);
 	if (e < b + len || strncasecmp(b, prefix, len))
 		return (0);
 	b += len;
+	if (sep != NULL) {
```
nigoroll: use `strpbrk()`?

nigoroll (approved):
Just a minor suggestion...

bin/varnishncsa/varnishncsa.c
dridi: You have 3 legitimate separators with the new prefix definition in varnishncsa: `':'`, `' '` and `'\t'` (VSL prefix and field separators).

nigoroll (approved):

bin/varnishncsa/varnishncsa.c
nigoroll: Sorry @dridi, I agree that we should be consistent. But are we really using `/t` anywhere in VSL?

Issue #4398, opened 2025-09-23
vtest2 integration revisited


nigoroll:
Yesterday, @rfc1036 raised an issue on irc that apparently our vtest2 integration as a git submodule has broken the workflow of some packagers, who now assume that a project could be completely built from just a git clone:

> I have just discovered that now the varnish repository uses the vtest2 submodule, which is a huge pain in a Debian contest. is there any chance to persuade you to use git-subtree instead?
> more specifically, Debian is (slowly) moving from the old tar archives source packages to git-first workflows (multiple ones of course, it's Debian...) which can be used to directly build and generate the source packages, and submodules are hard to implement there
> of course packages cannot download a submodule at build time because no network access is allowed

I have some reservations against the _git-subtree_ approach, because I fear it would bring us back to the diverging trees as before (despite better tooling, the problem is basically the same as before - having more than one source of truth), so I would like to ponder alternative approaches - all under the premise to _help packagers_ and specifically not make their lives harder than absolutely necessary.

#### Build from the dist archive

From my perspective, this is the baseline: We as a project produce a distribution archive, and it is complete and self-contained. The problem here is that apparently some packagers now have additional needs (requirements?):

> some people like me like to have the full upstream history in their package repository, which I would not have if I just imported the tar archives
> yes, there is a bit of a debate about packaging (and verifying the signature of) tar archives vs. git trees. but I think that the future will obviously be about preferring git trees as much as possible

#### Package vtest2

This option was also mentioned by Md:

> if vtest2 had releases then I would just package it by itself, but since it is basically a source-only library I can only create an intermediate upstream branch in the Debian repository where I manually merge the vtest2 repository

This option was already on our list as of #3983 

>  Split vtest into a library and a program

I maintain the position that we do not need this option to fulfill our own requirements, but if it helps packagers, why not add the option to use an external `vtest` for `make check`?

This would conflict with the plan to have certain vinyl-cache specific vtest extensions in our tree and not in the vtest tree, but that idea was already shelved by the argument that vtest should be able to use all components which run vtest.

#### Create a new kind of "dist archive"

If packagers now need the complete git history, why not create a _gist archive_ (pun intended) containing all the `git-bundle` s of all needed projects?

Then checking out the vinyl repo would not require much more than (pseudo code)

```bash
if -f ../vtest2.bundle ; then
git submodule set-url bin/varnishtest/vtest2 ../vtest2.bundle
git -c protocol.file.allow=always submodule update --init
fi
```

[...]

PR #4406, opened 2025-10-15
support for externally built vtest with libvtc_varnish.so and libvtc_builtwith.so

nigoroll:
This PR adds the option to use an externally built vtest binary as agreed in #4398.

#### Install VTest2 with extension support

* Build and install the `ext` branch from https://github.com/vtest/VTest2/pull/7 such that vtest is found in `$PATH` (the example here is intended to show a "canonical" way as with external packaging, I would not normally install unter `/usr`):

```bash
slink@haggis21:~/Devel/varnish-git/VTest2 (ext)$ sudo make install PREFIX=/usr
umask 022 && \
mkdir -p /usr/bin /usr/include/vtest /usr/lib/pkgconfig && \
cp vtest /usr/bin && chmod 755 /usr/bin/vtest && \
cp src/vtc.h src/cmds.h lib/vdef.h lib/miniobj.h lib/vas.h lib/vsb.h /usr/include/vtest && \
cp vtest.pc /usr/lib/pkgconfig && \
chmod 644 /usr/include/vtest/*.h /usr/lib/pkgconfig/vtest.pc
```

quick check:

```bash
slink@haggis21:~/Devel/varnish-git/VTest2 (ext)$ type vtest
vtest is hashed (/usr/bin/vtest)
slink@haggis21:~/Devel/varnish-git/VTest2 (ext)$ vtest -h 2>&1 | grep exten
    -E extension.so              # Load extenstion
slink@haggis21:~/Devel/varnish-git/VTest2 (ext)$ pkg-config --cflags vtest
-I/usr/include/vtest 
```

#### Build vinyl-cache in out-of-tree vtest mode"

This can be _later_ be done by not cloning the submodule (because the TODO items are not complete, see below), but for now we simulate a missing submodule by deleting the file which we look for to determine if the submodule is present:

```bash
slink@haggis21:~/Devel/varnish-git/varnish-cache (vtest_ext_varnish)$ rm bin/varnishtest/vtest2/src/vtc_main.c 
```

Now build as always, e.g.

```
./autogen.sh && ./configure && make -j40 check
```

All tests should succeed

### Implementation / Explanation

https://github.com/vtest/VTest2/pull/7 adds to VTest the option to load a shared object at runtime, which can then extend vtest functionality by registering additional commands.

#### Moving vinyl test functionality to `libvtc_varnish.so`

The first commit changes the build to create such a shared object with the vinyl-specific test mechanics in `vtc_varnish.c`,  `vtc_vsm.c` and `vtc_logexp.c`.

With this, calling `vtest -E libvtc_varnish.so` gives us a the same functionality we had before with the static build.

To keep the rest of the build simple, a wrapper program called ... (you guessed it) ... `varnishtest` is added which simply calls `vtest` with the extension argument. This is used in favor of the simple shell script 2-liner which could also easily fulfill the task in order to be able to get the `libtool` wrapper script for uninstalled libraries. This in turn requires linking to an empty dummy library.

#### Moving compile-time flag checks / macros to `libvtc_builtwith`

`varnishtest` already works for the most part with just the above, but we have just broken much of the `feature` functionality and the `pkg_version` and `pkg_branch` macros: These can only work if the respective code is built together with the code base it is supposed to reflect on - otherwise we get behavior based on however `vtest` was compiled, but not how vinyl-cache was compiled.

So we apply the same idea again and move these checks to a vtest extension with we _do_ compile each time. The outcome is a working `varnishtest` as demonstrated.

### Alternatives considered

We could do without the extra `libvtc_builtwith` and just move the respective code into `libvtc_varnish`, but I think this functionality might also be of general interest to other projects and I would add this as an example to VTest2 itself if we decide that we want to go this route.

The `builtwith` checks could also be replaced with compile time configuration of the tests to execute. I actually have working code which moves all the `-spersistent` checks to a subdirectory of `test` and includes these tests only if `--with-persistent-storage` is given to `configure`.

But in particular with negated tests I think this quickly becomes really tedious, and the information which tests to run is no longer nicely contained in the tests itself (unless we add something like magic comments which we grep for...). All in all, the "feature test" way is much clearer and easier to maintain.

In general, another alternative would be to invert the sense of what is an executable and what is a library: Create a `libvtest.so` in the `VTest2` project and then an executable in vinyl which only has minimal code to call the library and extend it. In my mind this simply seemed like the wrong way around and I found the extension model very clear, but opinions might vary and at this point, without having written the actual code, I would think that the alternative should also work.

### TODOs

A main TODO item is that that switch to the new model is not complete, and deliberately so:

For one, I think that keeping vtest bundled as a submodule will make the life for most of the developers easier. The problem to solve is to _enable_ a build from git with an external vtest, and I see no strong reason to destroy the existing workflow.

Secondly, the vinyl specific code still needs to be checked into the vinyl-cache repo again and removed from Vtest2.

Thirdly, the `feature` tests which have now been added to `builtwith` need to be removed from the former.

But all of this only makes sense after the changes are accepted.

### Notes

`make dist` still requires the submodule, and rightly so: Again, this addition is to support builds from git only, not to destroy previous workflows.

### DIFF
--- a/bin/varnishtest/Makefile.am
+++ b/bin/varnishtest/Makefile.am
@@ -127,12 +127,14 @@ else
 ################################################################################
 ## BEGIN vtc_varnish ext
 
-lib_LTLIBRARIES = libvtc_varnish.la libdummy.la
+lib_LTLIBRARIES = libvtc_varnish.la libvtc_builtwith.la libdummy.la
 
 # XXX TODO
 # - for actual move, the sources need to be relocated from VTest2 back to here
 # - for actual move, remove -DVTEST_WITH_...
-# - solve WITH_PERSISTENT_STORAGE
+# - remove feature tests now in builtwith from Vtest2 core
+# - remove pkg_version pkg_branch from Vtest2 core
+# - add libvtc_builtwith to Vtest2 as an example
 libvtc_varnish_la_SOURCES = \
 	vtest2/src/vtc_varnish.c \
 	vtest2/src/vtc_vsm.c \
@@ -150,6 +152,10 @@ libvtc_varnish_la_LIBADD = \
 	$(top_builddir)/lib/libvarnish/libvarnish.la \
 	$(top_builddir)/lib/libvgz/libvgz.la \
 	${PTHREAD_LIBS} ${NET_LIBS} ${LIBM}
+
+libvtc_builtwith_la_SOURCES = vtc_ext_builtwith.c
+
+libvtc_builtwith_la_CFLAGS = @VTEST_CFLAGS@
 
 libdummy_la_SOURCES = libdummy.c
 
--- a/bin/varnishtest/tests/b00068.vtc
+++ b/bin/varnishtest/tests/b00068.vtc
@@ -1,6 +1,6 @@
 varnishtest "Check timeout_linger"
 
-feature !workspace_emulator
+builtwith !workspace_emulator
 
 # XXX this test exploits the fact that the struct waited is
 # left near the free pointer of the session ws when a session
--- a/bin/varnishtest/tests/c00057.vtc
+++ b/bin/varnishtest/tests/c00057.vtc
@@ -2,7 +2,7 @@ varnishtest	"test sigsegv handler"
 
 # Under ASAN, the stack layout is different and STACK OVERFLOW is
 # never printed.
-feature !asan
+builtwith !asan
 
 server s1 {
 	rxreq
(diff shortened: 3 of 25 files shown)


nigoroll:
I noticed that I was making the life of reviewers unnecessarily hard by not providing at least a draft of a version without the submodule. So here it is:

https://github.com/nigoroll/varnish-cache/tree/vtest_ext_varnish_nosubmodule

PR #4407, opened 2025-10-15
[ci] update target distributions

gquintard:
remove `ubuntu:focal`, add `debian:trixie` and `almalinux:10`

closes: https://github.com/varnishcache/varnish-cache/pull/4407


gquintard:
I'm looking at this soon, but the goal is to migrate to https://github.com/varnish/all-packager , which is already producing packages (notably for `almalinux:10` and `debian:trixie`)

You can find the instructions here: https://varnish.org/docs/install-guide/

Issue #4409, opened 2025-10-29
SIGTERM should not log as Error


carste98:
Hi!

When running varnish in for example a Kubernetes environment, the container is regularly restarted as part of upgrades.

I have noticed that everytime the container receives a SIGTERM signal (which it gets everytime it is restarted) it prints.
> Error: Manager got SIGTERM

I would not classify it as an "application error" when the container receives a SIGTERM signal during an expected rolling upgrade.

This "Error" is then also sent to the logging stack and creates noise in the logging dashboards. 

It is fully possible to get around this with some custom log filtering, but it would be nice if the log level got changed at source!

nigoroll:
You could implement an explicit `varnishadm stop` instead of sending the signal.

dridi:
If it were up to me I would have already pushed the diff above. Please note that MGT complaints are also syslog'd, but I'm not sure what that entails in a kube environment.

PR #4411, opened 2025-11-03
mgt: Refine SIGHUP and SIGTERM handling

dridi:
This started as a mere fix for #4409 but a doc update was requested during bugwash. I felt that I should mention SIGHUP for the foreground mode but it turns out we don't honor it so the scope expanded a little.


nigoroll:
Picked just one commit, reworded commmit message slightly and merged as 58d43d2dffc6fb3dbfc927746509d1d39f48cb78

Issue #4410, opened 2025-11-03
vcl: Cannot have unreferenced probes


dridi:
### Expected Behavior

Disabling the `vcc_feature::err_unref` flag should allow unused probes.

### Current Behavior

Currently this test case passes, but it shouldn't:

```vcl
varnishtest "unreferenced probe"

varnish v1 -cliok "param.set vcc_feature -err_unref"

varnish v1 -errvcl "error=unused-const-variable" {
	backend be none;

	probe p { }
}

varnish v1 -errvcl "error=array-compare" {
	backend be none;

	probe p { }

	sub vcl_init {
		if (p == p) { } # not even a workaround
	}
}
```

### Context

Tested in trunk.

### Varnish Cache version

git describe: varnish-8.0.0-27-g7e9b65c62c

walid-git:
[7a88cab](https://github.com/varnishcache/varnish-cache/commit/7a88cab9fe0ea567e3ceac9fa9a9cd7108e85504) does not fix the `self-comparison always evaluates to true` error.

nigoroll:
> [7a88cab](https://github.com/varnishcache/varnish-cache/commit/7a88cab9fe0ea567e3ceac9fa9a9cd7108e85504) does not fix the `self-comparison always evaluates to true` error.

[Yes](https://github.com/varnishcache/varnish-cache/pull/4418#issuecomment-3590048382)

PR #4421, opened 2025-12-02
VCC: New 'unused' action

walid-git:
This allows marking individual VCL symbols as intentionally unused, allowing a fine grained control compared to the _vcc_feature_ `err_unref` parameter, and avoiding the need for other error prone workarounds.

Refs #4410
Refs #4418


walid-git:
It should be possible to have it at the top level. Do we want to keep both or just the top level one ?

PR #4427, opened 2025-12-10
Add std.rfc_ttl()

nigoroll:
From the added documentation:

Re-calculate the object timers (``beresp.ttl``, ``beresp.grace`` and ``beresp.grace``) and cacheability status based on the current state of ``beresp`` as if it had been processed by core code before ``vcl_backend_response`` was called.

This is useful to get the default ttl calculations after modifications of relevant properties like ``beresp.status``, ``beresp.http.Date``, ``beresp.http.Age`` or ``beresp.http.Cache-Control``.

### DIFF
--- a/vmod/vmod_std.vcc
+++ b/vmod/vmod_std.vcc
@@ -696,6 +696,19 @@ $Function DURATION timed_call(SUB)
 Call the given SUB and return a high precision measurement of the
 execution time.
 
+$Function VOID rfc_ttl()
+
+$Restrict vcl_backend_response vcl_backend_refresh
+
+Re-calculate the object timers (``beresp.ttl``, ``beresp.grace`` and
+``beresp.keep``) based on the current state of ``beresp`` as if it had been
+processed by core code before ``vcl_backend_response`` was called. This does not
+change ``beresp.uncacheable``.
+
+This is useful to get the default ttl calculations after modifications of
+relevant properties like ``beresp.status``, ``beresp.http.Date``,
+``beresp.http.Age`` or ``beresp.http.Cache-Control``.
+
 SEE ALSO
 ========
 
--- a/vmod/tests/std_b00015.vtc
+++ b/vmod/tests/std_b00015.vtc
@@ -1,0 +1,87 @@
+varnishtest "std.rfc_ttl()"
+
+server s1 {
+	rxreq
+	txresp -status 202 -hdr "Cache-Control: public, s-maxage=4711"
+
+	rxreq
+	expect req.url == "/200"
+	txresp -status 200 \
+	       -hdr "Cache-Control: public, s-maxage=0, stale-while-revalidate=99" \
+	       -hdr {Etag: "foo"} \
+	       -bodylen 64
+
+	rxreq
+	expect req.url == "/200"
+	expect req.http.If-None-Match == {"foo"}
+	txresp -status 304 -hdr "Cache-Control: s-maxage=11, stale-while-revalidate=77"
+} -start
+
+varnish v1 -vcl+backend {
+	import std;
+
+	sub vcl_backend_refresh {
+		set beresp.http.Cache-Control = "s-maxage=10, stale-while-revalidate=42";
+		std.rfc_ttl();
+	}
+
+	sub vcl_backend_response {
+		if (beresp.status == 202) {
+			set beresp.status = 200;
+			std.rfc_ttl();
+			set beresp.status = 202;
+		}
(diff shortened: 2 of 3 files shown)


vmod/vmod_std.vcc
```diff
@@ -696,6 +696,19 @@ $Function DURATION timed_call(SUB)
 Call the given SUB and return a high precision measurement of the
 execution time.
 
+$Function VOID rfc_ttl()
+
+$Restrict vcl_backend_response
```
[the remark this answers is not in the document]
nigoroll: Not sure because we will go through `vcl_backend_response` anyway, on the other hand I so no harm caused either. Do you have a particular scenario in mind? Like only using this on 304s?

nigoroll:
> maybe we should also offer the opposite?

Yes, some standard function to create `Cache-Control` based on the state of an object could be useful. Bug please let's keep this separate.

nigoroll:
I have updated the PR to also allow `std.rfc_ttl()` from `vcl_backend_refresh`, for which most of the work was to adjust the test case, because I got lured into a rabbit hole regarding `ttl <= 0s` being uncacheable and whether or not that still makes sense. But I got out before turning the hole into a burrow.

vmod/vmod_std.vcc
```diff
@@ -696,6 +696,19 @@ $Function DURATION timed_call(SUB)
 Call the given SUB and return a high precision measurement of the
 execution time.
 
+$Function VOID rfc_ttl()
+
+$Restrict vcl_backend_response vcl_backend_refresh
+
+Re-calculate the object timers (``beresp.ttl``, ``beresp.grace`` and
+``beresp.grace``) and cacheability status based on the current state of
```
SashankBhamidi: Doc typo: `beresp.grace` mentioned twice

```suggestion
Re-calculate the object timers (``beresp.ttl``, ``beresp.grace`` and
``beresp.keep``) and cacheability status based on the current state of
```

vmod/vmod_std.vcc
```diff
@@ -696,6 +696,19 @@ $Function DURATION timed_call(SUB)
...
+``beresp`` as if it had been processed by core code before
+``vcl_backend_response`` was called.
+
+This is useful to get the default ttl calculations after modifications of
+relevant properties like ``beresp.status``, ``beresp.http.Date``,
+``beresp.http.Age`` or ``beresp.http.Cache-Control``.
```
SashankBhamidi: This function rewrites `t_origin` and timers, which affects Age/freshness semantics. A short warning would prevent misuse.

```suggestion
This is useful to get the default ttl calculations after modifications of
relevant properties like ``beresp.status``, ``beresp.http.Date``,
``beresp.http.Age`` or ``beresp.http.Cache-Control``.

.. warning::
   This rewrites the object's internal origin time (``t_origin``) as well as
   TTL/grace/keep. Call it only after you have finished rewriting the relevant
   inputs (``Date``, ``Expires``, ``Cache-Control``, ``Age``), and avoid calling
   it repeatedly unless you really mean to.
```

SashankBhamidi (commented):
Hi, I've had a look at your pull request and I have a few suggestions. I'm new to this codebase, so please feel free to correct me if I'm wrong. 

Overall, I think this is a great addition to the std VMOD.

vmod/vmod_std.vcc
```diff
@@ -696,6 +696,19 @@ $Function DURATION timed_call(SUB)
...
 execution time.
 
+$Function VOID rfc_ttl()
+
+$Restrict vcl_backend_response vcl_backend_refresh
+
+Re-calculate the object timers (``beresp.ttl``, ``beresp.grace`` and
+``beresp.grace``) and cacheability status based on the current state of
+``beresp`` as if it had been processed by core code before
+``vcl_backend_response`` was called.
```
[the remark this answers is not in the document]
nigoroll: also correct, thank you

vmod/vmod_std.vcc
nigoroll: I would avoid this complication and do not see reason for a warning - which also can be irritating and unsettling: `t_origin` is implied in `beresp.ttl` and `beresp.http.Age`. We generally try to avoid exposing internals to users, as the abstractions in VCL have been chosen deliberately. They should be consistent, and should not need explanations from the implementation layer below.

SashankBhamidi: Thanks for explaining, it helps me better understand the project's design philosophy

Issue #4372, opened 2025-07-30
improve docs regarding builtin.vcl


nigoroll:
https://github.com/varnishcache/varnish-cache/blob/master/doc/sphinx/users-guide/vcl-built-in-code.rst might need an update.
During bugwash, consensus was that the purpose of "extensible" builtin subs should be explained in builtin.vcl. The rst referenced above should probably explain the concepts and reference these comments.

#3994 could make some steps in this direction

PR #4430, opened 2025-12-19
docs: Improve documentation for builtin VCL and pipe behavior

SashankBhamidi:
This addresses issue #4372 by improving documentation for the built-in VCL, based on feedback from @nigoroll and @walid-git.

The changes are split into two commits:

1.  **`docs: Document vcl_pipe default behavior change`**/[843ef0f](https://github.com/varnishcache/varnish-cache/pull/4430/commits/843ef0feea623c89e390b54ffe2c41d954fa5729): Updates the user guide to explain that Varnish no longer pipes unknown methods by default. It adds examples for re-enabling `pipe` for specific methods or requests and clarifies how to view the builtin VCL with `varnishd -x builtin`.
2.  **`docs: Explain extensible builtin subs concept in builtin.vcl`**/[142d735](https://github.com/varnishcache/varnish-cache/pull/4430/commits/142d735e4c6290901f0d91aaae01fda5584eb90b): Adds a comprehensive comment block to the top of `builtin.vcl` explaining the two patterns for extending default behavior (`vcl_builtin_*` subs vs. `vcl_req_*` hooks).

### DIFF
--- a/bin/varnishd/builtin.vcl
+++ b/bin/varnishd/builtin.vcl
@@ -28,7 +28,25 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 # SUCH DAMAGE.
 #
-# This is the builtin VCL code
+# This file contains the builtin VCL code.
+#
+# All subroutines in this file can be overridden by user-provided VCL.
+#
+# There are two main patterns for overridable subroutines:
+#
+# 1. `vcl_builtin_*` subroutines: These wrap core Varnish logic. When you
+#    override these in user VCL, your code runs first, followed by the
+#    builtin VCL code (unless you `return()`). Note that calling
+#    `vcl_builtin_*` from your own `vcl_builtin_*` does not work (triggers
+#    recursion check).
+#
+# 2. `vcl_req_*`, `vcl_beresp_*`, etc. subroutines: These act as specific
+#    "hooks" within the default Varnish behavior. Overriding these allows
+#    you to modify a narrow aspect of behavior without reimplementing the
+#    entire parent `vcl_builtin_*` logic.
+#
+# Consult the Varnish documentation for details on each subroutine and
+# its intended use.
 
 vcl 4.0;
 


nigoroll (commented):
I have mixed feedback about this suggestion.

the pipe mode explanation makes sense, and I am 👍🏽 on it specifically

the explanation of the `vcl_builtin_*` subs looks like an improvement at first, but I am not sure how helpful it really is: The reference to the naming scheme is incomplete, and the main motivation is subtly different to the added explanation: All subs of the `builtin.vcl` can be overridden, that is the whole point of appending the built-in VCL to the user VCL. IIUC, the main point about the `vcl_builtin_*` subs is to make them available to calls from custom VCL, while subs like `vcl_req_*` allow specific extensions of parts of the built-in behavior only.

Repeating the same comment all over the place in the built-in VCL does not seem to make any sense to me at all because, again, there is nothing special about the possibility to override _some_ subs, they can _all_ be overridden.

Maybe @gquintard or @dridi should document their original design principles? My understanding might still not match what they had in mind (Ref: 3f3d67d9da491155bda9f1f1d4986b5228db5fd6)

Similar thing with the comment on git: Sure, all we do is available in git, but that is no news.

Regarding the overall structure of the PR, I would suggest to split out at least the `pipe` example into a separate commit.

doc/sphinx/users-guide/vcl-built-in-code.rst
```diff
@@ -176,7 +176,8 @@ Built-in VCL reference
 
 A copy of the ``builtin.vcl`` file might be provided with your Varnish
 installation but :ref:`varnishd(1)` is the reference to determine the code
-that is appended to any loaded VCL.
+that is appended to any loaded VCL. The ``builtin.vcl`` file is also
```
walid-git: This part is unrelated to the rest of the commit, and as @nigoroll already mentioned, I don't think it adds much to say that it's available in git since it's the case for everything else. Instead, a more useful information would be to mention how to read it with `varnishd -x builtin`.

nigoroll (approved):

